PR #395, opened 2014-05-18
Add flycheck-ghc-packages, a list of visible GHC packages

gracjan:


gracjan:
Hej, what up with this one?

lunaryorn:
@gracjan I'm not sure that we still need it.  It's only for the rare case of conflicting identifiers in different packages, and you can already use `flycheck-ghc-args` to add the required arguments directly.

lunaryorn:
@gracjan Besides, sorry for the delay, I had completely forgotten about this PR in the aftermaths of the new generic checkers.

gracjan:
Generic args revolution? How could I have missed that!

I remember arguing for generic args some time ago and you were against. What has changed your mind?

Closing as not useful anymore.

lunaryorn:
GHC and some other checkers have just too many special cases, which would require special options.  Also, it's easier for extensions to extract settings from build systems, if they can just add compiler options from the build system directly.

I'd still like to have dedicated options for common settings, though, like include directories and warning levels and stuff like that.  It's easier to use dedicated options as file local variables, and we can provide a better customise interface for dedicated options.

PR #512, opened 2014-11-08
feature/R

dholm:
This pull request is incomplete!

I wrote these checkers while taking a Coursera class on R programming earlier this year.  The reason I never submitted the pull request was because I got stuck adding the test cases because they required R packages to be installed and I was unable to figure out how to do that with puppet.

In the time between adding these checkers and submitting this pull request I accidentally removed my original flycheck directory and apparently I had never commited the test cases so they have since been lost.  I can write them again but I will need help figuring out how to install those R packages and it looks like puppet is no longer used to manage test case dependencies?

I'm currently a bit constrained on time so expect my response time to be slow.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -221,6 +221,7 @@ attention to case differences."
     python-pylint
     python-pycompile
     r-lint
+    r-svtools
     racket
     rpm-rpmlint
     rst
@@ -6572,6 +6573,22 @@ See URL `http://cran.r-project.org/web/packages/lint/'."
   ((info line-start "Lint: " (minimal-match (message))
          ": found on lines " line (optional (one-or-more ", " line))
          line-end))
+  :modes ess-mode
+  :next-checkers (r-svtools))
+
+(flycheck-define-checker r-svtools
+  "An R syntax checker using the svTools package.
+
+See URL `http://cran.r-project.org/web/packages/svTools/'."
+  :command ("R" "--slave" "--restore" "--no-save" "-e"
+            (eval (concat
+                   "library(svTools);"
+                   "try(lint(commandArgs(TRUE), "
+                   "type = \"flat\", sep = \":\"))"))
+            "--args" source)
+  :error-patterns
+  ((warning "warning:" line ":" column ":" (message) line-end)
+   (error "error:" line ":" column ":" (message) line-end))
   :modes ess-mode)
 
 (flycheck-define-checker racket
--- a/flycheck.el
+++ b/flycheck.el
@@ -221,6 +221,7 @@ attention to case differences."
     python-pylint
     python-pycompile
     r-lint
+    r-lintr
     r-svtools
     racket
     rpm-rpmlint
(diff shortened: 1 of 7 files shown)


flycheck.el
```diff
@@ -6558,6 +6561,56 @@ See URL `https://docs.python.org/3.4/library/py_compile.html'."
           line ", " column ", " (one-or-more not-newline) line-end))
   :modes python-mode)
 
+(flycheck-define-checker r-lint
```
lunaryorn: Do we really need this syntax checker?  It seems as if it's not maintained anymore (last release almost two years ago, no public repo).  To me it looks as if `lintr` is the way to go.

flycheck.el
```diff
@@ -6558,6 +6561,56 @@ See URL `https://docs.python.org/3.4/library/py_compile.html'."
...
+  ((info line-start (file-name) ":" line ":" column ": style: " (message)
+         line-end)
+   (warning line-start (file-name) ":" line ":" column ": warning: " (message)
+            line-end)
+   (error line-start (file-name) ":" line ":" column ": error: " (message)
+          line-end))
+  :modes ess-mode
+  :next-checkers ((warning . r-svtools)))
+
+(flycheck-define-checker r-svtools
```
lunaryorn: @dholm Can you explain the difference between `svtools` and `lintr`?  The svtools site isn't very informative about the purpose and the features of this package.  Also, it doesn't seem to be well-maintained either:  svtools seems to be a side-kick of a GUI library, rather than a fully-fledged independent linter.

flycheck.el
dholm: Agreed.  I found it after I had completed the implementation for the other two.

flycheck.el
dholm: I tried to figure out the differences myself but was unable to.  It's fine by mine if you only want to support lintr.

dholm:
@lunaryorn Add the following line to your ~/.Rprofile and see if it works:

    options(repos=structure(c(CRAN="http://cran.rstudio.com/")))

The domain automatically redirects to the "closest" CRAN mirror.

lunaryorn:
@dholm What `~/.Rprofile`?  That of my user, or that of `root`.  Note that I used `sudo` to install the package.

Besides, can we somehow create the same effect *without* modifying any file?  E.g. with a special argument to `package.install` or so?  Modifying a file is just an extra complexity that I'd like to avoid.

dholm:
@lunaryorn For the user that will run `R` and the command to install `lintr`.  I googled it and it's also possible to make it part of the package installation command, i.e. `R --slave --restore --no-save -e 'install.packages("svTools", repos="http://cran.rstudio.com/")'`.

lunaryorn:
@dholm Thanks, I had already discovered that.  It's working well, but unfortunately lintr takes a long time to install to a large number of R dependencies.

lunaryorn:
@dholm Besides, I also discovered that the lintr author wrote [Flycheck integration](https://github.com/jimhester/lintr/blob/master/inst/flycheck/lintr.el) themselves.  I'll like the option for linters, but I'm not sure about the caching thing.  Do you have any idea?

Inviting @jimhester into this discussion, so that he's aware that we want to upstream lintr support in Flycheck.

jimhester:
the caching parameter is caching internal to lintr (basically you cache all the expressions in the file and only re-lint those that have changed since the last time).  It is fairly important as linting can take a fair amount of time on files > 5-10k lines or so.

I based my flycheck integration on @dholm's versions for the older linters.  Feel free to use it directly if that is easiest.  I have not done a ton of testing on it other than to verify it works for a few simple cases (as emacs is not my primary editor) so I would welcome feedback from those who do use emacs frequently.

In terms of features I think `lintr` lints most if not all of the same things as both `svTools` and `lint`, with many more additional linters as well.  It is probably safe to drop support for them, as neither is actively maintained.

Let me know if there is anything I can do to help and I will gladly do what I can.

lunaryorn:
@jimhester Thank you for joining this discussion, and for your input.

If I understand your comment aright, caching is really important for large files, but I wonder why you made it optional.  Why would a user want to disable caching?  Are there any drawbacks?

In other words, shouldn't we just unconditionally enable caching?  Would that cause any harm?

lunaryorn:
@dholm @jimhester I'd just like to inform you that feature is ready to be merged to `master`, I'm justing waiting for @jimhester's repsonse re. caching, i.e. whether we should have an option for caching, or can safely enable it globally.

jimhester:
I have an options for enabling caching because there are some cases where the cache does not properly account for dependencies between expressions, so the cache gets out of sync and must be cleared to provide accurate linting.  These situations are rare, however I feel like giving the user the option to disable caching if they are worried about this to be beneficial.  However as these situations are rare in practice having caching being enabled by default seems to be the right thing as well.

lunaryorn:
@dholm Merged.  Many thanks for this contribution!

@jimhester Thank you for your explanations.  I enabled caching by default, but added an option to disable it.

I merged the new syntax checker to `master`, and it'll be part of Flycheck 0.23.  I have no ETA for the release, but if you use MELPA, you should get the new syntax checker within some hours.

You may want to update the lintr README accordingly.

dholm:
@lunaryorn Awesome, thank you.

@jimhester Thanks for your linter.  It is much nicer than svTools/lint!

PR #521, opened 2014-11-16
[Fix #515] Add option for system include paths for Clang and GCC

### LINKED ISSUE #515: [C/C++] Please support specifying arguments to "-isystem"
The `-isystem` option is used for example in the build system for the Linux kernel, and is useful for suppressing warnings in system headers, please see: http://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-in-system-headers

There are a few more such options, like `-nostdinc`, `-isysroot`, `--system-header-prefix` etc., which are very useful in some specific cases, but might not warrant their own specific user options. Could we perhaps have a single "catch-all" user option, to specify such settings?

Small note: in Flymake's Info manual, the entry for `User Option: flycheck -clang-standard-library` has a broken link:

```
Refer to the Clang manual at URL
‘http://clang.llvm.org/docs/UsersManual.html’(3) for more
information about the standard library.
```

Clicking on the URL (note: not the footnote) tries to open `http://clang.llvm.org/docs/UsersManual.html%E2%80%99`, which gives a "Not Found" error.
The same link as shown by `C-h v flycheck-clang-standard-library <RET>` works without problems.

```
$ emacs --version
GNU Emacs 24.4.1
...
```

ackalker:
This allows suppressing most warnings for files which are not
relevant for the current project.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4698,6 +4698,17 @@ Relative paths are relative to the file being checked."
   :safe #'flycheck-string-list-p
   :package-version '(flycheck . "0.14"))
 
+(flycheck-def-option-var flycheck-clang-system-include-path nil c/c++-clang
+  "A list of system include directories for Clang.
+
+The value of this variable is a list of strings, where each
+string is a directory to add to the include path of Clang.
+For files in this directory, most warnings will be suppressed.
+Relative paths are relative to the file being checked."
+  :type '(repeat (directory :tag "System include directory"))
+  :safe #'flycheck-string-list-p
+  :package-version '(flycheck . "0.22"))
+
 (flycheck-def-option-var flycheck-clang-includes nil c/c++-clang
   "A list of additional include files for Clang.
 
@@ -4811,6 +4822,7 @@ See URL `http://clang.llvm.org/'."
             (option-flag "-fno-rtti" flycheck-clang-no-rtti)
             (option-flag "-fblocks" flycheck-clang-blocks)
             (option-list "-include" flycheck-clang-includes)
+            (option-list "-isystem" flycheck-clang-system-include-path)
             (option-list "-W" flycheck-clang-warnings concat)
             (option-list "-D" flycheck-clang-definitions concat)
             (option-list "-I" flycheck-clang-include-path)
@@ -4861,6 +4873,17 @@ Relative paths are relative to the file being checked."
   :type '(repeat (directory :tag "Include directory"))
   :safe #'flycheck-string-list-p
   :package-version '(flycheck . "0.20"))
+
+(flycheck-def-option-var flycheck-gcc-system-include-path nil c/c++-gcc
+  "A list of system include directories for GCC.
+
+The value of this variable is a list of strings, where each
+string is a directory to add to the include path of gcc.
+For files in this directory, most warnings will be suppressed.
+Relative paths are relative to the file being checked."
+  :type '(repeat (directory :tag "System include directory"))
+  :safe #'flycheck-string-list-p
+  :package-version '(flycheck . "0.22"))
 
 (flycheck-def-option-var flycheck-gcc-includes nil c/c++-gcc
   "A list of additional include files for GCC.
@@ -4941,6 +4964,7 @@ Requires GCC 4.8 or newer.  See URL `https://gcc.gnu.org/'."
             (option-flag "-fno-rtti" flycheck-gcc-no-rtti)
             (option-flag "-fopenmp" flycheck-gcc-openmp)
             (option-list "-include" flycheck-gcc-includes)
+            (option-list "-isystem" flycheck-gcc-system-include-path)
             (option-list "-W" flycheck-gcc-warnings concat)
             (option-list "-D" flycheck-gcc-definitions concat)
             (option-list "-I" flycheck-gcc-include-path)


lunaryorn:
@ackalker @Fanael To pick this discussion up again: I wouldn't like to introduce two different options for include paths, so my question is still whether we need this at all, and if so, whether it's not a better idea to use `-isystem` right away for the whole include path?

sbroberg:
Just my two cents: I very much want the ability to specify includes using both -I and -isystem.  My default set of warnings used on our project triggers all sorts of warnings in 3rd party headers that we have no control over, and some of those are template expansions, which show up as errors in the file itself.  I'm not sure why you believe it's burdensome to have an additional line like:

            (option-list "-isystem" flycheck-clang-include-system-path)

in flycheck-define-checker c/c++-clang.

lunaryorn:
@sbroberg Do me a favour and grep Flycheck for `flycheck-clang-include-path`.  You'll find that a new option involves more than a single line, and as long as I am to maintain these lines in my free time, I'll want to see a better reason than “I very much want…”.

But I'm sure that since you want this option so very much, you'll find the time to prepare a concrete example which demonstrates why you need both options, even though Flycheck does not actually show errors from included files anywhere.

That said, you *do* already have the ability to specify both kinds of includes, just like any other flag you would want to pass to Clang.  I'm sure that it's not too burdensome for you to find out how.

sbroberg:
I apologize for sounding confrontational - that was not my intention at all.  "burdensome" is common parlance in the groups I work with.

The fault is mine, in fact, for raising this issue.  I had worked on configuring flycheck about a year ago, and changes you've made since then (including suppressing header warnings the new flycheck-clang-args variable) have addressed the issue.  The -isystem flag in clang is one that is not widely supported (including Xcode, incredibly), so my first position is to advocate support of it.

Unfortunately (for me), my knowledge of elisp ends at "setq", so it was quite burdensome for me to get flycheck working in my old configuration.  I broke the cardinal rule of manually editing flycheck.el and adding the feature myself, which, of course, was overwritten when I updated, my hacked-in variable ceased to function.  and I stupidly left no notes for myself about how I got it working.

tl;dr: I rescind my request.

tomjakubowski:
> But since you want this option so very much, I'm sure you'll find the time to prepare a concrete example which demonstrates why you need both options, even though Flycheck does not actually show errors from included files anywhere. That might convince me to add the option.

This isn't really true, or maybe I misunderstand you:

screen shot 2015-06-06 at 1 20 48 pm

(Sorry for the redaction, but better to be safe than sorry!)

Each of those comes from a different compiler warning. The project's build passes those warningful paths as `-isystem` to the compiler (clang++) to avoid just this thing.

tomjakubowski:
I would be happy to take a crack at patching this in.

tomjakubowski:
FWIW, the spew of compiler warnings from headers in that system directory is 2300 lines long (as measured in `c/c++-clang`). I think that just handling that volume of output is also significantly slowing down Emacs when I have flycheck enabled in these projects.

tomjakubowski:
Those warnings are also almost always unfixable, because they come from some third party dependency (several boost headers don't pass `-Wall -Wextra`, for example). The `-isystem` flag is pretty important to keep C and C++ developers sane. :-)

tomjakubowski:
> To pick this discussion up again: I wouldn't like to introduce two different options for include paths, so my question is still whether we need this at all, and if so, whether it's not a better idea to use -isystem right away for the whole include path?

That's a good question. From the GCC docs:

```
-Idir

    Add the directory dir to the head of the list of directories to be
    searched for header files. This can be used to override a system
    header file, substituting your own version, since these directories
    are searched before the system header file directories. However, you
    should not use this option to add directories that contain
    vendor-supplied system header files (use -isystem for that). If you
    use more than one -I option, the directories are scanned in
    left-to-right order; the standard system directories come after.
```

lunaryorn:
@tomjakubowski I see.  Although I'd say that this is mostly a bug in how Flycheck aggregates warnings from headers.  Ideally—that's what I intended—there'd just be a *single* message for each affected `#include`, saying that there were warnings/errors in the header file.

In this case, though, it wouldn't help much, since aggregating well over 2k warnings will just take way too long.  I agree that we should suppress these warnings initially, but this is already possible: You can just add `-isystem` arguments to `flycheck-gcc-args`.

The only question I'd still like to see answered is whether it makes sense to have a *separate* options for system includes.  I don't think so, and I'd really like to avoid redundant options.

Thank you for quoting the GCC docs, but I don't see how the quoted paragraph would answer this question.  I see that we should use `-isystem` for, well, vendor includes, but can we use `-isystem` for *all* includes?

tomjakubowski:
> In this case, though, it wouldn't help much, since aggregating well over 2k warnings will just take way too long. I agree that we should suppress these warnings initially, but this is already possible: You can just add -isystem arguments to flycheck-gcc-args.

Ah! I didn't know about that variable (I ought to RTM, or RTS at least :-)).

>  I see that we should use -isystem for, well, vendor includes, but can we use -isystem for all includes?

flycheck could use it for all include directories, but GCC (and clang, I assume) searches for header files in this order: directories passed with `-I` flags, directories passed with `-isystem` flags, and then directories in the standard system directories. This might have surprising results in projects that have both `-I` and `-isystem` in their build system configuration and depend on that ordering.

lunaryorn:
@tomjakubowski I don't think that the order would be a problem.  You need explicitly map compiler flags to this option after all, which let's you choose whatever order you need.  And if you want to use the build system flags verbatim, you'd just add them to `flycheck-gcc-args` and skip the include directories option.

lunaryorn:
@Fanael @ackalker @tomjakubowski Is there any new input on this issue?  If not, I'll change Clang and GCC to use `-isystem` instead of `-I` for all include paths…

tomjakubowski:
> @Fanael @ackalker @tomjakubowski Is there any new input on this issue? If not, I'll change Clang and GCC to use -isystem instead of -I for all include paths…

I'm not sure this is the right thing to do, especially in the case that the header file contains static inline functions or templates. I made a project like this:
 
```
.
├── bar.cc
└── incl
    └── warn.hh

1 directory, 2 files
```

```
/* foo.c */
#include "warn.hh"

int main()
{
  warn<int>();
}
```

```
/* incl/warn.h */
#pragma once

#include <iostream>

template <typename T>
void warn()
{
  T x;
  std::cout << x << std::endl; // for some types T, warns about uninitialized variable
```

If I build bar.cc passing the `incl` directory as `-I`, I see a warning for that bad code in the header file:
```
% g++ -o /dev/null  -Wall -I incl bar.cc
In file included from bar.cc:1:0:
incl/warn.hh: In function ‘void warn() [with T = int]’:
incl/warn.hh:9:13: warning: ‘x’ is used uninitialized in this function [-Wuninitialized]
   std::cout << x << std::endl;
             ^
```

If I build with `-isystem`, I don't:

```
% gcc -o /dev/null  -Wall -isystem incl foo.c
```

(this actually surprises me a little, I'd expect that since the template is instantiated in a "local" translation unit that the warnings would appear in spite of the `-isystem`. But that's something to take up with GCC maintainers not here)

It's important to be warned of these kinds of mistakes.  But for many installed libraries (like the stdlib, or boost, or the library in question), being bombarded with (often frivolous) warnings that are in headers largely outside of your control is a massive nuisance. It's important to have control over which headers you do and don't care about seeing warnings from.

lunaryorn:
Since Flycheck now lets you pass arbitrary arguments to clang and GCC I think this issue has become redundant.  Thank you for your feedback.

PR #543, opened 2014-12-13
Add -pedantic, -pendantic-errors to c/c++ checkers

stephenjlewis:
The -pedantic compiler flags throws warning on the use of any language
extensions, and is used when strict ISO-compliance is desired.

-pedantic-errors is equivilient, but throws errors rather than warnings.

### DIFF
--- a/doc/guide/languages.rst
+++ b/doc/guide/languages.rst
@@ -76,6 +76,12 @@ C/C++
    .. option:: flycheck-clang-language-standard
       :auto:
 
+   .. option:: flycheck-clang-pedantic
+      :auto:
+
+   .. option:: flycheck-clang-pedantic-errors
+      :auto:
+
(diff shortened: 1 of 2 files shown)


stephenjlewis:
I noticed you've removed this from the 0.23 milestone; is there any issue in particular? The most recent commits just add docstrings.

lunaryorn:
@stephenjlewis No, no issues.  I just removed all milestones entirely: I was just experimenting whether I could use them to organize my work on Flycheck, but I noticed that they don't really work for me.  I make releases in a mostly arbitrary way anyway, and I can't really plan how much time I can spend on Flycheck, so they simply don't make sense to me.

But this PR will make it into the next Flycheck release.

nickbroon:
:+1:

PR #556, opened 2014-12-30
Use source-inplace for chef-foodcritic checker

petere:
Otherwise foodcritic won't find related files such as templates and will
report errors.


lunaryorn:
@petere Well, temporary files inside the source tree can potentially interfere with all sorts of other tooling, e.g. version control, automatic reloaders of web frameworks, etc.  Keeping files in the system temporary directory just avoids all this hassle.

But if there's no other way, I'll merge your PR.

PR #566, opened 2015-01-14
Added symbolic names to Pylint output

kstrauser:
Pylint can use symbolic names for errors. I tried changing:

~~~
(flycheck-define-checker python-pylint
  :command ("pylint" "-r" "n"
            "--msg-template" "{path}:{line}:{column}:{C}:{symbol}:{msg}"
~~~

but that messed with the formatting of the `*Flycheck errors*` buffer and had me in over my head. I changed that to:

~~~
(flycheck-define-checker python-pylint
  :command ("pylint" "-r" "n"
            "--msg-template" "{path}:{line}:{column}:{C}:{msg_id}:({symbol}) {msg}"
~~~

and now get errors like:

<pre>
    1   1 info     C0111  (missing-docstring) Missing module docstring... (python-pylint)
    2   1 error    E0012  (bad-option-value) Bad option value 'C0110'... (python-pylint)
    9   1 info     C0111  (missing-docstring) Missing class docstring... (python-pylint)
    9   1 warning  R0903  (too-few-public-methods) Too few public methods (1/2)... (python-pylint)
</pre>

This is very useful because now I know to write `# pylint: disable=missing-docstring` instead of `# pylint: disable=C0111`, which will be a big help a month from now when I'm wondering exactly what tests I've disabled.

It's simple enough for me to modify my own init.el to add the custom definition, but I think this is generally useful (and backward compatible!) enough that others might like it, too.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6550,7 +6550,7 @@ This syntax checker requires Pylint 1.0 or newer.
 See URL `http://www.pylint.org/'."
   ;; -r n disables the scoring report
   :command ("pylint" "-r" "n"
-            "--msg-template" "{path}:{line}:{column}:{C}:{msg_id}:{msg}"
+            "--msg-template" "{path}:{line}:{column}:{C}:{msg_id}:({symbol}) {msg}"
             (config-file "--rcfile" flycheck-pylintrc)
             ;; Need `source-inplace' for relative imports (e.g. `from .foo
             ;; import bar'), see https://github.com/flycheck/flycheck/issues/280
--- a/flycheck.el
+++ b/flycheck.el
@@ -6550,7 +6550,7 @@ This syntax checker requires Pylint 1.0 or newer.
 See URL `http://www.pylint.org/'."
   ;; -r n disables the scoring report
   :command ("pylint" "-r" "n"
-            "--msg-template" "{path}:{line}:{column}:{C}:{msg_id}:({symbol}) {msg}"
+            "--msg-template" "{path}:{line}:{column}:{C}:{msg_id}:{msg} ({symbol})"
             (config-file "--rcfile" flycheck-pylintrc)
             ;; Need `source-inplace' for relative imports (e.g. `from .foo
             ;; import bar'), see https://github.com/flycheck/flycheck/issues/280
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -4590,6 +4590,8 @@ Why not:
      "checkers/python/test.py" 'python-mode
      '(1 1 info "Missing module docstring (missing-docstring)" :id "C0111" :checker python-pylint)
      '(4 1 error "Unable to import 'spam' (import-error)" :id "F0401" :checker python-pylint)
+     '(5 1 error "No name 'antigravit' in module 'python' (no-name-in-module)" :id "E0611"
+         :checker python-pylint)
      '(5 1 warning "Unused import antigravit (unused-import)" :id "W0611"
          :checker python-pylint)
      '(7 1 info "Missing class docstring (missing-docstring)" :id "C0111" :checker python-pylint)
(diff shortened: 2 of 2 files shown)


lunaryorn:
@kstrauser Thanks, I didn't know that.  Still, I'd prefer if the name appeared *after* the error message, because I think that the message is more important in most cases.  Would you mind to update the PR accordingly?  And if you have some spare time, please also take a look at the test cases.

kstrauser:
@lunaryorn I've switched {msg} and {symbol} as requested. After looking at it for a few minutes, I've decided that it really does look better that way. Good call!

I *think* I've updated the test cases correctly, but this is my first time with Cask and I think I'm doing something wrong. The output for error backtraces is truncated narrowly like:

<pre>
Checker definition probably flawed.
Test flycheck-define-checker/c/c++-gcc/error backtrace:
  flycheck-ert-should-errors((8 17 error "aggregate ‘test<false> t’ ha
  apply(flycheck-ert-should-errors ((8 17 error "aggregate ‘test<false
  flycheck-ert-should-syntax-check("checkers/c_c++-error.cpp" c++-mode
</pre>

so I can't actually see what the errors are. I also get a lot of non-Pylint-related errors on a clean checkout and I'm having a hard time separating the wheat from the chaff. I don't want to be a total newb, but I'd appreciate any tips you might have.

lunaryorn:
@kstrauser  Well, you don't need to run the entire test suite, but only the tests for pylint:

```
$ cask install   # To install dependencies
$ test/run.el '(checker python-pylint)' 
```

This will only run tests affected by changes to the pylint checker, which should be considerably less (about ten or so) than the whole test suite.

kstrauser:
Well, today I learned something! Thanks. I updated the tests with the symbolic names so that they pass as expected, but noticed some odd things:

* I had to remove two `No name 'antigravit' in module 'python'` tests because Pylint (version 1.4.0) on my system doesn't emit those lines. The those tests fail on master for me. I'm waiting to see what Travis says about that, because apparently the versions in master are passing there.
* I consider this to be a bug in Pylint, but the `{msg}` token in the bad-whitespace test (and *only* that test) outputs the bad line of code with a caret to mark the error, so `(bad-whitespace)` is a couple of lines below the rest of the message line. Therefore the unit test doesn't find the symbolic name after the message, so I had to leave that just one line as-is. For example:
~~~
test.py:12:0:C:C0326:No space allowed around keyword argument assignment
    def with_ham(self, blubb = None):
                             ^ (bad-whitespace)
~~~

lunaryorn:
@kstrauser Ad 1) I can't reproduce that:

```console
$ pylint --version
pylint --version
No config file found, using default configuration
pylint 1.4.0,
astroid 1.3.2, common 0.63.2
Python 2.7.9 (default, Dec 15 2014, 10:01:34)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.56)]
$ test/run.el '(checker python-pylint)'
test/run.el '(checker python-pylint)'
Running tests on Emacs 25.0.50.1, built at 2015-01-04
Loading /Users/swiesner/Developer/Projects/flycheck/flycheck.el (source)...
Loading /Users/swiesner/Developer/Projects/flycheck/flycheck-ert.el (source)...
Loading /Users/swiesner/Developer/Projects/flycheck/test/flycheck-test.el (source)...
Finding all versions of R on your system...
Sorry, no version of R could be found on your system.
Running 5 tests (2015-01-15 21:05:51+0100)
   passed  1/5  flycheck-define-checker/python-pylint/default
   passed  2/5  flycheck-define-checker/python-pylint/disabled-warnings
   passed  3/5  flycheck-define-checker/python-pylint/syntax-error
   passed  4/5  flycheck-select-checker/selecting-runs-a-syntax-check
   passed  5/5  flycheck-select-checker/unselecting-a-checker-goes-back-to-automatic-selection

Ran 5 tests, 5 results as expected (2015-01-15 21:05:54+0100)
```

As you can see, the tests in current `master` pass, even with the `antigravit` line.

Ad 2) I see.  We should probably change the output format to put the error name to a different place, and use an error filter to re-format the message.

It's probably even better, though, to make the symbolic name part of the error ID.

kstrauser:
That's certainly an option, too. I tried using it *as* the ID at first, but it made the error buffer look goofy and I'm already in over my experience with elisp.

lunaryorn:
Superseded by #714

PR #570, opened 2015-01-17
Fix #569, don't enable flycheck in special modes.

### LINKED ISSUE #569: Global flycheck mode should not enable itself in *grep* buffers
In a grep buffer the mode is `grep-mode`, which is derived from parent mode `compilation-mode`, which is in turn *not* derived from `special-mode`, but has `'special` mode-class.

So when `flycheck-may-enable-mode` runs, it uses `derived-mode-p` and thinks it's okay to return `t` because no parent of `grep-mode` is `special-mode`. 

    flycheck-version "0.23snapshot (package: 20150115.1159)"
    GNU Emacs 24.4.1 (x86_64-apple-darwin13.4.0, NS apple-appkit-1265.21)

In my grep buffers this led to some weirdness where I couldn't follow links to files, but I'm not sure how that happened or how flycheck interferes with it. I don't think it should be enabled at all.

I think this line is the problem: https://github.com/flycheck/flycheck/blob/master/flycheck.el#L2469

ikirill:
Check the mode-class property as well as derived-mode-p.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2460,6 +2460,7 @@ Return t if so, or nil otherwise."
 Flycheck mode is not enabled for
 
 - buffers in a `special-mode' or `comint-mode'
+- buffers in a mode with `mode-class` set to `special`,
 - ephemeral buffers (see `flycheck-ephemeral-buffer-p'),
 - encrypted buffers (see `flycheck-encrypted-buffer-p'),
 - remote files (see `file-remote-p'),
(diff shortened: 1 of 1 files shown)


lunaryorn:
Duplicate of #569

Sarcasm:
@lunaryorn this is not a duplicate but the pull request that fixes #569 as said in the title.

lunaryorn:
@Sarcasm Uhm, yes, I am quite aware, thank you.

PR #575, opened 2015-01-22
Fix flags passed to `rustc`.

utkarshkukreti:
The flag `--no-trans` was replaced by `-Z no-trans` recently: https://github.com/rust-lang/rust/commit/953f294

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6932,10 +6932,10 @@ Relative paths are relative to the file being checked."
 (flycheck-define-checker rust
   "A Rust syntax checker using Rust compiler.
 
-This syntax checker needs Rust 0.10 or newer.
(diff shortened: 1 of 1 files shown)


lunaryorn:
@utkarshkukreti Is that change part of any Rust release already?  If so, please update the docstring accordingly to state the new requirement, otherwise I'd like to wait for the next rust release before merging this.

utkarshkukreti:
Just checked that it is available in the latest release, [1.0.0 alpha](http://blog.rust-lang.org/2015/01/09/Rust-1.0-alpha.html).
I have updated the comment and rebased.

lunaryorn:
@utkarshkukreti Thanks, merged.

Issue #531, opened 2014-11-28
Test error


lunaryorn:
Some test cases fail with a very strange `(file-error "Cannot open load file" "no such file or directory" "cask-cli")`:

```
Test flycheck-define-checker/d-dmd/missing-import backtrace:
  c-get-lang-constant(c-mode-menu (cc-langs) d-mode)
  byte-code("\301\30\303\304\305\306\307\310#B$\207" [d-mode-map eas
  d-mode()
  flycheck-ert-should-syntax-check("checkers/d/src/dmd/warning.d" d-mo
  (closure (python-indent-guess-indent-offset js3-mode-show-parse-erro
  ert--run-test-internal([cl-struct-ert--test-execution-info [cl-struc
  ert-run-test([cl-struct-ert-test flycheck-define-checker/d-dmd/missi
  ert-run-or-rerun-test([cl-struct-ert--stats t [[cl-struct-ert-test f
  ert-run-tests(t #[385 "\306\307\"\203D\211\211G\310U\203\211@\20
  ert-run-tests-batch(t)
  ert-run-tests-batch-and-exit(t)
  (let ((selector (car-safe (prog1 argv (setq argv (cdr argv)))))) (if
  flycheck-run-tests-batch-and-exit()
  (let ((debug-on-error t)) (flycheck-run-tests-batch-and-exit))
  flycheck-run-tests-main()
  (progn (flycheck-run-tests-main))
  (if (and noninteractive (flycheck-runs-this-script-p)) (progn (flych
  eval-buffer(#<buffer  *load*> nil "/flycheck/test/run.el" nil t)
  load-with-code-conversion("/flycheck/test/run.el" "/flycheck/test/ru
  load("/flycheck/test/run.el" nil t t)
  command-line-1(("-scriptload" "test/run.el" "t"))
  command-line()
  normal-top-level()
Test flycheck-define-checker/d-dmd/missing-import condition:
    (file-error "Cannot open load file" "no such file or directory" "cask-cli")
   FAILED  147/545  flycheck-define-checker/d-dmd/missing-import
```

@rejeep Any idea?  Seems to be similar to https://github.com/cask/cask/issues/256…

PR #583, opened 2015-02-10
Add flycheck-dmd-string-import-path variable

atilaneves:
I tried adding a test to `flycheck-test.el`, however:

1. The existing D tests are disabled due to #531 
2. `make test` failed for me, so I wouldn't be able to verify they did their job
3. I figured the vagrant box included would be able to run the tests but that wasn't the case either

So I tested manually.

### DIFF
--- a/doc/guide/languages.rst
+++ b/doc/guide/languages.rst
@@ -205,6 +205,9 @@ D
    .. option:: flycheck-dmd-include-path
       :auto:
 
+   .. option:: flycheck-dmd-string-import-path
+      :auto:
+
 .. seealso::
 
    The flycheck-d-unittest_ extension adds a syntax checker that runs D unit
--- a/flycheck.el
+++ b/flycheck.el
@@ -5538,6 +5538,17 @@ Relative paths are relative to the file being checked."
   :safe #'flycheck-string-list-p
   :package-version '(flycheck . "0.18"))
 
+(flycheck-def-option-var flycheck-dmd-string-import-path nil d-dmd
+  "A list of string import directories for dmd.
+
+The value of this variable is a list of strings, where each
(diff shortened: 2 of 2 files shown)


lunaryorn:
@atilaneves What is the “import path”?

atilaneves:
I was the one who added the `flycheck-dmd-include-path` variable as well. In D this is analogous to `-I` flags for clang or gcc and lets the compiler know where to search for dependencies. In addition to that, D also has "string imports". They work like `#include` in C and do textual insertion. Due to security issues, string imports are only allowed from white-listed directories. So it's kind of a 2nd type of `-I` flag and in dmd is specified with `-J`. Without this new variable, it's impossible for flycheck to know how to call dmd for projects that use string imports. With it, the errors go away (or don't if still doesn't compile).

lunaryorn:
@atilaneves Are these imports used a lot?

atilaneves:
Often enough. The average D project, no. But anybody using them will not get proper on-the-fly highlighting. An alternative is to change `dmd-include-path` to have explicit `-I` or `-J` flags in it, but that would be a breaking change.

atilaneves:
Bump. I'm still using my flycheck fork since this hasn't been merged yet.

lunaryorn:
@atilaneves I don't think that I'll merge this.  It's too exotic a use case, imho.

I'd much rather add a catch-all `flycheck-dmd-flags` (similar to `flycheck-clang-flags`) option, which allows for arbitrary settings.  What do you think?

atilaneves:
Sure, that works too.

lunaryorn:
@atilaneves Closing this PR then.  Would you mind to open a new pull request for a catch-all args variable as discussed, or make me a new issue so that I don't forget about it?

PR #599, opened 2015-03-03
Make `flycheck-{gcc,clang}-args' safe buffer local variables

YorkZ:
Variables `flycheck-{gcc,clang}-args` need to be safe buffer local variables
because different project may needs different values.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5053,6 +5053,7 @@ See URL `http://www.methods.co.nz/asciidoc'."
   :modes adoc-mode)
 
 (flycheck-def-args-var flycheck-clang-args c/c++-clang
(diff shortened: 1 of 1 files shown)


lunaryorn:
@YorkZ These variables are *deliberately* unsafe, because wrong or incompatible settings could break syntax checking badly.  You really need to check local settings for these variables before applying them.

However, they aren't risky, so you can just press `!` in the prompt, and Emacs will remember the values as safe for any future session.  Or even customize `safe-local-variable-values` directly.  See [Safe File Variables](https://www.gnu.org/software/emacs/manual/html_node/emacs/Safe-File-Variables.html#Safe-File-Variables) for details.

PR #605, opened 2015-03-18
add support for handlebars checking in web-mode

egh:
Allow web-mode users to get handlebars syntax checking.

Thank you for the wonderful package!

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6082,7 +6082,16 @@ See URL `http://handlebarsjs.com/'."
           "Error: Parse error on line " line ":" (optional "\r") "\n"
           (zero-or-more not-newline) "\n" (zero-or-more not-newline) "\n"
           (message) line-end))
-  :modes (handlebars-mode handlebars-sgml-mode))
+  :modes (handlebars-mode handlebars-sgml-mode web-mode)
+  :predicate (lambda ()
+               (if (eq major-mode 'web-mode)
+                   (and (boundp 'web-mode-engine-file-regexps)
+                        ;; check if this is a handlebars file since
+                        ;; web-mode does not store the non-canonical
+                        ;; engine name
+                        (string-match-p (cdr (assoc-string "handlebars" web-mode-engine-file-regexps))
+                                        (buffer-file-name)))
+                 t)))
 
 (defconst flycheck-haskell-module-re
   (rx line-start (zero-or-more (or "\n" (any space)))
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -4255,7 +4255,7 @@ The term \"1\" has type \"nat\" while it is expected to have type
 
 (flycheck-ert-def-checker-test handlebars handlebars nil
   (flycheck-ert-should-syntax-check
-   "checkers/handlebars-error.hbs" 'handlebars-mode
+   "checkers/handlebars-error.hbs" '(handlebars-mode web-mode)
    '(2 nil error "Expecting 'ID', 'STRING', 'NUMBER', 'BOOLEAN', 'DATA', got 'INVALID'"
        :checker handlebars)))
 


lunaryorn:
@egh According to the Web Mode website, `ctemplate` is for more than just handlebars, namely also for mustache, ember, hapax, ngtemplate, meteor, and blaze.

I don't know any of these save ngtemplate, so I'm unsure whether its acceptable to check them with handlebars.  ngtemplate for instance has neither `#each` nor `#if`.

In the past I removed Web Mode from other checkers because it's a mess, and I don't want to make the same mistake twice.

And in any case, please merge your syntax checker with the existing handlebars checker.  There's no reason to keep the same checker twice.

egh:
Thank you for your comment & your great test suite!

It's kind of ugly now, because web-mode has some strange detection features. But it seems to work, passes tests, and compiles cleanly.

One thing that bothers me is that the code path where `:predicate` is false is not tested. Is there a test helper to ensure that a given flycheck checker is *not* enabled on a file? Thanks.

flycheck.el
```diff
@@ -6079,7 +6084,17 @@ See URL `http://handlebarsjs.com/'."
...
+  :predicate (lambda ()
+               (or (and (eq major-mode 'web-mode)
+                        (string-equal web-mode-engine "ctemplate")
+                        (string-equal "handlebars"
+                                      (catch 'done
+                                        (dolist (elt web-mode-engine-file-regexps)
+                                          (when (string-match-p (cdr elt) (buffer-file-name))
+                                            (throw 'done (car elt)))))))
+                   (eq major-mode 'handlebars-mode)
+                   (eq major-mode 'handlebars-sgml-mode))))
```
lunaryorn: Just return `t` here, since these checks are essentially already covered by `:modes`.

flycheck.el
```diff
@@ -6079,7 +6084,17 @@ See URL `http://handlebarsjs.com/'."
...
-  :modes (handlebars-mode handlebars-sgml-mode))
+  :modes (handlebars-mode handlebars-sgml-mode web-mode)
+  :predicate (lambda ()
+               (or (and (eq major-mode 'web-mode)
+                        (string-equal web-mode-engine "ctemplate")
+                        (string-equal "handlebars"
+                                      (catch 'done
+                                        (dolist (elt web-mode-engine-file-regexps)
+                                          (when (string-match-p (cdr elt) (buffer-file-name))
+                                            (throw 'done (car elt)))))))
```
lunaryorn: What does this `dolist` do, and what purpose does it serve?  It looks to me as if you have duplicated some logic from Web Mode itself…

flycheck.el
```diff
@@ -45,6 +45,11 @@
   (require 'sh-script)      ; `sh-shell' for sh checker predicates
   (require 'jka-compr)      ; For JKA workarounds in `flycheck-temp-file-system'
   (require 'pcase)          ; `pcase-dolist' (`pcase' itself is autoloaded)
+  (if (featurep 'web-mode)
+      (require 'web-mode)
+    (defvar web-mode-engine nil)
+    (defvar web-mode-engine-file-regexps '()
+      "Engine file extensions."))
```
lunaryorn: This is a no-go.  Please don't introduce new `require`s or variable definitions from 3rd party packages.  Instead, use `boundp` at appropriate places to check whether the variables are defined and bound.

flycheck.el
```diff
@@ -6079,7 +6084,17 @@ See URL `http://handlebarsjs.com/'."
           "Error: Parse error on line " line ":" (optional "\r") "\n"
           (zero-or-more not-newline) "\n" (zero-or-more not-newline) "\n"
           (message) line-end))
-  :modes (handlebars-mode handlebars-sgml-mode))
+  :modes (handlebars-mode handlebars-sgml-mode web-mode)
+  :predicate (lambda ()
+               (or (and (eq major-mode 'web-mode)
```
lunaryorn: Please turn the whole conditional into an `(if (eq major-mode 'web-mode) … t)`, since the checks below are really only relevant if the major mode is Web Mode.  Currently the entire conditional is a little hard to follow.

flycheck.el
egh: Thanks!

flycheck.el
egh: Hm, I didn't know `boundp` could prevent against compile time warnings. Will fix.

flycheck.el
egh: Because web-mode does not store the engine name `"handlebars"`, only the "canonical" name `"ctemplate"`, this is checking agains the buffer file name. It works, but I have a much better solution in mind, which I'll push.

egh:
let me know if you'd prefer a rebase single commit

flycheck.el
```diff
@@ -6079,7 +6079,18 @@ See URL `http://handlebarsjs.com/'."
           "Error: Parse error on line " line ":" (optional "\r") "\n"
           (zero-or-more not-newline) "\n" (zero-or-more not-newline) "\n"
           (message) line-end))
-  :modes (handlebars-mode handlebars-sgml-mode))
+  :modes (handlebars-mode handlebars-sgml-mode web-mode)
+  :predicate (lambda ()
+               (if (eq major-mode 'web-mode)
+                   (and (boundp 'web-mode-engine)
+                        (boundp 'web-mode-engine-file-regexps)
+                        (string-equal web-mode-engine "ctemplate")
```
lunaryorn: Do we still need this check?  Isn't this covered by the next check anyway?

lunaryorn:
@egh Yes, I'd prefer a single commit.

flycheck.el
egh: Both `boundp` checks are necessary to prevent compilation warning about free variables.

egh:
Single commit pushed. Thanks for all your comments on this!

flycheck.el
lunaryorn: I meant the `string-equal` test, actually.

egh: Ah, yes, thanks. You're right, that is unnecessary.

egh: Won't need that `boundp` either.

flycheck.el
```diff
@@ -6082,7 +6082,16 @@ See URL `http://handlebarsjs.com/'."
...
           (message) line-end))
-  :modes (handlebars-mode handlebars-sgml-mode))
+  :modes (handlebars-mode handlebars-sgml-mode web-mode)
+  :predicate (lambda ()
+               (if (eq major-mode 'web-mode)
+                   (and (boundp 'web-mode-engine-file-regexps)
+                        ;; check if this is a handlebars file since
+                        ;; web-mode does not store the non-canonical
+                        ;; engine name
+                        (string-match-p (cdr (assoc-string "handlebars" web-mode-engine-file-regexps))
```
lunaryorn: @egh By the way, is there a reason why you chose `assoc-string` here instead of the standard `assoc`?

egh:
I just read the docs for `assoc-string` and thought it was considered more appropriate for alists with strings. I have no preference one way or the other.

egh:
Thanks for all your help @lunaryorn.

PR #606, opened 2015-03-18
Don't overwrite minibuffer if cursor is not moved

yudai:
This patch fixes a problem that flycheck always overwrites messages shown in the minibuffer immediately after running commands and users cannot get enough time to read the messages.
For example, when I point at an error by the cursor and find that the error says something like "not enough arguments in call to funcX", I would run the `go-def` command, which is provided by the go-mode, to check the definition of the "funcX" from `C-c C-d`. However, the `go-def` command shows the definition of the function in the minibuffer, and it immediately disappears as flycheck puts the same error message again in the minibuffer.
To solve this problem, this patch introduces a new local variable that stores the position of the cursor and  makes `flycheck-display-error-at-point-soon` invoke the timer only when the position is changed since the last time it was called.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3747,7 +3747,9 @@ non-nil."
   ;; errors
   (with-demoted-errors "Flycheck error display error: %s"
     (flycheck-cancel-error-display-error-at-point-timer)
-    (when flycheck-mode
+    (when (and flycheck-mode
+               (not this-command)
+               (flycheck--message-command-p last-command))
       (-when-let (errors (flycheck-overlay-errors-at (point)))
         (flycheck-display-errors errors)))))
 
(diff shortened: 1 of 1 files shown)


lunaryorn:
@yudai I don't like this change.  It's another complication in what is already not the simplest thing to understand by reading the code.

Can't you just increase `flycheck-display-errors-delay`?

yudai:
Hi @lunaryorn, thank you for the comment. I've updated my code to improve its readability.

I'm afraid I don't think increasing`flycheck-display-errors-delay` is a good idea. If I increase it to the value long enough to read messages  (about 10 secs?), I need to wait for  error messages to be shown for that seconds when I point other errors.

lunaryorn:
@yudai The echo area is not the right place for messages that take 10 seconds to comprehend.  It's for *short* and *easy* messages that are relevant only for the current context and can vanish quickly.

If you need 10 seconds to read whatever `C-c C-d` shows, I'd really recommend you to customize it so that it shows its content in a separate buffer, for Flycheck won't be the only thing that will conflict with `C-c C-d`.  There are many other types of messages that can override the echo area any time (e.g. `eldoc`, `message`s from signal handlers, etc.).

yudai:
@lunaryorn Thank you for the reply.

Indeed, even with the default delay, it's really hard to read any messages posted to the echo area because of flycheck's overwriting.

Simple conflicts of messages from different tools in the echo area, like flycheck overwrites messages from `eldoc`, make no problem indeed, because we can set their priorities by setting their delays. However, flycheck makes things more complicated because it rewrites the same message agin in the echo area even after users intentionally run commands that show messages.

In fact, `eldoc` has a mechanism to avoid this problem. [When the `post-command-hook` is called by non-editing commands, `eldoc` simply ignores the call](https://github.com/emacs-mirror/emacs/blob/0537943561a37b54467bec19d1b8afbeba8e1e58/lisp/emacs-lisp/eldoc.el#L358-L362). This works really well to share the echo area with other tools.

It seems the implementation of `eldoc` is more reasonable than my proposal, so I tried porting it to flycheck and updated the branch. Placement of functions and variables is provisional.

lunaryorn:
@yudai Uh, that's even more “hacky” code than the first.

I understand that it improves the user experience, but I'm not so sure whether *I* want to maintain that.  It's prone to subtle bugs.

For now, I'd prefer if you just used your own `flycheck-display-errors-function`, which could than handle all this logic and only show an error message under certain conditions.

yudai:
@lunaryorn I got it. I indeed agree with you that the code is hacky, even if it works well (Maybe Emacs itself should have had a native function to return whether the buffer was changed by the `last-command`).
I'm going to close this PR for now. Thanks again for your time and comments.

PR #609, opened 2015-03-21
Add luacheck syntax checker

ptrv:
Adding `luackeck` to the test VM is very simple:

After lua52 is installed you need following steps
1. Download and unpack latest luacheck from https://github.com/mpeterv/luacheck/archive/0.10.0.tar.gz
2. Run `sudo lua install.lua /usr/local` inside luacheck source folder

Now the `luacheck` executable is in `/usr/local/bin` which is probably in `PATH`

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -208,6 +208,7 @@ attention to case differences."
     javascript-gjslint
     json-jsonlint
     less
+    luacheck
     lua
     perl
     perl-perlcritic
@@ -6355,6 +6356,28 @@ See URL `http://lesscss.org'."
           ", column " column ":"
           line-end))
   :modes less-css-mode)
+
+(flycheck-def-config-file-var flycheck-luacheckrc luacheck ".luacheckrc"
+  :safe #'stringp
+  :package-version '(flycheck . "0.23"))
+
+(flycheck-define-checker luacheck
+  "A Lua syntax checker using luacheck.
+
+See URL `https://github.com/mpeterv/luacheck'."
+  :command ("luacheck"
+            (config-file "--config" flycheck-luacheckrc)
+            "--formatter" "plain"
+            "--codes"                   ; Show warning codes
+            "--no-color"
+            source)
+  :error-patterns
+  ((warning line-start (file-name)
+            ":" line ":" column ": (" (id (and "W" (one-or-more digit)))
+            ") " (message) line-end)
+   (error line-start (file-name)
+          ":" line ":" column ":" (message) line-end))
+  :modes lua-mode)
 
 (flycheck-define-checker lua
   "A Lua syntax checker using the Lua compiler.
--- a/playbooks/roles/language-lua/tasks/luacheck.yml
+++ b/playbooks/roles/language-lua/tasks/luacheck.yml
@@ -1,0 +1,10 @@
+- name: Download luacheck
+  get_url: url=https://github.com/mpeterv/luacheck/archive/{{luacheck_version}}.zip
+           dest=/usr/src/luacheck-{{luacheck_version}}.zip
+- name: Extract luacheck
+  unarchive: src=/usr/src/luacheck-{{luacheck_version}}.zip copy=no
+             dest=/opt
+- name: Install luacheck
+  command: /usr/bin/lua /opt/luacheck-{{luacheck_version}}/install.lua /usr/local
+  args:
+    chdir: /opt/luacheck-{{luacheck_version}}
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -4418,11 +4418,35 @@ Why not:
    "checkers/less-syntax-error.less" 'less-css-mode
    '(1 1 error "Unrecognised input" :checker less)))
 
-(flycheck-ert-def-checker-test lua lua nil
+(flycheck-ert-def-checker-test luacheck lua syntax-error
   (flycheck-ert-should-syntax-check
    "checkers/lua-syntax-error.lua" 'lua-mode
-   '(5 nil error "unfinished string near '\"oh no'"
-       :checker lua)))
+   '(5 7 error "unfinished string"
+       :checker luacheck)))
+
+(flycheck-ert-def-checker-test luacheck lua warnings
+  (flycheck-ert-should-syntax-check
+   "checkers/lua-warnings.lua" 'lua-mode
+   '(1 1 warning "setting non-standard global variable 'global_var'"
+       :id "W111" :checker luacheck)
+   '(3 16 warning "unused function 'test'"
+       :id "W211" :checker luacheck)
+   '(3 21 warning "unused argument 'arg'"
+       :id "W212" :checker luacheck)
+   '(4 11 warning "variable 'var2' is never set"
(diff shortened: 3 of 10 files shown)


ptrv:
Added also the ansible part. Extended the lua language role to install luacheck.

flycheck.el
```diff
@@ -6330,6 +6331,118 @@ See URL `http://lesscss.org'."
...
+                        ,(nth 5 flycheck-luacheck-std-globals-list))
+                 (const :tag "Globals of LuaJIT 2.0"
+                        ,(nth 6 flycheck-luacheck-std-globals-list))
+                 (const :tag "Intersection of globals of Lua 5.1, 5.2 and LuaJIT 2.0"
+                        ,(nth 7 flycheck-luacheck-std-globals-list))
+                 (const :tag "Union of globals of Lua 5.1, 5.2 and LuaJIT 2.0"
+                        ,(nth 8 flycheck-luacheck-std-globals-list))
+                 (const :tag "No standard globals"
+                        ,(nth 9 flycheck-luacheck-std-globals-list)))
+  :safe (lambda (value) (member value flycheck-luacheck-std-globals-list))
```
lunaryorn: This is way too detailed and too complex. Please make this option a simple string.

flycheck.el
```diff
@@ -6330,6 +6331,118 @@ See URL `http://lesscss.org'."
...
+            "--no-config"
+            (option "--std" flycheck-luacheck-std-globals)
+            (option-flag "--no-global" flycheck-luacheck-no-globals)
+            (option-flag "--no-unused" flycheck-luacheck-no-unused)
+            (option-flag "--no-redefined" flycheck-luacheck-no-redefined)
+            (option-flag "--no-unused-args" flycheck-luacheck-no-unused-args)
+            (option-flag "--no-unused-secondaries" flycheck-luacheck-no-unused-secondaries)
+            (option-flag "--no-unused-globals" flycheck-luacheck-no-unused-globals)
+            "--formatter" "plain"
+            "--codes"
```
lunaryorn: What does `--codes` do?

flycheck.el
```diff
@@ -6330,6 +6331,118 @@ See URL `http://lesscss.org'."
...
+                         (string :tag "Global name")))
+  :safe #'flycheck-string-list-p
+  :package-version '(flycheck . "0.23"))
+
+(flycheck-define-checker luacheck
+  "A Lua syntax checker using luacheck.
+
+See URL `https://github.com/mpeterv/luacheck'."
+  :command ("luacheck"
+            "--no-config"
```
lunaryorn: What does this option do?

test/flycheck-test.el
```diff
@@ -4423,11 +4423,42 @@ Why not:
...
    '(1 1 error "Unrecognised input" :checker less)))
 
-(flycheck-ert-def-checker-test lua lua nil
+(flycheck-ert-def-checker-test luacheck lua syntax-error
   (flycheck-ert-should-syntax-check
    "checkers/lua-syntax-error.lua" 'lua-mode
-   '(5 nil error "unfinished string near '\"oh no'"
-       :checker lua)))
+   '(5 7 error "unfinished string"
+       :checker luacheck)))
```
lunaryorn: Why did you remove the test for the `lua` checker?

playbooks/roles/language-lua/tasks/luacheck.yml
```diff
@@ -0,0 +1,10 @@
+- name: Download luacheck
+  get_url: url=https://github.com/mpeterv/luacheck/archive/{{luacheck_version}}.zip
+           dest=/usr/src/luacheck-{{luacheck_version}}.zip
+- name: Extract luacheck
+  unarchive: src=/usr/src/luacheck-{{luacheck_version}}.zip copy=no
+             dest=/opt
+- name: Install luacheck
+  command: /usr/bin/lua /opt/luacheck-{{luacheck_version}}/install.lua /usr/local
+  args:
+    chdir: /opt/luacheck-{{luacheck_version}}
```
lunaryorn: Just being curious: How long does it take to build?

flycheck.el
ptrv: Ok, I will simplify this

flycheck.el
ptrv: It adds the error code, e.g. `W212`. For errors there is no code, so this makes it possible to use the  `id` in the error format and distinct between errors and warnnings.

flycheck.el
ptrv: This option disables the option to load an external config file. I did not introduce the option of specifying an external config file on purpose because you could set the `luacheck` output format or remove support for `codes` which would make the syntax checker fail.

test/flycheck-test.el
ptrv: I haven't removed the test for the `lua` checker. It's just below the new `luacheck` tests

playbooks/roles/language-lua/tasks/luacheck.yml
ptrv: It's pretty fast because the `install.lua` script basically just copies the files to the right places without any building.

ptrv:
Updated the commit according to your requests

flycheck.el
lunaryorn: @ptrv Add a config file nonetheless.  Other config files (e.g. Pylint) allow for the same, but we still provide them.

flycheck.el
lunaryorn: @ptrv Would you mind to add a small comment that explains the use of `--codes` here?  This would make it easier for me to understand what's going on later.

playbooks/roles/language-lua/tasks/luacheck.yml
lunaryorn: Cool, thanks

test/flycheck-test.el
lunaryorn: @ptrv Ah, sorry, I overlooked it.  Sorry for the noise.

flycheck.el
ptrv: @lunaryorn well, ok. I'll add the config file options. Should I remove some of the options then? I am not sure how the `luacheck` behaves when setting both. Guess I have to check the case having stuff both in the config and via arument

mpeterv: @ptrv command-line options always override config.

ptrv: @mpeterv Thanks for clarifying. The, specifying `--formatter` and `--codes` as arguments will never be overwritten from config.

lunaryorn: @ptrv I'm not sure.  Generally, I prefer configuration files, since they can be shared among all contributors to a project and are independent of a specific editor or extension.  I think we only need to keep those options that the user is really likely to set individually.

ptrv:
Simplified the checker by removing all options and adding the `config-file` and since command-line options override the config file options the checker output format cannot be overridden.

ptrv:
@lunaryorn  Added requests and rebased onto master.

lunaryorn:
@ptrv Thanks, merged!

ptrv:
@lunaryorn Thanks for merging!

PR #621, opened 2015-04-10
Add option to specify Perl include path

mbunkus:
This adds support for specifying include paths for Perl with the `-I` command line option.

I've tried to write a test case for this, but as I've never used ERT before I've simply failed to do so properly. Sorry.

In order to reproduce this you can use two files like this: have a Perl package `TestBase/TestPackage.pm` and a script `TestBase/test_script.pl`. In `test_script.pl` simply `use TestBase::TestPackage;`.

This will generate a warning on the `use` line that the module is not found unless the directory that `TestBase` is located it is given as an include directory.

### DIFF
--- a/doc/languages.texi
+++ b/doc/languages.texi
@@ -450,6 +450,13 @@ interpreter})
 @uref{https://metacpan.org/pod/Perl::Critic, Perl::Critic})
 @end enumerate
 
+The @flyc{perl} checker provides the following options:
+
+@table @asis
+@flycoption flycheck-perl-include-path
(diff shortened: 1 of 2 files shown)


lunaryorn:
@mbunkus I'll add tests, but please add this option to the documentation in `languages.texi` as well.

mbunkus:
Will do, hopefully by tomorrow.

mbunkus:
I've added the documentation and amended my commit.

lunaryorn:
@mbunkus Thanks, merged :)

PR #628, opened 2015-04-20
Don't run .exs files through elixirc

### LINKED ISSUE #627: Elixir checker will eval .exs files
While `elixirc` will generate output errors, this is more of a side-effect for its main purpose of generating beam bytecode. Elixir also has `.exs` files which are not meant to be compiled, and running `elixirc` on them will instead execute them.

From digging around in the options for both `elixir` and `elixirc` it seems there's no way to get the errors for an `.exs` file. While I'm going to bring this up for the Elixir project, in the meantime we shouldn't eval code just because it's open.

I'm working up a PR for this, but in the meantime wanted to post the issue.

mattly:
The elixirc tool does not compile .exs files, it evaluates them.  While
ultimately this is a problem with elixir, in the meantime I would like
to not evaluate code just because I have it open in a buffer.

Closes #627

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5643,7 +5643,7 @@ See URL `http://elixir-lang.org/'."
   :modes elixir-mode
   :predicate
   (lambda ()
-    (not (string-match "\.exs$" buffer-file-name))))
+    (not (string-equal "exs" (file-name-extension buffer-file-name)))))
 
 (defconst flycheck-this-emacs-executable
   (concat invocation-directory invocation-name)
--- a/flycheck.el
+++ b/flycheck.el
@@ -5640,7 +5640,10 @@ See URL `http://elixir-lang.org/'."
    (warning line-start (file-name) ":" line ": warning:" (message) line-end)
    ;; Warnings from older Elixir versions
    (warning line-start (file-name) ":" line ": " (message) line-end))
-  :modes elixir-mode)
+  :modes elixir-mode
+  :predicate
+  (lambda ()
+    (not (string-match "\.exs$" buffer-file-name))))
 
 (defconst flycheck-this-emacs-executable
   (concat invocation-directory invocation-name)


flycheck.el
```diff
@@ -5640,7 +5640,10 @@ See URL `http://elixir-lang.org/'."
    (warning line-start (file-name) ":" line ": warning:" (message) line-end)
    ;; Warnings from older Elixir versions
    (warning line-start (file-name) ":" line ": " (message) line-end))
-  :modes elixir-mode)
+  :modes elixir-mode
+  :predicate
+  (lambda ()
+    (not (string-match "\.exs$" buffer-file-name))))
```
lunaryorn: You need to guard against `buffer-file-name` being `nil`, and please don't use a regular expression here.  Instead, use `file-name-extension` to get the extension and check it with `string-equal`.  That's clearer.

Besides, use `string-match-p` to check whether a regular expression matches a specific string.  `string-match` changes the match data, which isn't needed here.

lunaryorn:
Thank you.  Please amend this pull request to address the issues in my comment.

flycheck.el
mattly: ah, thanks. I'm still kind of new to emacs, so this is one of my first times actually diving into elisp beyond basic config stuff.

mattly:
BTW, after [this conversation](https://github.com/elixir-lang/elixir/issues/3282#issuecomment-94381550) with Elixir creator Jose Valim, I'm going to see about amending this even further at some point soon

lunaryorn:
@mattly So, should I merge it now, or not?

mattly:
Hey @lunaryorn, sorry for the delayed response, I've gotten busy with the week.

So, interesting fact: Compiling Elixir code will actually evaluate it. This is I think in part to handle macros and whatnot, but it has the potential downside of well, doing anything you can think of.

Jose suggested a safer way to syntax-check Elixir files, and I'm going to look at having flycheck do that, but it's a bit more involved than simply using `elixirc`. It might not provide compiler warnings.

lunaryorn:
@mattly Yes, uhm, and what does this mean for *this* pull request?  Should I merge it? Or close it? Or wait for some update?  I'm not sure what direction to take…

mattly:
@lunaryorn I've updated the PR addressing the original concerns about buffer-file-name being nil. This guards against my original issue of `.exs` scripts being executed by flycheck. I'll open a new issue addressing the broader concern of `elixirc` and code execution.

lunaryorn:
I'll close this issue in favor of #630.  Given what you said over there, I don't think that I want to keep this checker.

PR #634, opened 2015-04-30
Add JSCS checker

papaeye:
[JSCS](http://jscs.info/) is a code style linter for JavaScript.

JSHint has deprecated code style options and they will be removed in the next major release of JSHint.
JSCS is referred by [JSHint Options](http://jshint.com/docs/options/) page.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -55,6 +55,7 @@
 (require 'rx)                    ; Regexp fanciness in `flycheck-define-checker'
 (require 'help-mode)             ; `define-button-type'
 (require 'find-func)             ; `find-function-regexp-alist'
+(require 'json)                  ; `json-read-file'
 
 ;; Tell the byte compiler about autoloaded functions from packages
 (declare-function pkg-info-version-info "pkg-info" (package))
@@ -206,6 +207,7 @@ attention to case differences."
     javascript-jshint
     javascript-eslint
     javascript-gjslint
+    javascript-jscs
     json-jsonlint
     less
     luacheck
@@ -6309,7 +6311,8 @@ See URL `http://www.jshint.com'."
             source)
   :error-parser flycheck-parse-checkstyle
   :error-filter flycheck-dequalify-error-ids
-  :modes (js-mode js2-mode js3-mode))
+  :modes (js-mode js2-mode js3-mode)
+  :next-checkers ((warning . javascript-jscs)))
 
 (flycheck-def-option-var flycheck-eslint-rulesdir nil javascript-eslint
   "The directory of custom rules for ESLint.
@@ -6357,7 +6360,8 @@ See URL `https://github.com/eslint/eslint'."
                                  (flycheck-error-message err))))
                         (flycheck-sanitize-errors errors))
                   errors)
-  :modes (js-mode js2-mode js3-mode))
+  :modes (js-mode js2-mode js3-mode)
+  :next-checkers ((warning . javascript-jscs)))
 
 (flycheck-def-config-file-var flycheck-gjslintrc javascript-gjslint ".gjslintrc"
   :safe #'stringp)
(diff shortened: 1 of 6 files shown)


flycheck.el
```diff
@@ -6297,6 +6298,27 @@ See URL `https://github.com/w3c/tidy-html5'."
...
+            source)
+  :error-parser flycheck-parse-checkstyle
+  :error-filter flycheck-dequalify-error-ids
+  :modes (js-mode js2-mode js3-mode)
+  :predicate
+  ;; JSCS fails with exit code 1 if it cannot find a .jscsrc file.
+  ;; This :predicate prevents such cases.
+  (lambda ()
+    (flycheck-locate-config-file flycheck-jscsrc 'javascript-jscs))
+  :next-checkers (javascript-jshint javascript-eslint javascript-gjslint))
```
lunaryorn: Style linting should come last.  The other javascript checkers should have this one in their `:next-checkers`.

lunaryorn:
@papaeye Thank you very much!

I think we should reverse the order, though.  Imho style checking should come last, after syntactic and semantic checking, so ideally Flycheck would use jshint, eslint or gjslint first, and then jscs afterwards.

Would you mind to adjust this pull request accordingly?

papaeye:
Thank you for your quick feedback!
I've fixed it now.

papaeye:
@lunaryorn I'm sorry, I introduced a bug in the amended commit.
In :predicate function, We should check the existence of `jscsConfig` key in package.json.

> If there is no --config option specified, jscs it will consequentially search for jscsConfig option in package.json file then for .jscsrc (which is a just JSON with comments) and .jscs.json files in the current working directory then in nearest ancestor until it hits the system root.

Or is it possible to skip a checker if the command fails with exit code non-zero?

papaeye:
Anyway, I fixed the bug.  Test cases were ported from JSCS.

flycheck.el
```diff
@@ -6372,7 +6376,47 @@ See URL `https://developers.google.com/closure/utilities'."
...
+  (let (file)
+    (locate-dominating-file
+     default-directory
+     (lambda (dir)
+       (or (and (file-exists-p (setq file (expand-file-name "package.json" dir)))
+                (assq 'jscsConfig (json-read-file file)))
+           (file-exists-p (setq file (expand-file-name ".jscsrc" dir)))
+           (file-exists-p (setq file (expand-file-name ".jscs.json" dir)))
+           (setq file nil))))
+    file))
```
lunaryorn: Why exactly do we need this?  Does jscs not work without a configuration file?

lunaryorn:
@papaeye I appreciate your effort, but I think you've gone a little over the top on the test cases :) Remember that the tests should make sure that *Flycheck* behaves well.  It's not their job to test JSCS as well. :wink: 

Generally, I'm not sure whether I like the whole approach, though.  We shouldn't search and read a configuration file on every syntax check just to make sure that JSCS doesn't error.

I'd rather prefer if JSCS worked without a configuration file, probably with a minimal set of enabled checks.  Could we discuss this with the JSCS developers?

And if that's impossible or takes to much effort to discuss with upstream, I think Flycheck should just try to run JSCS and catch the error instead.  That's much less complicated in my opinion.  Take a look at the scss-lint syntax checker: It already does something similar to handle the case of a missing plugin.

Actually, I think that's the best short-term solution.

flycheck.el
papaeye: JSCS works without --config option, but it fails with exit code 1 when it cannot find a config file by itself.

papaeye:
@lunaryorn To prevent exit code 1 for no configuration error, this PR implements a function that searches a config file in JSCS's way.  The test cases are unit tests for the function.

But "run JSCS and catch the error" is what I was looking for before sending this PR!  I'll update this PR in scss-lint's way.

> I'd rather prefer if JSCS worked without a configuration file, probably with a minimal set of enabled checks.  Could we discuss this with the JSCS developers?

It has already discussed at jscs-dev/node-jscs#1053 and adding --ignore-missing-config option is rejected.  I think JSCS cannot provide a minimal set of enabled checks since JavaScript doesn't have official code style.  Although IMO, failing with exit code 1 makes sense, but JSCS should also provide --ignore-missing-config option to become automation tool friendly.

papaeye:
@lunaryorn Done.  I also used flycheck-remove-error-ids to remove static "jscs" ids.

papaeye:
Thanks for merging and fixing!

lunaryorn:
@papaeye You're welcome.  Many thanks for your contribution and your patience with my nit-picking :)  You're awesome :+1:

PR #636, opened 2015-04-30
Add django-mode to the list of modes supported by flake8/pylint/pycompile

brighid:
Django-mode files are pretty much always just Python files, so the same checkers work for that.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6681,7 +6681,7 @@ Requires Flake8 2.0 or newer. See URL
             (id (one-or-more (any alpha)) (one-or-more digit)) " "
             (message (one-or-more not-newline))
             line-end))
-  :modes python-mode)
+  :modes (python-mode django-mode))
 
 (flycheck-def-config-file-var flycheck-pylintrc python-pylint
                               ".pylintrc"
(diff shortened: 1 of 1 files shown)


brighid:
Pretty sure the Travis failures are related to other ongoing work, not this commit. Should I have based the diff on one of the version-number tags? It probably applies cleanly onto those.

lunaryorn:
What is Django Mode?  Do you have a link?

brighid:
Major mode for working with Django, which is a Python-based CMS. https://github.com/myfreeweb/django-mode It's packaged up in MELPA, which is where I got it from.

lunaryorn:
@brighid I'm not so sure whether this mode is something I'd like to support.  It seems to be unmaintained, and it's a questionable implementation—I see no reason why that should be a major mode on it's own and not just a minor mode. [Pony Mode](https://github.com/davidmiller/pony-mode) looks like a better alternative.  

I'd prefer if you kept this in your personal configuration, with

```cl
  (dolist (checker '(python-pylint python-flake8 python-pycompile))
    (flycheck-add-mode checker 'django-mode)))
```

brighid:
Oh, I'd missed the existence of `flycheck-add-mode`. That solves my problem in the moment, and if Django becomes some kind of world-straddling colossus that's a problem for another day. Thanks.

PR #641, opened 2015-05-08
Made scss-lint install task install scss_lint

iainbeeston:
The scss-lint ruby gem has changed name to scss_lint (note the
underscore instead of a hyphen). The binary command (with the hyphen)
has stayed the same, but the name of the gem to install has changed.

See https://github.com/brigade/scss-lint/blob/master/CHANGELOG.md for
details

This commit just updates the install task to install the gem using the
new name.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@ master (in development)
   - Add `flycheck-perl-include-path` to set include directories for Perl
     [GH-621]
 
+- Improvements:
+
+  - Updated `scss-lint` install task to install scss_lint rather than scss-lint [GH-641]
+
 0.23 (Apr 6, 2015)
 ==================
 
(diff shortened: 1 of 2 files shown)


lunaryorn:
@iainbeeston Thanks for bringing this to my attention.  Please excuse that I did not merge this PR, and instead fix the issue myself.  However, the changelog entry is wrong:  In Flycheck, the changelog only documents *user-visible* changes, but no internal changes, specifically not anything from the test suite or environment.  And since it's such a small thing, I chose to fix it myself instead of bothering you to update the PR.  I hope that's ok for you.

iainbeeston:
I'm just happy to see it fixed. Thanks!

lunaryorn:
@iainbeeston You're welcome.

PR #644, opened 2015-05-14
WIP: Add support for JavaScript Standard Style

lvillani:
**NOTE**: This is a work-in-progress. The pull-request is being temporarily
used to trigger Travis builds.

This commit introduces support for "standard", a tool to ensure code conforms
to the "JavaScript Standard Style" (https://github.com/feross/standard).

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -206,6 +206,7 @@ attention to case differences."
     javascript-eslint
     javascript-gjslint
     javascript-jscs
+    javascript-standard
     json-jsonlint
     less
     luacheck
@@ -6388,6 +6389,14 @@ See URL `http://www.jscs.info'."
   :error-filter (lambda (errors)
                   (flycheck-remove-error-ids
                    (flycheck-sanitize-errors errors)))
+  :modes (js-mode js2-mode js3-mode))
+
+(flycheck-define-checker javascript-standard
+  "A code style linter for the JavaScript Standard Style.
+
+See URL `https://github.com/feross/standard'."
+  :command ("standard" source)
+  :error-patterns ((error line-start "  " (file-name) ":" line ":" column ":" (message) line-end))
   :modes (js-mode js2-mode js3-mode))
 
 (flycheck-define-checker json-jsonlint
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -4489,6 +4489,23 @@ Why not:
      '(4 3 error "Expected indentation of 2 characters"
          :checker javascript-jscs))))
 
+(flycheck-ert-def-checker-test javascript-standard javascript error
+  (let ((flycheck-checker 'javascript-standard))
+    (flycheck-ert-should-syntax-check
+     "checkers/javascript-style.js" '(js-mode js2-mode js3-mode)
+     '(3 9 error "Missing space before function parentheses."
+         :checker javascript-standard)
+     '(4 2 error "Expected indentation of 2 characters."
+         :checker javascript-standard)
+     '(4 5 error "foo is defined but never used"
+         :checker javascript-standard)
+     '(4 12 error "Strings must use singlequote."
+         :checker javascript-standard)
+     '(4 27 error "Extra semicolon."
+         :checker javascript-standard)
+     '(5 5 error "Extra semicolon."
+         :checker javascript-standard))))
+
 (flycheck-ert-def-checker-test json-jsonlint json nil
   (flycheck-ert-should-syntax-check
    "checkers/json-jsonlint-error.json" 'text-mode
--- a/doc/languages.texi
+++ b/doc/languages.texi
@@ -390,6 +390,9 @@ Or @flyc{javascript-gjslint}
 @end itemize
 @item
 @flyc{javascript-jscs} (@uref{http://jscs.info/,JSCS})
+@item
+@flyc{javascript-standard}
+(@uref{https://github.com/feross/standard,JS Standard Style})
 @end enumerate
 
 The @flyc{javascript-eslint} checker provides the following option:
--- a/playbooks/roles/language-javascript/tasks/main.yml
+++ b/playbooks/roles/language-javascript/tasks/main.yml
@@ -3,6 +3,7 @@
   with_items:
     - jshint
     - eslint
+    - standard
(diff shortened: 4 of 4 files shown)


lunaryorn:
@lvillani Is this still WIP?  It's missing unit tests…

lvillani:
Yep, I'm trying to fix the outstanding issues with `all-checkers-are-documented` before moving on with tests and lifting the WIP status.

lvillani:
I rebased against master and added a couple of tests. Travis reports a failure but I'm seeing a similar one on master, tests seem to pass locally.

flycheck.el
```diff
@@ -6390,6 +6391,14 @@ See URL `http://www.jscs.info'."
                    (flycheck-sanitize-errors errors)))
   :modes (js-mode js2-mode js3-mode))
 
+(flycheck-define-checker javascript-standard
+  "A code style linter for the JavaScript Standard Style.
+
+See URL `https://github.com/feross/standard'."
+  :command ("standard" source)
```
lunaryorn: I read that `standard` is based on eslint, so I was wondering whether it probably has an option for checkstyle output, too?

test/flycheck-test.el
```diff
@@ -4489,6 +4489,30 @@ Why not:
      '(4 3 error "Expected indentation of 2 characters"
          :checker javascript-jscs))))
 
+(flycheck-ert-def-checker-test javascript-standard javascript nil
+  :tags '(checkstyle-xml)
+  (let ((flycheck-checker 'javascript-standard))
+    (flycheck-ert-should-syntax-check
+     "checkers/javascript-standard.js" '(js-mode js2-mode js3-mode))))
```
lunaryorn: I think this test case is redundant.  We don't need to test that a checker did nothing.  It's just important to know that it works correctly (e.g. runs, and has its output parsed) if there are any errors in a file (since a silent failure is pretty fatal if you rely on Flycheck reporting correct warnings).

flycheck.el
lvillani: It exposes very little on the command line, the following is all that's available:

```
      -F  --format    Automatically format code. (using standard-format)
      -v, --verbose   Show error codes. (so you can ignore specific rules)
          --stdin     Read file text from stdin.
          --version   Show current version.
      -h, --help      Show usage information.
```

lunaryorn: That's sad :(  Would you mind to ask the `standard` developers whether they could add an option for checkstyle XML output?  eslint supports it, so it shouldn't be too difficult for them, hopefully.

Doesn't stand against merging, though:  I'll merge it with patterns, too, but checkstyle would be nice to have, since it's less fragile.

lvillani: Opened https://github.com/feross/standard/issues/142

lvillani:
Pushed again to remove the unnecessary test.

test/flycheck-test.el
```diff
@@ -4489,6 +4489,24 @@ Why not:
      '(4 3 error "Expected indentation of 2 characters"
          :checker javascript-jscs))))
 
+(flycheck-ert-def-checker-test javascript-standard javascript error
+  :tags '(checkstyle-xml)
```
lunaryorn: Please remove this line.  This test isn't related to checkstyle currently, so the tag is wrong.

lunaryorn:
Looks good now, but I've got a couple of last questions:

- Is this syntax checker compatible with [semistandard](https://github.com/Flet/semistandard), too?  If it is, I'll add some documentation about how to use it (by changing `flycheck-javascript-standard-executable` accordingly), and also add `semistandard` to the unit tests, after merging this PR.
- How does this checker stand in relation to the existing Javascript checkers?  I understand that it replaces jscs—since both check style— but as far as I understand `standard` does only check for stylistic issues, but not for semantic or syntactic problems.  Hence, wouldn't you want to use `standard` *together* with eslint or jshint, i.e. use eslint or jshint first to check for errors, and then `standard` for style checks?  If so, we should probably add `javascript-standard` to the `:next-checkers` of eslint and jshint.

test/flycheck-test.el
lvillani: Removed `:tags` keyword in the latest push.

lvillani:
I'm not sure whether `semistandard` is compatible with `standard`. I typically use `standard` in conjunction with `standard-format`. In my case `standard-format` would remove semicolons just to cause `semistandard` to complain about them (I guess). In the end I came to like the lack of semicolons and it never occurred to me to look for something like `semistandard`.

As for the second question, I never used `jscs` nor `eslint`. In my experience `jshint` overlaps a great deal with `standard` to the point that I removed `jshint` altogether from my toolchain. `standard` correctly reports misspelled variables names, use-before-declaration, etc. I use it standalone.

lunaryorn:
@lvillani OK, I'll try for myself then.

lvillani:
:+1: Thanks!

lunaryorn:
I tried to use `semistandard`, but it breaks due to https://github.com/Flet/semistandard/issues/17

Flet:
Should be fixed now, reinstall `semistandard` and try again :)

Flet:
To comment on some other things in this thread, `semistandard`'s goal is to implement all the exact same rules as `standard` except require semicolons. It should be fully compatible in every way. Also, there is a `semistandard-format` command that can be installed or used via `semistandard -F` just like `standard` (this was just recently added).

Both `standard` and `semistandard` are really just wrappers around a set of `eslint` rules.
https://github.com/feross/eslint-config-standard/blob/master/index.js

In the beginning, `standard` was running `eslint` and `jscs`, but most of the pertinent `jscs` rules were added into `eslint` over the last few months. So, `jscs` was removed from `standard`.

lunaryorn:
@Flet Thanks for fixing this issue, and for joining this discussion to shed some light on things :)

Flet:
No problem! 
@lvillani I would encourage you to send a PR to https://github.com/feross/standard to add this plugin to the [Editor Plugins](https://github.com/feross/standard#editor-plugins) section of the README (when you're ready of course :grin:)

lvillani:
@Flet @lunaryorn Submitted https://github.com/feross/standard/pull/145

Issue #11, opened 2012-11-03
Viewing error message with flycheck


azizmb:
I installed flycheck and tried it out with some python files. I can see errors and warnings highlighted, but cant see the actual message anywhere. I'm most likely doing something wrong, but any pointers would be appreciated.

PR #647, opened 2015-05-22
Add support for external Rust crates

andschwa:
@lunaryorn This helped me work-around flycheck-rust's [#11](https://github.com/flycheck/flycheck-rust/issues/11).

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7062,6 +7062,19 @@ Relative paths are relative to the file being checked."
   :safe #'flycheck-string-list-p
   :package-version '(flycheck . "0.18"))
 
+(flycheck-def-option-var flycheck-rust-extern nil rust
+  "A list of external libraries (crates) for Rust.
+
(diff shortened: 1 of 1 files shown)


lunaryorn:
@andschwa Thanks, but I think a dedicated option is overkill for what I think is just a small cornercase.  Only few crates actually need this, for most the name is sufficiently unique.

However, with the lastest master you can pass arbitrary command line arguments to the rust checker via `flycheck-rust-args`, so if there are issues you can just copy arguments from `cargo build -v`.

andschwa:
`flycheck-rust-args` is just fine by me, and now I know a bit more about how Flycheck works. Thanks @lunaryorn!

lunaryorn:
@andschwa You're welcome!  Thanks for your feedback and your patience with me fixing this.

I really wish for a `cargo metadata` command to fix this kind of stuff once and for all :)

PR #660, opened 2015-06-10
Add c/c++-clang-check checker

stsquad:
The clang project provides a clang-check binary which can utilise a
compilation database. In a transitional Makefile based system you create
it with a tool called bear:

  bear make

Which will leave a compile_commands.json file in the root of your build.
When clang-check is called it will re-use the compile directives from
the json database. This eliminates the need to manually add include
patch to your flycheck setup.

As clang-check uses the path of the file to find the json database this
checker uses `source_original` and adds the :predicate
flycheck-buffer-saved-p.

Signed-off-by: Alex Bennée <alex.bennee@linaro.org>

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -205,6 +205,7 @@ attention to case differences."
 (defcustom flycheck-checkers
   '(ada-gnat
     asciidoc
+    c/c++-clang-check
     c/c++-clang
     c/c++-gcc
     c/c++-cppcheck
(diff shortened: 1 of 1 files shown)


lunaryorn:
@stsquad Did you use that checker?  I wonder whether it “feels” ok, because for modified (i.e. unsaved) buffers Flycheck would still fall back to the standard Clang checker, ignoring the compilation database and consequently reporting spurious errors.

stsquad:
Hmm something has gone wrong. Now whenever I look at flycheck-current-syntax-checker it's not there. I shall read up on debugging the checkers and see what happened.

lunaryorn:
@stsquad `flycheck-current-syntax-checker`?  I'm not aware of any such variable in Flycheck… 

Can I somehow help you with this issue?

stsquad:
@lunaryorn I'm not sure. It looks like the checker got stuck (flycheck-last-status-chane was running) so didn't kick off any additional checks (as observed by strace -e execve). Once I toggled global-flycheck-mode off and on again and set flycheck-check-syntax-automatically => '(save) it seems to be a lot more reliable. I'll re-push once:

* I've updated languages.texi for the new checker
* Figured out if it can be a bit more graceful reporting:
    Error while checking syntax automatically: (user-error "Selected syntax checker c/c++-clang-check cannot be used")
* Figured out how to add some new tests (or re-use the existing C checker tests)

stsquad:
New pull request #661

PR #661, opened 2015-06-11
Add clang check

stsquad:
Here is v2 of the pull request (github wouldn't let me re-open the old one).

Changes:
  - adds some documentation
  - don't chain to the "legacy" checkers (which would get confusing)
  - add a proper predicate for testing for the compilation database

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -205,6 +205,7 @@ attention to case differences."
 (defcustom flycheck-checkers
   '(ada-gnat
     asciidoc
+    c/c++-clang-check
     c/c++-clang
     c/c++-gcc
     c/c++-cppcheck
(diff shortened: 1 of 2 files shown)


stsquad:
I have no idea why emacs-snapshot is breaking the tests.

flycheck.el
```diff
@@ -5239,6 +5240,53 @@ This function determines the directory by looking at
...
+    (let ((errors (flycheck-sanitize-errors errors)))
+      (dolist (err errors)
+        ;; Clang will output empty messages for #error/#warning pragmas without
+        ;; messages.  We fill these empty errors with a dummy message to get
+        ;; them past our error filtering
+        (setf (flycheck-error-message err)
+              (or (flycheck-error-message err) "no message")))
+      (flycheck-fold-include-levels errors "In file included from")))
+  :modes (c-mode c++-mode)
+  :predicate flycheck-can-run-clang-check-p)
```
lunaryorn: Turn this into a `lambda` please.  There's no use in named functions for predicates that are only used at one place.

lunaryorn:
My question from the old PR still stands: Does that work well?  Afaics Flycheck would fall back to clang in  for modified buffers, which would be very confusing.

stsquad:
@lunaryorn: yeah pretty well. I've tested with emacs -Q and get the errors highlighted.

There is an anomaly with QEMU as you have multiple targets so you end up getting some errors twice. I haven't worked out how to test a new filter function interactively (i.e. how do I get the list of `filter-error` objects to process?).

I'm afraid I'm a little lost on expanding the tests. You would need to do a make pass on the C files to build the compilation database but I'm not sure how to hook that into the make system as I don't think any of the example tests get built. Also I'm not overly familiar with ert.

stsquad:
@lunaryorn AIUI by not adding a :next-checkers to the definition it shouldn't fall back. Of course it might still choose the other clang checker if the buffer isn't saved.

lunaryorn:
You don't need to add tests to this PR. Let's focus on the checker itself first :)

> Of course it might still choose the other clang checker if the buffer isn't saved.

That is exactly what I meant.  Wouldn't that feel awkward, since a buffers would then have different errors when unsaved and saved?

stsquad:
@lunaryorn: is that because flycheck will see that clang-check doesn't meet the predicate and select the next from the `flycheck-checkers`? What's the mechanism to stop that? Remove the other checkers from the list if we have ever run?

lunaryorn:
@stsquad There's no method to stop that other than disabling all other checkers via `flycheck-disabled-checkers` (modifying `flycheck-checkers` is not recommended).

stsquad:
@lunaryorn where would the best place to do this be? Inside :error-filters? Add a new :successful-run method?

lunaryorn:
There's no place to do this in Flycheck. `flycheck-disabled-checkers` is a user option.  

I'm not so sure whether this checker is a good fit for Flycheck proper.  I think it should be in a separate extension package, e.g. `flycheck-cxx`.  It won't work with manual generation of the database, which goes against the idea of automatic syntax checking of Flycheck proper, and it really needs `idle-change` being removed from `flycheck-check-syntax-automatically` to work fine, because even if you disabled other C/C++ checkers, you'd still have the problem that Flycheck would still attempt a syntax check on unmodified buffers, which would clear all errors reported by this checker.

In other words, this checker doesn't really work well with Flycheck's default configuration, which more or less blocks it from being part of Flycheck itself.

In a separate package, you'd be free to write a hook function which would check for the existence of a database upfront, and then explicitly enable the checker via `flycheck-checker` and disable `idle-change` buffer locally to entirely any issues with conflicting checkers.

lunaryorn:
I've seen that you've created a separate package for this checker.  I'm closing this PR then.

PR #669, opened 2015-06-23
add correct info level for `compilation-error-regexp-alist'

kolewu:
without this any info-level message gets a _TYPE_ of `nil` and this means the same as `2` (error):

> TYPE is 2 or nil for a real error or 1 for warning or 0 for info.

(from the documentation of `compilation-error-regexp-alist`)


lunaryorn:
@kolewu Thank you.

PR #671, opened 2015-06-26
Use stdin for process checker

strawhatguy:
Adds an option to flycheck-start-command-checker to pass the buffer's contents into the lint process instead of creating a temporary file on disk.

Change eslint to use this new stdin process.

The rationale for this was the fact that we run a grunt watch, that gets confused with temporary files flycheck creates.  You can change the filename flycheck creates, and grunt watch ignore files with a leading dot "." (i.e. a dotfile), however there is NO way for eslint to lint a dotfile.  It's actually hard-coded to not allow reading dotfiles.

In any case, this avoids creating temporary files at all, and thus solves the problem.


lunaryorn:
Wow, I didn't know that this is possible! I'll review and try later, but isn't a custom start function a little over the top?  Why not just a simple `:stdin t` option?

strawhatguy:
I think I did add it as an optional third argument to the ```flycheck-start-command-checker``` function, but I thought perhaps exposing the ability for a linter configuration to add their own functions would be the most flexible.  But I'm willing to do whatever is thought best of course, so please review it and tell me what you think!
flycheck is pretty awesome!

lunaryorn:
First of all, I'd like to thank you for this pull request, which really opened my eyes for what is possible in Emacs.  I never knew that `process-send-region` could be used this way!  Great work, truly!

I'll definitely merge this pull request, and even go through the nasty work of migrating as many checkers as possible to standard input, but I'm not sure yet about your proposed interface.

I think that I'd rather prefer to have a `:standard-input` option to `flycheck-define-checker`, i.e.

```
(flycheck-define-checker foo
   :command …
   :standard-input t)
```

`flycheck-start-command-checker` would then take the value of this property from the given `checker` (like it does for `:command` and friends), and send the buffer via `process-send-region` if it's non-nil.

I think that's as simple as possible, and I'd really like to keep the interface of `flycheck-define-checker` as simple as possible, because that's the single one function of Flycheck that's used the most by people who know only little Emacs Lisp.

For flexibility, there's `flycheck-define-generic-checker` :)

What do you think?  If you agree, would you mind to refactor your code this way, and update this PR by pushing to the branch?  If not, do you have another idea for an even simpler interface?

strawhatguy:
Hey it's your show. I saw ```:error-parser``` and ```:error-filter``` are both functions that can be defined in a checker's configuration, and thought following suit with the start, verify, etc. functions would be the way to go.  It depends on how much power to give the emacs user in defining some future checker, or in modifying an existing one.

In any case I opened another pull-request ( #673 ) that may be more to your liking.  Merge the one you like better.

lunaryorn:
@strawhatguy Thanks.  I'm closing this one as duplicate, and will merge #673 eventually.

strawhatguy:
Cool!  Know a guy that'll need this too.

PR #673, opened 2015-06-29
Add :standard-input option to define-checker

strawhatguy:
Merge either this PR or #671 , not both.

Adds :standard-input option to pass the buffer's contents into the lint process instead of creating a temporary file on disk.

Change eslint to use this new stdin process.

The rationale for this was the fact that we run a grunt watch, that gets confused with temporary files flycheck creates. You can change the filename flycheck creates, and grunt watch ignore files with a leading dot "." (i.e. a dotfile), however there is NO way for eslint to lint a dotfile. It's actually hard-coded to not allow reading dotfiles.

In any case, this avoids creating temporary files at all, and thus solves the problem.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3937,6 +3937,15 @@ of command checkers is `flycheck-sanitize-errors'.
      or as special symbol or form for
      `flycheck-substitute-argument', which see.
 
+`:standard-input BOOLEAN'
+     Whether or not to pass the buffer to COMMAND via standard input
+
+     When true, the current buffer's contents shall be sent
+     to the process via standard input.
+
+     This property is optional.  If omitted, it defaults to
+     false.
+
 `:error-patterns PATTERNS'
      A list of patterns to parse the output of the `:command'.
 
@@ -3993,7 +4002,8 @@ function."
         (patterns (plist-get properties :error-patterns))
         (parser (or (plist-get properties :error-parser)
                     #'flycheck-parse-with-patterns))
-        (predicate (plist-get properties :predicate)))
+        (predicate (plist-get properties :predicate))
+        (via-standard-input (plist-get properties :standard-input)))
 
     (unless command
       (error "Missing :command in syntax checker %s" symbol))
@@ -4016,7 +4026,8 @@ function."
                             (or (not predicate) (funcall predicate))))))
 
     (apply #'flycheck-define-generic-checker symbol docstring
-           :start #'flycheck-start-command-checker
+           :start (lambda (checker callback)
+                    (flycheck-start-command-checker checker callback via-standard-input))
            :interrupt #'flycheck-interrupt-command-checker
            :print-doc #'flycheck-command-checker-print-doc
            :verify #'flycheck-verify-command-checker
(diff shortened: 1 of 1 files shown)


P-Seebauer:
Hello, I also looked practically at the same issue. Is there a reason why you used `send-process-region` instead of `(send-buffer-string (buffer-contents))`? 
Both equally fail on narrowed buffers, so I see no difference there

lunaryorn:
@P-Seebauer `buffer-contents` allocates an additional string object, which is needlessly inefficient compared with `process-send-region`.

Thank you for bringing narrowing to my mind.  Indeed, we cannot send `point-min` and `point-max` to the subprocess.  Instead, the boundaries must be `0` and `(1+ (buffer-size))`.

flycheck.el
```diff
@@ -4288,6 +4304,10 @@ symbols in the command."
           ;; example for such a conflict.
           (setq process (apply 'start-process (format "flycheck-%s" checker)
                                nil program args))
+          (when (and (flycheck-checker-standard-input-p checker)
+                     (process-live-p process))
+            (process-send-region process (point-min) (point-max))
```
lunaryorn: Please use `0` and `(1+ (buffer-size))` instead of `point-min` and `point-max` respectively, to make sure that Flycheck always sends the entire buffer contents to the subprocess, even in narrowed buffers.

strawhatguy:
Two things: 
1. The boundaries start at 1, not 0
2.  I tried 1 (1+ (buffer-size)) as my bounds, but it will complain with index out-of-bounds errors when the region is narrowed.

By contrast (point-min) (point-max) gracefully handle both narrowed and widened buffers, is used already in several other places in flycheck.el, and is kinda a neat to see the errors change by what's visible.  

Seems better now as it is.

How would the temporary file based method work with narrowed regions?  That would make the lint process report line numbers that don't match the narrowed buffer's line numbers.

P-Seebauer:
This would work, but isn't that nice:

    (save-restriction (widen)
      (process-send-region (point-min) (point-max))
     )

And I'm not sure what happens to `point` then, maybe a `save-excursion` is also needed.

lunaryorn:
@strawhatguy I did not know that `process-send-region` checks for boundaries.  You'll have to use the pattern suggested by @P-Seebauer then, which is actually the _standard_ pattern to temporarily remove buffer restrictions.  But please format it properly, i.e. no hanging brace and `(widen)` on a line on its own.

> How would the temporary file based method work with narrowed regions? That would make the lint process report line numbers that don't match the narrowed buffer's line numbers.

Flycheck always writes the entire buffer contents to the temporary file, and computes buffer positions for errors with all restrictions removed.  You'll see the necessary `(save-restriction (widen) …)` pattern all over the place in Flycheck, for example in `flycheck-error-line-region`.

We cannot feed restricted buffer contents to syntax checkers: There's no guarantee that the narrowed contents are actually valid, and we might send partial code to the linter, leading to many false errors.

flycheck.el
lunaryorn: Ok, since `process-send-region` checks for boundaries, you'll need to use the following instead.

```
(save-restriction
  (widen)
  (process-send-region (point-min) (point-max)))
```

strawhatguy:
Ah, didn't know about save-restriction, that is better.

lunaryorn:
@strawhatguy Superseded by #728

lunaryorn:
@strawhatguy So, after too long a delay from my side, this has finally landed!  Many many thanks!  See #728 for the progress in converting individual syntax checkers.

:+1: :fireworks:

strawhatguy:
Awesome!  Yeah I had noticed other syntax checkers starting to use it. Great to know it's in!  Glad I could help

PR #694, opened 2015-07-17
Fix for #675, Environment Names in puppet could contain underlines too

### LINKED ISSUE #675: Problem in parsing Puppet Messages
This the parser for puppet doesnt catch all error Statements, the Message in the Log is:
`Suspicious state from syntax checker puppet-parser: Checker puppet-parser returned non-zero exit code 1, but no errors from output: Error: Could not parse for environment jahe_dev: Syntax error at 'stage' at /tmp/flycheck19432d0B/init.pp:16:9`

The Error could be triggered by the following Statement:

    class { 'base::root'
     stage => init,
    }

The Error is the missing ':' after the Ressource Name

Ladeburger:
This would fix #675
The Name of Environments should match the regular Expression `\A[a-z0-9_]+\Z` (Ruby-notation) according to https://docs.puppetlabs.com/puppet/latest/reference/environments_creating.html#allowed-environment-names but `(one-or-more words)` doesnt allow underlines but whitespaces.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6691,10 +6691,10 @@ See URL `http://puppetlabs.com/'."
   (
    ;; Patterns for Puppet 4
    (error line-start "Error: Could not parse for environment "
-          (one-or-more word) ":"
+          (one-or-more (in "a-z" "0-9" "_")) ":"
           (message) " at " (file-name) ":" line ":" column line-end)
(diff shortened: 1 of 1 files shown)


lunaryorn:
@Ladeburger Merged.  Thanks for the fix, and I'm really sorry for the long delay, I've been pretty busy over the last time.

PR #700, opened 2015-07-25
Add a compilation-level property to error levels

### LINKED ISSUE #699: Extend `flycheck-checker-pattern-to-error-regexp` to support custom levels
When passed a custom error level, the pcase in `flycheck-checker-pattern-to-error-regexp` returns `nil`, causing messages at that level to be highlighted as errors. This makes it tricky to add custom levels that are not errors.

I can think of an easy fix: we could add an extra keyword arg to define-checker, maybe `compilation-error-level`, which would be any of `('error 'warning 'info)`. The we would replace `(level-no (pcase level` by `(level-no (pcase (flycheck-get-compilation-error-level level)`. This would be backwards-compatible, as existing custom levels would return `nil`, causing the whole `pcase` to return `nil` just as before.

Would such a patch be useful enough to warrant inclusion? If so, I'll provide it.

cpitclaudel:
This property indicates which level to use for each error level when
building compilation-error-regexp-alist. This lets users define their
own levels and have them highlighted as info or warnings.

Closes #699

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2901,6 +2901,16 @@ The following PROPERTIES constitute an error level:
      The severity is used by `flycheck-error-level-<' to
      determine the ordering of errors according to their levels.
 
+`:compilation-level LEVEL
+
+     A number indicating the broad class of messages that errors
+     at this level belong to: one of 0 (info), 1 (warning), or
+     2 or nil (error). Defaults to nil.
+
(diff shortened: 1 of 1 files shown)


lunaryorn:
@cpitclaudel Oh, that's a great PR, with docstrings and all 😍  Many thanks!

cpitclaudel:
:)

PR #701, opened 2015-07-25
Add a 'filter' command to the error list

### LINKED ISSUE #698: Add something similar to navigation-minimum-level for the error list
Hi lunaryorn,

It would be cool to be able to specify a minimum error level for the error list; that would make it easier to deal with checkers that return lots of `info`-level messages.

I expect that implementing this should be a simple patch adding a `defcustom` and a level of filtering to `flycheck-error-list-entries`. If you think this would be a good feature to add, I can make a pull request.

Cheers,
Clément.

cpitclaudel:
Pressing "f" in the error list now prompts for an error level. All
errors whose severity level is below that level are hidden. Pressing "F"
resets the filter. The current filter is shown in the mode line. The
docs mention the two new commands.

Changing filter levels scrolls back to the top of the error list. This
is to avoid confusing behavior when filters are removed. Filtering out
the first entry of the error list, for example, moves the second entry
to the top. If we didn't scroll back up, removing the filter would leave
the first entry invisible, causing confusion.

Closes #698.

Should closing and reopening the error list reset the filter? If so 
```
  (unless (get-buffer flycheck-error-list-buffer)
    (with-current-buffer (get-buffer-create flycheck-error-list-buffer)
      (flycheck-error-list-mode)))
```
should be changed to 
```
  (unless (get-buffer flycheck-error-list-buffer)
    (setq flycheck-error-list-minimum-level nil)
    (with-current-buffer (get-buffer-create flycheck-error-list-buffer)
      (flycheck-error-list-mode)))
```

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -521,19 +521,42 @@ enabled.  Changing it will not affect buffers which already have
   :package-version '(flycheck . "0.15")
   :safe #'booleanp)
 
+(define-widget 'flycheck-minimum-level 'lazy
+  "A radio-type choice of minimum error levels.
+
+See `flycheck-navigation-minimum-level' and
+`flycheck-error-list-minimum-level'."
+  :type '(radio (const :tag "All locations" nil)
+                (const :tag "Informational messages" info)
+                (const :tag "Warnings" warning)
+                (const :tag "Errors" error)
+                (symbol :tag "Custom error level")))
+
 (defcustom flycheck-navigation-minimum-level nil
   "The minimum level of errors to navigate.
 
 If set to an error level, only navigate errors whose error level
 is at least as severe as this one.  If nil, navigate all errors."
   :group 'flycheck
-  :type '(radio (const :tag "All locations" nil)
-                (const :tag "Informational messages" info)
-                (const :tag "Warnings" warning)
-                (const :tag "Errors" error)
-                (symbol :tag "Custom error level"))
+  :type 'flycheck-minimum-level
   :safe #'flycheck-error-level-p
   :package-version '(flycheck . "0.21"))
+
+(defcustom flycheck-error-list-minimum-level nil
+  "The minimum level of errors to display in the error list.
+
+If set to an error level, only displays errors whose error level
+is at least as severe as this one in the error list.  If nil,
+display all errors.
+
+This is the default level, used when the error list is opened.
+You can temporarily change the level using
+\\[flycheck-error-list-set-filter], or reset it to this value
+using \\[flycheck-error-list-reset-filter]."
+  :group 'flycheck
+  :type 'flycheck-minimum-level
+  :safe #'flycheck-error-level-p
+  :package-version '(flycheck . "0.24"))
 
 (defcustom flycheck-completion-system nil
   "The completion system to use.
@@ -826,7 +849,8 @@ Set this variable to nil to disable the mode line completely."
 (defcustom flycheck-error-list-mode-line
   `(,(propertized-buffer-identification "%12b")
     " for buffer "
-    (:eval (flycheck-error-list-propertized-source-name)))
+    (:eval (flycheck-error-list-propertized-source-name))
+    (:eval (flycheck-error-list-mode-line-filter-indicator)))
   "Mode line construct for Flycheck error list.
 
 The value of this variable is a mode line template as in
@@ -1254,6 +1278,7 @@ FILE-NAME is nil, return `default-directory'."
 
 Show PROMPT and read one of CANDIDATES, defaulting to DEFAULT.
 HISTORY is passed to `ido-completing-read' or `completing-read'."
+  ;; TODO: Could use a small cleanup
   (pcase flycheck-completion-system
     (`ido (ido-completing-read prompt candidates nil
                                'require-match nil
@@ -1302,6 +1327,18 @@ a default on its own."
       (unless (flycheck-valid-checker-p checker)
         (error "%S is no Flycheck syntax checker" checker))
       checker)))
+
+(defun read-flycheck-error-level (prompt)
+  "Reads an error level from the user.
+
+Only offers level for which errors currently exist, in addition
+to the default levels."
+  (let* ((levels (mapcar #'flycheck-error-level
+                         (flycheck-error-list-current-errors)))
+         (levels-with-defaults (append '(info warning error) levels))
+         (uniq-levels (-uniq levels-with-defaults))
+         (level (flycheck-completing-read prompt uniq-levels nil)))
+    (and (stringp level) (intern level))))
 
 
 ;;; Checker API
@@ -3418,6 +3455,8 @@ the beginning of the buffer."
 
 (defvar flycheck-error-list-mode-map
   (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "f") #'flycheck-error-list-set-filter)
+    (define-key map (kbd "F") #'flycheck-error-list-reset-filter)
     (define-key map (kbd "n") #'flycheck-error-list-next-error)
     (define-key map (kbd "p") #'flycheck-error-list-previous-error)
     (define-key map (kbd "g") #'flycheck-error-list-check-source)
@@ -3527,12 +3566,17 @@ Return a list with the contents of the table cell."
                    (format "(%s)" checker)
                    'flycheck-error-list-checker-name)))))
 
+(defun flycheck-error-list-current-errors ()
+  "Read the list of errors in `flycheck-error-list-source-buffer'."
+  (when (buffer-live-p flycheck-error-list-source-buffer)
+    (buffer-local-value 'flycheck-current-errors
+                        flycheck-error-list-source-buffer)))
+
 (defun flycheck-error-list-entries ()
   "Create the entries for the error list."
-  (when (buffer-live-p flycheck-error-list-source-buffer)
-    (let ((errors (buffer-local-value 'flycheck-current-errors
-                                      flycheck-error-list-source-buffer)))
-      (mapcar #'flycheck-error-list-make-entry errors))))
+  (-when-let* ((errors (flycheck-error-list-current-errors))
+               (filtered (flycheck-error-list-apply-filter errors)))
+    (mapcar #'flycheck-error-list-make-entry filtered)))
 
 (defun flycheck-error-list-entry-< (entry1 entry2)
   "Determine whether ENTRY1 is before ENTRY2 by location.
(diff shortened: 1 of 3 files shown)


cpitclaudel:
I'd like input on the key choice. <kbd>f</kbd> for filter seems pretty straightforward, but <kbd>F</kbd> for resetting the filter may be less obvious. Paradox uses <kbd>q</kbd>, but that's used for closing the error list in flycheck. <kbd>g</kbd> was another candidate, but it's already taken too.

Resetting the filter when the list is closed and reopen would probably works, but it makes it hard (impossible?) for a package that depends on flycheck to set a default filter. If we do want to reset when closing and reopening the list, then we could introduce another variable `default-minimum-level`. Closing and reopening would reset `minimum-level` to `default-minimum-level` instead of nil (and we can even make `default-minimum-level` a `defcustom`).

lunaryorn:
@cpitclaudel I'm not sure either.  At some places Flycheck uses a prefix argument to reset state, e.g. for manual syntax checker selection, but I'm not sure whether `C-u f` is really better than `F`.

cpitclaudel:
I added a commit to solve the customization issue. There are now two settings:

* `flycheck-error-list-minimum-level` is the default level that gets
  applied every time the error list is reopened or the filtering level
  is reset.
* `flycheck-error-list-current-filter` is the current filtering level,
  which gets reset when the error list is reopened or "F" is pressed.

I think that works nicely. It also allows users to set a default filter.

flycheck.el
```diff
@@ -521,20 +521,43 @@ enabled.  Changing it will not affect buffers which already have
...
+(defconst flycheck-minimum-level-choices
+  '((const :tag "All locations" nil)
+    (const :tag "Informational messages" info)
+    (const :tag "Warnings" warning)
+    (const :tag "Errors" error)
+    (symbol :tag "Custom error level"))
+  "Choices for minimum error level options.
+
+See `flycheck-navigation-minimum-level' and
+`flycheck-navigation-minimum-level'.")
```
lunaryorn: I guess this is a copy & paste error, isn't it?  One mention of this variable would suffice, wouldn't it? 😉

flycheck.el
```diff
@@ -521,20 +521,43 @@ enabled.  Changing it will not affect buffers which already have
   :package-version '(flycheck . "0.15")
   :safe #'booleanp)
 
+(defconst flycheck-minimum-level-choices
```
lunaryorn: I'm not sure that a variable is the right way to do this.  Did you check whether we can [define a custom type](https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-New-Types.html#Defining-New-Types) for this one?  Probably even one that dynamically provides a choice for all defined levels in the current Emacs session?

flycheck.el
```diff
@@ -1249,6 +1273,26 @@ FILE-NAME is nil, return `default-directory'."
 (defvar read-flycheck-checker-history nil
   "`completing-read' history of `read-flycheck-checker'.")
 
+(defun flycheck-completing-read (prompt candidates default &optional history)
```
lunaryorn: Thanks for extracting this one.  It's a first step, and I think it needs further cleanup.  Not part of this PR, of course, but it's something that I'll need to remember.  If you want, you can drop me a `;; TODO` comment here :)

flycheck.el
```diff
@@ -826,7 +849,8 @@ Set this variable to nil to disable the mode line completely."
 (defcustom flycheck-error-list-mode-line
   `(,(propertized-buffer-identification "%12b")
     " for buffer "
-    (:eval (flycheck-error-list-propertized-source-name)))
+    (:eval (flycheck-error-list-propertized-source-name))
+    (:eval (flycheck-error-list-mode-line-filter-indicator)))
```
lunaryorn: A cool, mode line indication of the current filter is an awesome idea! :+1:

flycheck.el
```diff
@@ -3610,6 +3650,55 @@ list."
       ;; highlighting
       (flycheck-error-list-highlight-errors preserve-pos))))
 
+(defvar flycheck-error-list-current-filter nil
```
lunaryorn: I'm not sure about this variable.  Do we need an extra one in addition to `flycheck-error-list-minimum-level`?  Is there a reason why you don't just set the buffer local variable of `flycheck-error-list-minimum-level`?

flycheck.el
```diff
@@ -3610,6 +3650,55 @@ list."
...
+is at least as severe as this one in the error list.  If nil,
+display all errors.")
+
+(defun flycheck-error-list-mode-line-filter-indicator ()
+  "Create a string representing the current error list filter."
+  (if flycheck-error-list-current-filter
+      (format " [>= %s]" (symbol-name flycheck-error-list-current-filter))
+    ""))
+
+(defun flycheck-error-list-query-level (prompt)
```
lunaryorn: Please rename this one to `read-flycheck-error-level`, for consistency with similar functions and with `read-flycheck-checker`.  Also, please move it below the definition of `read-flycheck-checker`, so that all minibuffer input functions are in one section.

flycheck.el
```diff
@@ -3610,6 +3650,55 @@ list."
...
+  "Reads an error level from the user.
+
+Only offers level for which errors currently exist."
+  (-when-let* ((error-levels (mapcar #'flycheck-error-level
+                                     (flycheck-error-list-current-errors)))
+               (uniq-error-levels (-uniq error-levels))
```
lunaryorn: I think we should always offer all levels.  A user might want to “proactively” filter warnings and infos, for instance, if they are about to make a larger refactoring, and fix up compilation errors first, before dealing with the new warnings.

lunaryorn:
@cpitclaudel Thanks for this great PR!  Awesome work!  I've added a couple of questions and remarks, but overall this one is definitely good for merging.

I'd be happy if you could address my remarks and comments, but it's also totally fine if you don't want to.  In this case, I'll do it myself, as soon as possible.

Thanks!

lunaryorn:
@cpitclaudel Please excuse my excessive (ab-)use of emojis, but it's just so much fun and pleasure to review great PRs and constructively work together on a new useful and important feature!  I wish every PR I get was like yours!  Many many thanks!

cpitclaudel:
Awesome, thanks for the very detailed review and the great feedback! I'll have time to look at this in more detail tomorrow :)
Flycheck really has a beautiful codebase, so it was a pleasure to navigate it and to make these changes. Thanks for this wonderful piece of software!

flycheck.el
cpitclaudel: Great suggestion, thanks! I changed to a custom widget. If might be good to add an `:offset` to it, but I'll leave that to your judgement :)

I'm not sure how to make the widget list all currently defined levels though; I don't think we currently keep track of that list. It would be easy to add such bookkeeping to `flycheck-define-error-level`, but it might be slightly confusing if the user selects a level that is defined by an autoloaded package, as the corresponding `severity` will not be known until that package is loaded.

flycheck.el
cpitclaudel: Nice catch, thanks!

flycheck.el
cpitclaudel: Ah, brilliant idea!

flycheck.el
cpitclaudel: Done.

flycheck.el
cpitclaudel: I've edited the code to include the default levels (`'error`, `'warning`,  and `'info`). We could display all defined levels if we kept track of them, but do we want to list unrelated levels too? Maybe package `a` and package `b` define different levels, and it's not clear that we want to offer both options at the same time; I'll let you judge!

cpitclaudel:
All right, I pushed a commit with changes addressing the points that we discussed! The one place that could definitely use a second pair of eyes is the `flycheck-list-errors` function, which I refactored slightly to avoid duplicating the call to `with-current-buffer`. Let me know if you want me to squash these commits, especially the last three.

lunaryorn:
@cpitclaudel I'll take a look asap, but I'm out tonight and tomorrow night, and might not get to it before Saturday.  I hope that's ok for you, and please excuse the delay!

cpitclaudel:
No problem at all! And no hurry. Cheers!

flycheck.el
```diff
@@ -3727,10 +3814,13 @@ non-nil."
...
-  ;; Create and initialize the error list
-  (unless (get-buffer flycheck-error-list-buffer)
-    (with-current-buffer (get-buffer-create flycheck-error-list-buffer)
-      (flycheck-error-list-mode)))
+  ;; Create the error list and reset the filter
+  (with-current-buffer (get-buffer-create flycheck-error-list-buffer)
+    (unless (eq 'major-mode 'flycheck-error-list-mode)
+      (flycheck-error-list-mode))
+    (set (make-local-variable 'flycheck-error-list-minimum-level)
+         (default-value 'flycheck-error-list-minimum-level)))
```
lunaryorn: I would not make the variable at this place, and rather rely on the global value as long as the user did not press `f`.

flycheck.el
```diff
@@ -3610,6 +3663,40 @@ list."
...
+      (set (make-local-variable 'flycheck-error-list-minimum-level)
+           (and (flycheck-error-level-p level) level)))
+    (flycheck-error-list-refresh)
+    (flycheck-error-list-recenter-at (point-min))))
+
+(defun flycheck-error-list-reset-filter ()
+  "Remove filters and show all errors in the error list."
+  (interactive)
+  (flycheck-error-list-set-filter
+   (default-value 'flycheck-error-list-minimum-level)))
```
lunaryorn: I think we can just use `(kill-local-variable 'flycheck-error-list-minimum-level)` here, to remove the buffer-local binding, which makes Emacs fall back to the global value, i.e. the default value, automatically.

flycheck.el
```diff
@@ -3610,6 +3663,40 @@ list."
...
+    ""))
+
+(defun flycheck-error-list-set-filter (level)
+  "Restrict the error list to errors at level LEVEL or higher."
+  (interactive
+   (list (read-flycheck-error-level
+          "Minimum error level (errors at lower levels will be hidden): ")))
+  (-when-let (buf (get-buffer flycheck-error-list-buffer))
+    (with-current-buffer buf
+      (set (make-local-variable 'flycheck-error-list-minimum-level)
```
lunaryorn: Please `setq-local` instead `(set (make-local-variable …) …)`.

flycheck.el
```diff
@@ -3610,6 +3663,40 @@ list."
...
+           (and (flycheck-error-level-p level) level)))
```
lunaryorn: I think we should not silently reset any filter if the `level` is invalid, but instead signal an error.   The current behaviour will be confusing in my opinion, because users can easily reset filters accidentally:  As far as I can see `read-flycheck-error-level` does not require matching input and returns whatever symbol the user enters, which in turn will pass an invalid `level` to this function, causing the `and` clause to return `nil` and thus resetting the filter.

flycheck.el
lunaryorn: @cpitclaudel I thought that you could easily define widgets that construct their `:type` dynamically, but going back to the docs I see that this is rather convoluted—it requires a custom `:create` function I think.  And I didn't think of the auto-loading issue.

You're right, let's leave it as is.  It's a good compromise between convenience and extensibility, and above all, it's simple :)

flycheck.el
lunaryorn: :+1:

flycheck.el
lunaryorn: We do not track them, but we can find all levels the same way we find all syntax checkers at some places: By iterating over all symbols and testing for special properties.

Assuming that we'll at some point have dozen of packages with custom error types, that's not a good idea, but I don't think that we ever will.  Actually, I'm not aware of any extension that actually defines custom levels, and in fact I only know of a single external user of `flycheck-define-error-level`: Spacemacs, which uses it to redefine the built-in levels to change their fringe indicators.

But then again, under this assumption we don't need to spend much care for custom levels either, so I think that your approach of including the built-in levels and the currently present levels, and allowing the user to input arbitrary symbols is a good compromise.

Let's keep it that way.  If it doesn't work well for our users, we can still fix it afterwards.

lunaryorn:
@cpitclaudel It's almost good for merging :+1:  We can simplify the local variable handling a bit, but that's mostly just a stylistic issue.  I'd be happy if you could address it, as by my comments, but I can also fix it myself after merging.

But prior to merging I'd like to ask you to squash your commits into a single one, to keep Flycheck's history clean.

Many thanks for this cool feature!  It's a great idea, and a great implementation!

flycheck.el
cpitclaudel: Done :)

flycheck.el
cpitclaudel: Done!

flycheck.el
cpitclaudel: Done! Do all emacs version that we support have it? I saw that the rest of flycheck used `(set (make-local-variable ...) ...)`, so I initially did the same.

flycheck.el
cpitclaudel: That's smart! I didn't think about that. Done.

cpitclaudel:
Alrighty, I've addressed the points that you mentioned, and squashed it back to two commits. Should that be just one? 

Thanks for the great reviewing!

flycheck.el
lunaryorn: Yes.  I've recently raised the minimum Emacs version to 24.3, which has `setq-local`, but even before Flycheck provided a fallback definition for this macro on older versions.

lunaryorn: @cpitclaudel There's only two occurrences of `make-local-variable` in Flycheck.  One of these was a mistake (probably written before I knew about `setq-local`), and the other is necessary, because at this place we compute the variable name dynamically, and `setq-local` wants a literal symbol.

lunaryorn:
@cpitclaudel No, it's perfect that way.  I'll have another look at it tonight, and merge it.

lunaryorn:
@cpitclaudel Thanks, awesome feature!

Sarcasm:
Sorry to shim in a little bit late, I was just wondering about the keys, if it would make sense to have:

* `f` to toggle filtering: using the default (user-customizable via defcustom) level filter, so `f` to filter and reset
* `F` to filter with a special filter (`f` to reset it again).

What do you think?

Issue #447, opened 2014-07-30
javascript-eslint has trouble finding the right config files


xendk:
I was getting different errors from flycheck running eslint and running it manually, so I looked into what flycheck and eslint does.

When not explicitly supplied a config file as an argument, eslint reads .eslintrc files from the directory of the file, and all parent directories and merge the settings together. Flycheck looks for the config file in the same way, but only supplies the first one to eslint.

So I'd suggest that the `flycheck-eslintrc` variable is defaulted to `nil` and that the `javascript-eslint` checker is changed to use `source-inplace` instead of `source`, to let eslint collect its configuration itself.

The alternative would be to have flycheck find all the configuration files, and supply them on the command line.

azangru:
@jasonkuhrt So have you figured out how to disable jshint and enable eslint? I am now in the exact situation as you were: using jshint by default but would like to try eslint.

jasonkuhrt:
```
  (setq flycheck-disabled-checkers '(javascript-jshint))
  (setq flycheck-checkers '(javascript-eslint))
```

jasonkuhrt:
This no longer works, I'm not sure why.

PR #703, opened 2015-07-28
Fix source arguments of JSHint and JSCS checkers

papaeye:
It is necessary to use `source-inplace` instead of `source` to support [.jshintignore](http://jshint.com/docs/cli/#ignoring-files) file in JSHint and [excludeFiles](http://jscs.info/overview.html#excludefiles) option in JSCS, for example.

This is same fix as ESLint checker (see #447).

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6359,7 +6359,7 @@ See URL `http://jade-lang.com'."
 See URL `http://www.jshint.com'."
   :command ("jshint" "--checkstyle-reporter"
             (config-file "--config" flycheck-jshintrc)
-            source)
+            source-inplace)
   :error-parser flycheck-parse-checkstyle
   :error-filter flycheck-dequalify-error-ids
   :modes (js-mode js2-mode js3-mode)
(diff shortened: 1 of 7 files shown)


lunaryorn:
@papaeye Would you mind to check whether these checkers can read source code from standard input, for #673?

papaeye:
Both can read from standard input:

http://jshint.com/docs/cli/#specifying-input

> If a file path is a dash (-) then JSHint will read from standard input.

http://jscs.info/overview.html#installation

> You can also pipe input into jscs:
> ```
> cat myfile.js | jscs
> ```

But I counldn't use the :standard-output option for JSHint and JSCS checkers.
Here is my attempt: https://github.com/flycheck/flycheck/compare/master...papaeye:wip/pr-673%2B703

lunaryorn:
Flycheck doesn't yet support that, but will quite soon, because it will save us a lot of trouble I think.  I'm currently just compiling a list of checkers that support standard input, though, to be prepared for converting them.

PR #705, opened 2015-07-29
may-use-checker is no longer an enemy of debugging

brighid:
Previously, if you wanted to find out why a seemingly-usable checker
could not be enabled, flycheck-may-use-checker hated you and wanted
you to suffer. With this patch, flycheck-may-use-checker loves you and
wants to help you find out why the checker could not be enabled.

flycheck-may-use-checker has been expanded to speak accurately of its
error conditions rather than making them all look identical, including
the condition where the executable that a checker needs is not present
on exec-path (previously this was handled by a lambda, implanted in
checkers' :predicate property by flycheck-define-command-checker, that
only ran the real :predicate if executable-find came up non-nil for the
checker's executable: this behavior is preserved).

flycheck-may-use-checker now accepts an optional quiet-mode
parameter. When quiet-mode is non-nil, the function returns to its old
no-messages behavior. This lets flycheck-get-checker-for-buffer avoid
dumping hundreds of lines into `*Messages*` when it loops over the entirety
of flycheck-checkers. That loop now passes the quiet-mode parameter each
time, and does so with a more efficient and readable cl-remove-if loop
that replaces a confusing while loop.

`make tests` gives the same results as for parent (8a691012).

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1634,17 +1634,35 @@ A valid checker is a symbol defined as syntax checker with
        (= (or (get checker 'flycheck-generic-checker-version) 0)
           flycheck-generic-checker-version)))
 
-(defun flycheck-may-use-checker (checker)
+(defun flycheck-may-use-checker (checker &optional quiet-mode)
   "Whether a generic CHECKER may be used.
 
 Return non-nil if CHECKER may be used for the current buffer, and
-nil otherwise."
+nil otherwise.  Prints a message about CHECKER's suitability
+unless QUIET-MODE is non-nil."
   (let ((modes (flycheck-checker-get checker 'modes))
-        (predicate (flycheck-checker-get checker 'predicate)))
-    (and (flycheck-valid-checker-p checker)
-         (not (flycheck-disabled-checker-p checker))
-         (or (not modes) (memq major-mode modes))
-         (funcall predicate))))
+        (predicate (flycheck-checker-get checker 'predicate))
+        (executable (flycheck-checker-executable checker)))
+    (cl-flet
+        ((qmsg (message-string &rest args)
+               (unless (or noninteractive quiet-mode)
+                 (apply #'message message-string args)) nil))
+      (cond
+       ((not (flycheck-valid-checker-p checker))
+        (qmsg "%s is not a valid checker." checker))
+       ((flycheck-disabled-checker-p checker)
+        (qmsg "Checker %s is disabled." checker))
+       ((not modes)
+        (qmsg "No modes are associated with checker %s." checker))
+       ((not (memq major-mode modes))
+        (qmsg "Checker %s can handle %s, but this buffer is using %s."
+              checker modes major-mode))
+       ((not (executable-find executable))
+        (qmsg "Checker %s needs to run '%s', which is not on `exec-path'."
+              checker executable))
+       ((not (funcall predicate))
+        (qmsg "Calling :predicate of checker %s returned nil." checker))
+       (t t)))))
 
 (defun flycheck-may-use-next-checker (next-checker)
   "Determine whether NEXT-CHECKER may be used."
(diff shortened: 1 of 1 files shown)


lunaryorn:
Thank you for this pull request, but I must admit that it is not an approach that I would like to merge.  My preference is to keep functions pure.  Introducing side-effecting printing and—even worse—debugging code (that's what it is, after all) into a pure predicate is a no-go for me.  It's going to end up in a maintenance nightmare sooner or later.

The go-to tool for debugging syntax checkers in Flycheck is `flycheck-verify-setup` at `C-c ! v` which prints a list of all available checkers and potential issues with each of them.  As of yesterday evening, it also includes the explicitly selected checker, if any, and shows why it would not work.

As such, I'm not sure whether we still need this change.  We should probably rather popup the verification buffer in case of failure, because it's the nicer interface and provides even more detail on failures.

brighid:
All right, that's a reasonable stance to have. Part of why I wrote this change the way I did is that I wanted to have a small diff that minimized the chance of introducing bugs. Replying "this is a frequently-used predicate: it should go fast and operate without side effects" is _entirely_ reasonable. 

What if I try implementing this change again, using something more like the functional style that you're recommending? I feel strongly that since `flycheck-may-use-checker` is, effectively, the user's first point of contact with "can I check this buffer or not," it should **not** emit a useless error message when the answer is "no you can't check it." It should be user-compassionate and either provide more information or directly point to a way to get that information.

Right now, when you run `flycheck-mode` and thereby ask flycheck whether you can use a checker, all of the following answers lead to the same (very unhelpful) message: `"Selected syntax checker FOO cannot be used"`.
* The checker needs external program Foo to exist on `exec-path`, and it doesn't
* The checker doesn't handle the kind of file you asked me to check with it
* The checker is so broken I can't parse it
* The checker is disabled
* The checker ran its `:predicate` function and got nil

Some of those situations are trivially fixable, and some are not. It is very good that `flycheck-verify-setup` exists, but if it's not discoverable it's not usable or useful (as I write this, the official documentation's [page about errors](http://www.flycheck.org/manual/latest/Errors.html) is completely blank, so I'm deeply dubious about the discoverability of `flycheck-verify-setup`). The user experience I was trying to create with this patch is that when a user tries to turn on checking and flycheck says "No, I can't check this buffer," flycheck _immediately_ provides actionable information about why the answer was "no." Right now, the user experience is that instead of that information, the message sent to the user is "and I refuse to tell you why," which is a profoundly suboptimal user experience. It is entirely okay that utilities like flycheck sometimes fail, but they should fail in ways that minimize the likelihood of the user being left confused and unable to mend the failure. Trying to figure how to proceed in the face of a determinedly unhelpful error message is _expensive,_ it's one of the paradigmatic Bad User Experiences.

This is F/OSS. We can do better. :)

lunaryorn:
I do agree that the user experience is not good in this situation, but I'm not sure about the best way to improve it.  I've put a task on my Trello board to improve the situation, and I'll experiment with various ways of poping up the verify buffer if things go wrong.   I think that's the better approach than a message, because it's more detailed, has a nicer UI, and is interruptive and not easily lost (whereas messages can get lost quickly if Emacs produces other output).

I would like to point out, though, that the page about “Errors”, which you have linked, is entirely unrelated to the issue at hand.  If you'd like to take a look at the navigation bar for this page, you'll realize that it's a section in the chapter “Flycheck API”, which is intended to provide a documentation of Flycheck's API, not for users, but for developers who intend to write custom Flycheck extensions.  The section “Errors” specifically will at some point provide an overview of all functions and types that Flycheck provides for errors reported for the current buffer, i.e. how many are there, what levels do they have, etc.  

You should rather refer to [Checking buffers](http://www.flycheck.org/manual/latest/Checking-buffers.html#Checking-buffers), and you'll find `flycheck-verify-setup` to be documented at the bottom of the page.  I do agree, though, that it should be placed more prominently.

lunaryorn:
@brighid `flycheck-select-checker` now pops up a buffer that tells you why when you try to select a checker that is not applicable for the current buffer.

Does this address the issue for you?  Or are the other places, where Flycheck does not provide sufficiently good error reporting?

PR #709, opened 2015-07-31
add source-directory

edvorg:


lunaryorn:
@edvorg I thank you for this contribution, but I'd appreciate if you could tell me what this is actually about.  Why do you need this feature?  What specific problem does it solve for you?

edvorg:
I'm trying to implement a protobuf support for flycheck.
I finally managed, but with this simple modification of flycheck.
The issue is that protoc requires path to proto files directory to find imported proto files.
Please, look at the example
https://github.com/edvorg/flycheck-protobuf/blob/master/flycheck-protobuf.el

lunaryorn:
@edvorg Well, if it's for just one external syntax checker, please use `eval` instead.  I'm unsure anyway whether `source-directory` would be the right approach for this situation:  Written as is, wouldn't this syntax checker only find protobuf files from the very same directory?  And fail if the protobuf files are organized in some kind of hierarchic structure?

edvorg:
I tried to use eval for it. But I'm not surely understand, how to get current file path.
I tried something like
```
(eval (file-name-path source-inplace))
```

edvorg:
Also typically protoc files stored in same single directory

lunaryorn:
@edvorg As documented `flycheck-substitute-arguments` does not recursively apply substitutions, and specifically does not substitute _inside_ `eval`.  You need to use `(buffer-file-name)` or `default-directory`.

> Also typically protoc files are stored in the same single directory.

Is that just “typical”, or actually _required_?

edvorg:
Thanks. It works now, so changes in flycheck are no longer required.
It's not required. But because of its imports system, it's hard to store protocols in hierarchic directory structure.

Issue #702, opened 2015-07-26
Questions regarding executables which use wrappers


purcell:
Let's say that instead of `hlint`, I would like to run `stack exec hlint`. Currently I can set `flycheck-haskell-hlint-executable` to either `nil` or a string, but `(executable-find "stack exec hlint")` naturally fails. Would it make sense to allow executables to be set to lists?

The downside of this is that a wrapper script might be executable, while the command it wraps is not, so that the wrapper will fail every time it is run. But this is also a (semi-related) problem for us `rbenv` users: for example, `~/.rbenv/bin/rubocop` will always be on the path, but will fail noisily inside any ruby project directory which does not have that gem installed.

So a second thought/suggestion is that it would be nice to be able to specify a per-checker sanity check for executables which goes beyond just verifying that they are present on the `$PATH`. For example, "run with `--version` and expect a zero exit code", or even "check that the output of `checker --version` contains a version string greater than 0.2.2".

lunaryorn:
@purcell I'm not sure how to deal with such wrappers.  You can use `:predicate` to add custom checks, and Flycheck does that already for some checkers (e.g. `go-vet`), but I'm not sure whether that's the best approach.

For `stack` specifically, though, you can run `stack install hlint`, which will install `hlint` globally in `~/.local/bin/`.

purcell:
If I recall correctly from working on it recently, `syntastic` always does a version check by default. I guess it trades off that extra `exec` for some assurance that the checker will actually work when needed. I think this is the best trade-off in the end.

Re. `stack`, the example of `hlint` wasn't the best one for me to give. More specifically, `stack exec -- ghc -fno-code -Wall ...` conveniently makes ghc inherit the stack library environment, cutting `cabal` out of the picture. So as stack gains in popularity I'd quite like to figure out how to use flycheck seamlessly with it. I think that allowing the executable var to be set to `("stack" "exec" "--" "ghc")` might do the trick.

lunaryorn:
@purcell I'll add a separate syntax checker for `stack ghc` (there's a branch for it already, see `feature/stack-ghc`), because that works better for the user in my opinion.  Flycheck could automatically use `stack ghc` for projects with `stack.yml` and standard `ghc`—if installed—for project without, without any manual configuration required.

[...]

Issue #629, opened 2015-04-22
Add flycheck-substituted-command-wrapper-function to mangle command after substitution


lunaryorn:
>  It'll allow extensions to transform the command after the argument substitution. IOW, it'll get the command as given to call-process, and can do whatever it wants to do with the command.

Primarily for Nix support, see #610 for discussion.

purcell:
Nice.

svenkeidel:
I want to help implementing this feature. I'm also a NixOS user, so I can test this first hand. I started looking through the source where this function could fit in. I would change [these lines](https://github.com/flycheck/flycheck/blob/master/flycheck.el#L4490-L4496) and pass the results to the `flycheck-substituted-command-wrapper-function`.

A problem are calls of `executable-find`. In NixOS executables like `hlint` are made available on a project level, not system-wide. One way to check if a executable is available in a project is to execute the following command: `nix-shell --command 'whereis hlint'`. This is the same way you would wrap every call to a checker with `flycheck-substituted-command-wrapper-function`. My suggestion would be to add a `flycheck-executable-find` function that uses the wrapper function and `whereis` to check if a executable is available in a project or not. The second step would be to replace every occurence of `executable-find` with `flycheck-executable-find`.

What do you think of my suggestions? I upload a prototype of this shortly.

Best,
Sven

svenkeidel:
I have ignored the `executable-find` problem for now. The prototype is [here](https://github.com/flycheck/flycheck/compare/master...svenkeidel:command-wrapper-function)

lunaryorn:
I like the implementation, though there are some minor issues, or things that I'd like to do differently.  For instance while I understand the motivation for `cons` being the default, it's a rather strange thing from a user's perspective.  I'd rather expect `nil` or something along the lines here.  Please do feel free to open a pull request where we can discuss the details of the implementation.  I'll definitely merge it.

I'm not particularly convinced of a `flycheck-executable-find` function that calls `whereis`.  That smells like a very bad hack.  We could add a `flycheck-executable-find` option, though, defaulting to `executable-find`.  You could then override it with a custom function that goes through nix.  Again, I'd welcome a pull request.

That's the most I'd be willing to do for Nix though.  I understand that it's the hot new cool thing(TM), but it's also a nuisance because it doesn't play well with standard Unix patterns (i.e. a `$PATH` where all executables are), and do not want to bloat Flycheck's code to work around these limitations of Nix.

svenkeidel:
[...]

PR #711, opened 2015-08-01
Add a syntax checker for stack ghc

lunaryorn:
TODO

- [x] Install stack on VM
- [x] Add some unit tests?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ master (in development)
   - Javascript with `jscs` [GH-634] and `standard` [GH-644]
   - Jade [GH-686]
   - SQL with `sqllint` [GH-691]
+  - Haskell with `stack ghc` [GH-711]
 
 - New features:
 
--- a/doc/languages.texi
+++ b/doc/languages.texi
@@ -338,11 +338,18 @@ functions for format string problems.
 
 @enumerate
 @item
+@itemize @bullet
+@item
+@flyc{haskell-stack-ghc} (syntax and type checker using
+@uref{https://github.com/commercialhaskell/stack, Stack}) in Stack
+projects, or
+@item
 @flyc{haskell-ghc} (syntax and type checker with
-@uref{http://www.haskell.org/ghc/,GHC}), and
+@uref{http://www.haskell.org/ghc/,GHC}),
+@end itemize
 
 @item
-@flyc{haskell-hlint} (style checker with
+and @flyc{haskell-hlint} (style checker with
 @uref{https://github.com/ndmitchell/hlint,hlint}), with the
 following options:
 
--- a/flycheck.el
+++ b/flycheck.el
@@ -193,6 +193,7 @@ attention to case differences."
     go-errcheck
     haml
     handlebars
+    haskell-stack-ghc
     haskell-ghc
     haskell-hlint
     html-tidy
@@ -6278,7 +6279,7 @@ is added to the GHC search path via `-i'."
   :package-version '(flycheck . "0.16"))
 
 (flycheck-def-option-var flycheck-ghc-language-extensions nil haskell-ghc
-  "Language extensions for GHC.
+  "Language extensions for GHC and Stack.
 
 The value of this variable is a list of strings, where each
 string is a Haskell language extension, as in the LANGUAGE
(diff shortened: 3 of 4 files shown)


purcell:
Will this be both a `haskell-stack-ghc` and a `haskell-stack-hlint`?

lunaryorn:
@purcell I see no point in `haskell-stack-hlint`.  You can just use `stack install hlint`, can't you?

purcell:
Probably, but I kind of assumed that different projects could use different `hlint` versions, and that the right way to exec the correct one for the current project would be with `stack exec hlint`. Maybe I'm misunderstanding how the whole thing should work.

lunaryorn:
@purcell I don't know, but I was under the impression that you'd just install `hlint` globally with `stack install`…  I wasn't even aware that you could install `hlint` into the per-project environment.  How do you even do that?

purcell:
Yeah, I guess you're right. It doesn't look like there's any project-specific handling of executables. In that case, just ignore me. :-)

purcell:
Awesomesauce - will have to give this a go. :-)

conklech:
Stack sort of has project-specific executables. If, say, in a project using the `lts-3.0` snapshot, you run `stack build hlint`, the `hlint` binary will be installed in the snapshot directory at `stack/snapshots/arch/snapshot/compiler/bin`, and can be run within the project directory using `stack exec hlint`.

For binary packages that aren't in a snapshot, I believe the binary would be placed in the project-local `.stack-work`, but would be run the same way.

lunaryorn:
@conklech Do people actually use project-local `hlint` variants?  I'd expect most Haskell developers to install a basic toolset (e.g. stylisch, hindent, hlint, etc.) globally…

conklech:
That's my expectation as well. Perhaps there are compiler version specific
tools. I just wanted to observe that it's possible.

lunaryorn:
@conklech Great, thanks for your input.  I think we don't need to provide a special `stack exec hlint` syntax checker then, and an implementation of #629 should cover this case anyway should the need arise.

slpopejoy:
Hi, I opened the above issue, but I also want to question, why was this created as a completely separate checker? 

The definition copy-pastes all of the warning and error parsing, meaning any future changes must maintain both.

Also, the docs were changed in a way to imply that Stack and GHC are somehow different, e.g. `flycheck-ghc-language-extensions` now has "Language extensions for GHC and Stack". The extensions are for GHC only -- stack is not a compiler.

I would argue instead that `haskell-ghc` should support stack similarly to how it detects a cabal file or a sandbox.

Cheers,
Stuart

purcell:
@slpopejoy See related discussion in https://github.com/flycheck/flycheck/issues/702. I'm still not sure the current situation is ideal, but IIRC it was the expedient choice at the time. Now that `stack` is becoming very much more widely used, it could be that we have a clearer understanding of how things should work. Any suggestions would be welcome.

lunaryorn:
@slpopejoy Summarizing the discussion in #702, a separate syntax checker lets us re-use all the existing features for syntax checker selection.  Specifically, it 

- lets you individually disable Stack for specific projects via `flycheck-disable-checker` or `flycheck-select-checker`,
- lets you see whether stack is actually used via `flycheck-verify-setup` or in the error list,
- lets us re-use the existing logic for checking for executables and predicates,
- lets us easily account for the subtle difference between invoking plain GHC and stack GHC (e.g. pass a package database or not)
- and—most important to me—puts the use of Stack in an nice declaration definition that's easy to understand and to follow.

On the other hand, discovering stack and manipulating `:command` to support a single syntax checker definition moves the logic for using stack far away from the syntax checker definition, is less transparent to the user because there wouldn't be an easy way to see what's used anymore, and would re-invent some wheels (e.g. finding the stack executable, disabling stack).

For instance, we'd have to provide an option and interface to disable Stack, something that we get for free with a nice interface if we have a separate syntax checker definition.

To me these advantages outweight the small amount of copy-pasted code, the more since the likeliness of GHC changing their output format in incompatible ways appears to be rather low, doesn't it?  And should that—against all odds—happen we can still easily overcome this by adding support for shared error pattern definitions to Flycheck.

Summarizing, making Stack a separate syntax checker tends to play best with how Flycheck works.  We are open to better ideas, though—if you have a suggestion or a better implementation please share it.

slpopejoy:
To my mind, the different-checker thing isn't as nice as `auto` for `haskell-process-type`, which does a great job selecting `stack-ghci` or `cabal-repl` appropriately. 

Unless I'm mistaken, I'll have to do a dir-locals hack (which I haven't had to do in a while, thanks to endless improvement of `haskell-mode`) if I want a project to select the right checker. This is indeed why I reported #847, as I've gotten used to emacs doing the right thing vis-a-vis stack or not. Presumably we can find a solution for that, and this will be suitably automatic.

Thanks for summarizing #702, I agree with the thinking there overall.

lunaryorn:
@slpopejoy Flycheck should select the appropriate syntax checker automatically as well.  If it doesn't, that's a bug, i.e. #847.

PR #714, opened 2015-08-04
Add pylint error symbol to message

donkopotamus:
This pull request adds the pylint message symbol to the pylint checker.  After this change messages appear in emacs as:

    Unable to import 'spam' (import-error) [F0401]

instead of

    Unable to import 'spam' [F0401]

The motivation is following ... I sometimes need to add pylint directives to my code to suppress particular messages around particular snippets.  And its better practice to add those directives using meaningful message symbols rather than message ids, for the benefit of readers.  For example:

    ### pylint: disable=import-error
    <code>

With this change, users can see the symbol they want to disable right there in their emacs, without having to look it up.

### DIFF
--- a/doc/languages.texi
+++ b/doc/languages.texi
@@ -590,9 +590,14 @@ The maximum length of lines for Flake8.
 @flycconfigfile{flycheck-flake8rc,Flake8}
 @end table
 
-@flyc{python-pylint} reads a configuration file:
+@flyc{python-pylint} provides the following options:
 
 @table @asis
+@flycoption flycheck-pylint-use-symbolic-id
+A boolean indicating whether to report symbolic or numeric message identifiers.
+For example, whether to report a message type as no-name-in-module, or E0611.
+On by default.
+
 @flycconfigfile{flycheck-pylintrc,Pylint}
 @end table
 
--- a/flycheck.el
+++ b/flycheck.el
@@ -7101,6 +7101,16 @@ Requires Flake8 2.0 or newer. See URL
                               ".pylintrc"
   :safe #'stringp)
 
+(flycheck-def-option-var flycheck-pylint-use-symbolic-id t python-pylint
+  "Whether to use pylint message symbols or message codes.
+
+A pylint message has both an opaque identifying code (such as `F0401') and a
+more meaningful symbolic code (such as `import-error').  This option governs
+which should be used and reported to the user."
+  :type 'boolean
+  :safe #'booleanp
+  :package-version '(flycheck . "0.24"))
+
 (flycheck-define-checker python-pylint
   "A Python syntax and style checker using Pylint.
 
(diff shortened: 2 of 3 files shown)


lunaryorn:
@donkopotamus Can we have these symbolic names in place of the numeric error ID?   Ideally, we could make that configurable, by changing the error patterns to accept both, and changing the Pylint format according to the value of a Flycheck option…

donkopotamus:
@lunaryorn Perhaps something like this?  I've now added a `flycheck-pylint-use-symbolic-id` option, defaulting to `t`.  The `msg-template` is then altered depending on its value.

lunaryorn:
@donkopotamus That's great.  Would you mind to add a single test case for the alternate setting of the new option.  Just copy one of the pylint tests, bind `flycheck-pylint-use-symbolic-id` to `nil` and test that the errors now have numeric IDs.  Just to make sure that both settings work.

Thanks :smile:

donkopotamus:
@lunaryorn Just to confirm ... is this as simple as adding something like the following to `flycheck-test.el` ?

```
(flycheck-ert-def-checker-test python-pylint python syntax-error
  (let ((flycheck-disabled-checkers '(python-flake8))
        (python-indent-guess-indent-offset nil)    ; Silence Python Mode
        (flycheck-pylint-use-symbolic-id nil)) 
     "checkers/python-syntax-error.py" 'python-mode
     '(3 1 error "invalid syntax" :id "E0001" :checker python-pylint))))
```

I currently don't have the tools available to actually run the test suite ... and my understanding is that Travis CI currently does not run the syntax checkers themselves?

lunaryorn:
@donkopotamus Yes, but you need to change the name of the test (i.e. `syntax-error-with-numeric-ids` or so), and I'd like to have a large test copied, where warnings, infos and errors are tested, just to make sure that all patterns work.

lunaryorn:
@donkopotamus Also, please add this new option to `doc/languages.texi` so that it is documented in the manual.

lunaryorn:
@donkopotamus Great, looks good to me now.  Please squash all commits into a single one, so that I can merge with a clean history.

donkopotamus:
@lunaryorn Has been squashed and rebased

lunaryorn:
@donkopotamus Merged.  Thanks a lot :heart_eyes:

jtamagnan:
Would there be opposition to modifying flycheck-pylint-use-symbolic-id to add an option to show both the {symbol} and the {msg_id}. Depending on the option it would show:
-``` Unable to import 'spam' [F0401]```
-```Unable to import 'spam' (import-error) ```
-```Unable to import 'spam' (import-error) [F0401] ```

lunaryorn:
@jtamagnan I'd merge a PR that adds a third value to that option.  As long as it's a simple change I have no objections.  Still, I'd like to know why you need both…

PR #728, opened 2015-08-17
Feed syntax checkers from standard input

lunaryorn:
Supersedes #673 

# Remaining tasks

- [ ] Change syntax checkers to use standard input as far as possible
- [ ] Add documentation
- [x] Adapt `flycheck-compile`
- [ ] Fix tests

Test failure:

```
Test flycheck-define-checker/sh-bash/default backtrace:
  signal(error ("Process flycheck-sh-bash not running"))
  (condition-case err (let* ((checker (flycheck-get-checker-for-buffer
  (if (flycheck-running-p) nil (run-hooks (quote flycheck-before-synta
  (if flycheck-mode (if (flycheck-running-p) nil (run-hooks (quote fly
  flycheck-buffer()
  flycheck-ert-buffer-sync()
```
 
# Updating syntax checkers

Checked syntax checkers use standard input, checkers which can't use standard input are struck through, with reasons explained.

- [x] ~~ada-gnat~~ (can't read from standard input)
- [ ] asciidoc
- [x] c/c++-clang
- [ ] c/c++-gcc
- [ ] c/c++-cppcheck
- [ ] cfengine
- [ ] chef-foodcritic
- [ ] coffee
- [ ] coffee-coffeelint
- [ ] coq
- [ ] css-csslint
- [ ] d-dmd
- [x] ~~emacs-lisp~~ (can't read from standard input)
- [x] emacs-lisp-checkdoc (can't read from standard input)
- [ ] erlang
- [ ] eruby-erubis
- [ ] fortran-gfortran
- [ ] go-gofmt
- [ ] go-golint
- [ ] go-vet
- [ ] go-build
- [ ] go-test
- [ ] go-errcheck
- [ ] groovy
- [ ] haml
- [ ] handlebars
- [x] ~~haskell-stack-ghc~~ (GHC can't read from standard input)
- [x] ~~haskell-ghc~~ (GHC can't read from standard input)
- [ ] haskell-hlint
- [ ] html-tidy
- [ ] jade
- [ ] javascript-jshint
- [ ] javascript-eslint
- [ ] javascript-gjslint
- [ ] javascript-jscs
- [ ] javascript-standard
- [ ] json-jsonlint
- [ ] less
- [ ] luacheck
- [ ] lua
- [ ] perl
- [ ] perl-perlcritic
- [ ] php
- [ ] php-phpmd
- [ ] php-phpcs
- [ ] puppet-parser
- [ ] puppet-lint
- [ ] python-flake8
- [ ] python-pylint
- [ ] python-pycompile
- [ ] r-lintr
- [ ] racket
- [ ] rpm-rpmlint
- [ ] rst
- [ ] rst-sphinx
- [x] ruby-rubocop
- [x] ~~ruby-rubylint~~ (can't read from stdin)
- [x] ruby
- [x] ruby-jruby
- [ ] rust
- [ ] sass
- [ ] scala
- [ ] scala-scalastyle
- [ ] scss-lint
- [ ] scss
- [ ] sh-bash
- [ ] sh-posix-dash
- [ ] sh-posix-bash
- [x] ~~sh-zsh~~ (zsh doesn't include column numbers when reading from standard input :disappointed:)
- [ ] sh-shellcheck
- [ ] slim
- [ ] sql-sqlint
- [ ] tex-chktex
- [ ] tex-lacheck
- [ ] texinfo
- [ ] verilog-verilator
- [ ] xml-xmlstarlet
- [ ] xml-xmllint
- [ ] yaml-jsyaml
- [ ] yaml-ruby

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,11 +83,11 @@ This is a model commit message:
 
     - Use a hanging indent
 
-[Git Commit Mode][] and [Magit][] provide a major mode for Git commit messages,
-which helps you to comply to these guidelines.
+[Git Commit][] and [Magit][] provide Emacs mode for Git commit messages, which
+helps you to comply to these guidelines.
 
 [Tim Pope's guidelines]: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
-[Git Commit Mode]: https://github.com/magit/git-modes/
+[Git Commit]: https://github.com/magit/magit/
 [Magit]: https://github.com/magit/magit/
 
 Contributing syntax checkers
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -37,6 +37,7 @@ substantial code to Flycheck:
 - [Mario Rodas](https://github.com/marsam)
 - [Mark Karpov](https://github.com/mrkkrp)
 - [Mark Hellewell](https://github.com/markhellewell)
+- [Matthew Curry](https://github.com/strawhatguy)
 - [Matthias Dahl](https://github.com/BinaryKhaos)
 - [Miro Bezjak](https://github.com/mbezjak)
 - [Mitch Tishmack](https://github.com/mitchty)
--- a/flycheck.el
+++ b/flycheck.el
@@ -4246,6 +4246,14 @@ of command checkers is `flycheck-sanitize-errors'.
      `flycheck-parse-with-patterns'.  In this case,
      `:error-patterns' is mandatory.
 
+`:standard-input t'
+     Whether to send the buffer contents on standard input.
+
+     If this property is given and has a non-nil value, send the
+     contents of the buffer on standard input.
+
+     Defaults to nil.
+
 Note that you may not give `:start', `:interrupt', and
 `:print-doc' for a command checker.  You can give a custom
 `:verify' function, though, whose results will be appended to the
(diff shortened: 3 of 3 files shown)


cpitclaudel:
One comment about this: we may want to consider adding a setting to revert to writing files to disk; emacs subprocesses have particularly bad performance in some cases.  This is an issue I've run into while writing code for `company-coq` and `proof-general`, and when using Flycheck in connection with a persistent verification server process in the context of `boogie-friends`. The issue is mostly that passing large strings over a pipe is slow, so although stdin works well for small files, if someone is editing a large file things might get slow. The problem is not too perceptible right now, since error output is typically smaller than the input file, but it might pop up if we move many checkers to use stdin.

cpitclaudel:
(Side note: I think this is a very exciting feature; we should definitely have it as the default :)

lunaryorn:
@cpitclaudel Is writing buffer contents to a pipe slower than writing them to disk?

cpitclaudel:
Sometimes, at least. I'm not entirely sure why. Poor buffering performance (and SSDs)? Your implementation of `flycheck-save-buffer-to-temp` is pretty smart, too (it sets `write-region-inhibit-fsync`).

lunaryorn:
@cpitclaudel That'd be unfortunate.  I don't think that we can easily make standard input customizable, as it obviously affects syntax checker definitions, and there's no easy and generic mapping from an invocation with standard input to an invocation with a temporary file that'd fit all syntax checkers.  Some tools want a `-` to read from standard input, others always read from standard input when not given a file, and error patterns need to be changed as well to handle different files.

I'd rather try to make it the default, and push Emacs upstream for improvements when we see performance regressions.

tomjakubowski:
The basic rust checker can use stdin (`rustc -Z no-trans -- - < lib.rs`) but that'll still only work (in all cases) for the file that corresponds to the "crate root" module.  It also needs to be run from the same directory as the Rust source file for module resolution to work.

cpitclaudel:
@lunaryorn Ok :) Seems reasonable.

PR #733, opened 2015-08-20
Never show 0/0 in the modeline

cpitclaudel:
This is a very small change, so I'm not sure if it deserves a prior discussion in a separate issue. The idea is that if there are only info messages, then Flycheck will show `FlyC:0/0` instead of `FlyC`:

```elisp
(flycheck-ert-with-temp-buffer
  (let ((flycheck-current-errors
         (list (flycheck-error-new-at 1 1 'info "info"))))
    (should (string= (flycheck-mode-line-status-text 'finished) " FlyC"))))

(flycheck-ert-with-temp-buffer
  (let ((flycheck-current-errors
         (list (flycheck-error-new-at 1 1 'info "info")
               (flycheck-error-new-at 1 1 'error "error"))))
    (should (string= (flycheck-mode-line-status-text 'finished) " FlyC:1/0"))))
```

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2995,10 +2995,10 @@ nil."
                 (`running "*")
                 (`errored "!")
                 (`finished
-                 (if flycheck-current-errors
-                     (let-alist (flycheck-count-errors flycheck-current-errors)
-                       (format ":%s/%s" (or .error 0) (or .warning 0)))
-                   ""))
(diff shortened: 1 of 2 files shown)


lunaryorn:
@cpitclaudel Great. Can you just add these tests to the test suite?

cpitclaudel:
Sure! Done.

lunaryorn:
@cpitclaudel Thanks, merged.

PR #742, opened 2015-08-31
Fix Coq error pattern.

mgudemann:
This fixes an error with the error pattern for Coq. Before, the matching
was done on `Error: ' as start of error message, but there can also be a
newline directly after `Error:'.Now it matches on `Error:' and `Syntax
error:' and removes all leading whitespace before in error-filter.


lunaryorn:
@mgudemann Thank you.

Issue #743, opened 2015-09-01
Replace the GIF with a annotated screenshot of Flycheck at work


lunaryorn:
I'd like to replace the README "video" with an annotated static screenshot of Flycheck.   Take a look at the [screenshots of the Napkin App](https://itunes.apple.com/us/app/id581789185?mt=12&ign-mpt=uo%3D4) to get an idea of what I mean.

I think the video does a bad job of demonstrating what specific features Flycheck actually provides, and the "animation" very quickly tends to feel "noisy" when looking at the Github page.  A static image, which has "annotations" and highlights for each specific feature of Flycheck (i.e. the error list, the mode line reporting, the fringe indicators, the underlines, etc.) provides more information with less noise.

@tonini, @purcell, @cpitclaudel Your opinions?

tonini:
I like the idea, I had the same thought too for Alchemist's README (to remove these noisy GIF's). 
But I find it kind of hard to bring all features and highlights on one image at once. 

I think it's worth the effort, maybe have a look at jetbrains product presentation images, I quite like the loupe effect. https://www.jetbrains.com/webstorm/?fromMenu

cpitclaudel:
I think it's a great idea. I like the gif, but I think in the case of Flycheck features could be conveyed just as clearly by a static image. And if we wanted to highlight specific features, we could add smaller screenshots further down the page.

purcell:
:+1:

cpitclaudel:
Here's some code that I use before taking screenshots:

```elisp
(defun prepare-for-screenshot ()
  (interactive "P")
  (load-theme 'tango t)
  (menu-bar-mode -1)
  (tool-bar-mode -1)
  (fringe-mode '(4 . 4))
  (setq-default cursor-type nil))
```

lunaryorn:
@tonini @cpitclaudel @purcell I've made a first draft of an annotated screenshot.

![annoted flycheck screenshot 19 09 15 1 40 59 nachm](https://cloud.githubusercontent.com/assets/224922/9976029/109c1a42-5ed4-11e5-9a99-52d48058c8a1.png)

You can see it in action in [this preliminary branch](https://github.com/flycheck/flycheck/tree/annotated-screenshot-for-readme#readme).  If you'd like to work on it yourself, I can send you the original screenshot and the Napkin source file.

[...]

PR #750, opened 2015-09-20
Use custom fringe indicators [#743]

lunaryorn:
New uniform indicators, inspired by Syntastic's indicators:

napkin 20 09 15 12 41 41 nachm

I find these much nicer than the default ones, and a uniform indicator make it easier to tell Flycheck indicators apart from other fringe icons.

Inviting @purcell @cpitclaudel @tonini and @syl20bnr for your feedback :)


cpitclaudel:
Looks beautiful! One small thing is that giving them the same shape might make it trickier for people with bad eyesight to tell them apart. Since the underlines have the same color as the indicators, this change might be removing a shape-based clue that some might have depended on.

Can we fit three `>` signs in the fringe? So we'd have `>`, `≫`, and `⋙` for info, warning, and error. Alternatively, we could just have `>` for info and `≫` for warning and errors (then the only risk would be confusing warnings and errors, which isn't that bad). This would also be consistent with indicators used by some mail programs. 

But yes, these look great!

cpitclaudel:
Ah, looking at the implementation it might be tricky to fit three `>`s. Then maybe just moving `info` messages to a single `>` would do?

lunaryorn:
@cpitclaudel I've thought about this point as well, but I don't think it's a real issue.  I guess even colour-blind people would rather customize their theme to use differentiable colours than relying on the fringe indicators.  These are just not helpful for this purpose:  They are too small, and they're not actually related to the error _at point_.

Syntastic and Spacemacs have been using uniform indicators ever since, and I'm not aware of any issues ever coming up with regards to the shape.

I think we can just try and use uniform indicators.  If people actually relied on the indicator shape they'll come here and report an issue, and then we can still try to find a better solution :relaxed:

cpitclaudel:
Seems quite reasonable :)

syl20bnr:
I like it and it makes the UI less heavy for the eyes.

Even for not color-blind people there is the issue of not being able to
distinguish the different type of indicators if a theme decide to do so.

In Spacemacs with leuven theme the dots are all the same color and this is
even worst in Spacemacs because the lighter in the modeline relies also on
the colors. This is a real problem.

What I planed in Spacemacs is to add an option for color blind people that
replaces the dots in modeline by e, w, and i.

Jenkins ships with blue/red status colors instead of green/red to fix the
issue for color blind people. In Flycheck blue is already taken for info so
it may be tricky to do this.

I'm not sure this is an issue in pure flycheck because the lighter is more
specific.

Le dimanche 20 septembre 2015, Clément Pit--Claudel <
notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>> a écrit :

> Seems quite reasonable :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/flycheck/flycheck/pull/750#issuecomment-141783208>.
>


-- 
-syl20bnr-

lunaryorn:
> I like it and it makes the UI less heavy for the eyes.

Great, thank you :)

> Even for not color-blind people there is the issue of not being able to
> distinguish the different type of indicators if a theme decide to do so.

Did you actually get issue reports or feedback from colour-blind people for Spacemacs?

> Jenkins ships with blue/red status colors instead of green/red to fix the
> issue for color blind people. In Flycheck blue is already taken for info so
> it may be tricky to do this.

Flycheck does not explicitly choose any colours.  By default, its faces inherit from the standard `error`, `warning` and `info` faces respectively, and anything beyond that is up to the theme.   IOW, it's the theme (solarized probably?) that “takes” blue, not Flycheck itself.

And I think colour-blind people will already have chosen a theme which is friendly to colour-blindness, simply because the issue arises at any lace where Emacs uses colours, i.e. virtually anywhere.

purcell:
Big thumbs-up from me.

tonini:
Looks really great! :+1:

lunaryorn:
Thank you all for your feedback and for the discussion.  I've merged this branch now, let's see whether folks enjoy the new indicators.

syl20bnr:
> Did you actually get issue reports or feedback from colour-blind people for Spacemacs?

Nope I haven't, rather surprising. You are right, color-blind people will select appropriate themes for them.

Nice to see these indicators merged :+1:

PR #758, opened 2015-10-06
Add json checking via python json.tool module

marsam:
Python `json.tool` module  is available in both python2 and python3.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -206,6 +206,7 @@ attention to case differences."
     javascript-jscs
     javascript-standard
     json-jsonlint
+    json-python-json
     less
     luacheck
     lua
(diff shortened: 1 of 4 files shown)


lunaryorn:
@marsam Thank you for your contribution.  It's a good idea to have JSON checking without additional tools, and I'll definitely merge it, but there are a couple of minor issues that need to be solved first.  

Please address my remarks in the diff.  Furthermore, please document the new syntax checker in `doc/languages.texi`, and add a unit test to `test/flycheck-tests.el`.

flycheck.el
```diff
@@ -206,6 +206,7 @@ attention to case differences."
     javascript-jscs
     javascript-standard
     json-jsonlint
+    json-pythonjson
```
lunaryorn: Please rename to `json-python-json`.  That's just easier to read :relaxed:

flycheck.el
```diff
@@ -6872,6 +6873,22 @@ See URL `https://github.com/zaach/jsonlint'."
      (and (buffer-file-name)
           (string= (file-name-extension (buffer-file-name)) "json")))))
 
+(flycheck-define-checker json-pythonjson
+  "A JSON syntax checker using Python json.tool module.
+
+See URL `https://docs.python.org/3.5/library/json.html#command-line-interface'."
+  :command ("python" "-m" "json.tool" source-inplace null-device)
```
lunaryorn: Why do you use `source-inplace` here?  Is that really necessary?  Can't we use `source` instead?

I presume `null-device` is for the (pretty-printed?) output?   If so, please add a little comment that explains it, so that I don't need to look up documentation if I need to understand this syntax checker in half a year from now or so :relaxed:

flycheck.el
```diff
@@ -6872,6 +6873,22 @@ See URL `https://github.com/zaach/jsonlint'."
...
+  :error-patterns
+  ((error line-start
+          (message) ": line " line " column " column (one-or-more not-newline)
```
lunaryorn: What does `(one-or-more not-newline)` match?

marsam: it matchs the char of the error, e.g: ` (char 57)`. I was not sure if included or not, but given that all the `error-patterns` include the `line-end` regex, I preferred to add it

marsam:
Hi:
I've updated the pr with your comments

lunaryorn:
@marsam Cherry-picked on `master` :)  Thank you very much for this contribution!

PR #760, opened 2015-10-12
Fix regex stack overflow in haskell-hlint

drvink:
Some background: Emacs's regex engine is ["not very clever"](https://groups.google.com/forum/#!original/gnu.emacs.help/pxfMC5f7JHg/lhMNJ0tx6nwJ); in particular, the `.` atom induces recursion, which when combined with a long enough string, will overflow the [regex failure stack](http://vovick.blogspot.com/2013/04/fixing-stack-overflow-in-regexp-matcher.html). This is particularly bad when working with [multi-line strings](http://www.emacswiki.org/emacs/MultilineRegexp), as Flycheck often is.

The `haskell-hlint` checker was susceptible to this when given a big enough string; we solve the issue by using `[^\n]` instead of `.` when we want to match any non-newline character.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6681,14 +6681,12 @@ See URL `https://github.com/ndmitchell/hlint'."
   ((warning line-start
             (file-name) ":" line ":" column
             ": Warning: "
-            (message (one-or-more not-newline)
-                     (one-or-more "\n" (one-or-more not-newline)))
+            (message (one-or-more (and (one-or-more (not (any ?\n))) ?\n)))
             line-end)
    (error line-start
           (file-name) ":" line ":" column
           ": Error: "
-          (message (one-or-more not-newline)
-                   (one-or-more "\n" (one-or-more not-newline)))
+          (message (one-or-more (and (one-or-more (not (any ?\n))) ?\n)))
           line-end))
   :modes (haskell-mode literate-haskell-mode))
 


drvink:
If you'd like to see the "before and after", `Main.hs` in [ginsu](http://hackage.haskell.org/package/ginsu) will allow you to reproduce the issue.

flycheck.el
```diff
@@ -6681,15 +6681,13 @@ See URL `https://github.com/ndmitchell/hlint'."
   ((warning line-start
             (file-name) ":" line ":" column
             ": Warning: "
-            (message (one-or-more not-newline)
-                     (one-or-more "\n" (one-or-more not-newline)))
-            line-end)
+            (message (+ (not (any ?\n))))
+            (| ?\n string-end))
```
lunaryorn: Please use or instead of |, for the sake of readability ☺️

flycheck.el
```diff
@@ -6681,15 +6681,13 @@ See URL `https://github.com/ndmitchell/hlint'."
   ((warning line-start
             (file-name) ":" line ":" column
             ": Warning: "
-            (message (one-or-more not-newline)
-                     (one-or-more "\n" (one-or-more not-newline)))
-            line-end)
+            (message (+ (not (any ?\n))))
```
lunaryorn: Please use on-or-more instead of +.  It's easier to read that way. ☺️

flycheck.el
drvink: OK, will fix it up shortly.

BTW, please don't merge this yet--I may be on some serious crack, but in my terminal scrollback, I can clearly see that the code was working correctly, but I opened the same file that I'd been using to test just now and it's behaving differently, so I need to get to the bottom of this first (hurrrr durrr).

drvink:
I made the changes you requested and pushed the correct version (orz).

lunaryorn:
@drvink Thanks, merged :relaxed: :+1:

Issue #767, opened 2015-10-25
Convert syntax checkers to use standard input


lunaryorn:
- [x] ~~ada-gnat~~ (can't read from standard input)
- [x] asciidoc
- [x] c/c++-clang
- [x] c/c++-gcc
- [x] ~~c/c++-cppcheck~~ (can't read from stdin)
- [x] ~~cfengine~~ (can't read from stdin)
- [x] ~~chef-foodcritic~~ (can't read from stdin)
- [x] coffee
- [x] coffee-coffeelint
- [x] ~~coq~~ (can't read from stdin)
- [x] ~~css-csslint~~ (no stdin)
- [x] ~~d-dmd~~ (no stdin)
- [x] ~~emacs-lisp~~ (no stdin)
- [x] ~~emacs-lisp-checkdoc~~ (no stdin)
- [x] ~~erlang~~ (no stdin)
- [x] ~~eruby-erubis~~ (behaves differently when using standard input)
- [x] ~~fortran-gfortran~~ (can't read from stdin without language option, don't know the proper value)
- [x] go-gofmt
- [x] ~~go-golint~~ (no stdin, wants a package)
- [x] ~~go-vet~~ (no stdin)
- [x] ~~go-build~~ (no stdin)
- [x] ~~go-test~~ (no stdin)
- [x] ~~go-errcheck~~ (no stdin, wants a package)
- [x] ~~groovy~~ (no stdin)
- [x] haml
- [x] handlebars
- [x] ~~haskell-stack-ghc~~ (GHC can't read from standard input)
- [x] ~~haskell-ghc~~ (GHC can't read from standard input)
- [x] ~~haskell-hlint~~ (needs file extension to determine source format)
- [x] html-tidy
- [x] jade
- [x] javascript-jshint
- [x] javascript-eslint
- [x] ~~javascript-gjslint~~ (doesn't read from stdin)
- [x] javascript-jscs
- [x] javascript-standard
- [x] ~~json-jsonlint~~ (doesn't print errors when using standard input?!)
- [x] ~~less~~ (no stdin support)
- [x] luacheck
- [x] lua
- [x] perl
- [x] perl-perlcritic
- [x] php
- [x] ~~php-phpmd~~ (can't read from stdin)
- [x] php-phpcs
- [x] puppet-parser
- [x] ~~puppet-lint~~ (can't read from stdin)
- [x] python-flake8
- [x] ~~python-pylint~~ (doesn't read from stdin)
[...]

PR #768, opened 2015-10-26
Use stdin for sqlint checker

purcell:
See #767


purcell:
Filed this as a PR so that travis would run the tests for me. :-)

Will go ahead and merge now.

purcell:
(I updated the list in #767)

PR #776, opened 2015-11-01
Add faces for non-zero error / warning counts in the mode line

purcell:
I've been working on this change, but while the face properties are added to (at least) flycheck's mode line string, the colours don't show up in the modeline even when I customize the faces to use bright colours. Can you see what I'm doing wrong?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -831,6 +831,18 @@ This variable is a normal hook.  See Info node `(elisp)Hooks'."
   :package-version '(flycheck . "0.15")
   :group 'flycheck-faces)
 
+(defface flycheck-mode-line-errors
+  '((t :inherit mode-line))
+  "Flycheck face to highlight a non-zero error count in the mode line."
+  :package-version '(flycheck . "0.25")
+  :group 'flycheck-faces)
+
+(defface flycheck-mode-line-warnings
+  '((t :inherit mode-line))
+  "Flycheck face to highlight a non-zero warning count in the mode line."
+  :package-version '(flycheck . "0.25")
+  :group 'flycheck-faces)
+
 (defvar flycheck-command-map
   (let ((map (make-sparse-keymap)))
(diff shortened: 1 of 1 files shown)


purcell:
Yep, absolutely reasonable. :-)

lunaryorn:
@purcell I do not know.  If I had to guess I'd say that Emacs overrides the `face` of minor mode entries in the minor mode list, but I really don't know, not even where to look.  At least, I've never seen a coloured minor mode lighter in the minor mode list, which I conveniently consider anecdotical evidence for my theory :joy:

In my old Emacs configuration I [had a custom mode line with colours](https://github.com/lunaryorn/.emacs.d/blob/0607578bfb9545c796fada0c06e2cf21155c5eab/lisp/lunaryorn-flycheck.el#L76) for Flycheck, but at its own place in `mode-line-format` and not as part of the minor mode list.   For debugging purposes you could try to add `flycheck-mode-line` to `global-mode-string` and check whether the colours appear at this place.  That'd at least tell you whether your code is wrong or whether Emacs ignores `face` properties for minor mode lighters.

----

Anyhow, I don't think that this is a feature I'd like to merge to Flycheck core.   I'm sorry but I'd prefer to keep the mode line lighter as simple as possible.  Not that I'm against this change in particular, but if I merge this one, I'd have to merge the next one for custom colours for the "no checker found" case, and the next one for a customisable mode string, and the next one…  I'd have no arguments left to prevent mode line customisation from escalating into a messy beast of customisation code.

I'd prefer if you would keep fancy coloured Flycheck lighter in your personal configuration, or publish it as a separate package, like https://github.com/liblit/flycheck-status-emoji.

lunaryorn:
Closing as `wontfix`, because I do not want to merge this feature.  But please let us continue the discussion about the implementation, I'm curious why it doesn't work :)

PR #777, opened 2015-11-02
Use stdin for perl-perlcritic checker

juiko:
perlcritic documentation (https://metacpan.org/pod/distribution/Perl-Critic/bin/perlcritic#USAGE-EXAMPLES) says that reading input from a file is the same as reading it from stdin, so i think this changes should be enough.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7121,8 +7121,8 @@ the `--severity' option to Perl Critic."
 See URL `https://metacpan.org/pod/Perl::Critic'."
   :command ("perlcritic" "--no-color" "--verbose" "%f/%l/%c/%s/%p/%m (%e)\n"
             (option "--severity" flycheck-perlcritic-severity nil
-                    flycheck-option-int)
-            source)
+                    flycheck-option-int))
+  :standard-input t
   :error-patterns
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -7121,8 +7121,8 @@ the `--severity' option to Perl Critic."
 See URL `https://metacpan.org/pod/Perl::Critic'."
   :command ("perlcritic" "--no-color" "--verbose" "%f/%l/%c/%s/%p/%m (%e)\n"
             (option "--severity" flycheck-perlcritic-severity nil
-                    flycheck-option-int)
-            source)
+                    flycheck-option-int))
+  :standard-input t
   :error-patterns
   ((info line-start
          (file-name) "/" line "/" column "/" (any "1") "/"
```
lunaryorn: What about the `(file-name)` here?  I think it's unlikely that it's going to be correct when it reads from standard input.

juiko:
Yeah i forgot to delete (file-name) :/, it is done now.

juiko:
I am not sure if the tests should be changed, i looked at them and they do not seem to care about the file name in the message from perlcritic.

Issue #506, opened 2014-10-25
Can't use python-pylint checker with custom major-mode


angrybacon:
For educationnal purposes, I'm writing an Emacs major mode for Python. First step for me is to underline unused imports, then add a shortcut to remove those unused imports.

``` elisp
(defun yapm-delete-unused-import ()
  (message "It works!"))


(defvar yapm-mode-map
  (let ((map (make-sparse-keymap)))
    (define-key map "[C-c d]" 'yapm-delete-unused-import)
    map)
  "Keymap for `yapm-mode'.")


    ;;;###autoload
(define-derived-mode yapm-mode python-mode "yapm"
  "Yet another Python mode.")

(add-hook 'yapm-mode-hook 'flycheck-mode)
(flycheck-select-checker python-pylint)


(provide 'yapm-mode)
```

I assume Flycheck doesn't know (by default) that he may use `python-piylint` to check `yapm-mode` buffers. So I tried <kbd>M-x</kbd> `flycheck-select-checker` <kbd>RET</kbd> `python-pylint`.

```
user-error: Can't use syntax checker python-pylint in this buffer
```

Syntax checking works fine in a `python-mode` buffer.

ackalker:
@lunaryorn Please document `flycheck-add-mode` (in the info file and perhaps even in the `README.md` file).
I was about to make a feature request along the line of "Please add support for checking in derived modes" when I happened to stumble across thisvery issue. It fits my needs precisely: enable Bash syntax checking in pkgbuild-mode.

lunaryorn:
@ackalker Indeed, the documentation for this function is missing.  Please file a separate issue that we can keep track of this flaw.

ackalker:
@lunaryorn done, see #869

PR #778, opened 2015-11-02
Add support for js2-jsx-mode

jacksonrayhamilton:
`js2-mode.el` recently added a derived mode with support for JSX (an extension to ECMAScript). But when trying to use a JavaScript checker in this mode, I got the error `user-error: Can't use syntax checker javascript-eslint in this buffer`. Using `(flycheck-add-mode 'javascript-eslint 'js2-jsx-mode)` fixed this, but since `js2-jsx-mode` is defined alongside `js2-mode`, I think it is just as deserving of a spot in the official list.

An alternative might be for Flycheck to use the checker of a parent mode by default. But I read [this](https://github.com/flycheck/flycheck/issues/506), and I also suppose it could cause issues if a child mode is too far removed from its parent.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6861,7 +6861,7 @@ See URL `http://www.jshint.com'."
   (lambda (errors)
     (flycheck-remove-error-file-names
      "stdin" (flycheck-dequalify-error-ids errors)))
-  :modes (js-mode js2-mode js3-mode)
+  :modes (js-mode js2-mode js2-jsx-mode js3-mode)
   :next-checkers ((warning . javascript-jscs)))
 
 (flycheck-def-option-var flycheck-eslint-rulesdir nil javascript-eslint
@@ -6908,7 +6908,7 @@ See URL `https://github.com/eslint/eslint'."
                                  (flycheck-error-message err))))
                         (flycheck-sanitize-errors errors))
                   errors)
-  :modes (js-mode js2-mode js3-mode)
+  :modes (js-mode js2-mode js2-jsx-mode js3-mode)
   :next-checkers ((warning . javascript-jscs)))
 
 (flycheck-def-config-file-var flycheck-gjslintrc javascript-gjslint ".gjslintrc"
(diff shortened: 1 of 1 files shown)


lunaryorn:
@jacksonrayhamilton Thank you, that is a nice thing.  I'll definitely add it, it's much better than the various kludges with Web Mode that float around the inter webs…

However, are you sure that all of these Javascript checkers actually support JSX?  I've checked standard, jscs and eslint, but there's still an [open issue for jsint](https://github.com/jshint/jshint/pull/2167), and I'm really unsure about gjslint.

jacksonrayhamilton:
Neither jshint nor gjslint support JSX. But I expect many `js2-jsx-mode` users will do a sweeping `(add-to-list 'auto-mode-alist '("\\.js\\'" . js2-jsx-mode))` for all of their JavaScript code, JSX or not. And that's not a bad idea, as JS is a subset of JSX.

I imagine that if we were strict about which modes the linters supported, we'd have to instruct users to do this:

```elisp
(defun js2-or-jsx-mode ()
  (cond
   ((use-jsx-p) (js2-jsx-mode))
   (t (js2-mode))))
(add-to-list 'auto-mode-alist '("\\.js\\'" . js2-or-jsx-mode))
```

Also, I forgot to add `js-jsx-mode`, the equivalent mode also recently added to Emacs master.

lunaryorn:
> Neither jshint nor gjslint support JSX. 

Then please remove JSX from their `:modes`.  Syntax checkers should not declare modes that they don't actually support fully.

> But I expect many js2-jsx-mode users will do a sweeping `(add-to-list 'auto-mode-alist '("\\.js\\'" . js2-jsx-mode))` for all of their JavaScript code, JSX or not.  And that's not a bad idea, as JS is a subset of JSX.

I'd argue that this is wrong, and in any case it's not a reason for Flycheck checkers to “lie” about what they modes they support.

`jshint` _will_ break if a buffer doesn't contain only the subset of JSX that it can handle.  And that alone is reason enough to _not_ advertise JSX in its `:modes`, regardless of what users could possibly add to their init file.

> Also, I forgot to add js-jsx-mode, the equivalent mode also recently added to Emacs master.

Would you mind to add this one, too?

lunaryorn:
@jacksonrayhamilton I've applied this change, sans the checkers that don't support JSX.

PR #779, opened 2015-11-04
Shut up byte compiler about unused var

issue-dispenser:
It doesn't seem like it's being used right now. It was annoying me when updating.


cpitclaudel:
Why not remove that line entirely instead?

issue-dispenser:
That could make more sense. I assumed that perhaps it was left in for later use.

lunaryorn:
I've removed the binding.  Thank you very much for reporting this, sorry for the flaw.

PR #783, opened 2015-11-07
(flycheck-fold-include-levels): Find include file first, then find next

xwl:
error in the same file.

gcc(e.g, 4.8.3, 4.9.1) sometimes seems not printing included files
immediately after error. Here is a small testcase:
	// foo.cpp
	#include "foo.h"
	int main(int argc, char *argv[]) {
	    CHECK(Hi::b);
	    xx;
	}

	// foo.h
	namespace Hi { int a; };
	#define CHECK(arg) arg

gcc gives below errors:
	In file included from foo.cpp:1:0:
	foo.cpp: In function ‘int main(int, char**)’:
	foo.cpp:6:11: error: ‘b’ is not a member of ‘Hi’
	     CHECK(Hi::b);
		   ^
	foo.h:5:20: note: in definition of macro ‘CHECK’
	 #define CHECK(arg) arg
			    ^
	foo.cpp:8:5: error: ‘xx’ was not declared in this scope
	     xx;
	     ^
flycheck will complain:
    Error from syntax checker c/c++-gcc: Undefined error level: nil

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3357,17 +3357,23 @@ included file."
     (while remaining-errors
       (let* ((current-error (pop remaining-errors)))
         (when (funcall sentinel current-error)
-          ;; We found an error denoting errors in the included file, so process
-          ;; all subsequent errors until an error has the current file name
-          ;; again, and find the most severe error level
+          ;; We found an error denoting errors in the included file:
+          ;; 1. process all subsequent errors until faulty include file is found
+          ;; 2. process again all subsequent errors until an error has the
+          ;;    current file name again
+          ;; 3. find the most severe error level
           (let ((current-filename (flycheck-error-filename current-error))
                 (current-level nil)
-                (faulty-include-filename (flycheck-error-filename
-                                          (car remaining-errors))))
-            (while (and remaining-errors
-                        (not (string= (flycheck-error-filename
-                                       (car remaining-errors))
-                                      current-filename)))
+                (faulty-include-filename nil)
+                (filename nil)
+                (done (null remaining-errors)))
+
+            (while (not done)
+              (setq filename (flycheck-error-filename (car remaining-errors)))
+              (unless faulty-include-filename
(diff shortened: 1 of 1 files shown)


lunaryorn:
@xwl Sorry, but I have a hard time to understand what this PR is supposed to fix.  Please edit your original post to fix the formatting and provide more details.

xwl:
Actually I explain in the commit message, the content here is auto formated by github.  Can you look at this commit message, which has the original format: https://github.com/xwl/flycheck/commit/72072a47f75750f48dc6c9aa73903eedfce7cf22

I give a testcase there, the error message generated by gcc can't be correctly parsed by flycheck.

flycheck.el
```diff
@@ -3357,17 +3357,25 @@ included file."
...
-                                          (car remaining-errors))))
+                                          (car remaining-errors)))
+                (faulty-include-filename-found nil)
+                (filename nil))
             (while (and remaining-errors
-                        (not (string= (flycheck-error-filename
-                                       (car remaining-errors))
-                                      current-filename)))
+                        (progn
+                          (setq filename (flycheck-error-filename (car remaining-errors)))
```
lunaryorn: Well, I do not really like this pattern of setting variables in conditions of `while`.  I know that it's somewhat common in Emacs Lisp, but do you think that we could refactor this to not change a variable here?  It's just easier to understand.

lunaryorn: But if it's too complex, I'll merge this code as well.

lunaryorn:
@xwl Ah, I see, thanks for explaining.

flycheck.el
xwl: It can be improved. I will update.

xwl:
i did a force push: https://github.com/xwl/flycheck/commit/bedf2986ecd4288a113f6c5d6a5eb5e820aadd48

Issue #784, opened 2015-11-11
C-c ? ! doesn't provide a default


jech:
`C-c ? ! RET` should describe the checker used in the current buffer.

cpitclaudel:
The relevant bit of code is this: 

```
(defun flycheck-describe-checker (checker)
  "Display the documentation of CHECKER.

CHECKER is a checker symbol.

Pop up a help buffer with the documentation of CHECKER."
  (interactive
   (let* ((checker (flycheck-checker-at-point))
          (enable-recursive-minibuffers t)
          (prompt (if (symbolp checker)
                      (format "Describe syntax checker (default %s): " checker)
                    "Describe syntax checker: "))
          (reply (read-flycheck-checker prompt)))
```

At the moment it uses the checker at point, if there's one. As a side note, maybe this code should pass `checker` to `read-flycheck-checker` as the default argument.

jech:
> At the moment it uses the checker at point, if there's one.

Right, that wasn't obvious to me.

The usage scenario is as follows -- I wanted to know if there's a checker
for Java (yeah, I know, I'm sorry).  So I opened /tmp/Foo.java, and typed
C-c ? ! RET, expecting to find out.

M-x flycheck-verify-setup is well documented, so I found out in the end,
but I was surprised by the behaviour of C-c ? ! RET.

lunaryorn:
@jech You are aware that we have a comprehensive [online documentation](http://www.flycheck.org/manual/latest) which includes a [list of supported languages](http://www.flycheck.org/manual/latest/Supported-languages.html#Supported-languages)? :wink:

The current behaviour of `flycheck-describe-checker` is modelled after `describe-function` and friends, which default to the symbol at point.  But I see no reason against picking the syntax checker of the current buffer as default, if the symbol at point is no syntax checker.

@jech @cpitclaudel What do you think?

cpitclaudel:
I just learnt about `flycheck-describe-checker` actually :) 
Defaulting to the checker of the current buffer if there's no checker at point seems quite fine to me. We should also make sure to pass that default to `read-flycheck-checker`, as otherwise it doesn't play well with `ido-ubiquitous`.

lunaryorn:
@cpitclaudel Would you take care of this?

cpitclaudel:
Sure! Btw, why does `checker-at-point` return 0 if there is no checker, instead of nil?

lunaryorn:
@cpitclaudel Does it? :wink:

[...]

PR #787, opened 2015-11-11
Fix 784

cpitclaudel:
Edit: Connects to #784

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1790,11 +1790,10 @@ nil otherwise."
 (defun flycheck-checker-at-point ()
   "Return the Flycheck checker found at or before point.
 
-Return 0 if there is no checker."
+Return nil if there is no checker."
   (let ((symbol (variable-at-point 'any-symbol)))
-    (if (and (symbolp symbol) (flycheck-valid-checker-p symbol))
-        symbol
-      0)))
+    (when (flycheck-valid-checker-p symbol)
+      symbol)))
 
 (defun flycheck-describe-checker (checker)
   "Display the documentation of CHECKER.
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -1803,14 +1802,15 @@ CHECKER is a checker symbol.
 
 Pop up a help buffer with the documentation of CHECKER."
   (interactive
-   (let* ((checker (flycheck-checker-at-point))
+   (let* ((checker-at-pt (flycheck-checker-at-point))
           (enable-recursive-minibuffers t)
```
lunaryorn: I wonder what I added this binding for?  Do we actually use recursive mini buffers here?!

flycheck.el
```diff
@@ -1803,14 +1802,15 @@ CHECKER is a checker symbol.
 
 Pop up a help buffer with the documentation of CHECKER."
   (interactive
-   (let* ((checker (flycheck-checker-at-point))
+   (let* ((checker-at-pt (flycheck-checker-at-point))
```
lunaryorn: I think this binding is redundant.  It's only used at one place, so we can just as well inline it there, can't we?

flycheck.el
```diff
@@ -1803,14 +1802,15 @@ CHECKER is a checker symbol.
...
-     (list (or reply checker))))
-  (if (or (null checker) (not (flycheck-valid-checker-p checker)))
+          (default (or checker-at-pt
+                       (ignore-errors (flycheck-get-checker-for-buffer))))
+          (prompt (if default
+                      (format "Describe syntax checker (default %s): " default)
+                    "Describe syntax checker: ")))
+     (list (read-flycheck-checker prompt default))))
+  (if (not (flycheck-valid-checker-p checker))
       (message "You didn't specify a Flycheck syntax checker.")
```
lunaryorn: While we are at it, shouldn't this be a `user-error`?  I think that's the better choice here, isn't it?

lunaryorn:
@cpitclaudel Great implementation, just a couple of minor nit-picks from myself :blush:  And, geez, I wonder what had come to my mind at some of this code… :flushed: :wink:

flycheck.el
cpitclaudel: Not sure; I don't really use recursive minibuffers myself :/

flycheck.el
cpitclaudel: Done!

flycheck.el
cpitclaudel: Yup. Updated it.

lunaryorn:
@cpitclaudel I've cherry-picked your changes into `master`.  Thanks :+1:

PR #792, opened 2015-11-16
Clean up dependency list

syohex:
cl-lib was bundled since Emacs 24.3.


lunaryorn:
@syohex Thank you for catching this :+1:

PR #806, opened 2015-11-24
Add Scribble checking

marsam:
Hi:
I've added a flycheck checker for [scribble](http://docs.racket-lang.org/scribble/)

Thanks in advanve.

### DIFF
--- a/doc/languages.texi
+++ b/doc/languages.texi
@@ -632,6 +632,13 @@ selects the linters to use.
 
 @end itemize
 
+@flyclanguage{Scribble}
+
+@itemize
+@item
+@flyc{scribble} (@uref{http://docs.racket-lang.org/scribble/,Scribble})
+@end itemize
+
 @flyclanguage{RPM Spec}
 
 @itemize
--- a/flycheck.el
+++ b/flycheck.el
@@ -222,6 +222,7 @@ attention to case differences."
     python-pylint
     python-pycompile
     r-lintr
+    scribble
     rpm-rpmlint
     rst-sphinx
     rst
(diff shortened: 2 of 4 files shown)


cpitclaudel:
Hi,

Thanks for this contribution! Is there a way to feed input directly to this checker's stdin (instead of using a file)? And can you confirm that this checker won't actually run code, but just do a syntax check?

Clément.

lunaryorn:
@marsam Did you get a chance to look into the question

marsam:
Hi:
Sorry for the late response
> Is there a way to feed input directly to this checker's stdin (instead of using a file)? 

not without a custom script

> And can you confirm that this checker won't actually run code, but just do a syntax check?

This checker works by converting the scribble to text, and by scribble nature evaluates code. I think this invalides this checker to be part of flycheck. Feel free to close this pull request.

cpitclaudel:
Thanks for the answers. @lunaryorn's policy on checkers is that they shouldn't run code not intended to be run at compile time (so executing a program to check for run-time exceptions is not a valid checker definition, for example). Macro expansion, on the other hand, is ok. Does this checker fall in the second category?

If not, and if you find this checker useful in your day to day operation, please do feel free to distribute it as a separate package, with the appropriate security warnings. In that case, a suggestion to the scribble people to include a syntax-only checking facility might be in appropriate, too :)

marsam:
Sorry for the delay. Scribble can be use only for prose, but can also [evaluate](http://docs.racket-lang.org/scribble/eval.html) code so I think falls in the first category.

> If not, and if you find this checker useful in your day to day operation, please do feel free to distribute it as a separate package, with the appropriate security warnings. In that case, a suggestion to the scribble people to include a syntax-only checking facility might be in appropriate, too :)

I'll make a separate package. Thanks for your patience!

Issue #312, opened 2014-01-19
Add working directory property


rejeep:
As discussed in https://github.com/flycheck/flycheck/issues/308, we should add a property called `:cwd` that will change the `default-directory` when running a checker. The property takes a single argument, which is a function that will be called **each** time the checker runs and it will return a string path.

I guess it makes to fallback and not change the default directory if the function returns `nil`.

Example:

```lisp
(flycheck-define-checker foo
  :cwd (lambda ()
         (f-traverse-upwards
          (lambda (path)
            (f-dir? (f-expand ".cask" path)))
          (f-parent (f-this-file)))))
```

Note that I have noticed that setting the `default-directory` for a sub-process did not always work unless you specified the absolute path with an ending slash. The function `f-full` does exactly this.

rbraunstein:
+1

s3bs:
+1, I'm using flycheck with a custom checker. I modified my flycheck sources such that it changes default-directory in the (let ... scope where the checker is executed. For me a static :cwd is sufficient at the moment. Also as discussed in #308 I added functionality to change the process-environment temporary to add environment variables needed by the checker, which are also static.

lunaryorn:
@s3bs I'm sorry that you have to go so far as to modify our sources, but currently this feature has a rather low priority on our backlog.  To me, #774 and #800 are much more important.  They are also much larger and complex than this one, so it'll take a long time before I get to this issue.

I'd love to see a PR for this feature :relaxed:

s3bs:
@lunaryorn I actually like to tinker once in a while with elisp code. If I'd know more elisp I'd give you a hand. I put my changes that work for me as a fork in my github repository in the cwd branch if that helps anyone in the meantime. I'm not quite sure what you mean with 'PR'?!

rbraunstein:
PR is short for pull request.  Check github help or google. You are pretty
close.

BernardNotarianni:
I did modify a checker to change directory using the `:predicate` property.
Any idea to improve this to make the working directory coming back to the original location?

```elisp
(with-eval-after-load "flycheck"
  (flycheck-define-checker elixir-mix
    "An Elixir syntax checker using the Elixir interpreter.
     See URL `http://elixir-lang.org/'."
    :command ("mix"
              "compile"
              source)
    :predicate is-mix-project-p
    :error-patterns
    ((error line-start "** (" (zero-or-more not-newline) ") "
[...]

Issue #594, opened 2015-02-26
invalid-function on :predicate in with-eval-after-load


nathankot:
Following [these guidelines](http://flycheck.readthedocs.org/en/latest/dev/extending.html#advanced-syntax-checker-definitions) I'm getting getting this error:

![screen shot 2015-02-26 at 5 21 00 pm](https://cloud.githubusercontent.com/assets/1400871/6386321/dd7df9d0-bddb-11e4-9edc-fd3722ed3b03.png)

With this code:

``` elisp
(flycheck-define-checker jsxhint
  "A JSX syntax and style checker based on JSXHint."
  :command ("jsxhint" source)
  :error-patterns ((error line-start (1+ nonl) ": line " line ", col " column ", " (message) line-end))
  :predicate (lambda ()
               (and (executable-find "jsxhint")
                    (buffer-file-name)
                    (string-match ".*\.jsx?$" (buffer-file-name))))
  :modes (web-mode))
```

Even when I swap `:predicate` out for the code on the docs I still get a similar error :disappointed: 

I'm using the latest flycheck from MELPA, wondering if I'm doing anything wrong?

lunaryorn:
@nathankot Your Emacs version?

nathankot:
GNU Emacs 24.4.1
ᐧ

On Thu, Feb 26, 2015 at 9:07 PM, Sebastian Wiesner <notifications@github.com

lunaryorn:
@nathankot Can please try to reproduce this error after `M-x toggle-debug-on-error`, and paste the entire stacktrace.  Also, please check whether the issue also occurs in `emacs -Q`.

nathankot:
Hey @lunaryorn here's the stacktrace:

```
Debugger entered--Lisp error: (invalid-function (function (lambda nil (and
(executable-find "jsxhint") (buffer-file-name) (string-match ".*.jsx?$"
(buffer-file-name))))))
  signal(invalid-function ((function (lambda nil (and (executable-find
"jsxhint") (buffer-file-name) (string-match ".*.jsx?$"
(buffer-file-name)))))))
  #[257 "\300 \210\301@A\"\207" [flycheck-report-failed-syntax-check
signal] 4 "\n\n(fn ERR)"]((invalid-function (function (lambda nil (and
(executable-find "jsxhint") (buffer-file-name) (string-match ".*.jsx?$"
(buffer-file-name)))))))
  funcall(#[257 "\300 \210\301@A\"\207"
[...]

Issue #298, opened 2014-01-10
flycheck has conflicts with tabbar-mode


ghost:
flycheck has conflicts with tabbar-mode, when save file disappears the others tabs and only shows current tab, blink each time when write code.

how fix this?

PR #813, opened 2015-11-30
Add current working directory [:cwd] property

s3bs:
This is a first iteration of trying to implement #312. It give the possibility to add :cwd to the flycheck-define-checker. :cwd can be a fixed path or a function that returns a absolute path. The function is called just before the checker. The default-directory is only changed in the lexical environment of the checker execution. 
In addition there is a :environment property, where one can define additional environment variables. 
ie ("var1=value1" "var2=value2")

Edit: Connects to #312


lunaryorn:
@s3bs Thank you.  Unfortunately that's only about a third of what's actually needed.  Particularly we need to “remember” the working directory and use it later when we expand relative filenames from the syntax checker output.   Otherwise Flycheck will fail to parse relative filenames in the syntax checker output.

Please do remove the `:environment` property from this pull request, by amending your commits and force-pushing to your branch.  It's a different and unrelated feature that needs to be added independently after discussion (for I'm not convinced that we need it).

lunaryorn:
See also https://github.com/flycheck/flycheck-haskell/issues/50#issuecomment-160329289.  Inviting @mrkkrp

mrkkrp:
@s3bs, Thank you for working on this. In fact, I also started to work on the issue, but I don't have anything to commit yet. I hope this PR will evolve to the point when it will be merged. Would be really great.

s3bs:
a) My use case for the :environment is that I don't start emacs from the command line, so it doesn't have the env variables set as you would have by a login shell. I need those variables to set the path for a license file.  ... I will removed this from the pull request

b) as for the implementation of the functionality for :cwd I would propose to attach a buffer to the process as it is done in async.el. With that we can pass the information of 'default-directory easily. I seen a comment in the code referencing #298 about side effects attaching a process to a buffer. From what I can see is that the current-buffer was used to be attached to the process and this caused problems. What would stop us from using a complete new buffer that will be hidden in the background with 'generate-new-buffer. Let me know if I missed something in regards to the side effects #298. 

@mrkkrp. Great that you are working on this as well. Lets try to not duplicate the effort here and code it twice.

lunaryorn:
@s3bs We track the source buffer in a process property already, but we definitely can't change the `default-directory` of the source buffer.  That's a no-go, for it's global side-effects alone.  We have to track our syntax checking working directory separately.  And I'm definitely not in favour of just `let` binding `default-directory` around the whole error parsing code.  I'd rather take this opportunity to get rid of this particular piece of global state in Flycheck, and explicitly pass the directory when we need it (e.g. to `expand-file-name`).

And since I'd like to avoid attaching ever more properties to the process I proposed the idea of a syntax checking context that keeps track of all of these properties in a single `cl-defstruct`.

----

As for 1., take a look at https://github.com/purcell/exec-path-from-shell which lets you copy environment variables from your shell.  But even if you don't want to use this package, I don't see why you could just call `setenv` for the variable in your Emacs configuration.

Thank you for removing it from this pull request.

s3bs:
@lunaryorn I was not talking about the default-directory of the source-buffer, hence why I referenced async.el as code example. They use 'generate-new-buffer, attach that new buffer to the process. I was proposing to set this new buffer's default-directory. 

I'll have a look into cl-defstruct. 

-- 
As for the environment variables, you should probably open a new issue. I'm not using set-env because I only want to temporarily modify my environment variables and not clutter the environment, which could potentially cause problems with other things.

lunaryorn:
@s3bs I'm not sure whether I understand you: Why do we need a fresh buffer to track the working directory?  Why can't we just attach the directory to a process property?

lunaryorn:
@s3bs As for environment variables, please do open a new issue, but I must admit that I'm very sceptical whether it's worth adding to Flycheck.

s3bs:
@lunaryorn I must have misunderstood you. I was not quite sure with what you meant with " I'd rather take this opportunity to get rid of this particular piece of global state in Flycheck, and explicitly pass the directory when we need it "

Inviting @rejeep to this discussion as he started #312

lunaryorn:
@s3bs I'm sorry, I was apparently not clear in my comments.

I'm less concerned with how we pass the working directory of the checker along with the process: We can just attach it to the process as a property.

However, after the process is finished we have to use this directory to expand file names at some places.  And instead of changing `default-directory` here, I'd prefer if we could explicitly give it as function argument where needed.

s3bs:
@lunaryorn I think I'm a bit lost in your code. One part that confuses me is `flycheck-error-filename`. It is probably some sort of closure, but I just don't get it at the moment how that one works. Loosing my elisp foo. :(

lunaryorn:
`flycheck-error-filename` is just an accessor for the `filename` slot of the `flycheck-error` struct—search the code for `cl-defstruct` to find the definition (unfortunately struct definitions are not indexed in the imenu) and read more about [Structures](http://www.gnu.org/software/emacs/manual/html_node/cl/Structures.html) in the manual :)  Likewise there's `flycheck-error-line` for the `line` slot, or `flycheck-error-level` for the `level` slot, etc.

Please don't hesitate to ask further questions.

s3bs:
@lunaryorn As I see it we need the information in which directory the checker is started (cwd), everywhere were a error struct is passed, since all filenames in errors can be relative to that calling path. Currently all this happens under a `with-current-buffer` and therefore we have the `default-directory` of the buffer that the syntax-checker is called from. 
I would propose the following solution:
1) when starting the process that executes the syntax checker, we create a buffer local variable in a format like `flycheck-<checkername>-context`, that contains the cwd. This way its a unique identifier and supports multiple checker in one buffer. Since all the error structs have the checker name and the buffer reference in them, we can easily access this information when the error filenames are processed. Once all the errors are process the buffer local variable is deleted.

lunaryorn:
Alternatively we can require that all paths in `flycheck-error` be absolute, and expand relative names right after parsing.   Then we'd only need the directory once.

s3bs:
I don't have a testcase were the checker produces relative file names in the error messages. Can someone proved such testcase.
@lunaryorn Is there a way to run travis-ci from the local repository before pushing it to github?

flycheck.el
```diff
@@ -4608,6 +4627,7 @@ symbols in the command."
           ;; example for such a conflict.
           (setq process (apply 'start-process (format "flycheck-%s" checker)
                                nil command))
+          ;; why is this using setf and not set-process-sentinal ??
```
lunaryorn: `setf` sets the value of a “generalized variable”.  Essentially it's just a consistent syntax to set many different types of variables, include process sentinels.  If you expand the macro you'll find that it actually expands to `(set-process-sentinel process …)`.

Whether you use generalized variables or not is ultimately a matter of taste; personally I prefer the consistency it gives.

If this answers your question, please remove this comment again :blush:

flycheck.el
```diff
@@ -4615,6 +4635,7 @@ symbols in the command."
           (process-put process 'flycheck-checker checker)
           (process-put process 'flycheck-callback callback)
           (process-put process 'flycheck-buffer (current-buffer))
+          (process-put process 'flycheck-cwd  default-directory)
```
lunaryorn: Would you mind to call this property `flycheck-default-directory` instead of `flycheck-cwd`?  It's just easier to understand if the property name preserves the analogy to `default-directory`.

flycheck.el
```diff
@@ -4315,8 +4329,11 @@ default `:verify' function of command checkers."
         (parser (or (plist-get properties :error-parser)
                     #'flycheck-parse-with-patterns))
         (predicate (plist-get properties :predicate))
-        (standard-input (plist-get properties :standard-input)))
-
+        (standard-input (plist-get properties :standard-input))
+        (cwd       (plist-get properties :cwd)))
+    ;; FIXME add more to the cwd checker
+    ;; (unless (stringp (car cwd))
+    ;;   (error "Current working directory must be a string"))
```
lunaryorn: What about this `FIXME`?

flycheck.el
```diff
@@ -5087,7 +5110,7 @@ tool, just like `compile' (\\[compile])."
 
 
 ;;; General error parsing for command checkers
-(defun flycheck-parse-output (output checker buffer)
+(defun flycheck-parse-output (output checker buffer)
```
lunaryorn: What is this change?!

lunaryorn: Please be careful to not introduce spurious whitespace changes.  Run `M-x delete-trailing-whitespace` on this buffer after editing, and take a look at `whitespace-cleanup-mode` or `ws-butler` to automate this process.

flycheck.el
```diff
@@ -4214,6 +4214,20 @@ universal prefix arg, and only the id with normal prefix arg."
...
+  "Function returning apropriate current working directory.
+
+Work out the working directory from which the CHECKER command
+will be called.  The default is the `default-directory' of the source-buffer
+from which flycheck is calling the checker.
+:cwd can be a list that contains a fixed path string or a function that will
+return a path.  The function needs to return an absolut path not a relative path."
+  (let ((cwd (flycheck-checker-get checker 'cwd)))
+    (or (and (functionp cwd) (funcall cwd))
+        (and (listp cwd) (stringp (car cwd)) (file-truename (car cwd)))
```
lunaryorn: Would you mind to explain the purpose of this line?  As far as I can see this expects a single element list containing a string.  But why a list, and not just a plain string right away? 

Also, what use case does a constant string as working directory address?

lunaryorn:
@s3bs Great work, this patch looks really good now! :+1:  There are a couple of minor issues on the diff, which I'd like you to address, but over all it's really a good implementation that I'll definitely merge!  Thank you very very much :+1: :heart_eyes: 

As for Travis CI, you can run the test suite locally with:

```console
$ cask install
$ test/run.el '(not (tag external-tool))'
```

This runs all tests that do not need specific syntax checker tools.  It's perfectly fine to rely on Travis CI, though :smiley: 

We also have a [Vagrant based VM](https://github.com/flycheck/flycheck-vm) that includes all syntax checker tools for a complete test, but that's too much for this change.  We'll run the complete tests before merging anyway, there's no need for you to go through the hassle of setting it up :relaxed:

flycheck.el
s3bs: thanks for beeing so detailed in your answers.

flycheck.el
s3bs: Agreed.

flycheck.el
s3bs: As in for the use case. I'm setting my working directory to a constant. I'm setting it to '~/temp' since my syntax checker creates new files everywhere it's executed and I don't want them to be all over the repository tree. 

I put it in a list for consistency reasons. As far as I can tell most of the other properties take lists as arguments, when defining a syntax checker. We could allow both as well if you'd like. (string and string in list)

flycheck.el
s3bs: ups. I need to still update this to check if the property is either a string (fixed path) or a function.

flycheck.el
lunaryorn: @s3bs I think that's a very obscure corner case, and I'm not sure whether it's reasonable to add an extra special case just for this purpose, the more since if you really need it you could always write a function that just returns the constant string.

I'd really prefer if we were able to remove this special case from this implementation, and only accept functions for the working directory.  That'd also somewhat simplify this code.

lunaryorn:
@s3bs I forgot one thing, sorry.  Could we rename the checker property from `:cwd` to `:default-directory` everywhere?  I think that's easier to understand; the docstring would essentially come down to "The value of `default-directory' when executing this syntax checker."  What do you think?

s3bs:
@lunaryorn yes, we can do that. It makes it more obvious and clearer. I will use `flycheck-default-directory` as variable names to not accidently assign the `default-directory`.

mrkkrp:
What is status of this PR? I see `blocked` label, what does this mean?

lunaryorn:
@mrkkrp It's blocked because we're still waiting for various remarks to be addressed.

s3bs:
Apologies for the delay, but I had been away over Christmas.

lunaryorn:
@s3bs It's all fine :) I hope you had good holidays :blush:  I'll review your changes.

mrkkrp:
ping (?)

s3bs:
@mrkkrp Until this gets merged and in case you have the time you could use the code and beta test and see if it works for your use case. Any feedback welcome.

aaronjensen:
Looking forward to this, @lunaryorn is there anything I could do to help get this merged?

lunaryorn:
@aaronjensen Test it :relaxed:

aaronjensen:
I tested with both a hard-coded directory and a function, both worked:

```elisp
  (defun mix-directory ()
    (locate-dominating-file default-directory "mix.exs"))
    (flycheck-define-checker elixir-dialyzer
      "Erlang syntax checker based on dialyzer."
      :command ("mix" "dialyzer")
      :default-directory mix-directory
      :error-patterns
      ((error line-start
              (file-name)
              ":"
              line
              ":"
              (message)
              line-end))
      :modes elixir-mode)
```

I also tested with a hard coded string. Both worked.

Some notes:

1. `file-name` in the `:error-patterns` is still relative to the file's directory whereas I would expect it to be relative to the new `default-directory`.
2. I tried using a `lambda` for `:default-directory` and that didn't work. I'm new to elisp, so maybe that was a silly thing to even try.

Other than those, this is awesome. I'm looking forward to this being in, it really helps w/ elixir and I'm sure other languages. Thanks @s3bs!

s3bs:
@aaronjensen Could you please put your example with the lambda function here so I could have a look at it. Cheers.

aaronjensen:
@s3bs I believe it was:

```elisp
(flycheck-define-checker elixir-dialyzer
      "Erlang syntax checker based on dialyzer."
      :command ("mix" "dialyzer")
      :default-directory (lambda () (locate-dominating-file default-directory "mix.exs"))
      :error-patterns
      ((error line-start
              (file-name)
              ":"
              line
              ":"
              (message)
              line-end))
      :modes elixir-mode)
```

I'm new to elisp though so there could be something very wrong w/ that. Thanks!

s3bs:
That lambda function looks good to me. You can always use M-x  flycheck-compile and run that on your buffer to see the raw/unparsed output of your checker.

aaronjensen:
The default directory is the directory of the file, the lambda does not appear to be invoked or its result is ignored.

```
-*- mode: compilation; default-directory: "~/Source/project/booking/lib/booking/" -*-
Compilation started at Mon Mar  7 20:09:23

mix dialyzer
[31m[1m** (Mix) The task "dialyzer" could not be found[0m

Compilation exited abnormally with code 1 at Mon Mar  7 20:09:23
```

s3bs:
@aaronjensen Looking at your lambda function and at the path "~/Source/project/booking/lib/booking/" your lambda function will return nil since there is no hierarchy level that is "mix.exs". I'm not sure why your hard coded function would have worked then.  The variable default-directory in your lambda function at this stage will return the path of you file minus the file name.

aaronjensen:
`~/Source/project/booking`

aaronjensen:
btw @s3bs the lambda is never called. I put a `message` in there and it is never invoked.

s3bs:
@aaronjensen I just tried that and it worked fine. Have you re-evaluated  flycheck-define-checker?  Can we take this to the flycheck gitter chat as I want to avoid cluttering this pull request with message debugging this issue.

aaronjensen:
Looks like this is actually a problem w/ using `flycheck-define-checker` inside `with-eval-after-load`. I dont' understand it, but naming the lambda works around it, I don't think this is a problem w/ flycheck or this branch.

lunaryorn:
@aaronjensen Ah, yes, that is a known issue, see #594.  I didn't consider it too important, but it appears to be a nuisance, so I'll try to fix it soon.

aaronjensen:
@lunaryorn ah, good to know. #emacs helped me figure out it was a macro expansion problem, I still don't quite understand the issue, but yes, a fix would be great, especially since configuring in `with-eval-after-load` is probably pretty common.

By the way, do you have a suggestion for @s3bs on my first comment:

> file-name in the :error-patterns is still relative to the file's directory whereas I would expect it to be relative to the new default-directory.

lunaryorn:
@aaronjensen No, unfortunately not.  I've still not reviewed and tested this patch.  Albeit being small, it's complex in its interactions with the rest of Flycheck, which makes testing somewhat time-consuming, and I simply didn't have the time to do that.

aaronjensen:
Got it, well let me know if there's more that I can do to help. I think that the `:error-patterns` issue should probably be a blocker.

purcell:
Anything I can do to help here? The change looks clean, the discussion here suggests it's close to being ready, and I see a wide variety of people excited about this change.

flycheck.el
```diff
@@ -4256,6 +4256,25 @@ universal prefix arg, and only the id with normal prefix arg."
...
+will be called.  The default is the `default-directory' of the source-buffer
+from which flycheck is calling the checker.
+:default-directory can be a string that contains a fixed directory path or a function
+that will return a path.  The function needs to return an absolut path not a
+relative path."
+  (let* ((cwd (flycheck-checker-get checker 'default-directory))
+         (default-directory (or (and (functionp cwd) (funcall cwd))
+                                (and (stringp cwd) (file-truename cwd))
+                                default-directory))
+         )
```
cpitclaudel: No need to put the close-parens `)` on a separate line here

flycheck.el
```diff
@@ -4357,8 +4385,8 @@ default `:verify' function of command checkers."
         (parser (or (plist-get properties :error-parser)
                     #'flycheck-parse-with-patterns))
         (predicate (plist-get properties :predicate))
-        (standard-input (plist-get properties :standard-input)))
-
+        (standard-input (plist-get properties :standard-input))
+        (cwd       (plist-get properties :default-directory)))
```
cpitclaudel: No need for the extra spaces

flycheck.el
```diff
@@ -4683,6 +4712,7 @@ and rely on Emacs' own buffering and chunking."
                (args (flycheck-checker-substituted-arguments checker))
                (command (funcall flycheck-command-wrapper-function
                                  (cons program args)))
+               (default-directory  (flycheck-default-directory-wrapper checker))
```
cpitclaudel: Same here, no need for extra spaces

flycheck.el
```diff
@@ -4706,6 +4736,7 @@ and rely on Emacs' own buffering and chunking."
           (process-put process 'flycheck-checker checker)
           (process-put process 'flycheck-callback callback)
           (process-put process 'flycheck-buffer (current-buffer))
+          (process-put process 'flycheck-default-directory  default-directory)
```
cpitclaudel: Same

flycheck.el
```diff
@@ -4807,12 +4839,12 @@ _EVENT is ignored."
...
                   files
-                  (flycheck-get-output process) callback)))
+                  (flycheck-get-output process) callback cwd)))
             ((debug error)
              (funcall callback 'errored (error-message-string err)))))))))
 
 (defun flycheck-finish-checker-process
-    (checker exit-status files output callback)
+    (checker exit-status files output callback cwd)
   "Finish a checker process from CHECKER with EXIT-STATUS.
```
cpitclaudel: Adding CWD to the docstring would be good

flycheck.el
```diff
@@ -5190,7 +5223,7 @@ CHECKER.  BUFFER is the buffer which was checked.
 Return the errors parsed with the error patterns of CHECKER."
   (funcall (flycheck-checker-get checker 'error-parser) output checker buffer))
 
-(defun flycheck-fix-error-filename (err buffer-files)
+(defun flycheck-fix-error-filename (err buffer-files cwd)
   "Fix the file name of ERR from BUFFER-FILES.
```
cpitclaudel: Same here, documenting CWD would be nice

flycheck.el
```diff
@@ -4256,6 +4256,25 @@ universal prefix arg, and only the id with normal prefix arg."
     (`(eval ,_) t)
     (_ nil)))
 
+(defun flycheck-default-directory-wrapper (checker)
+  "Function returning apropriate current working directory.
+
+Work out the working directory from which the CHECKER command
+will be called.  The default is the `default-directory' of the source-buffer
+from which flycheck is calling the checker.
+:default-directory can be a string that contains a fixed directory path or a function
```
cpitclaudel: This line could be re-wrapped :)

cpitclaudel:
I've added a few stylistic comments to the patch, which otherwise looks very nice. @aaronjensen raised a concern with `:error-patterns`, but I'm not sure I fully understand it.

aaronjensen:
> @aaronjensen raised a concern with :error-patterns, but I'm not sure I fully understand it.

The issue is that the path for a file generally is relative to the default-directory, but flycheck, with this patch, does not understand that.  In other words, say you have a project like:

```
.
├── acronym
│   ├── README.md
│   ├── acronym.exs
│   └── acronym_test.exs
├── anagram
    ├── README.md
    ├── anagram.exs
    └── anagram_test.exs
```

If the default-directory for `acronym.exs` gets set to whatever `.` is, then the error tool may report errors as having occurred in `acronym/acronym.exs`. However, flycheck will expect `file-name` to be `acronym.exs` and it will not associate errors from `acronym/acronym.exs` with the appropriate file.

Here is an example checker:

```
(flycheck-define-checker elixir-dialyzer
    "Erlang syntax checker based on dialyzer."
    :command ("mix" "dialyzer")
    :default-directory (lambda () (locate-dominating-file default-directory "mix.exs"))
    :predicate
    (lambda ()
      (and
       (buffer-file-name)
       (file-exists-p "mix.exs")
       (file-exists-p "deps/dialyxir")
       (file-exists-p ".local.plt")))
    :error-patterns
    ((error line-start
            (file-name)
            ":"
            line
            ":"
            (message)
            line-end))
    :modes elixir-mode)
```

and some sample output:

```
dialyzer --no_check_plt --plt .local.plt -Wno_opaque -Wno_return -Wunderspecs -Wunknown -Wrace_conditions --fullpath /Users/aaronjensen/Source/curadora/booking/_build/dev/lib/booking/ebin
  Compiling some key modules to native code... done in 0m0.26s
  Proceeding with analysis...
web/partner/request/hotel_details.ex:167: The call erlang:'+'(3,x@1::'atom') will never return since it differs in the 2nd argument from the success typing arguments: (number(),number())
web/partner/request/hotel_details.ex:171: The call 'Elixir.Juniper.Request.HotelDetails':foo('atom') will never return since it differs in the 1st argument from the success typing arguments: (number())
```

If my `default-directory` is *manually* set via `setq` to the directory containing the `web` directory, this works fine. However, if it is set with this new feature, it does not. It expects the `file-name` to be only `hotel_details.ex`

I hope this clears things up some, thanks!

lunaryorn:
@aaronjensen The problem is that `flycheck-fill-and-expand-error-file-names` in `flycheck-finish-current-syntax-check` isn't called with the proper working directory, but as things stand this is somewhat hard to add because Flycheck doesn't know about the working directory anymore at this point.

@purcell I'm sorry for everyone that's excited about this change, but I'm sure they'd be much less excited about a broken Flycheck. 

This change interacts with many parts of Flycheck, and what's worse, using the wrong directory for relative file names could just make Flycheck silently throw away valid errors for the current buffer, which is obviously a no-go.

That's why I want to test this change myself to make sure that it really works as intended.  I haven't had the time for this yet.

And then there's the issue discovered by @aaronjensen which @s3bs would need to fix first.  I don't have the time to do that myself.

lunaryorn:
Closing in favour of #973.  Thanks @sergv

Issue #648, opened 2015-05-26
cannot open a file when a long flycheck message is displayed


soonhokong:
Reproduce Steps
------------------------

1. Set up an example checker for text-mode:
    ```elisp
(flycheck-define-checker example-checker
                         "Example Checker"
                         :command ("echo" "1:1:"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\n"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\n"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\n"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
                                   "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
[...]

PR #822, opened 2015-12-16
Never show error buffer in currently active window

cpitclaudel:
Investigating #648 led me to another small bug, I think. This should fix it, but since I'm not too familiar with window configuration stuff I'd like a second opinion :)

In short, the error message buffer is sometimes opened in the current window, masking the source code buffer. To reproduce:

* Follow the instructions in #648 (including the checker definition) and open test1.txt
* Wait for the error buffer to pop up
* Move to the window displaying the error buffer with `C-x o`
* Display test1.txt in that buffer as well (using `C-x b`)
* Wait for the error buffer to pop up again.

On my machine, without the patch, the error buffer pops up and replaces the source buffer that the point was in:

bug

Btw, how do you record good gifs?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4135,7 +4135,7 @@ In the latter case, show messages in
   (when (and errors (flycheck-may-use-echo-area-p))
     (let ((messages (seq-map #'flycheck-error-format-message-and-id errors)))
       (display-message-or-buffer (string-join messages "\n\n")
-                                 flycheck-error-message-buffer))))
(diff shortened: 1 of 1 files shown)


lunaryorn:
I'll take a look.

lunaryorn:
Looks good to me, even though the meaning of this argument is pretty obscure :confused:  I've cherry-picked the commit to `master`, and deleted the branch.

Thank you very much for fixing all these glitches in with the error message buffer, that's really awesome work! :heart:

PR #825, opened 2015-12-20
Add a flycheck-jshint-extract option.

tealeg:
This option sets the value passed to --extract on execution of jshint,
allowing it to extract javascript embedded in an HTML file and therefore
avoid spurious linting errors on surrounding HTML elments.


lunaryorn:
Thank you for this patch.  But how would you use this option?  Flycheck does not use jshint in HTML files by default.

tealeg:
@lunaryorn I use this via [multi-web-mode](https://github.com/fgallina/multi-web-mode)  - but essentially any of the modes that allow multiple major modes when editing an HTML file with embedded javascript would do this.  Using mulit-web-mode, when I move the point into an area of the file bounded by `<script>` and `</script>` then js2-mode is invoked exactly as if it were a .js file.  The jshint output will then warn about the html tags (which are obviously invalid JS).  Setting this option to either `auto` or `always` cures that problem.

flycheck.el
```diff
@@ -6900,6 +6900,22 @@ See URL `http://jade-lang.com'."
...
 
+(flycheck-def-option-var flycheck-jshint-extract nil javascript-jshint
+  "Whether jshint should extract Javascript from HTML before linting.
+
+The value of this variable is one of 'auto', 'always' or 'never'.
+
+Refer to the jshint manual at the URL
+  `http://jshint.com/docs/cli/#flags'
+for more information."
+  :type '(choice (const :tag "No extraction rule" nil)
```
lunaryorn: @tealeg What would happen in this case, i.e. if no rule was specified?

lunaryorn:
@tealeg Ah, I see.  I didn't know that such a thing existed for Emacs :relaxed:

I'll merge this PR, but I'm unsure about the choices offered by the option.  Would you mind to answer on my comment I left in the diff?

flycheck.el
lunaryorn: @tealeg Sorry, did you miss my comment?  I would like to understand how this option works before I merge this pull request :relaxed:

tealeg: @lunaryorn sorry I didn't see it.  In this case it's the same as passing "never" which is the default.

tealeg:
@lunaryorn I've answered the comment now, sorry for the delay!

lunaryorn:
@tealeg Merged, thank you for this contribution.

cwhatley:
FYI, this broke for me when I updated. See: https://github.com/flycheck/flycheck/issues/871

lunaryorn:
@cwhatley Sorry, my bad :disappointed:

ShadyAlfred:
Is it possible to be used with web-mode?

PR #831, opened 2015-12-30
Make go-build compile only current file

Gnouc:
Currently, go-build checker treats all files in them same directory of current files as a Go project, `go-build` checker will fail to run if any C file present in directory:
```
$ ls
test.c test.go
$ go build -o /dev/null
can't load package: package .: C source files not allowed when not using cgo or SWIG: test.c
```

This pull request made `go-build` only compile the current file.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6502,7 +6502,11 @@ See URL `http://golang.org/cmd/go'."
             (option-flag "-i" flycheck-go-build-install-deps)
             ;; multiple tags are listed as "dev debug ..."
             (option-list "-tags=" flycheck-go-build-tags concat)
-            "-o" temporary-file-name)
+            "-o" temporary-file-name
+            ;; Only compile current file, go-build failed if if current
+            ;; directory contains any C file.
+            ;; See https://github.com/dominikh/go-mode.el/issues/19
+            source)
   :error-patterns
   ((error line-start (file-name) ":" line ":"
           (optional column ":") " "


flycheck.el
```diff
@@ -6502,7 +6502,11 @@ See URL `http://golang.org/cmd/go'."
             (option-flag "-i" flycheck-go-build-install-deps)
             ;; multiple tags are listed as "dev debug ..."
             (option-list "-tags=" flycheck-go-build-tags concat)
-            "-o" temporary-file-name)
+            "-o" temporary-file-name
+            ;; Only compile current file, go-build failed if if current
+            ;; directory contains any C file.
+            ;; See https://github.com/dominikh/go-mode.el/issues/19
```
lunaryorn: How's this pull request related to https://github.com/dominikh/go-mode.el/issues/19?

lunaryorn:
Thank for this patch.  I'm unsure though whether it's a good idea to apply this change to Flycheck.  If I remember correctly there was a reason for us to compile the whole project.

Please give us a bit more background information about this change.  In which situations does a Go project even contain C files?  How would you normally compile a Go project that includes C files?  Can't we do that for Flycheck as well, i.e. compile in a way that includes the C files?

lunaryorn:
I'm inviting @gfrey and @ptrv because the contributed a lot to these syntax checkers.  Specifically @ptrv contributed the original versions in #107 which as far as I can see already compiled the whole Go project.

In some sense only compiling the current file now would probably be a regression in Flycheck's behaviour.

ptrv:
I am not actively using go for a while now, so I am not up to the latest changes in the go toolchain. I am also not sure why we compiled the package. Do we get different error messages when we compile only the current file compared to compiling the whole package. From a naive perspective I would guess that compiling only the current file is a desired behavior also in terms of performance/compile-time.

gfrey:
Compiling single files is not sufficient, as something like the following is perfectly valid in go (as long as the files are in the same package):

file `a.go`
~~~go
package foo
import "fmt"
func a() {
  fmt.Println("a")
  b()
}
~~~

file `b.go`
~~~go
package foo
import "fmt"
func b() {
  fmt.Println("b")
}
~~~

Compiling only `a.go` will give the following error
~~~
user@dev:~/foo|⇒  go build -o /dev/null a.go
# command-line-arguments
./a.go:7: undefined: b
~~~

I've no solution for the OPs cgo problem, but restricting compilation to all `.go` files (aka `go build *.go`) could work.

lunaryorn:
@gfrey Thank you.  I'm closing this pull request then, as it'd break the current behaviour of Flycheck.

By the way, how come that `-o /dev/null` now works with `go build`?  I remember that it used to be broken.  When did they fix that?

@Gnouc I am sorry, but we need to come up with a different fix for your issue.

lunaryorn:
@Gnouc If you can come up with a different solution that does not break in the situation illustrated by @gfrey I would be very happy to re-open this pull request or to review a new pull request :relaxed:

Gnouc:
@lunaryorn Yeah, @gfrey pointed out the good point, compiling all `.go` files is better. I will go with new solution.

lunaryorn:
@Gnouc Please hold on.  Given what you wrote in #832 I am not sure that this really needs to be fixed in Flycheck.

If you can find a _simple_ solution, I'd merge it, but anything that goes so far as to enumerate all Go files in the current directory and pass them to `go build` is too complex for what is—as far as I understand it—essentially an incorrectly setup Go environment on your side, in my opinion.

PR #848, opened 2016-01-09
Fix haskell-module-re to exclude following paren

fizbin:
It is legal and not overly uncommon in haskell to start a module with:

    module Foo.Bar(
      quux,
      baz) where

That is, the module name can be terminated with an open paren as well as with a space.


lunaryorn:
Oh what a blunder of mine; thank you very much for finding and fixing this.  😔 ☺️

Would you mind to add a test case for this to `flycheck-test.el`?  It's totally fine if you don't want to; I'll add it then.

lunaryorn:
@fizbin I've merged your patch.  Thank you very much! :+1:

Issue #794, opened 2015-11-16
flycheck hangs with eslint


brunoqc:
I have 80 errors in my file if I run eslint from the command line. I tested with a smaller file (with less errors) and it was fine. Emacs has 0% cpu usage and I don't see eslint in the process manager. I think it's related to eslint since the file opens if I remove eslint from my PATH.

There's nothing in the message buffer.

GNU Emacs 24.5.1 (x86_64-w64-mingw32) of 2015-05-16 on KAEL

lunaryorn:
Please tell us your Flycheck version `M-x flycheck-version` and your eslint version, and please check whether you are able to reproduce the issue in `emacs -Q`.

brunoqc:
Flycheck version: 0.25snapshot (package: 20151113.858)
eslint v1.9.0
node v5.0.0

I can't reproduce with `emacs -Q`.

lunaryorn:
In other words, if you start `emacs -Q`, enable `global-flycheck-mode` and visit the affected file Flycheck works fine and highlights all issues in the file?

brunoqc:
Sorry I didn't realise you wanted me to load flycheck manually.

emacs hangs when I visit the affected file after loading flycheck manually like this :

```lisp
(add-to-list 'load-path "~/.emacs.d/elpa/dash-20151021.113")
(add-to-list 'load-path "~/.emacs.d/elpa/flycheck-20151113.858")
(require 'flycheck)
(global-flycheck-mode)
```

lunaryorn:
@brunoqc Uhm, yes, I wanted you to actually try and see whether Flycheck works in `emacs -Q`.  Sorry for the confusion.

Can you given me the file which shows this issue?  I'd like to try myself.  Alternatively, can you try to reproduce the problem on a Unix system?  That's somewhat crucial, for otherwise I'm unable to debug and fix the issue.  If it's a Windows-only problem, you'll have to debug that yourself.

Besides, you're not using the latest Flycheck build from MELPA, please consider updating.

brunoqc:
I have the problem with https://ajax.googleapis.com/ajax/libs/angularjs/1.4.7/angular.js . I also tested with an up-to-date flycheck.

I wasn't able to reproduce on Debian.
```
Debian GNU/Linux 8.2 (jessie)
GNU Emacs 24.5.1 (x86_64-pc-linux-gnu, GTK+ Version 3.16.4) of 2015-06-28 on trouble, modified by Debian
[...]

PR #850, opened 2016-01-11
Fix for 794 (chunking input on Windows)

cpitclaudel:
Here's the PR :)

Edit: Connects to #794

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4606,6 +4606,55 @@ symbols in the command."
          (seq-map (lambda (arg) (flycheck-substitute-argument arg checker))
                   (flycheck-checker-arguments checker))))
 
+(defun flycheck--process-send-buffer-contents-chunked (process)
+  "Send contents of current buffer to PROCESS in small batches.
+
+Send the entire buffer to the standard input of PROCESS in chunks
+of 4096 characters.  Chunking is done in Emacs Lisp, hence this
+function is probably far less efficient than
+`send-process-region'.  Use only when required."
+  (let ((from (point-min)))
+    (while (< from (point-max))
+      (let ((to (min (+ from 4096) (point-max))))
+        (process-send-region process from to)
+        (setq from to)))))
+
+(defvar flycheck-chunked-process-input
+  ;; Chunk process output on Windows to work around
+  ;; https://github.com/flycheck/flycheck/issues/794 and
+  ;; https://debbugs.gnu.org/cgi/bugreport.cgi?bug=22344.  The presence of
+  ;; `w32-pipe-buffer-size' denotes an Emacs version (> Emacs 25.1 )where pipe
+  ;; writes on Windows are fixed.
+  ;;
+  ;; TODO: Remove option and chunking when dropping Emacs 24 support, see
+  ;; https://github.com/flycheck/flycheck/issues/856
+  (and (eq system-type 'windows-nt) (not (boundp 'w32-pipe-buffer-size)))
+  "If non-nil send process input in small chunks.
+
+If this variable is non-nil `flycheck-process-send-buffer' sends
+buffer contents in small chunks.
+
+Defaults to nil, except on Windows to work around Emacs bug
+#22344.")
+
+(defun flycheck-process-send-buffer (process)
+  "Send all contents of current buffer to PROCESS.
+
+Sends all contents of the current buffer to the standard input of
+PROCESS, and terminates standard input with EOF.
+
+If `flycheck-chunked-process-input' is non-nil, send buffer
+contents in chunks via
+`flycheck--process-send-buffer-contents-chunked', which see.
+Otherwise use `process-send-region' to send all contents at once
+and rely on Emacs' own buffering and chunking."
+  (save-restriction
+    (widen)
+    (if flycheck-chunked-process-input
+        (flycheck--process-send-buffer-contents-chunked process)
+      (process-send-region process (point-min) (point-max))))
+  (process-send-eof process))
+
 (defun flycheck-start-command-checker (checker callback)
   "Start a command CHECKER with CALLBACK."
   (let (process)
@@ -4643,10 +4692,7 @@ symbols in the command."
           (setq flycheck-temporaries nil)
           ;; Send the buffer to the process on standard input, if enabled
           (when (flycheck-checker-get checker 'standard-input)
-            (save-restriction
-              (widen)
-              (process-send-region process (point-min) (point-max))
-              (process-send-eof process)))
+            (flycheck-process-send-buffer process))
           ;; Return the process
           process)
       (error
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,6 +47,8 @@ now for questions and discussions about development. [GH-820]
   * Fix checkdoc failure with some Emacs Lisp syntax [GH-833] [GH-845]
   * Correctly parse Haskell module name with exports right after the module name
     [GH-848]
+  * Don’t hang when sending buffers to node.js processes on Windows [GH-794]
+    [GH-850]
 
 [buttercup]: https://github.com/jorgenschaefer/emacs-buttercup
 


flycheck.el
```diff
@@ -4646,10 +4668,8 @@ symbols in the command."
           (setq flycheck-temporaries nil)
           ;; Send the buffer to the process on standard input, if enabled
           (when (flycheck-checker-get checker 'standard-input)
-            (save-restriction
-              (widen)
-              (process-send-region process (point-min) (point-max))
-              (process-send-eof process)))
+            (flycheck--process-send-buffer process)
+            (process-send-eof process))
```
lunaryorn: I think I'd prefer to make `--process-send-buffer` send the EOF as well.  Let's have everything at just one place :)

flycheck.el
```diff
@@ -4609,6 +4609,28 @@ symbols in the command."
...
+the threshold above which Emacs just hangs).
+
+See https://debbugs.gnu.org/cgi/bugreport.cgi?bug=22344 for more
+information about the Emacs issue, and see the original issue at
+https://github.com/flycheck/flycheck/issues/794 for information
+about the impact on Flycheck."
+  (save-restriction
+    (widen)
+    (if (eq system-type 'windows-nt)
+        (let ((from (point-min)))
```
lunaryorn: I think we should move all of this into a separate function, e.g. `flycheck--process-send-buffer-contents-chunked`.  The function name will be easier to understand than the code itself; it'll help us to understand what it's for in the future.

Furthermore, I think we should move the `system-type` test out of this function into a `defvar`, e.g. `flycheck-chunked-process-input` or so, that holds the value of the `system-type` test.  This lets users easily switch between both, which will help us to test the code on systems other than Windows (e.g. to fix a bug therein), and users to switch if we ask them to, e.g. to check whether their version of Emacs or Node.js is still affected by the bug.

What do you think?

lunaryorn: PS: I don't think that we should have a proper `defcustom` for this, for most users should never ever need to change it, or would want to.  A simple `defvar` should be sufficient, in my opinion.

cpitclaudel: I think your code reviews are great, as usual. Will update the PR.

cpitclaudel:
Done :)

lunaryorn:
@cpitclaudel Thanks.  I pushed a commit that shortens the docstrings.  

However, following the discussion in the corresponding Emacs bug, I'm not sure whether I still like to default to chunking generally.  We probably have to since we can't reasonably choose more specifically, but I do not like to degrade performance for all Windows users that aren't affect.  Python, for instance, seems to be fine.

Did you check whether there's a node.js issue?  I didn't find any.  Should we probably report it to node.js, too?

cpitclaudel:
> However, following the discussion in the corresponding Emacs bug, I'm not sure whether I still like to default to chunking generally. We probably have to since we can't reasonably choose more specifically, but I do not like to degrade performance for all Windows users that aren't affect. Python, for instance, seems to be fine.

Yeah, I'm not sure either. I don't really like this kind of bugs. I don't know if there's a way to detect this situation, either. I don't want this small windows tweak to start rippling through the codebase.

I can, however, test out the performance of that thing. Maybe it's not that bad.

> Did you check whether there's a node.js issue? I didn't find any. Should we probably report it to node.js, too?

I didn't find anything either :/ A report would probably be in order, but I'm not sure if now is the best time. We might want to wait until things are a bit clearer on the Emacs side; then (or if things are not making progress) we can probably ping the node.js people. What do you think? We could also ask them right now.

lunaryorn:
>  I don't know if there's a way to detect this situation, either. I don't want this small windows tweak to start rippling through the codebase though, so I won't champion per-checker settings :)

Yes, definitely not.  That'd go out of hand.

But I'm afraid that this kind of a hack will stick forever in our code because none ever looks at it again.  I don't want to end up there.

Hence, I think we should make it the default on Windows now, but add a TODO with a date to re-evaluate the situation in January 2017.  By that time it should have been reported to node.js and fixed.  If it's not, I'd just remove the fix again, and simply declare node.js on Windows broken.  That'd at least put up pressure on anyone with Windows and Node.js experience to properly track down and fix the issue, and we can't live with workarounds forever.

> I can, however, test out the performance of that thing. Maybe it's not that bad.

That'd be great :+1: 

> A report would probably be in order, but I'm not sure if now is the best time. We might want to wait until things are a bit clearer on the Emacs side; then (or if things are not making progress) we can probably ping the node.js people. What do you think? We could also ask them right now.

I think it's better to report early.   We'll probably get node.js people to add their opinion and research to the corresponding Emacs bug, or probably discover a bug in their own code.

Can you do that?  I don't have a Windows system to test anything on; I'm afraid I'm of little help in this issue :disappointed:

lunaryorn:
@cpitclaudel Given that the issue appears to have been fixed in Emacs 25 now, we can safely merge this workaround as there's a clear path for obsoleting and removing it again, namely as part of our [Emacs 25.1 cleanup milestone](https://github.com/flycheck/flycheck/milestones/Emacs%2025.1).

Based on [Eli's commit](http://git.savannah.gnu.org/cgit/emacs.git/commit/?h=emacs-25&id=58a622d473112f8ff5b4bdb3e49bc6573dfd3404) I'd adapt the condition for enabling the workaround though, in order to skip it if the issue is fixed upstream.

flycheck.el
```diff
@@ -4609,6 +4609,49 @@ symbols in the command."
...
+  (let ((from (point-min)))
+    (while (< from (point-max))
+      (let ((to (min (+ from 4096) (point-max))))
+        (process-send-region process from to)
+        (setq from to)))))
+
+(defvar flycheck-chunked-process-input
+  ;; See https://github.com/flycheck/flycheck/issues/794 and
+  ;; https://debbugs.gnu.org/cgi/bugreport.cgi?bug=22344
+  (eq system-type 'windows-nt)
```
lunaryorn: @cpitclaudel Following [Eli's commit](http://git.savannah.gnu.org/cgit/emacs.git/commit/?h=emacs-25&id=58a622d473112f8ff5b4bdb3e49bc6573dfd3404) that fixed the issue properly we should update the condition to not enable our work around on Emacs 25.1.

The commit appears to introduce a new `w32-pipe-buffer-size` option, which we could use as feature test for this fix, and thus change the condition to `(and (eq system-type 'windows-nt) (not (boundp 'w32-pipe-buffer-size)))`.

What do you think?  

Besides, what about cygwin?  Did any of the reporters of #794 observe the issue on cygwin?

cpitclaudel: This sounds great; feel free to go ahead with this if you have time before I do. I just tried on Cygwin quickly, but node.js was getting confused by cygwin's path format (`/cygdrive/c/...`).

cpitclaudel:
I opened an issue with Node.js at https://github.com/nodejs/node/issues/4697

lunaryorn:
@cpitclaudel I merged the fix into `master`, with modifications as discussed.

May I ask you to test the next MELPA build on Windows in a broken and a fixed Emacs version, and check whether the issue is gone, and whether `flycheck-chunked-process-input` is really `nil` on a fixed build of Emacs?  That'd be really awesome :blush:

I opened #856 for us to track the removal of the workaround once we drop support for Emacs 24, in our Emacs 25 cleanup milestone.

cpitclaudel:
Will do :)

cpitclaudel:
All checks passed :)

* ✓ Reproduced the bug in Emacs 24.5 before updating Flycheck
* ✓ Updated Flycheck and confirmed that the bug was gone and that flycheck-chunked-process-input was t there
* ✓ Made sure that the bug was gone in Emacs 25 and that flycheck-chunked-process-input was nil there

My, that was a tricky bug.

lunaryorn:
@cpitclaudel Thank you!  That's great news!  You did awesome work on tracking down and fixing this bug!  Thank you very much; Flycheck and I owe you a lot!

cpitclaudel:
:)

PR #858, opened 2016-01-22
Update flycheck.texi

cmarqu:
Fix wrong references to pylint and JavaScript that should be Ruby references.


lunaryorn:
@cmarqu Thank you very much, merged :relaxed:

Issue #859, opened 2016-01-22
Migrate to buttercup


swsnr:
Migrate all ERT unit tests to buttercup specs.  Keep all integration tests, specifically for syntax checkers, as ERT tests, but split them up and really denote them as integration tests.
# Motiviation

Our ERT test suite has gone out of hand, and mixes simple unit tests with large integration tests. 

Use buttercup because it's much nicer and sharpen the distinction between unit tests (which should run on Travis CI) and integration tests, which only run manually and occasionally.  Require the former, but be generous with the latter.  

Specifically, we shouldn't bother to demand good test coverage for syntax checkers, and we shouldn't bother to update them that regularly.  Integration tests may fail, if there's a reason for them to fail.
## Guidelines
- New specs in `test/specs`, to be run by `rake test:specs`.
- One file per suite.
- Keep suites small and focused, better split too much than too little.
- Care for a nice and clean structure.  Use sub-directories if necessary.
- Care for readable specs.  Ideally each spec only has a single matcher.  
- Avoid complicated tests.  Make generous use of setup and teardown features, and prefer to write compound matcher in `flycheck-buttercup.el` over making a lot of expectations in a single suite.
- Liberally use spies and mocks to avoid complicated environment setup.
## Migration path

We write no new ERT tests.  Whenever a component of Flycheck is touched and needs new tests or modifications of existing tests, first port the existing tests to buttercup before making the change.

lunaryorn:
@flycheck/maintainers 

Your opinion, specifically on the last paragraph?  I think that maintaining comprehensive tests for all syntax checkers is a lot of effort for little to no gain, the more since we can't run them on Travis CI anyway—the setup is just too expensive.

In future, I'd just like to have a little test per checker, and don't bother to make sure that it's actually supported on the testing VM, which I'd slowly retire by then.  The tests would then only serve to support larger refactorings, or for occasional debugging of reported issues, but not to proactively catch issues with syntax checker definitions.

What do you think?

cpitclaudel:
I think it's a good idea. It's hard to catch issues with linter updates, for example, since we end up running a specific version anyway.

Historically, have these checker tests caught many issues?

lunaryorn:
@cpitclaudel Probably, but I can't remember, and thus am inclined to believe that they did not :relaxed: :wink:

cpitclaudel:
:)

geraldus:
Hi!
[...]

PR #864, opened 2016-01-31
Handle warnings in included files properly.

xwl:
* Current code may mistake warnings in included files for errors, as
  demonstrated below:

        // foo.cpp
        #include "foo.h"
        int main(int argc, char *argv[]) {
            xx;
        }

        // foo.h
        #warning ok sfdsfsdf

* Improve included files' error message display in minibuffer. In above
  example, change "In include foo.h" to "In include foo.h: ok sfdsfsdf".

### DIFF
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -1911,7 +1911,7 @@ and extension, as in `file-name-base'."
                                              :filename "foo.cpp"))))
     (should
      (equal (flycheck-fold-include-levels errors "In file included from")
-            (list (flycheck-error-new-at 1 0 'error "In include foo.h"
+            (list (flycheck-error-new-at 1 0 'info "In include foo.h"
                                          :filename "foo.cpp")
                   (flycheck-error-new-at 6 11 'error "b is not a member of hi"
                                          :filename "foo.cpp")
--- a/flycheck.el
+++ b/flycheck.el
@@ -3380,11 +3380,12 @@ included file."
         (when (funcall sentinel current-error)
           ;; We found an error denoting errors in the included file:
           ;; 1. process all subsequent errors until faulty include file is found
-          ;; 2. process again all subsequent errors until an error has the
-          ;;    current file name again
+          ;; 2. process again all subsequent errors before the current file name appears again
           ;; 3. find the most severe error level
           (let ((current-filename (flycheck-error-filename current-error))
-                (current-level nil)
+                (saved-level nil)
+                (saved-message "")
+                (saved-filename "")
                 (faulty-include-filename nil)
                 (filename nil)
                 (done (null remaining-errors)))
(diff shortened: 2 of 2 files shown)


lunaryorn:
@xwl Thank you for this fix.  It'll take me a while to review though, and to write new tests to cover the change in behaviour.

lunaryorn:
Note to myself: We'll need to move the ERT tests for error folding to buttercup first, see #859 :)

lunaryorn:
@xwl Would you mind to show the output of Clang for your example?  

I can't really image how this fix works; furthermore I wonder what happens when there are are multiple warnings in the include file.  Which message gets picked up in this case?

xwl:
$ clang++ foo.cpp 

    In file included from foo.cpp:1:
    ./foo.h:1:4: warning: ok sfdsfsdf [-W#warnings]
      #warning ok sfdsfsdf
       ^
    foo.cpp:3:7: error: use of undeclared identifier 'xx'
          xx;
          ^
    1 warning and 1 error generated.

 $ g++ foo.cpp 

    In file included from foo.cpp:1:0:
    foo.h:1:4: warning: #warning ok sfdsfsdf [-Wcpp]
       #warning ok sfdsfsdf
        ^
    foo.cpp: In function 'int main(int, char**)':
    foo.cpp:4:10: error: 'xx' was not declared in this scope
              xx;
              ^

If there are more warnings in header file, we will show the first one.  But if there is an error in the heading, then error will be shown.

lunaryorn:
@xwl It occurred to me that this does not behave correctly actually.  In fact your change to the test is wrong.

test/flycheck-test.el
```diff
@@ -1911,7 +1911,7 @@ and extension, as in `file-name-base'."
                                              :filename "foo.cpp"))))
     (should
      (equal (flycheck-fold-include-levels errors "In file included from")
-            (list (flycheck-error-new-at 1 0 'error "In include foo.h"
+            (list (flycheck-error-new-at 1 0 'error "In include foo.h: b is not a member of hi"
```
lunaryorn: That's not actually correct.  If you look at the test errors above, you'll see that this message comes from `foo.cpp` and not from `foo.h`.

`in definition of macro CHECK` comes from `foo.h`.

xwl:
Well, for the existing testcase, below is output from g++, it is hard to decide how we should format the output.  It mentions header file first, then error itself, finally the actual header file.

    $ g++ foo.cpp 
    In file included from foo.cpp:1:0:
    foo.cpp: In function ‘int main(int, char**)’:
    foo.cpp:3:16: error: ‘b’ is not a member of ‘Hi’
              CHECK(Hi::b);
                    ^
    foo.h:2:28: note: in definition of macro ‘CHECK’
             #define CHECK(arg) arg
                                ^
    foo.cpp:4:10: error: ‘xx’ was not declared in this scope
              xx;
              ^
In contrast, clang++'s output is more clear. It mentions error in cpp file directly.

    foo.cpp:11:23: error: no member named 'b' in namespace 'Hi'
                CHECK(Hi::b);
                      ~~~~^
    ./foo.h:8:28: note: expanded from macro 'CHECK'
            #define CHECK(arg) arg
                               ^
    foo.cpp:12:13: error: use of undeclared identifier 'xx'
                xx;
                ^
    2 errors generated.
        
I think the problem is that g++ should not output "In file included from foo.cpp:1:0:" in this case, because error doesn't actually happen in the header file.  Maybe we could check this case and ignore this sort of false signal?

lunaryorn:
@xwl Well, I don't know how we should parse these errors; I don't know C++ and I must admit that I don't particularly care for it either.

But it seems that we should not merge errors from a file with errors from an include: Flycheck should obviously skip over the first error to the `foo.h` note in case of gcc.

xwl:
How about the udpated change? I only changed the error leve from 'error to 'info for this particular case.

lunaryorn:
I'm closing this pull request, as we're likely to remove all this logic entirely and instead just show errors from other files as well, see #972.

JackalopeGit:
It's still not fixed.

cpitclaudel:
This got fixed in #1427.  That warning is not shown at all by default, unless you customize `flycheck-relevant-error-other-file-minimum-level`

Issue #471, opened 2014-08-26
LESS syntax checker errors in function data-uri


albertgasset:
The LESS syntax checker shows this error when a [data-uri](http://lesscss.org/functions/#misc-functions-data-uri) function is used:

>  error evaluating function `data-uri`: ENOENT, no such file or directory '/tmp/flycheck13473hZv/logo.png'

It doesn't find the image, which is present in the original directory next to the less file. What's worse, errors after the data-uri function are not shown, because lessc stops at the first error found.

I've tried to change the syntax checker definition. It works when using `source-inplace` instead of `source`.

I'm using the latest version from MELPA (20140825.353)

PR #867, opened 2016-02-04
Use standard input for Less checker

stevenremot:
Make the Less checker use standard input. This prevent temporary files
to mess with Less watchers.

Standard input is supported in Less 1.4 and later, so this commit is
backward-compatible.

As far as I understand the issue #471 referenced in the deleted comment won't suffer from this change, as `lessc` will be executed in the buffer's working directory. I did some tests that confirm it.


lunaryorn:
@stevenremot Thank you very much :relaxed: Merged :+1:

Issue #799, opened 2015-11-19
New racket checker


cpitclaudel:
Re-routing the discussion from #786 for ease of tracking and inviting @soegaard and @takikawa.

To summarize:

* We're happy to take a PR adding a racket checker based on `raco expand`.
* I for one would welcome a sandboxed checker. The best place for it would probably be https://github.com/greghendershott/racket-mode (@greghendershott, opinions?)

Thanks for the help!

greghendershott:
Although I only read the discussion twice quickly:

From [this comment](https://github.com/flycheck/flycheck/issues/786#issuecomment-157996699) it sounds like @lunaryorn would be comfortable with replacing `racket -f` with `raco expand`?

If so that's simply:

1. Revert commit 173728f (as well as the two commits for e.g. doc changes)
2. Change the [`:command ("racket" "-f" source-inplace)`](https://github.com/flycheck/flycheck/commit/173728f6804525118db9657e4bcdaee74b3c7c8c#diff-374cdd9563b826c712b1dc6a373a0bc1L7493) to `:command ("raco" "expand" source-inplace)`.

I think??

---

Would that be sandboxed?  I don't know. But if it isn't:

1. That didn't sound must-have to @lunaryorn (and FWIW I agree with his reasoning).
2. Even if would be nice-to-have, it seems the logical and effective place to add that would be in the implementation of `raco expand` itself?

---

Having said all that: If the discussion somehow leads to needing to deliver some Racket code, I'd be happy to roll that into racket-mode.

---

p.s. If languages with "unsandboxed" macros are a problem, does that mean Rust needs to be removed, too?

cpitclaudel:
@greghendershott Indeed, I think we'll revert that commit and move to raco in core Flycheck. But it would be very nice for racket-mode to bundle the sanboxed version; what do you think?

I think @lunaryorn's policy is reasonable, and I'll abide by it :) So rust & al. can stay :)

lunaryorn:
I don't mind whether it's sandboxed or not.  If the compiler runs it Flycheck can run it, too, imho :relaxed:

I just don't want to run _all_ code.  That's too unsafe for me, because normal code is very likely to have side-effects, whereas code that is _intended_ to run at compilation time (e.g. macros, etc.) usually has not.

In other words, I'm fine with `raco expand`.  Is `raco` part of a standard Racket installation?

greghendershott:
[...]

PR #873, opened 2016-02-11
Racket checker w/ raco expand

wcummings:
Can't actually get the tests to run, output in flycheck-list-errors matches up, though.

Edit (@lunaryorn): Connects to #799

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -224,6 +224,7 @@ attention to case differences."
     python-pylint
     python-pycompile
     r-lintr
+    racket
     rpm-rpmlint
     rst-sphinx
     rst
@@ -7661,6 +7662,18 @@ See URL `https://github.com/jimhester/lintr'."
   :modes ess-mode
   ;; Don't check ESS files which do not contain R
   :predicate (lambda () (equal ess-language "S")))
+
+(flycheck-define-checker racket
+  "Racket syntax checker.
+
+See URL `https://racket-lang.org/'."
+  :command ("raco" "expand" source-original)
+  :error-filter
+  (lambda (errors)
+    (flycheck-sanitize-errors (flycheck-increment-error-columns errors)))
+  :error-patterns
+  ((error line-start (file-name) ":" line ":" column ":" (message) line-end))
+  :modes racket-mode)
 
 (flycheck-define-checker rpm-rpmlint
   "A RPM SPEC file syntax checker using rpmlint.
--- a/flycheck.el
+++ b/flycheck.el
@@ -7667,7 +7667,7 @@ See URL `https://github.com/jimhester/lintr'."
   "Racket syntax checker.
 
 See URL `https://racket-lang.org/'."
-  :command ("raco" "expand" source-original)
+  :command ("raco" "expand" source-inplace)
   :error-filter
   (lambda (errors)
     (flycheck-sanitize-errors (flycheck-increment-error-columns errors)))
--- a/flycheck.el
+++ b/flycheck.el
@@ -7663,11 +7663,22 @@ See URL `https://github.com/jimhester/lintr'."
   ;; Don't check ESS files which do not contain R
   :predicate (lambda () (equal ess-language "S")))
 
+(defun lines-contain (needle line)
+  (let ((line (car lines)))
+    (cond
+     ((not line) f)
+     ((string-match needle line) t)
+     (lines-contain (cdr lines)))))
+
 (flycheck-define-checker racket
   "Racket syntax checker.
 
 See URL `https://racket-lang.org/'."
   :command ("raco" "expand" source-inplace)
+  :predicate (lambda ()
+	       (let ((raco (flycheck-checker-executable 'racket)))
+		 (-any (apply-partially 'string-match "compiler-lib")
+		       (ignore-errors (process-lines raco "pkg" "show" "--all")))))
   :error-filter
   (lambda (errors)
     (flycheck-sanitize-errors (flycheck-increment-error-columns errors)))
--- a/flycheck.el
+++ b/flycheck.el
@@ -7663,13 +7663,6 @@ See URL `https://github.com/jimhester/lintr'."
   ;; Don't check ESS files which do not contain R
   :predicate (lambda () (equal ess-language "S")))
 
-(defun lines-contain (needle line)
-  (let ((line (car lines)))
-    (cond
-     ((not line) f)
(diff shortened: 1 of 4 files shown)


lunaryorn:
@wcummings I'll review the change and run the tests.   Thank you very much for this contribution, that's awesome :+1:

flycheck.el
```diff
@@ -7662,6 +7663,18 @@ See URL `https://github.com/jimhester/lintr'."
   ;; Don't check ESS files which do not contain R
   :predicate (lambda () (equal ess-language "S")))
 
+(flycheck-define-checker racket
+  "Racket syntax checker.
+
+See URL `https://racket-lang.org/'."
+  :command ("raco" "expand" source-original)
```
lunaryorn: Why do we need `source-original`?

wcummings: So it resolves modules correctly and tells if you fuck up a require or something. I figured it was fine since `raco expand` (afaik) shouldn't modify the input source. Is there a nicer way to do this?

lunaryorn: Wouldn't `source-inplace` work as well?

wcummings: You're right, it would :100:

lunaryorn:
@wcummings How to I install `raco expand`? 

I installed racket with `brew install racket`, and I do have a `raco` command, but:

```
$ raco expand
/usr/local/bin/raco: Unrecognized command: expand

Usage: raco <command> <option> ... <arg> ...

Frequently used commands:
  setup   install and build libraries and documentation
  pkg     manage packages

A command can be specified by an unambiguous prefix.
See `raco help' for a complete list of commands.
See `raco help <command>' for help on a command.
```

greghendershott:
> How to I install raco expand?
>
> I installed racket with brew install racket, and I do have a raco command, but:

Although I appreciate why homebrew chose a recipe using Minimal Racket, it's unfortunate for first-time users. Not only are none of Racket's wonderful batteries included, neither is a bulb or lens. :) I'm fairly familiar with Racket and I didn't realize something like `raco expand` isn't included.

You could `raco pkg install drracket` [as I suggest here for racket-mode users](https://github.com/greghendershott/racket-mode#minimal-racket). That's "heavier" than you need (maybe someone else can determine the exact package(s) for `raco expand` and chime in). Anyway you can `raco pkg remove drracket` later.

lunaryorn:
@greghendershott So do I understand correctly that it's possible and also likely (at least on OS X) to have `raco` but not `raco expand`?

In this case we'll need a `:predicate` that checks whether `raco expand` is present before we can safely add this syntax checker to Flycheck.  

@wcummings Would you mind to update your pull request accordingly?  Take a look at the `go-vet` checker to see how we do that in another similar case.

greghendershott:
> So do I understand correctly that it's possible and also likely (at least on OS X) to have raco but not raco expand?

In some other thread here I'd said I believed the `raco` `expand` command would be included even in Minimal Racket. Your recent comment made me think I'm wrong. But I'm not sure. I only ever use the full distribution of Racket. Although I won't have time today to experiment with Minimal Racket on OS X, I'll ping `#racket` with a link so hopefully someone who knows for sure can chime in.

lunaryorn:
@greghendershott As I've said, `brew install racket` doesn't include `raco expand`.  I haven't looked any further, as I don't use racket.

Generally we don't document requirements of individual syntax checkers in the README.  We list them in the manual, and use our "verification" feature to provide guidelines to the user.  That's why the checker needs to have a proper `:predicate`.

greghendershott:
I don't doubt what you're seeing. I doubt my own knowledge; I don't want to mislead you. I asked on `#racket`, hopefully someone will advise here soon.

samth:
The `raco expand` command is part of the `compiler-lib` package.

greghendershott:
Thanks @samth !

So IIUC:

1. You'll need a `:predicate`.
2. `raco pkg install compiler-lib` will install an `expand` command for `raco`. (People using the full download from `racket-lang.org` will already have this. )

wcummings:
@lunaryorn will do :+1:

lunaryorn:
@wcummings Thank you very much.  Please ping us when you have updated the pull request.  Github doesn't send notifications for pull requests; hence we might miss your update otherwise :relaxed:

wcummings:
@lunaryorn I added a predicate, works alright, but it does seem noticeably more sluggish to me with the predicate, does it get checked on every call?

lunaryorn:
@wcummings Yes, on every call, necessarily—Emacs sessions can run for days, weeks, even months, so it's somewhat likely that the environment changes underneath a running Emacs session.  We prefer to err on the side of caution to avoid confusion error messages.

However, an additional process call by itself shouldn't be sluggish.  It's probably `raco pkg show` that's slow.  How about this predicate instead which just checks whether `expand` is a valid command?

```
(lambda ()
  (let ((raco (flycheck-checker-executable 'racket)))
    (with-temp-buffer
      (call-process raco nil t nil "expand")
      (goto-char (point-min))
      (not (looking-at-p (rx bol (1+ not-newline) "Unrecognized command: expand" eol))))))
```

I couldn't install `compiler-lib`—`raco pkg install` failed to connect to `mirror.racket-lang.org`—so I couldn't test whether it works in the "good" case.  Would you mind to try it and check whether it works better?

wcummings:
@lunaryorn much better :+1:

lunaryorn:
@wcummings Does it work well?

wcummings:
@lunaryorn yeah, snappier than before, at least to my eyes.

lunaryorn:
@wcummings Thank you very much. :+1: I've cherry-picked your patch :) 

By the way, I noticed that you made your commits with an email that is not associated with your Github account.  Was that by accident?  Would you like me to amend the cherry-picked commits to use a different email address?

wcummings:
@lunaryorn will@wpc.io, thanks! :)

cpitclaudel:
@wcummings Wonderful work! Thanks :)

lunaryorn:
@wcummings I've picked that address now, but it doesn't seem to be associated to your account either, albeit being shown in your profile.  I would recommend you to check https://github.com/settings/emails and add this address (and probably the other one you used first as well) there to properly claim authorship of your commits on Github :)

lunaryorn:
@wcummings I applied your change to `master`, and with the next MELPA build Flycheck will have Racket support again.  Thank you very much for your contribution :+1:  Great work :clap:

PR #874, opened 2016-02-14
Accept suggestions from haskell-lint as 'info'

corngood:
In hint-1.9.28 a bunch of warnings were changed to 'suggestions'.  Without this change any suggestion will result in:

`
Suspicious state from syntax checker haskell-hlint: Checker haskell-hlint returned non-zero exit code 1, but no errors from output:
`

PR #875, opened 2016-02-16
hlint: Include suggestions as info

jbaum98:
Hlint issues three types of messages: Errors, Warnings, and Suggestions. Currently Flycheck only handles Errors and Warnings. This commit adds support for Suggestions.


jbaum98:
This was already addressed in #874

PR #889, opened 2016-02-27
set C locale and patch C, fortran and haskell tests

spwhitton:
gcc and gfortran output smartquotes in a Unicode locale and pure ASCII
in the C locale.  Flycheck's test suite assumes a Unicode locale, so the
test suite can unexpectedly fail when only the C locale is available.

This patch sets the test suite to always use the C locale and updates
expected gcc and gfortran output accordingly.  This makes the test suite
more portable, in particular to the Debian package autobuilder
infrastructure.

### DIFF
--- a/test/run.el
+++ b/test/run.el
@@ -108,6 +108,11 @@ Node `(ert)Test Selectors' for information about test selectors."
   (let* ((load-prefer-newer t)
          (source-directory (locate-dominating-file flycheck-runner-file "Cask"))
          (pkg-rel-dir (format ".cask/%s/elpa" emacs-version)))
+
+    ;; standardise on the C locale so we can be sure that none of
+    ;; gcc, ghc and gfortran will output smartquotes
+    (setenv "LC_ALL" "C")
+
     (setq package-user-dir (expand-file-name pkg-rel-dir source-directory))
     (package-initialize)
 
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3118,7 +3118,7 @@ of the file will be interrupted because there are too many #ifdef configurations
   (let ((flycheck-disabled-checkers '(c/c++-clang)))
     (flycheck-ert-should-syntax-check
      "language/c_c++/error.cpp" 'c++-mode
-     '(2 18 error "‘struct A’ has no member named ‘bar’"
+     '(2 18 error "'struct A' has no member named 'bar'"
          :checker c/c++-gcc))))
 
 (flycheck-ert-def-checker-test c/c++-gcc (c c++) fatal-error
@@ -3132,7 +3132,7 @@ of the file will be interrupted because there are too many #ifdef configurations
   (let ((flycheck-disabled-checkers '(c/c++-clang c/c++-cppcheck)))
     (flycheck-ert-should-syntax-check
      "language/c_c++/warning.c" 'c-mode
-     '(5 10 warning "unused variable ‘unused’" :checker c/c++-gcc)
+     '(5 10 warning "unused variable 'unused'" :checker c/c++-gcc)
      '(7 15 warning "comparison between signed and unsigned integer expressions"
          :checker c/c++-gcc)
      '(8 7 warning "#warning" :checker c/c++-gcc))))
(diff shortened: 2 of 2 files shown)


test/run.el
```diff
@@ -108,6 +108,11 @@ Node `(ert)Test Selectors' for information about test selectors."
   (let* ((load-prefer-newer t)
          (source-directory (locate-dominating-file flycheck-runner-file "Cask"))
          (pkg-rel-dir (format ".cask/%s/elpa" emacs-version)))
+
+    ;; standardise on the C locale so we can be sure that neither of
+    ;; gcc and gfortran will output smartquotes
+    (setenv "LC_ALL" "C")
```
lunaryorn: Is that enough?  Shouldn't we set `$LANG` and `$LC_MESSAGES` as well?  Sorry if that's a stupid question, but I'm not at all familiar with the semantics of all these locale variables.

lunaryorn:
Normalising the locale for our tests is a really good idea, thank you very much.

Please squash your commits and start the commit message with a capitalised letter (see our [commit message guidelines](http://www.flycheck.org/contributing.html#commit-guidelines)).

test/run.el
spwhitton: On Sat, Feb 27, 2016 at 12:06:58PM -0800, Sebastian Wiesner wrote:
> Is that enough? Shouldn't we set $LANG and $LC_MESSAGES as well? Sorry if
> that's a stupid question, but I'm not at all familiar with the semantics of all
> these locale variables.

My understanding of the locale variables is that LC_ALL is designated as
an override variable, so we only need to set that one.  The gcc manpage
suggests that LC_ALL is enough (and indeed, the test suite passes).

On Sat, Feb 27, 2016 at 12:21:01PM -0800, Sebastian Wiesner wrote:
> Please squash your commits and start the commit message with a capitalised
> letter (see our commit message guidelines).

Done.

-- 
Sean Whitton

spwhitton:
On Sat, Feb 27, 2016 at 12:06:58PM -0800, Sebastian Wiesner wrote:
> Is that enough? Shouldn't we set $LANG and $LC_MESSAGES as well? Sorry if
> that's a stupid question, but I'm not at all familiar with the semantics of all
> these locale variables.

My understanding of the locale variables is that LC_ALL is designated as
an override variable, so we only need to set that one.  The gcc manpage
suggests that LC_ALL is enough (and indeed, the test suite passes).

On Sat, Feb 27, 2016 at 12:21:01PM -0800, Sebastian Wiesner wrote:
> Please squash your commits and start the commit message with a capitalised
> letter (see our commit message guidelines).

Done.

-- 
Sean Whitton

cpitclaudel:
@lunaryorn Objections on merging this? I can cherry-pick it if you want :)

lunaryorn:
@cpitclaudel Mmh, I'm not sure about the approach; I'm no particular fan of globally changing the environment for a testsuite.  I'd rather prefer if the locale environment was overridden per testcase, probably in `flycheck-ert-should-syntax-check` or so.

OTOH, it's a small change and probably does no harm either, so I can live with merging it.

cpitclaudel:
@lunaryorn Ok, works for me :) Looks like setenv uses `process-environment`:
```
Variable: process-environment

    This variable is a list of strings, each describing one environment variable. The functions getenv and setenv work by means of this variable.

    process-environment
    ⇒ ("PATH=/usr/local/bin:/usr/bin:/bin"
        "USER=lewis"

        "TERM=xterm"
        "SHELL=/bin/bash"
        "HOME=/home/lewis"
        …)

    If process-environment contains “duplicate” elements that specify the same environment variable, the first of these elements specifies the variable, and the other “duplicates” are ignored. 
```

and 

```
Documentation:

List of overridden environment variables for subprocesses to inherit.
Each element should be a string of the form ENVVARNAME=VALUE.

Entries in this list take precedence to those in the frame-local
environments.  Therefore, let-binding ‘process-environment’ is an easy
way to temporarily change the value of an environment variable,
irrespective of where it comes from.  To use ‘process-environment’ to
remove an environment variable, include only its name in the list,
without "=VALUE".
```

@spwhitton, what do you think of adding something like 

```emacs-lisp
(let ((process-environment (cons "LC_ALL=C" process-environment)))
   ...)
```

to `flycheck-ert-should-syntax-check`? (@lunaryorn, that would work, right?)

spwhitton:
I considered modifying the process-environment variable with a let binding, but I decided against it because in my view the locale should be standardised across all tests.  Otherwise the test environment varies depending on what country the user is in, but the test suite is meant to be something uniform.  As lunaryorn put it in his first message, "Normalising the locale for our tests is a really good idea".

cpitclaudel:
I'm confused; doesn't the let-binding achieve essentially the same as set-env, except for not being a persistent mutation?

spwhitton:
Sorry, let me clarify what I have in mind.  There are three options:

(1) Call `(setenv)` once to set the environment (this pull request).

(2) Let-bind `process-environment` *once*, before the test suite runs.

(3) Let-bind `process-environment` for each test that gives different output in different locales.

My previous comment argued that you should choose (1) or (2), but not (3).  lunaryorn's most recent comment suggests that he would prefer (3).

If the choice is between (1) and (2), I think that (1) is better because the code is simpler, and the effect is identical.  There is no permanent environment change because a dedicated `emacs --batch` instance is spawned to run the test suite.  After the test suite run is complete, the `(setenv)` will have no effect on the shell that called `emacs --batch`.

lunaryorn:
@spwhitton Well, simplicity is in the eye of the beholder, isn't it?  I prefer the third option, i.e. a local `let`-binding, because I find it strange the test *runner* enforces global state upon the tests it runs and thus changes the test environment.  I think that tests should have a self-contained and local environment.

I understand that the difference is moot, though, and I'm fine with merging as is if need be.

PR #890, opened 2016-02-27
clear GPG agent env vars

spwhitton:
If a GPG agent is running, EasyPG ignores the setting
`epa-file-passphrase-alist' and prompts the user for a password
regardless.  So unset these environment variables before running the
test suite.

### DIFF
--- a/test/run.el
+++ b/test/run.el
@@ -108,6 +108,12 @@ Node `(ert)Test Selectors' for information about test selectors."
   (let* ((load-prefer-newer t)
          (source-directory (locate-dominating-file flycheck-runner-file "Cask"))
          (pkg-rel-dir (format ".cask/%s/elpa" emacs-version)))
+
+    ;; clear GPG agent environment variables so that EPG doesn't
+    ;; prompt for any passphrases during the test suite run
+    (setenv "GPG_TTY" "")
+    (setenv "GPG_AGENT_INFO" "")
+
     (setq package-user-dir (expand-file-name pkg-rel-dir source-directory))
     (package-initialize)
 


lunaryorn:
Thank you, but I'm sceptical about this pull request, for pretty much the same reasons as in https://github.com/flycheck/flycheck/pull/891#issuecomment-189716536.  

Still, in this case I can see the value of this change.  I'll merge it, like in #891 I would prefer if the environment was only reset for the tests that actually use gpg, and also in the spec tests.

Thank you very much :)

spwhitton:
Thanks for your feedback.  I'm packaging version 0.25.1 for Debian, but since you've moved the GPG tests out of `flycheck-test.el` in the master branch, a more specific patch changing the environment only for particular tests would not apply to master.  So this pull request is obsolete at least until the next release of flycheck.  I'll submit a better patch then.

I should note that this thread is not an example of "encoding the peculiarities of Debian's build infrastructure" into flycheck's test suite (something I agree you should avoid).  That's because the issue this patch fixes would affect anyone who runs the test suite without Cask who also happens to be running gpg-agent.  Since you're using the standard ERT toolchain, it ought to be possible to run the test suite without Cask, if you have all the dependencies available by other means.

lunaryorn:
@spwhitton Ah, so you're building on the last release?  Indeed, our tests have moved somewhat since then.

We're going to have [1.0 soon](https://github.com/flycheck/flycheck/milestones/Flycheck%201.0); it'd be great if you could take a look at the current tests to see whether they need refactoring so that the necessary changes can go into 1.0.  Ideally 1.0 would then ship a testsuite that works flawlessly on your build infrastructure.

By the way, I noticed that you seem to be concerned with the Haskell tests mostly.  Is there a particular reason why you'd like to run the Haskell tests specifically?  Or do you run the entire test suite?

lunaryorn:
Note that I've reopened this issue.  I think gpg-agent interferring with our tests really is a bug, and I'll fix it for our buttercup tests.

spwhitton:
Hello,

On Mon, Feb 29, 2016 at 04:52:44AM -0800, Sebastian Wiesner wrote:
> @spwhitton Ah, so you're building on the last release? Indeed, our tests have
> moved somewhat since then.
> 
> We're going to have 1.0 soon; it'd be great if you could take a look at the
> current tests to see whether they need refactoring so that the necessary
> changes can go into 1.0. Ideally 1.0 would then ship a testsuite that works
> flawlessly on your build infrastructure.

Okay.  I'll take a look this weekend.

> By the way, I noticed that you seem to be concerned with the Haskell
> tests mostly. Is there a particular reason why you'd like to run the
> Haskell tests specifically? Or do you run the entire test suite?

I run as much of the test as possible -- everything for which the
dependencies are already available in Debian.  We have packages for
haskell-mode, ghc and stack so I'm trying to get the Haskell tests
running.

-- 
Sean Whitton

spwhitton:
Okay.  I'll take a look at the new test suite this weekend.

Regarding the Haskell tests, I run as much of the test as possible -- everything for which the dependencies are already available in Debian.  We have packages for haskell-mode, ghc and stack so I'm trying to get the Haskell tests running.

spwhitton:
Just to keep you updated: I'm working on packaging buttercup for Debian.  Until I've done that I won't be able to try your new test suite.

PR #891, opened 2016-02-27
set epg-gpg-home-directory

spwhitton:
gpg immediately dies with an error when invoked if the current user's
home directory does not exist (as is the case on a Debian package
autobuilder).  This causes all tests that invoke gpg to be skipped.
Setting `epg-gpg-home-directory` when running the test suite fixes the
problem, making the test suite more portable.

### DIFF
--- a/test/run.el
+++ b/test/run.el
@@ -120,7 +120,8 @@ Node `(ert)Test Selectors' for information about test selectors."
       (load (expand-file-name "flycheck-test"
                               (file-name-directory flycheck-runner-file)))))
 
-  (let ((debug-on-error t))
+  (let ((debug-on-error t)
+        (epg-gpg-home-directory "/tmp"))
     (flycheck-run-tests-batch-and-exit)))
 
 ;;; run.el ends here


lunaryorn:
Thank you, but I am not sure whether I would like to merge this.  I don't think that it makes the test suite more portable.  Rather, it feels as if we encode the peculiarities of Debian's build infrastructure into our test suite.  It's very unusual—even for automated build systems—to have no `$HOME` directory in my experience.

In any case, I think I would prefer if this option was `let`-bound only for the relevant tests.  Please take a look at whether our new spec tests (see #859) in `tests/specs` need this option as well, too.  Furthermore, please do not hard-code `/tmp/`.  Use `make-temp-file` with a non-nil `DIR-FLAG` instead.

spwhitton:
Thanks for your feedback.  I'm packaging version 0.25.1 for Debian, but since you've moved the GPG tests out of `flycheck-test.el` in the master branch, a more specific patch let-binding `epg-gpg-home-directory` in particular tests would not apply to master.  So this pull request is obsolete at least until the next release of flycheck.  I'll submit a better patch then.

lunaryorn:
@spwhitton I've fixed that for the new buttercup tests.  Thank you for making me aware!

spwhitton:
Cool!

PR #895, opened 2016-02-28
Skip haskell stack tests when homedir nonexistent

spwhitton:
stack requires the user's home directory to exist.  Skip the tests of
stack if it doesn't.  (The assumption that it exists is unwarranted on
Debian's autobuilding infrastructure.)


spwhitton:
Quick note: I could of course create a temporary home directory as part of the test to make it run, but I thought that this more minimal change of just skipping would avoid "encoding the peculiarities of Debian's build infrastructure into our test suite" -- I agree that you shouldn't do that.

lunaryorn:
@spwhitton Thank you, merged :relaxed:

spwhitton:
\o/ thanks!

PR #900, opened 2016-03-02
Remove mention about flycheck-scalastyle-jar

chessman:
http://www.flycheck.org/2015/11/14/flycheck-0.25.html

"The scala-stylestyle syntax checker now expects a scalastyle executable in exec-path now. The flycheck-scalastyle-jar option which used to provide the location of the Scalastyle JAR file was removed."


lunaryorn:
@chessman Thank you very much!

Issue #676, opened 2015-06-30
go-build: bug if there are multiple packages in a directory


phst:
If a directory contains files belonging to more than one package (e.g. one file `package a` and one file `package b`), the go-build checker barfs:

```
Suspicious state from syntax checker go-build: Checker go-build returned non-zero exit code 1, but no errors from output: can't load package: package .: found packages a.go (a) and b.go (b) in /private/tmp

Checker definition probably flawed.
```

phst:
Actually I think that instead of adding a new error pattern the checker command could just be replaced by

```lisp
("go" "build" "-o" temporary-file-name "--" source)
```

to compile only a single file.

lunaryorn:
@phst I do not know go, but I was under the impression that there is just one go package per directory.  Is it common for multiple go packages to be in the same directory?

phst:
It's not common, but possible. Flycheck might either choose to not support this use case (then the error pattern should be fixed to detect the situation), or support it (then the go tool requires an explicit file name argument).

phst:
However, my suggestion wouldn't work out of the box if the package consists of more than one file. Maybe just give up if a directory with multiple packages is detected? Otherwise this would require parsing all Go files in the current directory to find out to which package they belong.

lunaryorn:
@phst I think I'm the wrong person to discuss.  I can't contribute any ideas, because I do not know Go, and I'll not fix this issue by myself, because fixing a corner case for a language that I neither know nor intent to ever use is not a priority for me.  Whatever the fix turns out to be, you or someone else has to come up with a pull request or patch.

All that I can do is to set the basic requirements for such a patch:  It _must_ not break anything that Flycheck already supports, and it should not be overly complex.

In other words, if giving an explicit file name argument to `go build` breaks packages consisting of multiple files, then that's not an acceptable fix for me, and if you need a 100 lines of Lisp code to find out how to check the current buffer, then that's not acceptable either.

I think, it'll probably come down to adding a new error pattern or a specific `:error-parser:` to deal with this situation, and just error out in this case.

phst:
Fair enough, I can produce a pull request for this.

phst:
What I would actually prefer is to disable the Go checker in that case, e.g. by returning an empty error list. The error from go build doesn't represent an error in the source file being checked, but a limitation in the go tool itself.

lunaryorn:
[...]

PR #904, opened 2016-03-08
Catch error message for multiple Go packages

phst:
If there are multiple Go packages in a single directory, `go build` will
bail out.  Catch the error message printed by the tool and turn it into
an informational message because Flycheck can’t really syntax-check such
packages.

Because the error message doesn’t contain a line number, install an
error filter to fake a line number.  Otherwise the message would be
ignored.

Fixes GH-676 and closes GH-676.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6644,7 +6644,24 @@ See URL `http://golang.org/cmd/go'."
           (optional column ":") " "
           (message (one-or-more not-newline)
                    (zero-or-more "\n\t" (one-or-more not-newline)))
-          line-end))
+          line-end)
+   ;; Catch error message about multiple packages in a directory, which doesn't
+   ;; follow the standard error message format.
+   (info line-start
+         (message "can't load package: package "
+                  (one-or-more (not (any ?: ?\n)))
+                  ": found packages "
+                  (one-or-more not-newline))
+         line-end))
+  :error-filter
+  (lambda (errors)
+    (dolist (error errors)
+      (unless (flycheck-error-line error)
+        ;; Flycheck ignores errors without line numbers, but the error
+        ;; message about multiple packages in a directory doesn't come with a
+        ;; line number, so inject a fake one.
+        (setf (flycheck-error-line error) 1)))
+    errors)
   :modes go-mode
   :predicate (lambda ()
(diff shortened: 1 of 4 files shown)


cpitclaudel:
Thanks for working on this. I made a small comment in the source; I'll let @lunaryorn comment further.

cpitclaudel:
By the way, I don't think yo need to say both "Fixes" and "Closes" in your commit message; just one is enough to close the issue.

lunaryorn:
> If this new type of errors is the only one from Go that doesn't include a line number, then maybe you could just add a line number to any error that doesn't have one, instead of (re-)matching every message?

Indeed, I think that'd be much simpler.

flycheck.el
```diff
@@ -6644,7 +6644,25 @@ See URL `http://golang.org/cmd/go'."
...
+          line-end)
+   (info line-start
+         (message "can't load package: package "
+                  (one-or-more (not (any ?: ?\n)))
+                  ": found packages "
+                  (one-or-more not-newline))
+         line-end))
+  :error-filter
+  (lambda (errors)
+    (dolist (error errors errors)
```
lunaryorn: Please return `errors` explicitly after the loop.  I am no particular fan of this behaviour of `dolist`.

test/flycheck-test.el
```diff
@@ -3462,6 +3462,18 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
...
+  (flycheck-ert-with-env
+      `(("GOPATH" . ,(flycheck-ert-resource-filename "checkers/go")))
+    (flycheck-ert-should-syntax-check
+     "checkers/go/src/multipkg/a.go" 'go-mode
+     `(1 nil info
+         ,(concat "can't load package: package multipkg: "
+                  "found packages a (a.go) and b (b.go) in "
+                  (flycheck-ert-resource-filename
+                   "checkers/go/src/multipkg"))
+         :checker go-build))))
```
lunaryorn: Thank you for adding a test case :+1:

phst:
Thanks for the review. Resolved the comments.

lunaryorn:
@phst Thank you very much, cherry-picked :relaxed: :+1:

PR #906, opened 2016-03-12
Clean up go-errcheck checker

dominikh:
errcheck has a new flag and fixed a bug that makes the checker implementation a lot simpler. The bug was fixed late February 2015, the flag was added late August 2015.

/cc @lunaryorn

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6695,7 +6695,7 @@ part of any package or if GOPATH is nil."
 Requires an errcheck version from commit 8515d34 (made on August 28 2015) or newer.
 
 See URL `https://github.com/kisielk/errcheck'."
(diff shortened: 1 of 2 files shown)


lunaryorn:
@dominikh Thank you very much, great change for the better!  Just would you mind to document the minimum version requirement for go errcheck (i.e. the first version that has this flag) in the syntax checker docstring?

dominikh:
Does a git commit sha constitute a version? errcheck doesn't have explicit releases/version numbers.

lunaryorn:
@dominikh Uuuh 😒 in this case, please include an abbreviated commit hash and the date when the change was pushed.

dominikh:
@lunaryorn PTAL.

lunaryorn:
@dominikh Thank you very much, cherry-picked :relaxed: :+1:

PR #909, opened 2016-03-16
Set cppcheck -x param based on major-mode

Claymore:
cppcheck treats a header file as a C header file by default. We can make
it recognize C++ header files with `-x` command line parameter so we set
up the parameter to `c++` when we visit a C++ file in c++-mode.


lunaryorn:
@Claymore Thank you very much for this change, cherry picked onto `master` :clap: :+1: :relaxed:

PR #911, opened 2016-03-18
Check whether R has installed the library lintr

marsam:


flycheck.el
```diff
@@ -7684,7 +7684,25 @@ See URL `https://github.com/jimhester/lintr'."
...
+  :verify (lambda (checker)
+            (let ((has-lintr (flycheck-R-has-lintr
+                              (flycheck-checker-executable checker))))
+              (list
+               (flycheck-verification-result-new
+                :label "lintr library"
+                :message (if has-lintr "present" "missing")
+                :face (if has-lintr 'success '(bold error)))))))
+
+(defun flycheck-R-has-lintr (R)
```
lunaryorn: Please use a lowercase `r`.  I know that the language name is `R`, but for the sake of consistency I'd like to stick to Emacs Lisp' naming conventions.

flycheck.el
```diff
@@ -7684,7 +7684,25 @@ See URL `https://github.com/jimhester/lintr'."
...
+  "Whether R has installed the `lintr' library."
+  (with-temp-buffer
+    (call-process R nil t nil
+                  "--slave" "--restore" "--no-save" "-e"
+                  "library('lintr')")
+    (goto-char (point-min))
+    (not (looking-at-p "Error in library(\"lintr\") : there is no package called ‘lintr’"))))
```
lunaryorn: Are these really smart quotes here?  Wouldn't that break on non-UTF-8 locales?

flycheck.el
lunaryorn: Also, please move this function before the syntax checker definition :)

lunaryorn:
@marsam Nice, thank you very much :+1:  There are a couple of minor issues which I'd like to ask you ti fix, but other than that it looks good, in my opinion :relaxed:

flycheck.el
marsam: yes they are smart quotes, but can be disabled adding to `LC_ALL=C` the process-environment

```elisp

(let ((process-environment (append '("LC_ALL=C") process-environment)))
  ;; ...
  ;; or simply check for "there is no package called"  
  (not (re-search-forward "there is no package called 'lintr'" nil 'no-error)) 
```

marsam:
Hi, I've updated the pull request, tell me if there is more issues

Issue #901, opened 2016-03-04
Drop in-emacs manual?


lunaryorn:
I'm really fed up with Texinfo.  I'd love to go back to [Sphinx](http://www.sphinx-doc.org) and readthedocs as build service.  ReStructuredText is so much easier to write than Texinfo, and I think it's much more friendly to us as maintainers and to newcomers.

However, Sphinx' support for Texinfo is inferior.  When we used Sphinx in the past we worked around this with a lot of extensions to Sphinx, which were tedious to write and maintain.  I don't want to do that again.

Hence we'd have to drop support for Info manuals essentially.  I for my part am totally for it; I'd rather wait and hope for Emacs to add support for a better manual format that we build to.  Texinfo is only used within Gnu and Emacs nowadays; it's a dead end and a rotten technology.  I'd just throw it out, rather than try and continue to cope with it.

But what do you think?

/cc @cpitclaudel @purcell @jwiegley @bbatsov

bbatsov:
I completely agree.

lunaryorn:
Sometimes I wished we had Google Analytics for Emacs… I'm dying to find out how many of my users actually ever looked at the manual in Emacs :grin:

tzz:
I'm sure many of you have seen https://lwn.net/Articles/625072/

And from me:

https://lists.gnu.org/archive/html/help-gnu-emacs/2010-05/msg00175.html

rule #1 of writing manuals:

They Won't Read It.

followed by rule #2:

They Still Won't Read It.

and then rule #3:

They'll Complain About It.

lunaryorn:
@tzz Thanks for your feedback, but… I must admit that I'm not sure what to make of it.  Do you agree with my point of view? Or not?  Or do you think that all efforts for a manual are vain?

tzz:
@lunaryorn I was trying to say that the vast majority of users don't read manuals, replying to your question about how many of your users look at your manual. Good UI is worth much more than a good manual. So I would use the documentation markup that's most comfortable for the developers with local project scope, which is probably RST.

I am especially skeptic of claims that any particular documentation markup language will solve software usability issues, or that it's causing them.

lunaryorn:
I appreciate your opinion, but this is not the place for it.  If you seek to continue the discussion about Emacs' own manual format, please do so on emacs-devel.

[...]

PR #923, opened 2016-03-28
New Sphinx website

lunaryorn:
This is the start of a new website and manual written with Sphinx, following #901.

[Readthedocs](https://readthedocs.org/projects/flycheck/) already builds this branch, a preview is available at http://flycheck.readthedocs.org/en/sphinx-doc/.

The site already contains our meta-documentation (CoC, contributing, etc.) and the list of supported languages, plus most of the extensions to document Emacs Lisp with Sphinx.

The manual itself is still missing, but a single-page HTML build of the old one is included, until it's been rewritten in Texinfo.

As things stand, I'd merge this pull request asap, and point `http://www.flycheck.org` to the new site, and then start to port the manual in `master`.


lunaryorn:
@cpitclaudel Do you have an account on ReadTheDocs?  I'll give you admin access to the Flycheck project there.

cpitclaudel:
I think it all looks good :) Great progress. I'm cpitclaudel on readthedocs.

I just noticed the very quick start section. It's a great way to get people to try the package, but maybe we should make it clearer that it's just for a test run? If I understand it correctly, it will install Flycheck but it won't register melpa persistently, so subsequent updates won't be downloaded, correct?

lunaryorn:
@cpitclaudel Indeed we should emphasise that it's not permanently installed this way.

I guess we better rename the section to "Try out".

lunaryorn:
@cpitclaudel I merged the branch, added you as a maintainer on RTD and updated the very quick start section.

lunaryorn:
And now http://www.flycheck.org points to the readthedocs site.  I'll remove flycheck.github.io this evening when the change has propagated.

Issue #741, opened 2015-08-30
Pass checkdoc variables to subprocess


kbauer:
Related to https://github.com/flycheck/flycheck/issues/53

Currently `checkdoc` as invoked by flycheck ignores customized variables. Relevant variables include all customization variables of `checkdoc` and at least `sentence-end-double-space`.

I currently solved this locally by a startup script setting `flycheck-emacs-args` to

    (list "-Q" "--batch" "--eval"
          (pp-to-string
           `(setq sentence-end-double-space
                  ',sentence-end-double-space))
          "--eval"
          (pp-to-string
           `(progn
              ,@(cl-loop for var being the symbols
                         if (and (custom-variable-p var)
                                 (string-match-p "^checkdoc-"
                                                 (symbol-name var)))
                         collect `(setq ,var ',(symbol-value var)))))))

lunaryorn:
@kbauer I don't see the need for this myself, because personally I use the default checkdoc settings, and encourage everyone else to do the same, as these settings represent the standard Emacs style for docstrings.

Closing as “wontfix” hence, because I won't implement this feature myself.

However, I'll happily merge pull requests or patches that add these variables to the checkdoc checker.  Note, though, that I do not like your approach of blindly passing all checkdoc variables.  I'd rather prefer if every variables was explicitly forwarded as an `option` to the checkdoc checker.

cpitclaudel:
@lunaryorn, I was looking at this and wondering: have we ever considered support for file-local variables as a way to handle this problem? This would be similar to how some checkers allow you to configure the warnings for a file by adding special comments (and I would consider it good practice to make such settings parts of a file, so that everyone in a project checks the file according to the same standards)

lunaryorn:
@cpitclaudel I'm not sure what you mean.  Do you think the checkdoc syntax checker should read a local variables file explicitly?

cpitclaudel:
Probably, yes; I was wondering what the checkdoc equivalent would be of (say) the following header (taken from a python file of mine):

```python
#!/usr/bin/env python2
# -*- coding: utf-8 -*-
# pylint: disable=too-few-public-methods
```

and it occurred to me that we might be interested in supporting something like

```elisp
;; Local Variables:
;; checkdoc-arguments-in-order-flag: nil
;; End:
```

for checkdoc.

Does that make sense?

lunaryorn:
[...]

PR #937, opened 2016-04-02
Pass relevant settings to checkdoc subprocess

cpitclaudel:
‘flycheck-emacs-lisp-checkdoc-variables’ contains all variables that are
passed to the inferior Emacs process running checkdoc (but only if they
are already bound in the parent process).

The main use of this is allowing for file- and directory-local checkdoc
settings.

Closes GH-741.

Edit: Connects to #741.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6368,11 +6368,33 @@ See Info Node `(elisp)Byte Compilation'."
             (princ (buffer-substring-no-properties (point-min) (point-max)))
             (kill-buffer)))))))
 
+(defconst flycheck-emacs-lisp-checkdoc-variables
+  '(checkdoc-symbol-words
+    checkdoc-arguments-in-order-flag
+    checkdoc-force-history-flag
+    checkdoc-permit-comma-termination-flag
+    checkdoc-force-docstrings-flag
+    checkdoc-package-keywords-flag
+    checkdoc-spellcheck-documentation-flag
+    checkdoc-verb-check-experimental-flag
+    checkdoc-max-keyref-before-warn)
+  "Variables inherited by the checkdoc subprocess.")
+
+(defun flycheck-emacs-lisp-checkdoc-variables-form ()
+  "Make a sexp to pass relevant variables to a checkdoc subprocess.
+
+Variables are taken from `flycheck-emacs-lisp-checkdoc-variables'."
+  `(progn
+     ,@(seq-map (lambda (opt) `(setq-default ,opt ,(symbol-value opt)))
+                (seq-filter #'boundp flycheck-emacs-lisp-checkdoc-variables))))
+
 (flycheck-define-checker emacs-lisp-checkdoc
   "An Emacs Lisp style checker using CheckDoc.
 
 The checker runs `checkdoc-current-buffer'."
   :command ("emacs" (eval flycheck-emacs-args)
+            "--eval" (eval (flycheck-sexp-to-string
+                            (flycheck-emacs-lisp-checkdoc-variables-form)))
             "--eval" (eval flycheck-emacs-lisp-checkdoc-form)
             "--" source)
   :error-patterns
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3285,6 +3285,11 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
    '(11 nil warning "All variables and subroutines might as well have a documentation string"
         :checker emacs-lisp-checkdoc)))
 
+(flycheck-ert-def-checker-test (emacs-lisp-checkdoc) emacs-lisp
+                               inherits-checkdoc-variables
+  (flycheck-ert-should-syntax-check
+   "language/emacs-lisp/local-checkdoc-variables.el" 'emacs-lisp-mode))
+
 (flycheck-ert-def-checker-test (emacs-lisp emacs-lisp-checkdoc) emacs-lisp
                                checks-compressed-file
   (flycheck-ert-should-syntax-check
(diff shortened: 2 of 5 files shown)


cpitclaudel:
Quick note: I added `(hack-local-variables)` to the test setup. An alternative would be let-binding a few variables around the call to `flycheck-should-syntax-check` in the test that I added, but as written the test is more self-contained, and should better reflect the use that people will make of that feature.

lunaryorn:
@cpitclaudel Thank you very much.  I'll take a look.

flycheck.el
```diff
@@ -6368,11 +6368,32 @@ See Info Node `(elisp)Byte Compilation'."
...
 
+(defconst flycheck-emacs-lisp-checkdoc-variables
+  '(checkdoc-symbol-words
+    checkdoc-arguments-in-order-flag
+    checkdoc-force-history-flag
+    checkdoc-permit-comma-termination-flag
+    checkdoc-force-docstrings-flag
+    checkdoc-package-keywords-flag
+    checkdoc-spellcheck-documentation-flag
+    checkdoc-verb-check-experimental-flag)
```
lunaryorn: Could we discover all this options automatically?  Perhaps by finding all options in the `checkdoc` custom group?

test/flycheck-test.el
```diff
@@ -3285,6 +3285,11 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
    '(11 nil warning "All variables and subroutines might as well have a documentation string"
         :checker emacs-lisp-checkdoc)))
 
+(flycheck-ert-def-checker-test (emacs-lisp-checkdoc) emacs-lisp
+                               inherits-checkdoc-variables
+  (flycheck-ert-should-syntax-check
+   "language/emacs-lisp/local-checkdoc-variables.el" 'emacs-lisp-mode))
```
lunaryorn: Did you perhaps forget to include this file?  I can't see it in this PR.

cpitclaudel: Woops. That's right! Fixed.

flycheck.el
cpitclaudel: Sure; I'll push the relevant code soon.

cpitclaudel: Actually, I'm not sure that will work :/ It is quite possible for checkdoc to not be loaded but for a given checkdoc variable to be set, and in that case we still want to propagate the variable, right?
I don't think we can find the list of checkdoc variables without loading the package. We don't want to do that in the parent process, right?

cpitclaudel:
Alright, pushed an updated version; thanks for the review!

flycheck.el
lunaryorn: @cpitclaudel Oh yes, I didn't think of this.

However, I think we should try to make sure that we don't miss a variable here.  Do you think that we can have a buttercup spec for this?

cpitclaudel: Done :)

cpitclaudel:
I've added a buttercup spec checking that the list is up-to-date.

test/specs/languages/test-emacs-lisp.el
```diff
@@ -0,0 +1,57 @@
...
+    (it "exports the right customizable variables"
+      (require 'checkdoc)
+      (let* ((checkdoc-custom-group
+              (get 'checkdoc 'custom-group))
+             (checkdoc-custom-pairs
+              (seq-filter #'tel--custom-var-p checkdoc-custom-group))
+             (checkdoc-custom-vars
+              (seq-map #'car checkdoc-custom-pairs))
+             (checkdoc-relevant-vars
+              (cl-set-difference checkdoc-custom-vars tel--excluded-checkdoc-vars)))
```
lunaryorn: Please use `seq-difference` instead of `cl-set-difference` here.  Just my personal preference for `seq` over `cl-lib` :relaxed:

test/specs/languages/test-emacs-lisp.el
```diff
@@ -0,0 +1,57 @@
...
+        ;; Test for subset, since all variables aren't defined in all Emacsen
```
lunaryorn: I'd prefer if we could test for actual equality, and instead limit the test to a particular Emacs version, i.e. with `(assume (version<= "25" emacs-version))`. That's safer, imho.

test/specs/languages/test-emacs-lisp.el
```diff
@@ -0,0 +1,57 @@
...
+
+;;; Commentary:
+
+;; Specs for Emacs Lisp support.
+
+;;; Code:
+
+(require 'flycheck-buttercup)
+
+(defconst tel--excluded-checkdoc-vars
```
lunaryorn: I don't think that we need a special prefix for each test file.  My other tests just use `flycheck/`, i.e. a trailing slash instead of dash to avoid conflicts with Flycheck itself.  That should be sufficient, and it's easier to read imho.  It took me a while to understand what `tel` actually stands for :blush:

cpitclaudel: Great, `flycheck/` is a good choice.

test/specs/languages/test-emacs-lisp.el
cpitclaudel: Ooh, I didn't realize that was in seq as well.

cpitclaudel: Is there anything for subsetp?

test/specs/languages/test-emacs-lisp.el
cpitclaudel: Got it. Let's do it for 25 though, since Emacs 25 is the one that has more checkdoc variables.

cpitclaudel:
Ok, build fixed :)

lunaryorn:
@cpitclaudel Merged, with some little refinements.  Specifically I noticed that we need to bind `enable-local-variables` to `:safe` around `(hack-local-variables)` to make sure that Emacs doesn't prompt for configuration when a resource file contains unsafe variables.  That'd make the test suite stuck.

cpitclaudel:
Awesome; nice catch!

Issue #899, opened 2016-03-02
A checker for textlint?


kozross:
I recently came across [this](https://github.com/textlint/textlint), and thought it would make for a really nice Flycheck checker in a place where Flycheck hasn't really gone before (ordinary text). I don't know if this falls outside its scope, but it'd be something I'd really love to have.

lunaryorn:
@kozross It's surely within our scope, and it's a great idea.  Please feel free to open a pull request :relaxed:

kozross:
I came across something *very* similar which could be integrated [here][1]. What should I do to make this part of 'core' Flycheck, if anything? 

[1]: https://github.com/amperser/proselint/blob/master/plugins/flycheck/flycheck-proselint.el

lunaryorn:
@kozross We would happily accept a pull request for this syntax checker.  Please follow our [Pull Request](https://github.com/flycheck/flycheck/blob/master/CONTRIBUTING.md#pull-requests) guide lines for further information.

We would like to ask that the pull request maintains proper authorship, though, i.e. commits should be attributed to @suchow.  And we prefer if @suchow explicitly agreed to add their syntax checker to Flycheck.

suchow:
You have my blessing to the extent that I am legally able to give it. A similar version of that code was originally posted on http://unconj.ca/blog/linting-prose-in-emacs.html, and then modified by @unhammer in https://github.com/amperser/proselint/issues/37.

unhammer:
Is textlint the same as proselint? (I don't think my .el changes were big enough to be copyrightable, but just in case I hereby allow licensing it under whatever flycheck uses.)

suchow:
@unhammer — they're different, but it seems this issue has expanded in scope a little to include proselint.

robstewart57:
I've written up an approach that works well for me: https://www.macs.hw.ac.uk/~rs46/posts/2018-12-29-textlint-flycheck.html

This implements textlint as a flycheck checker.

textlint doesn't apply any rules automatically, they must be explicitly installed with `npm` and specified explicitly with `--rule` when running `textlint`. This is a key design philosophy of textlint: as a __pluggable__ natural language linting tool.

Does this hold back the idea of building a textlint checker directly into flycheck?

fmdkdd:
> Does this hold back the idea of building a textlint checker directly into flycheck?

It would mean duplicating the rules in your Emacs config, so it would add friction.

Can you specify the rules in a config file instead?

[...]

Issue #836, opened 2016-01-02
New system for chaining checkers


swsnr:
Edit: Take a look at the [corresponding project](https://github.com/flycheck/flycheck/projects/1) for status updates.

Our current system with `:next-checkers` works fine, but it has a couple of serious drawbacks:
- It's not at all easy to understand.
- It's overly verbose and repetitive. Just take a look at `:next-checkers` of our go checkers :unamused: 
- It's very hard to extend correctly. `flycheck-add-next-checker` can break subtly if the order of calls or the order of insertions is wrong.

I think our current system was a nice try, but it's more or less a failure in my opinion.  I didn't anticipate Flycheck's success and the amount of languages and syntax checkers we'd support as well as the desire of people to add their own syntax checkers.  

We need to come up with a better system that is able to figure out automatically which syntax checkers to run and which to skip, based on a few simple declarations.

What I have in mind is this:
- Add a `:maximum-level` property to syntax checker declarations (the exact name to be determined) which gives the maximum error level at which the syntax checker still runs, i.e. with `:maximum-level error` the syntax checker will run if there are warnings in the current buffer but not if there are errors.
- Add a `:priority` which is a number that denotes when a checker will run.  Higher priority syntax checkers run first, and only the first syntax checker of a given priority runs.

I think that simpler, but I'm really open to other suggestions!

purcell:
Yes! It would be great to overhaul this aspect of flycheck. It's certainly a bit odd that individual checkers should specify which other checkers they should chain with. _How_ to check and _when_ to check seem like orthogonal concerns. For that matter, I'm not sure replacing `:next-checkers` with `:maximum-level` and `:priority` properties in checker definitions would improve matters.

Why not a separate nested chaining order list, or an optionally richer structure for `flycheck-checkers`?

lunaryorn:
@purcell I'm not sure. A richer structure for `flycheck-checkers` would make it harder to add new syntax checkers to the proper place, I presume.  Likewise for a separate list that denotes chains.

I'd like to put as much information as possible into the syntax checker definition, to make setup as easy as possible.

purcell:
I guess that if there were a way for users to update a checker's `:maximum-level` and `:priority` after it had been defined, then the same results would be possible.

lunaryorn:
I'm not sure why the user should be able to update these?  They are specific to the syntax checker; it makes no sense to update these settings afterwards.

[...]

PR #939, opened 2016-04-03
Add proselint checker

marsam:
Add support for http://proselint.com/

Related: #899

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -75,6 +75,7 @@
   )
 
 (require 'dash)
+(require 'json)
 
 (require 'seq)                   ; Sequence functions
 (require 'subr-x nil 'no-error)  ; Additional utilities, Emacs 24.4 and upwards
@@ -218,6 +219,7 @@ attention to case differences."
     php-phpmd
     php-phpcs
     processing
+    proselint
     puppet-parser
     puppet-lint
     python-flake8
@@ -7840,6 +7842,39 @@ Requires Sphinx 1.2 or newer.  See URL `http://sphinx-doc.org'."
   :modes rst-mode
   :predicate (lambda () (and (flycheck-buffer-saved-p)
                              (flycheck-locate-sphinx-source-directory))))
+
+(defun flycheck-proselint-parse-errors (output checker buffer)
+  "Parse proselint json output errors from OUTPUT.
+
+CHECKER and BUFFER denoted the CHECKER that returned OUTPUT and
+the BUFFER that was checked respectively.
+
+See URL `http://proselint.com/' for more information about proselint."
+  (let-alist (json-read-from-string output)
+    (mapcar (lambda (err)
+              (let-alist err
+                (flycheck-error-new-at
+                 .line
+                 .column
+                 (pcase .severity
+                   (`"suggestion" 'info)
+                   (`"warning"    'warning)
+                   (`"error"      'error)
+                   ;; Default to error
+                   (_             'error))
+                 .message
+                 :id .check
+                 :buffer buffer
+                 :checker checker)))
+            .data.errors)))
+
+(flycheck-define-checker proselint
+  "Flycheck checker using Proselint.
+
+See URL `http://proselint.com/'."
+  :command ("proselint" "--json" source)
+  :error-parser flycheck-proselint-parse-errors
+  :modes (text-mode markdown-mode))
 
 (flycheck-def-config-file-var flycheck-rubocoprc ruby-rubocop ".rubocop.yml"
   :safe #'stringp)
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -922,6 +922,12 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       Check SQL syntax with `Sqlint <https://github.com/purcell/sqlint>`_.
 
+.. supported-language:: Text
+
+   .. syntax-checker:: proselint
+
+      Check text prose with `Proselint <http://proselint.com/>`_.
+
 .. supported-language:: TeX/LaTeX
 
    Flycheck checks TeX and LaTeX with either `tex-chktex` or `tex-lacheck`.
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3845,6 +3845,19 @@ Why not:
    '(26 12 error "TRUE, FALSE and NULL must be lowercase; expected \"false\" but found \"FALSE\""
         :id "Generic.PHP.LowerCaseConstant.Found" :checker php-phpcs)))
 
+(flycheck-ert-def-checker-test proselint (text markdown) nil
+  (flycheck-ert-should-syntax-check
+   "language/text.txt" 'txt-mode
+   '(1 5 warning "Substitute 'damn' every time you're inclined to write 'very;' your editor will delete it and the writing will be just as it should be."
+       :id "weasel_words.very"
+       :checker proselint)
+   '(2 4 warning "Redundancy. Use 'associate' instead of 'associate together'."
+       :id "redundancy.garner"
+       :checker proselint)
+   '(3 5 warning "Gender bias. Use 'lawyer' instead of 'lady lawyer'."
+       :id "sexism.misc"
+       :checker proselint)))
+
 (flycheck-ert-def-checker-test processing processing syntax-error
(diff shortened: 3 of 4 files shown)


flycheck.el
```diff
@@ -7841,6 +7843,40 @@ Requires Sphinx 1.2 or newer.  See URL `http://sphinx-doc.org'."
...
 
+(defun flycheck-proselint-parse-errors (output checker buffer)
+  "Parse proselint json output errors from OUTPUT.
+
+CHECKER and BUFFER denoted the CHECKER that returned OUTPUT and
+the BUFFER that was checked respectively.
+
+See URL `http://proselint.com/' for more information about proselint."
+  (let* ((body (json-read-from-string output))
+         (errors (let-alist body .data.errors)))
```
lunaryorn: Can shrink this to a single outer `let-alist`? I.e.

```
(let-alist (json-read-from-string output)
   (mapcar … .data.errors))
```

lunaryorn:
@marsam Great, thank you :+1: Would you mind to add documentation for this syntax checker in `doc/languages.rst`?

marsam:
Updated

manuel-uberti:
Oh yes please! Great idea, thank you. :+1:

manuel-uberti:
Any update on this one?

lunaryorn:
@manuel-uberti I'll merge it when I find the time.

flycheck.el
```diff
@@ -7841,6 +7843,39 @@ Requires Sphinx 1.2 or newer.  See URL `http://sphinx-doc.org'."
...
+                 :checker checker)))
+            .data.errors)))
+
+(flycheck-define-checker proselint
+  "Flycheck checker using Proselint.
+
+See URL `http://proselint.com/'."
+  :command ("proselint" "--json" source)
+  :error-parser flycheck-proselint-parse-errors
+  :modes (text-mode markdown-mode))
```
mkcode: Consider adding `gfm-mode` here as well.

mkcode:
@lunaryorn - I opened a PR on the spacemacs syl20bnr/spacemacs#6012 to introduce something similar. I think that using the `modes` option is correct here, on spacemacs we would want to use the `predicate` option that is hooked up to a `spacemacs toggle`. 

When you get a chance, could you instruct the best way to override and change the modes option to a predicate within the spacemacs layer. Thanks!

lunaryorn:
@mkcode  There's no official way to add or override the `:predicate` of a syntax checker, by design.  A `:predicate` should check for the requirements of a syntax checker; you should not use it to check user options or toggles.   If you'd like to disable a syntax checker then please add it to `flycheck-disabled-checkers`.

mkcode:
> There's no official way to add or override the :predicate of a syntax checker, by design.

Appreciate the feedback. In that case, what arguments could I provide to `define-flycheck-checker so that a checker may be run everywhere (in all modes)? I do not see any examples of this being done.

Appreciate the feedback, perhaps we should continue in https://github.com/syl20bnr/spacemacs/pull/6012

lunaryorn:
@mkcode  I'm afraid that Flycheck doesn't support this case currently, and I'm not sure whether I'd like to add this feature to Flycheck.  I'm not sure what kind of use case your trying to cover, though.

Would you mind to elaborate why you need the checker behave as you said?  Why would you like to *remove* the list of modes?  A syntax checker *without* modes per se is something that I'll remove sooner or later because it makes many things harder than they should be.  IOW, at some point *every* Flycheck checker has to declare their major modes just like every buffer has a major mode.

I'm sorry but we'll have to discuss at this place; I'm not going to join Spacemacs' channels.

mkcode:
> Would you mind to elaborate why you need the checker behave as you said?

Sure. What I want out of the proselint checker is to have it run in all modes. And I also would like the regular modes checkers to run. For example, in Emacs-Lisp mode, I would like (proselint, emacs-lisp, and emacs-lisp-checkdoc) checkers to run. The reason for this, is I would like proselint feedback in my code comments as well.

I seem to have actually made it work actually: https://gist.github.com/d40dfe8fcd0e854b611055f63d276fd3 
In this gist, toggling the `prose-proselint-enabled` to non-nil will make the predicate function return true and enable proselint (in any mode). I then have some magic for adding the mode specific checkers to proselint's `next-checkers` in a `flycheck-before-syntax-check-hook`

> I'm sorry but we'll have to discuss at this place; I'm not going to join Spacemacs' channels.

Understood.

lunaryorn:
Can you use proselint in source code files?!

mkcode:
> Can you use proselint in source code files?!

As far as I can tell, yes - if you provide the filename directly to the `proselint` executable. Providing a dir will filter based on the file extension. Works for me if I explicitly give a file on the command line and more importantly - with flycheck.

screen shot 2016-05-13 at 12 59 30 pm

JAremko:
@lunaryorn 
>Can you use proselint in source code files?!

They have live demo http://proselint.com/write/ you can try it.  I tested .go and .el code and it looks ok except  "curly quotes"  warning is annoying.  Probably it can be disabled with a proselint configuration file.

lunaryorn:
@mkcode I see.  What you'd like to do is not supported by Flycheck yet.  Your hack is clever, but we don't actually give any guarantees that this works; it's a side-effect of how syntax checker selection is implemented currently.

However, I think we just need two changes to Flycheck:

* #836 for a more flexible chaining system which is already in progress in #961.
* More flexible rules for `:modes`, e.g. the ability to not only say `:modes (text-mode …)` but rather something like `:modes (derived text-mode) (derived prog-mode)` or so.

The latter is a rather minor addition to Flycheck as far as I can see, but there's no ticket yet.

@mkcode Do you think you could implement your change to Spacemacs on top of these features?  We could hold back this PR until both requirements are implemented and then update it accordingly.

JAremko:
@lunaryorn   
> something like :modes (derived text-mode) (derived prog-mode) or so.  

That will be really cool :+1:   But it also might be somewhat tricky to implement. For example `web-mode` is derived from `web-mode-prog-mode` - alias for `prog-mode`.  Or `typescript-mode` and `typescript-parent-mode`.  `derived-mode-p` will not detect it as a derived from `prog-mode`  and probably modes that aren't directly derived also won't be detected.

lunaryorn:
@marsam @mkcode @JAremko I'd like to hold this syntax checker back until #961 is merged, and then update this syntax checker right away as outline in my previous comment so that it works seamlessly with all kinds of modes.  Is that ok for you?

marsam:
Sure, no problem :)

jpfairbanks:
I am looking forward to this PR and noticed that issue #961 was superceded by issue #1072. I I just wanted to link to the new issue number here.

manuel-uberti:
Any progress on this? If there is something I can do to help, please tell. :)

lunaryorn:
@manuel-uberti We can't merge this checker before #1072 because it doesn't fit in our current system for chaining.

waltervargas:
Hi @mkcode and @lunaryorn, there is some update on this?, there is something where can I help?

cpitclaudel:
@waltervargas Not much of an update; we're focusing on simple PRs to merge before tackling larger changes like this one (which is somewhat dependent on #1072, which is currently on hold)

JAremko:
Any progress? :roll_eyes:

marsam:
Hi, given that #1072 is taking some time, can this PR be merged? 

P.S. A package with this checker to be included in melpa https://github.com/melpa/melpa/pull/4938, but I think is better to be part of flycheck

fmdkdd:
@marsam Sure, we can merge this.  LGTM.  You just need to rebase on master.

If we include this, then we should probably inform the ["official" plugin](https://github.com/amperser/proselint/blob/master/plugins/flycheck/flycheck-proselint.el) that they don't need to bundle it themselves.

mavit:
> If we include this, then we should probably inform the "official" plugin that they don't need to bundle it themselves.

I'm happy to do that, once this merges.

marsam:
I've made a new PR with these changes rebased in #1304.  Closing this because
having my old fork deleted, GitHub does not allow me to rebase this PR.

flycheck.el
gdkrmr: what about `org-mode`?

PR #959, opened 2016-05-08
add motion-mode support for rubocop + rubylint

jojojames:
motion-mode is a derived major mode from ruby for rubymotion

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8004,7 +8004,7 @@ See URL `http://batsov.com/rubocop/'."
    (error line-start (file-name) ":" line ":" column ": " (or "E" "F") ": "
           (optional (id (one-or-more (not (any ":")))) ": ") (message)
           line-end))
-  :modes (enh-ruby-mode ruby-mode)
+  :modes (enh-ruby-mode ruby-mode motion-mode)
   :next-checkers ((warning . ruby-rubylint)))
 
 ;; Default to `nil' to let Rubylint find its configuration file by itself, and
@@ -8029,7 +8029,7 @@ ruby-lint 2.0.2 or newer is required.  See URL
             (file-name) ":W:" line ":" column ": " (message) line-end)
    (error line-start
           (file-name) ":E:" line ":" column ": " (message) line-end))
-  :modes (enh-ruby-mode ruby-mode))
+  :modes (enh-ruby-mode ruby-mode motion-mode))
 
 (flycheck-define-checker ruby
   "A Ruby syntax checker using the standard Ruby interpreter.


lunaryorn:
Thanks! 👏 What about the other Ruby checkers?

jojojames:
I wasn't too familiar with the other checkers so I erred on the safe side.

chef-foodcritic, ruby-jruby were ones I wasn't sure about.

We can probably add motion-mode to the ruby checker. Should I do a second pull request for that or is there a way I can merge it with this pull request?

cpitclaudel:
@jojojames Thanks for your contributions! You can definitely add to the current pull request by pushing more commits to the same branch; and when comes the time to merge, we'll ask you to rebase your changes ("squash" them) into one commit, to keep a clean history :)

cpitclaudel:
Btw, @lunaryorn, have we considered checking whether the current mode is derived from one of the ones listed in the `:modes` annotation in the checker definitions, instead of checking for strict equality? Do we have examples where a checker that's appropriate for a parent mode isn't for a derived mode?

cpitclaudel:
(I'm guessing that part of the problem would be when you have competing checkers: one for the parent mode, and one for the derived one)

jojojames:
@cpitclaudel 

Thanks! Was there an opinion on including motion-mode with these two checkers?

chef-foodcritic, ruby-jruby

lunaryorn:
@cpitclaudel Sure, there are examples: JS and JSX Mode, for instance.  The latter is derived from the former, but not all Javascript checkers cope with JSX.

I thought about adding a `(derived 'foo-mode)` form for `:modes` a while back that'd let checkers explicitly opt in for derived modes, but there hasn't been any need for this feature yet.

@jojojames I'm sorry but you'll need to answer this question.  I don't even know what Motion Mode is.

Simplify:
@lunaryorn Motion mode is mode derived from ruby-mode for RubyMotion projects (building iOS, OS X and Android apps with Ruby). RubyMotion have his own Ruby implementation based om Apple's MacRuby with ahead-of-time compiler. I have my doubts about using `ruby` and/or `ruby-jruby`checkers with RubyMotion projects because of different Ruby implementation. Also, if you check code of [motion-mode](https://github.com/ainame/motion-mode/blob/master/motion-mode.el#L175) they use RubyMotion's own Ruby implementation with flymake. Error pattern looks differently than pattern standard MRI.
`chef-foodcritic` is definitely off the list for `motion-mode`.

lunaryorn:
@Simplify If it's that different can we even safely use it with rubocop and ruby-lint?  As far as I know the former uses MRI's parser, doesn't it?  /cc @bbatsov

Simplify:
@lunaryorn That is something that someone with more RubyMotion knowledge can answer. I used it only once, when it came out, just to try it, and never again.

jojojames:
I'm not too experienced with the different ruby implementations but rubocop is working fine for me in my RubyMotion project (as far as I can tell).

I wasn't a fan of using flymake with motion-mode so opted to used flycheck instead.

Maybe someone with more experience in this domain can point us in the right direction.

bbatsov:
> @Simplify If it's that different can we even safely use it with rubocop and ruby-lint? As far as I know the former uses MRI's parser, doesn't it? /cc @bbatsov

I'm not familiar with RubyMotion.

lunaryorn:
I'd appear that no one really knows in how far RubyMotion is compatible with Ruby, so I'm inclined to close this pull request, and rather refer people to `flycheck-add-mode` to add `motion-mode` themselves if needed, at their own risk.  @jojojames I'm sorry but if even @bbatsov himself doesn't know whether rubocop supports RubyMotion, I don't think Flycheck should go along and say otherwise.

PR #964, opened 2016-05-12
Port Checkstyle ERT tests to buttercup

Simplify:


test/specs/test-error-parsers.el
```diff
@@ -0,0 +1,99 @@
...
+    :buffer 'buffer
+    :line 1
+    :column nil
+    :level 'error
+    :message "Cannot read property 'id' of undefined"
+    :id "Foo1"))
+  "Errors to be parsed from `flycheck-checkstyle-xml'.")
+
+(describe "Error parsers"
+  (describe "Checkstyle XML parsing"
```
lunaryorn: Use "The checkstyle parser" as argument for `describe`.  That reads better in the test output imho.

lunaryorn:
@Simplify Great, thanks.

I don't think that we need `defconst` here, though.  As far as I can see there's no reason to have these names available as global variables, so a `let` binding around all `it`s should be enough.

Please use a `let` binding instead of `defconst`s.  I've just refactored the phpmd XML accordingly as well.

Simplify:
@lunaryorn Updated. tslint parser test need updating as well?

lunaryorn:
@Simplify It's buttercup, not butterfly :wink:

Simplify:
@lunaryorn hehehehe, fixed :)

lunaryorn:
@Simplify Thanks.  I've updated the tslint spec.

Please feel free to push simple refactorings in the tests directly to master.  Of course, pull requests are always fine as well, if you'd like to get a review or collect feedback :)

Simplify:
@lunaryorn Yes, I prefer review and feedback until I get more familiar with Flycheck and your vision for this project.

PR #970, opened 2016-05-17
Strawman: Dynamic adjustment based on buffers

chrisdone:
This is a strawman (not a proper cleaned up commit, NOT INTENDED TO BE MERGED), to enable discussion on the best UX for supporting additional buffers. The code is intentionally poorly formatted with dangling `(` and `)` so that the diff is a bit smaller. In the proper PR we'll format properly.

This is one of two strawman PRs I'm opening for the purpose of comparing.

This PR will, when given an error list that contains an error whose source buffer does not match the currently-being-flychecked buffer:

* Add another column, "Buffer", to the tabulate columns (`flycheck-error-list-format`).
* Re-generate that tabulation (`flycheck-retabulate`).
* Add that data to the output columns (`flycheck-error-list-make-entry`).

Copy/paste the code below and if you run it, you will get two buffers:

``` lisp
(flycheck-define-generic-checker 'cross-buffer-errors
  "A dummy checker that always returns an error for A.txt."
  :start 'cross-buffer-errors-check
  :modes '(text-mode)
  :next-checkers '())
(add-to-list 'flycheck-checkers 'cross-buffer-errors)

(defun cross-buffer-errors-check (checker cont)
  ;; Need to run with a timer because this function must be async.
  (run-with-timer 0 nil
                  'cross-buffer-errors-call-in-buffer
                  (current-buffer)
                  (lambda (checker cont)
                    (funcall cont
                             'finished
                             (list (flycheck-error-new-at
                                    1 1 'error
                                    "Something bad"
                                    :checker checker
                                    :buffer (get-buffer-create "A.txt")))))
                  checker
                  cont))

(defun cross-buffer-errors-call-in-buffer (buffer func &rest args)
  "Utility function which calls FUNC in BUFFER with ARGS."
  (with-current-buffer buffer
    (apply func args)))

(progn (let ((a (get-buffer-create "A.txt"))
             (b (get-buffer-create "B.txt")))
         (with-current-buffer a
           (text-mode)
           (flycheck-select-checker 'cross-buffer-errors)
           (flycheck-mode))
         (with-current-buffer b
           (text-mode)
           (flycheck-select-checker 'cross-buffer-errors)
           (flycheck-mode))
         (switch-to-buffer a)
         (call-interactively 'flycheck-list-errors)))
```

* A.txt
* B.txt

If you switch between the two buffers with your cursor, you should see that the "Buffer" column appears and disappears depending on whether you're seeing the error from A.txt or B.txt. Like this:

screenshot 2016-05-17 12 15 58

Personally, I'm not sure that this dynamic adjustment is a good idea. What do you think? While neat, it seems a bit jarring that the column appears and disappears. Perhaps the other option; that you just enable it always-on as a user option, might be better. I'm going to open a completely separate PR for that.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -746,6 +746,12 @@ This variable is a normal hook.  See Info node `(elisp)Hooks'."
   :group 'flycheck-faces
   :package-version '(flycheck . "0.16"))
 
+(defface flycheck-error-list-buffer-name
+  '((t :inherit buffer-menu-buffer))
+  "Face for buffer names in the error list."
+  :group 'flycheck-faces
+  :package-version '(flycheck . "0.16"))
+
 (defface flycheck-error-list-column-number
   '((t :inherit font-lock-doc-face))
   "Face for line numbers in the error list."
@@ -2919,6 +2925,7 @@ Return the created overlay."
   ;; We must have a proper error region for the sake of fringe indication,
   ;; error display and error navigation, even if the highlighting is disabled.
   ;; We erase the highlighting later on in this case
+  (when (eq (flycheck-error-buffer err) (current-buffer))
   (pcase-let* ((`(,beg . ,end) (flycheck-error-region-for-mode
                                 err (or flycheck-highlighting-mode 'lines)))
                (overlay (make-overlay beg end))
@@ -2939,6 +2946,7 @@ Return the created overlay."
                     level flycheck-indication-mode)))
     (overlay-put overlay 'help-echo (flycheck-error-message err))
     overlay))
+  )
 
 (defun flycheck-filter-overlays (overlays)
   "Get all Flycheck overlays from OVERLAYS."
@@ -3094,19 +3102,25 @@ the beginning of the buffer."
     map)
   "The keymap of `flycheck-error-list-mode'.")
 
-(defconst flycheck-error-list-format
-  [("Line" 4 flycheck-error-list-entry-< :right-align t)
-   ("Col" 3 nil :right-align t)
-   ("Level" 8 flycheck-error-list-entry-level-<)
-   ("Message" 0 t)
-   (" (Checker)" 8 t)]
-  "Table format for the error list.")
+(defun flycheck-error-list-format (&optional buffers-p)
+  "Table format for the error list.
+
+If we're displaying buffer names, then BUFFERS-P should be true."
+  (vconcat
+   (if buffers-p
+       [("Buffer" 6)]
+     [])
+   [("Line" 4 flycheck-error-list-entry-< :right-align t)
+    ("Col" 3 nil :right-align t)
+    ("Level" 8 flycheck-error-list-entry-level-<)
+    ("Message" 0 t)
+    (" (Checker)" 8 t)]))
 
 (define-derived-mode flycheck-error-list-mode tabulated-list-mode "Flycheck errors"
   "Major mode for listing Flycheck errors.
 
 \\{flycheck-error-list-mode-map}"
-  (setq tabulated-list-format flycheck-error-list-format
+  (setq tabulated-list-format (flycheck-error-list-format)
         ;; Sort by location initially
         tabulated-list-sort-key (cons "Line" nil)
         tabulated-list-padding 1
@@ -3114,6 +3128,12 @@ the beginning of the buffer."
         ;; `revert-buffer' updates the mode line for us, so all we need to do is
         ;; set the corresponding mode line construct.
         mode-line-buffer-identification flycheck-error-list-mode-line)
+  (tabulated-list-init-header))
+
+(defun flycheck-retabulate (buffers-p)
+  "To be run in the messages buffer: retabulate with (possibly)
+  new column configuration. BUFFERS-P to display buffer names."
+  (setq tabulated-list-format (flycheck-error-list-format buffers-p))
   (tabulated-list-init-header))
 
 (defvar-local flycheck-error-list-source-buffer nil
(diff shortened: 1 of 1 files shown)


chrisdone:
(I'm hoping after we come to a decision on the best approach, then I'll open a proper PR that handles all the edge-cases and such.)

lunaryorn:
@chrisdone I'm not sure whether this is the best approach.  

Why did you use the `buffer` slot, and not `filename`?

chrisdone:
> I'm not sure whether this is the best approach.

Me neither -- working on an alternate configuration-based version.

> Why did you use the `buffer` slot, and not `filename`?

Either seems to work, but from what I can tell the difference is that:

* `flycheck-error-filename` is used for filtering. As [here](https://github.com/chrisdone/flycheck/blob/dynamic-column-adjustment/flycheck.el#L2516..L2528). But seems to be optional.
* `flycheck-error-buffer` is used for jumping to the error. As [here](https://github.com/chrisdone/flycheck/blob/dynamic-column-adjustment/flycheck.el#L3313..L3333) and [here](https://github.com/chrisdone/flycheck/blob/dynamic-column-adjustment/flycheck.el#L2324..L2331). Without this, I don't expect things work well.

So I tried to follow that tradition of using the buffer as the canonical "place" of the errors.

Do you have an alternative suggestion? When to use the buffer name and when to use the filename is a bit unclear to me. One thing I suppose to consider: is Flycheck supposed to work on a buffer which as no filename? It seems to be the case; the example code above doesn't specify filenames for buffer A and B.

lunaryorn:
@chrisdone Ah, I see.  

Currently the line between buffer and filename is very blurry, simply because there wasn't any need for a strict difference anyway: We'd always stay within the same buffer :)

Now that we need to sharpen the distinction I'm not sure what's best.  

Instinctively I'd use `filename`, because that's what we already get from error parsing—unless the buffer has no filename, that is.  But if we were to use `buffer` what do we do with files that have no buffers?  Your original proposal used `find-file-noselect` to get hold of a buffer for a file, but that's not an option: Visiting arbitrary files as way too many side-effects to do non-interactively deep in a signal handler of a process.

chrisdone:
> But if we were to use buffer what do we do with files that have no buffers?

Are there checkers which generate errors without a buffer? The logic in `flycheck-jump-to-error` seems to assume that there's always a buffer:

``` lisp
(defun flycheck-jump-to-error (error)
  "Go to the location of the error."
  (let ((buffer (flycheck-error-buffer error)))
    (when (buffer-live-p buffer)
      (if (eq (window-buffer) (get-buffer flycheck-error-list-buffer))
          ;; When called from within the error list, keep the error list,
          ;; otherwise replace the current buffer.
          (pop-to-buffer buffer 'other-window)
        (switch-to-buffer buffer))
      (let ((pos (flycheck-error-pos error)))
        (unless (eq (goto-char pos) (point))
          ;; If widening gets in the way of moving to the right place, remove it
          ;; and try again
          (widen)
          (goto-char pos)))
      ;; Re-highlight the errors
      (flycheck-error-list-highlight-errors 'preserve-pos))))
```

> Your original proposal used find-file-noselect to get hold of a buffer for a file, but that's not an option: Visiting arbitrary files as way too many side-effects to do non-interactively deep in a signal handler of a process.

That's true, visiting a buffer can trigger all sorts of things, which is not the kind of thing ideally we would do from inside a process filter.

Perhaps, then, providing a filename and then updating `flycheck-jump-to-error` to automatically `find-file` (which IIRC are forced into being absolute elsewhere in flycheck), rather than assuming the buffer exists. Finding a file is an acceptable side-effect when the user requests to go to that file. I can change the code to go in this direction. Perhaps in PR #971?

lunaryorn:
@chrisdone I guess you refer to `flycheck-error-list-goto-error`? Flycheck can safely assume that because we actually discard all errors not reported for the current file—take a look at `flycheck-relevant-error-p` and its call sites.

> Perhaps, then, providing a filename and then updating flycheck-jump-to-error to automatically find-file (which IIRC are forced into being absolute elsewhere in flycheck), rather than assuming the buffer exists. Finding a file is an acceptable side-effect when the user requests to go to that file. I can change the code to go in this direction. Perhaps in PR #971?

Yes, something along this line would work, I think.  I haven't taken a look at #971 yet, and I don't think I'll get to it today, so I can't comment on whether it's the better approach yet.

But generally I think we should lift the requirement for errors to be from the current buffer only.  We shouldn't discard errors from other files early on and instead preserve those in `flycheck-current-errors`.  The error list could then show all errors reported from the checker, regardless from which file, and could navigate all of them, but we'd only add overlays for errors that a strictly from the current buffer.

I'm not too familiar with those parts of Flycheck anymore for it's long since I last touched them, but I'm pretty sure that this wouldn't be too complicated.  We'd have to modify `flycheck-error-list-goto-error` to visit files if necessary, and change `flycheck-add-overlay` to ignore errors that are not from the current buffer—we can't reasonably highlight these in the current buffer anyway.   As far as I can see this would be it.  In-buffer error navigation (`C-c ! n` and friends) work on the existing overlays and won't need any change.

In this case the error list would show a `"File name"` column, obviously, but in my opinion it could always be visible.  If we just show the base name of the file there it'd be short enough, wouldn't it?

What do you think?

chrisdone:
That all sounds proper to me. I'll open a final third PR that puts all this into practice. I think we're in agreement in the behaviour and the implementation. :+1:

cpitclaudel:
Just chiming in to say that this looks great to me. In particular, the filename column being always on seems quite fine. And opening the file on goto-error seems very reasonable (just like clicking on an error in a compilation buffer)

Issue #966, opened 2016-05-16
Displaying errors from other dependent buffers


chrisdone:
Sorry to bring up an old issue (#829). I'd assumed there was a way in flycheck to display messages for other files. I'm writing a mode for Haskell, [Intero](https://github.com/chrisdone/intero), which uses flycheck as its main means of type-checking. It's basically a fork of GHCi and uses that to get very quick feedback. 

I've been trying it out and it works nice, but sometimes if I change a type in e.g. `Types.hs` and switch to `Packages.hs`, the packages module no longer type checks because `Foo.hs` doesn't type check, but flycheck doesn't indicate it. A file that imports an invalid file is by definition not valid. It misleads me. 

I assumed there must be some way and asked @dysinger who's been using flycheck for a while. He said when he wants that, he hits `C-c C-l` which uses the haskell-mode interactive mode session.  I've no intention of abandoning flycheck due to this issue. I'd like to ride on the waves of Flycheck goodness rather than re-implementing this great functionality.

So my current thinking is that in [my checker Elisp](https://github.com/chrisdone/intero/blob/master/elisp/intero.el#L176):
- There's an error, but it comes from a file that this file imports.
- Generate a new error on line 1 column 1 that says "Error in another file: X.hs"

At this point the user knows to switch to X.hs and then flycheck will show the errors for it. At least this way there is a clear process when things go wrong. It looks like this:

<img width="1440" alt="screenshot 2016-05-16 18 09 47" src="https://cloud.githubusercontent.com/assets/11019/15295577/6705f8e2-1b91-11e6-87d9-2a28409725bf.png">

But it's a but unsatisfying. Ideally, there'd be an additional column like "File" that becomes visible when there're multiple files. Then when I'm trying to check "X.hs" I can quickly jump from there to the problem in "Y.hs" to fix the issue. It's sort of a mid-level functionality:
- Flycheck currently and, from reading this issue backlog (#829), only intends on, dealing with just this buffer
- In the middle, I want to check just one file, not the whole project, but I do care about dependencies.
- Compilation mode or a `stack build` deals with the whole project.

With haskell-interactive-mode, I do get this middle road solution. But that mode re-implements basically all the functionality in flycheck (line overlays, fringe displays, list of errors). A practical piece of work, considering at the time I wrote that code, flycheck didn't support generic checkers that could talk to a process. But now in my old age, I don't feel like re-implementing things from scratch if I can help it.

[...]

PR #971, opened 2016-05-17
Support errors from other buffers 

chrisdone:
Here's Take 2 of supporting errors from other buffers (#966). The other PR was https://github.com/flycheck/flycheck/pull/970.

This version provides a customization option: `flycheck-error-list-buffer-name` If you enable it, the "Buffer" column is *always* visible.

It also adds support to `flycheck-next-error` so that if there're no places left to jump to in this file, it looks at the general list of errors (`flycheck-current-errors`) and looks for any errors that aren't associated with this buffer. It takes the first one from that list and jumps to that other buffer and location and runs `flycheck-buffer` in that buffer. This supports a general "troubleshooting" workflow.

I'll comment on the commit itself with some questions/clarifications.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -504,6 +504,16 @@ This variable has no effect, if `idle-change' is not contained in
   :package-version '(flycheck . "0.13")
   :safe #'numberp)
 
+(defcustom flycheck-show-buffer-name-in-errors nil
+  "Show buffer name in the errors list.
+
+When you change this, close your Flycheck errors buffer. When it
+opens next time, the columsn in the error list will be updated."
+  :group 'flycheck
+  :type 'boolean
+  :package-version '(flycheck . "0.22")
+  :safe #'booleanp)
+
 (defcustom flycheck-google-max-messages 5
   "How many messages to google at once.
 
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -2919,26 +2935,27 @@ Return the created overlay."
...
-    (unless flycheck-highlighting-mode
-      ;; Erase the highlighting from the overlay if requested by the user
-      (overlay-put overlay 'face nil))
-    (when flycheck-indication-mode
-      (overlay-put overlay 'before-string
-                   (flycheck-error-level-make-fringe-icon
-                    level flycheck-indication-mode)))
-    (overlay-put overlay 'help-echo (flycheck-error-message err))
-    overlay))
+  (when (eq (flycheck-error-buffer err) (current-buffer))
```
chrisdone: This line just adds this `(when ..)` condition, there's no other change in here.

flycheck.el
```diff
@@ -3051,7 +3068,17 @@ Intended for use with `next-error-function'."
   (let ((pos (flycheck-next-error-pos n reset)))
     (if pos
         (goto-char pos)
-      (user-error "No more Flycheck errors"))))
+      (let ((result (cl-remove-if
```
chrisdone: This implements the "no more errors, so let's see if there are errors from another buffer" logic.

flycheck.el
```diff
@@ -3281,8 +3320,12 @@ list."
 
 POS defaults to `point'."
   (interactive)
-  (-when-let* ((error (tabulated-list-get-id pos))
-               (buffer (flycheck-error-buffer error)))
+  (-when-let* ((error (tabulated-list-get-id pos)))
```
chrisdone: I factored out the functionality from this function into `flycheck-jump-to-error` so that it could be used above.

flycheck.el
cpitclaudel: Tip: adding w=1 to a github diff URL ignores spaces (e.g. https://github.com/flycheck/flycheck/pull/971/files/142c1bd73170beb48403d12f55614076b66f8f53?w=1)

chrisdone:
Closing in favor of a final PR putting into practice the plan in here: https://github.com/flycheck/flycheck/pull/970#issuecomment-219734788

PR #976, opened 2016-05-20
Sphinx manual improvements

cpitclaudel:
The first commit fixes index entries; they should have 5 fields, not 4 (could this be a new API?).

The second one adds support for parallel builds. Not sure how useful that is, so we could take it or leave it :)

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -15,7 +15,7 @@
 # this program.  If not, see <http://www.gnu.org/licenses/>.
 
 PIP = pip3
-SPHINXOPTS =
+SPHINXOPTS = -j4
 SPHINXBUILD = sphinx-build
 SPHINXAUTOBUILD = sphinx-autobuild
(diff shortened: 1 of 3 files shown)


lunaryorn:
@cpitclaudel  Thanks!  Your work on supporting parallel builds is great! 😍 

 I'm not sure about the index change, though.  Are you sure that this still builds with the current stable release of Sphinx?

cpitclaudel:
I'm not entirely sure. I noticed this while working on the `info` build, which broke without that change. I think we can hold it off for now (I stashed it locally and removed it from the PR)

lunaryorn:
@cpitclaudel LGTM now, merged.

I think at some point we should extract the obarray and keymap "concepts" in `elisp.py` into proper classes to make the whole code easier to comprehend, but I guess it's fine to do that next time I'll touch `elisp.py`.  For now, things work nicely with this extension :blush:

PR #979, opened 2016-05-26
Enable Racket checker when Geiser uses Racket

dkim:
[Geiser](http://www.nongnu.org/geiser/) enhances `scheme-mode`'s support for Racket as a minor mode (`geiser-mode`), not as a new major mode. However, we cannot just check `minor-mode-alist` for `geiser-mode` because Geiser supports other Scheme implementations as well as Racket. `.rkt` is the standard file extension for Racket source code.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7827,7 +7827,11 @@ See URL `https://racket-lang.org/'."
   :command ("raco" "expand" source-inplace)
   :predicate
   (lambda ()
-    (flycheck-racket-has-expand-p (flycheck-checker-executable 'racket)))
+    (and (or (eq major-mode 'racket-mode)
+             (and (eq major-mode 'scheme-mode)
+                  (boundp 'geiser-impl--implementation)
+                  (eq geiser-impl--implementation 'racket)))
+         (flycheck-racket-has-expand-p (flycheck-checker-executable 'racket))))
   :verify
   (lambda (checker)
     (let ((has-expand (flycheck-racket-has-expand-p
@@ -7842,7 +7846,7 @@ See URL `https://racket-lang.org/'."
     (flycheck-sanitize-errors (flycheck-increment-error-columns errors)))
   :error-patterns
   ((error line-start (file-name) ":" line ":" column ":" (message) line-end))
-  :modes racket-mode)
+  :modes (racket-mode scheme-mode))
 
 (flycheck-define-checker rpm-rpmlint
   "A RPM SPEC file syntax checker using rpmlint.


lunaryorn:
@dkim I'm not sure whether I'd like to merge this change.  As far as I see it, it's a workaround for a flaw in scheme-mode/geiser-mode which doesn't expose the scheme dialect in anyway.  We should not have to check file extensions to determine the language of a buffer.  After all, that's what major modes are for.

dkim:
@lunaryorn Geiser appears to be implemented that way because Racket used to be PLT Scheme. So is it difficult to let FlyCheck in the official repository support Geiser unless Geiser's support for Racket turns into a major mode? Or is it acceptable to register the Racket checker as a syntax checker for `scheme-mode` as well as `racket-mode`? `:modes (racket-mode scheme-mode)`

dkim:
I have written a new patch https://github.com/dkim/flycheck/commit/48048c877dc81dff467744ba6e910ddfb1a927a7 that does not check file extensions although it is more specific to Geiser than the version above.

cpitclaudel:
@dkim Thanks for your help and efforts :) However, I'm not sure I like the new patch either; depending on an explicitly private variable (as indicated by `--` in the name) sounds like a bad idea :/

lunaryorn:
@dkim Generally the syntax checker should have `:modes '(racket-mode scheme-mode)` in either case because it lets Flycheck select checkers for efficiently.  With some changes ahead `:modes` will actually become mandatory for all syntax checkers.

We do not strictly require major modes for syntax checker selection but we should not duplicate the major mode selection of Emacs or the language guessing of certain modes like Geiser Mode by dispatching on the file extension.  That's too flaky and too limited.

We can certainly dispatch on variables, though.  For instance we use `sh-shell` to dispatch to the proper shell in Sh Mode.

As such I prefer your second patch, though I share @cpitclaudel's concerns about using a private variable from Geiser Mode.  I wonder if there's no public API in Geiser Mode to find out what Scheme implementation is being used in the current buffer.  Would you mind to contact the Geiser Mode developers to find out?  Or suggest that they add a public variable/function which we can use to get the underlying scheme implementation?

dkim:
@cpitclaudel @lunaryorn The private variable was the best that I could find. Yesterday I asked geiser-users@nongnu.org if there is a more established way of doing it. I attached to the mail a link to this PR. I hope that we will get a reply soon.

jaor:
hi, i'm geiser's author.  it's perfectly fine to use `geiser-impl--implementation`.  the reason it uses `--` is that it doesn't belong to the interactive, user-level interface of geiser, that's all.  it's guaranteed to remain there and be stable.

dkim:
Thank you for your reply and confirmation, @jaor!

@lunaryorn @cpitclaudel  I plan to add the `:modes (racket-mode scheme-mode)` line to the second patch, as @lunaryorn suggested, and make a new PR. Please let me know if there is any other thing that I can improve.

lunaryorn:
@jaor Thank you for joining this discussion and helping us out 👍 

@dkim I'm fine with using `geiser-impl--implementation` in the `:predicate` then.  I'd be great if you could submit your second patch with an appropriate `:modes` declaration.

Please don't open a second pull request, though.  Just `git push --force` to the `geiser` branch of your fork to update this pull request.  That's easier for us to keep track of things :relaxed:

Thanks!

lunaryorn:
@dkim Thanks, merged :+1:

dkim:
Thank you all! You guys are as much supportive as Flycheck and Geiser. 😃

PR #984, opened 2016-06-06
Pass epg-gpg-home-directory to gpg invocation

spwhitton:
Otherwise, the temporary GPG home directory that the test suite sets up
doesn't actually get used when gpg is invoked to encrypt the file.


lunaryorn:
@spwhitton Thank you very much!

PR #990, opened 2016-06-12
Detect DMD continuation lines

CyberShadow:
When a template declared in another module fails to instantiate, DMD produces errors such as:

```
lib.d(3,12): Error: undefined identifier 'invalid'
lib.d(8,12): Error: template instance lib.T!() error instantiating
src.d(3,1):        instantiated from here: U!()
```

Previously, Flycheck would only detect the first two lines as error messages. However, since they both occur outside of the edited file, this gives the impression that the edited file has no issues.

This patch detects continuation lines (such as the third line above) as Flycheck errors with the "info" severity.

----

I hope I got the tests right, because I haven't been able to run them locally. Running `make SELECTOR='(language d)' integ` tries to run three tests and fails them all, with the message `(file-error "Cannot open load file" "no such file or directory" "cask-cli")`. I found https://github.com/flycheck/flycheck/issues/531, however from that I gather that this is a problem specific to Emacs 24.3 and below, and I'm running Emacs 24.5.1.

Also, I guess this isn't relevant to the changes here, but when running `make specs`, the "Utilities" / "flycheck-encrypted-buffer-p" / "recognizes an encrypted buffer" test seems to still pop up a dialog box prompting me for a password. Entering "spam with eggs" (from looking at the code) does get it to continue.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6187,7 +6187,9 @@ Requires DMD 2.066 or newer.  See URL `http://dlang.org/'."
           (file-name) "(" line "," column "): Error: " (message)
           line-end)
    (warning line-start (file-name) "(" line "," column "): "
-            (or "Warning" "Deprecation") ": " (message) line-end))
+            (or "Warning" "Deprecation") ": " (message) line-end)
+   (info line-start (file-name) "(" line "," column "): "
+            (one-or-more " ") (message) line-end))
   :modes d-mode)
 
 (defconst flycheck-this-emacs-executable
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -2950,6 +2950,19 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
    '(4 8 error "module external_library is in file 'external_library.d' which cannot be read"
        :checker d-dmd)))
 
+(flycheck-ert-def-checker-test d-dmd d continuation-line
+  (unless (version<= "24.4" emacs-version)
+    (ert-skip "Skipped because CC Mode is broken on 24.3.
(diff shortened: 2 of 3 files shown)


lunaryorn:
> Also, I guess this isn't relevant to the changes here, but when running make specs, the "Utilities" / "flycheck-encrypted-buffer-p" / "recognizes an encrypted buffer" test seems to still pop up a dialog box prompting me for a password. Entering "spam with eggs" (from looking at the code) does get it to continue.

That's interesting… I haven't been able to reproduce it.  Could you give me a little more information about your OS and your desktop environment?  Do you have a special GPG setup?

lunaryorn:
@CyberShadow I merged your PR.  Thanks for the contribution! :blush: :+1:

CyberShadow:
> Could you give me a little more information about your OS and your desktop environment?

Arch Linux, no DE (i3).

>  Do you have a special GPG setup?

I don't think so... I do have it set up with a key pair, nothing unusual as far as I know.

lunaryorn:
@CyberShadow What kind of dialog box pops up?  Is that some pinentry tool?

CyberShadow:
Hmm... I tried to reproduce it just now and I don't see the dialog box any more. Now it just hangs at "recognizes an encrypted buffer". There is a gpg process running - killing it causes the test to fail.

IIRC it was the standard GPG dialog ("Please enter your passphrase to unlock ..."). I also get it in Emacs when I open a .gpg file or a sudo path via TRAMP.

lunaryorn:
Mmh, I'm not a GPG expert at all… setting up commit signing for Github was almost beyond my skills :wink:  I did my best to configure Emacs' GPG support to be non-interactive for that test, but I don't know how to do better.

PR #992, opened 2016-06-16
Enable ms-compatibility-version for clang

trolleyyy:
Hi,

I started today to configure flycheck for myself on windows and I tried to make it work with windows. Unfortunately clang reports errors for Microsoft Visual C++ headers. There is option to avoid it - ms-compatibility-version but unfortunately it was not available in flycheck.

Now you only need to call:
`(setq flycheck-clang-ms-compatibility-version "19")`
to fix this problem.

You may see how problem looks like in the following image (trying to include <string> header)
error

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5639,14 +5639,14 @@ When non-nil, enable Microsoft extensions to C/C++ via
   :package-version '(flycheck . "0.16"))
 
 (flycheck-def-option-var flycheck-clang-ms-compatibility-version nil c/c++-clang
-  "The Microsoft compatibility version to use with clang
+  "The Microsoft compatibility version to use with Clang.
 
 When non-nil, enable Microsoft compatibility to C/C++ via
 `-fms-compatibility-version'."
   :type '(choice (const :tag "No compatibility version" nil)
                  (string :tag "Compatibility version"))
   :safe #'stringp
-  :package-version '(flycheck . "0.29"))
+  :package-version '(flycheck . "29"))
 
 (flycheck-def-option-var flycheck-clang-no-exceptions nil c/c++-clang
   "Whether to disable exceptions in Clang.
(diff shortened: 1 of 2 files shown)


flycheck.el
```diff
@@ -5638,6 +5638,16 @@ When non-nil, enable Microsoft extensions to C/C++ via
   :safe #'booleanp
   :package-version '(flycheck . "0.16"))
 
+(flycheck-def-option-var flycheck-clang-ms-compatibility-version nil c/c++-clang
+  "The Microsoft compatibility version to use with clang
```
cpitclaudel: You're missing a dot at the end of this line, I think.

flycheck.el
```diff
@@ -5638,6 +5638,16 @@ When non-nil, enable Microsoft extensions to C/C++ via
...
+
+When non-nil, enable Microsoft compatibility to C/C++ via
+`-fms-compatibility-version'."
+  :type '(choice (const :tag "No compatibility version" nil)
+                 (string :tag "Compatibility version"))
+  :safe #'stringp
+  :package-version '(flycheck . "0.29"))
```
cpitclaudel: We use whole version numbers now, so this should be "29", not "0.29".

flycheck.el
cpitclaudel: Also `clang` → `Clang` would be nice for consistency.

trolleyyy:
I've updated things you've commented out.

lunaryorn:
@trolleyyy Did you try to add the flag to `flycheck-clang-args`?  Wouldn't that work?

trolleyyy:
You are right, setting flycheck-clang-args works as well. Sorry for your trouble.

lunaryorn:
@trolleyyy Good.  I'll close this issue then.  I think this covers just a special case and does not warrant a dedicated option.

PR #996, opened 2016-06-22
install: add (package-refresh-contents)

moy:
At least on Emacs 24.4.1, the instructions didn't work and instead the
line (package-install 'flycheck) was raising the error:

    package-compute-transaction: Package `flycheck-' is unavailable

Indeed, the index of available packages wasn't refreshed after
'package-archives was updated. Add (package-refresh-contents) to make
sure it is.


lunaryorn:
Oh dear, I'm sorry for this blunder.  Thank you for fixing it :+1:

PR #1007, opened 2016-07-09
Define default directory for haskell-stack-ghc checker

sergv:
Default directory is detected either by looking for `stack.yaml` or asking `stack path --project-root` for `haskel-stack-ghc` checker or by looking for cabal file in case of `haskell-ghc` checker.

@mrkkrp Please comment, if you think something should be done differently.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6977,16 +6977,15 @@ Otherwise return the previously used cache directory."
             (make-temp-file "flycheck-haskell-ghc-cache" 'directory))))
 
 (defun flycheck-haskell--find-default-directory (checker)
-  "Come up with a sutable default directory for Haskell to run
-checker in.
+  "Come up with a suitable default directory for Haskell to run CHECKER in.
 
 In case of `haskell-stack-ghc' checker it is directory with
-stack.yaml file. If there's no stack.yaml file in any parent
+stack.yaml file.  If there's no stack.yaml file in any parent
 directory, it will be the directory that \"stack path --project-root\"
-command returs.
+command returns.
 
 For all other checkers, it is the closest parent directory that
-contains cabal file."
+contains a cabal file."
   (pcase checker
     (`haskell-stack-ghc
      (or
@@ -7001,19 +7000,14 @@ contains cabal file."
                      (file-directory-p stack-dir))
             stack-dir)))))
     (_
-     (if (fboundp 'haskell-cabal-find-file)
-         (let ((cabal-file (haskell-cabal-find-file)))
-           (when cabal-file
-             (file-name-directory cabal-file)))
-       (locate-dominating-file (file-name-directory (buffer-file-name))
-                               (lambda (dir)
-                                 (not
-                                  (null
-                                   (directory-files dir
-                                                    nil ;; full
-                                                    ".+\\.cabal\\'"
-                                                    t ;; nosort
-                                                    )))))))))
+     (locate-dominating-file
+      (file-name-directory (buffer-file-name))
+      (lambda (dir)
+        (directory-files dir
+                         nil ;; use full paths
+                         ".+\\.cabal\\'"
+                         t ;; do not sort result
+                         ))))))
 
 (flycheck-define-checker haskell-stack-ghc
   "A Haskell syntax and type checker using `stack ghc'.
--- a/flycheck.el
+++ b/flycheck.el
@@ -6976,6 +6976,45 @@ Otherwise return the previously used cache directory."
         (or flycheck-haskell-ghc-cache-directory
             (make-temp-file "flycheck-haskell-ghc-cache" 'directory))))
 
+(defun flycheck-haskell--find-default-directory (checker)
+  "Come up with a sutable default directory for Haskell to run
+checker in.
+
+In case of `haskell-stack-ghc' checker it is directory with
+stack.yaml file. If there's no stack.yaml file in any parent
+directory, it will be the directory that \"stack path --project-root\"
+command returs.
+
+For all other checkers, it is the closest parent directory that
+contains cabal file."
+  (pcase checker
+    (`haskell-stack-ghc
+     (or
+      (locate-dominating-file (buffer-file-name) "stack.yaml")
+      (when (executable-find "stack")
+        (let* ((stack-output
+                (process-lines "stack" "path" "--project-root"))
+               (stack-dir
+                (when stack-output
+                  (car stack-output))))
+          (when (and stack-dir
+                     (file-directory-p stack-dir))
+            stack-dir)))))
+    (_
+     (if (fboundp 'haskell-cabal-find-file)
+         (let ((cabal-file (haskell-cabal-find-file)))
+           (when cabal-file
+             (file-name-directory cabal-file)))
+       (locate-dominating-file (file-name-directory (buffer-file-name))
+                               (lambda (dir)
+                                 (not
+                                  (null
+                                   (directory-files dir
+                                                    nil ;; full
+                                                    ".+\\.cabal\\'"
+                                                    t ;; nosort
+                                                    )))))))))
+
 (flycheck-define-checker haskell-stack-ghc
   "A Haskell syntax and type checker using `stack ghc'.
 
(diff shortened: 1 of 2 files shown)


flycheck.el
```diff
@@ -6976,6 +6976,36 @@ Otherwise return the previously used cache directory."
         (or flycheck-haskell-ghc-cache-directory
             (make-temp-file "flycheck-haskell-ghc-cache" 'directory))))
 
+(defun flycheck-haskell--default-directory (checker)
+  "Come up with a sutable default directory for Haskell to run checker in."
```
lunaryorn: Come up sounds strange. Why not simply "find"?

Please also extend the docstring to explain what a "suitable" directory would be, i.e. document the behaviour of the function as well.

flycheck.el
```diff
@@ -6976,6 +6976,36 @@ Otherwise return the previously used cache directory."
...
+  (pcase checker
+    (`haskell-stack-ghc
+     (or
+      (locate-dominating-file (buffer-file-name) "stack.yaml")
+      (when (executable-find "stack")
+        (let ((stack-dir
+               (replace-regexp-in-string
+                (rx (* (any " \t\n")) eos)
+                ""
+                (shell-command-to-string "stack path --project-root"))))
```
lunaryorn: Please use process-lines instead of shell-command-to-string.  We should not use shell commands without need.

flycheck.el
```diff
@@ -6976,6 +6976,36 @@ Otherwise return the previously used cache directory."
...
+(defun flycheck-haskell--default-directory (checker)
+  "Come up with a sutable default directory for Haskell to run checker in."
+  (pcase checker
+    (`haskell-stack-ghc
+     (or
+      (locate-dominating-file (buffer-file-name) "stack.yaml")
+      (when (executable-find "stack")
+        (let ((stack-dir
+               (replace-regexp-in-string
+                (rx (* (any " \t\n")) eos)
```
lunaryorn: Why's this? Please add a comment explaining the reason for this substitution.

lunaryorn:
Thanks.  I've done a cursory review, will look at the details tomorrow.

lunaryorn:
@flycheck/haskell @purcell Would you also take a look? Maybe also test this PR on different Haskell projects?

sergv:
@lunaryorn I've amended commit with your suggestions.

mrkkrp:
@sergv, Looks good, thank you for writing this! I'm looking forward to using it.

purcell:
LGTM. I tried this patch on top of flycheck HEAD with a fairly vanilla stack project and all was well, so I'm pretty confident this will work.

flycheck.el
```diff
@@ -6976,6 +6976,45 @@ Otherwise return the previously used cache directory."
         (or flycheck-haskell-ghc-cache-directory
             (make-temp-file "flycheck-haskell-ghc-cache" 'directory))))
 
+(defun flycheck-haskell--find-default-directory (checker)
+  "Come up with a sutable default directory for Haskell to run
```
cpitclaudel: `sutable` → `suitable`

flycheck.el
```diff
@@ -6976,6 +6976,45 @@ Otherwise return the previously used cache directory."
...
+checker in.
+
+In case of `haskell-stack-ghc' checker it is directory with
+stack.yaml file. If there's no stack.yaml file in any parent
+directory, it will be the directory that \"stack path --project-root\"
+command returs.
```
cpitclaudel: `returs` → `returns`

flycheck.el
```diff
@@ -6976,6 +6976,45 @@ Otherwise return the previously used cache directory."
...
+
+For all other checkers, it is the closest parent directory that
+contains cabal file."
```
cpitclaudel: "a cabal file" (maybe :)

flycheck.el
```diff
@@ -6976,6 +6976,45 @@ Otherwise return the previously used cache directory."
...
+    (_
+     (if (fboundp 'haskell-cabal-find-file)
+         (let ((cabal-file (haskell-cabal-find-file)))
+           (when cabal-file
+             (file-name-directory cabal-file)))
+       (locate-dominating-file (file-name-directory (buffer-file-name))
+                               (lambda (dir)
+                                 (not
+                                  (null
+                                   (directory-files dir
```
cpitclaudel: I think `(not (null x))` is the same as `x` in this case

flycheck.el
```diff
@@ -6976,6 +6976,45 @@ Otherwise return the previously used cache directory."
...
+                                                    nil ;; full
+                                                    ".+\\.cabal\\'"
+                                                    t ;; nosort
+                                                    )))))))))
+
```
cpitclaudel: Do we really need this double implementation? Nothing in the docstring mention using haskell mode to find the cabal file; could we just use `locate-dominating-file`? Or is there a benefit to `haskell-cabal-find-file`? (If so, what happens if it returns null? Should we fall back to locate-dominating-file?)

flycheck.el
```diff
@@ -6976,6 +6976,45 @@ Otherwise return the previously used cache directory."
...
+         (let ((cabal-file (haskell-cabal-find-file)))
+           (when cabal-file
+             (file-name-directory cabal-file)))
+       (locate-dominating-file (file-name-directory (buffer-file-name))
+                               (lambda (dir)
+                                 (not
+                                  (null
+                                   (directory-files dir
+                                                    nil ;; full
+                                                    ".+\\.cabal\\'"
```
cpitclaudel: I don't think you need the `+` in that regexp. (Btw, can a cabal file be called just `.cabal`?)

flycheck.el
```diff
@@ -6976,6 +6976,45 @@ Otherwise return the previously used cache directory."
...
+                  (car stack-output))))
+          (when (and stack-dir
+                     (file-directory-p stack-dir))
+            stack-dir)))))
+    (_
+     (if (fboundp 'haskell-cabal-find-file)
+         (let ((cabal-file (haskell-cabal-find-file)))
+           (when cabal-file
+             (file-name-directory cabal-file)))
+       (locate-dominating-file (file-name-directory (buffer-file-name))
```
cpitclaudel: Maybe we could rewrap this to make the lambda more readable? (by adding a newline before `(file-name...`

cpitclaudel:
I've added quick review notes as well. Thanks for this PR!

flycheck.el
purcell: I think `haskell-cabal-find-file` doesn't add much beyond stopping upwards traversal if file/directory ownership changes. Which sounds questionable to me anyway. Probably best to avoid referring to anything in haskell-mode here IMO.

flycheck.el
purcell: Yep. And this docstring wouldn't pass `checkdoc`, so it needs a little work.

flycheck.el
purcell: `haskell-mode` uses `(directory-files dir t ".\\.cabal\\'")`, so it admits `a.cabal`, but not `.cabal`.

purcell: (See https://github.com/haskell/haskell-mode/blob/master/haskell-cabal.el#L243-L284)

cpitclaudel:
@purcell Thanks for the review-of-the-review :))

flycheck.el
sergv: That's true, but I tried to express the idea that we're only interested in whether the returned list is non-empty and not the list itself

lunaryorn: @sergv It's fine to just return the list in this case.  The double negation doesn't really add anything imho.

sergv:
@cpitclaudel I've rebased on master and added commit that fixes review notes

flycheck.el
```diff
@@ -6976,6 +6976,39 @@ Otherwise return the previously used cache directory."
...
+  (pcase checker
+    (`haskell-stack-ghc
+     (or
+      (locate-dominating-file (buffer-file-name) "stack.yaml")
+      (when (executable-find "stack")
+        (let* ((stack-output
+                (process-lines "stack" "path" "--project-root"))
+               (stack-dir
+                (when stack-output
+                  (car stack-output))))
```
lunaryorn: @sergv The `when` guard is redundant.  `(car nil)` is perfectly safe and evaluates to `nil` so it's perfectly safe to just do `(car (process-lines …))`.

lunaryorn:
@sergv Would you add a note about this change to `CHANGES.rst`?  I think it falls under `Improvements:` :blush:

lunaryorn:
LGTM. Thanks for the change 😊 

I think though that #1012 should go in first.

lunaryorn:
@sergv I think you forgot to rename `:default-directory` to `:working-directory` here before merging this PR.  I fixed it in https://github.com/flycheck/flycheck/commit/ceaf8c460dea7d999858d53c5eab70317b329fce.  Could take another look?

sergv:
@lunaryorn Indeed, I missed it. Thank you!

PR #1013, opened 2016-07-13
Add slim-lint checker definition

ibrahima:
Slim-lint is a third-party linter for slim templates.

See https://github.com/sds/slim-lint

I hope this was formatted correctly, first time contributing to flycheck! Currently there's no option to read from stdin so I couldn't set that, which is kind of annoying. Also, not sure if I set :next-checkers correctly or if the order matters. I think slim-lint also includes the basic slim errors so not sure if having both enabled causes problems, or what happens if the executable is not found (I assume no errors are thrown if it's not installed, right?)

Thanks!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,16 +1,16 @@
 29-cvs (in development)
 =======================
 
+- New syntax checkers:
+
+  - Slim with ``slim-lint`` [GH-1013]
+
 - New features:
 
   - Add ``:working-directory`` option to ``flycheck-define-command-checker``
     [GH-973] [GH-1012]
   - ``flycheck-go-build-install-deps`` turns on dependency installation for ``go test``
     as well as ``go build`` [GH-1003]
-
-- New syntax checkers:
-
-  - Slim with ``slim-lint`` [GH-1013]
 
 - Improvements:
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@
   - ``flycheck-go-build-install-deps`` turns on dependency installation for ``go test``
     as well as ``go build`` [GH-1003]
 
+- New syntax checkers:
+
+  - Slim with ``slim-lint`` [GH-1013]
+
 - Improvements:
 
   - Add default directory for ``haskell-stack-ghc`` and ``haskell-ghc`` checkers
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -972,6 +972,10 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       Check Slim using the `Slim <http://slim-lang.com/>`_ compiler.
 
+   .. syntax-checker:: slim-lint
+
+      Check Slim best practices using the `slim-lint <https://github.com/sds/slim-lint>`_ linter.
+
 .. supported-language:: SQL
 
    .. syntax-checker:: sql-sqlint
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -974,7 +974,8 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
    .. syntax-checker:: slim-lint
 
-      Check Slim best practices using the `slim-lint <https://github.com/sds/slim-lint>`_ linter.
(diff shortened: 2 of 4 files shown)


flycheck.el
```diff
@@ -8630,6 +8631,16 @@ See URL `http://slim-lang.com'."
           line-end))
   :modes slim-mode)
 
+(flycheck-define-checker slim-lint
+  "A SLIM linter
+
+See URL `https://github.com/sds/slim-lint'."
+  :command ("slim-lint" "--reporter=checkstyle"
+            source-original)
```
cpitclaudel: Must this be `source-original`? Wouldn't `source` (or even `source-inplace`) work?

flycheck.el
```diff
@@ -8630,6 +8631,16 @@ See URL `http://slim-lang.com'."
...
+  :error-parser flycheck-parse-checkstyle
+  :modes slim-mode
+  :next-checkers ((info . slim)))
```
cpitclaudel: @lunaryorn will confirm this, but I would expect `slim-lint` to be the `next-checker` of `slim`, not the other way around.

cpitclaudel:
Great, thanks! Please check my inline comments.

flycheck.el
ibrahima: I have no idea, I just found an example for defining custom checkers online and modified it. Probably should have gone to the source rather than random online examples though ;). It seems like source works, so that the display isn't lagging behind until you save. Will change!

flycheck.el
ibrahima: That would seem correct to me as well but I wasn't sure how the precedence works and how it works if you don't have one or the other installed.

ibrahima:
Thanks! Addressed your comments and force pushed. Any idea what the CI failure is about? I added a doc string correctly I think (again, just following what I see elsewhere in the file) so I don't understand what it's saying (seems like it expects slim-lint to *not* have a doc string?).

Simplify:
Hi @ibrahima,

CI failure comes because documentation for this checker is missing in `doc/languages.rst` file. 
Can you fix that? There is already `slim` section in that file, just add extra checker and short description with URL that points to slim-lint page.

ibrahima:
Oh cool, done!

doc/languages.rst
```diff
@@ -972,6 +972,10 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       Check Slim using the `Slim <http://slim-lang.com/>`_ compiler.
 
+   .. syntax-checker:: slim-lint
+
+      Check Slim best practices using the `slim-lint <https://github.com/sds/slim-lint>`_ linter.
```
lunaryorn: Please wrap this line with `Alt+Q` (i.e. `fill-paragraph`), as it's beyond our 80 character limit.

lunaryorn:
@ibrahima Please add this change to `CHANGES.rst` (take a look at release 27 to see how we document syntax checkers in the changelog).

If you'd like please also feel free to add yourself to the `Contributors` section in `doc/community/people.rst` :blush:

ibrahima:
Done!

CHANGES.rst
```diff
@@ -8,6 +8,10 @@
   - ``flycheck-go-build-install-deps`` turns on dependency installation for ``go test``
     as well as ``go build`` [GH-1003]
 
+- New syntax checkers:
```
lunaryorn: Ah, thanks, but could you just move this item above `- New features`?  Generally we list breaking changes, then new syntax checkers, and then other new features, because I think that support for new languages is more important to most users than other options.

@ibrahima Could you just make this final change?

lunaryorn:
@ibrahima Finally, please sign our [CLA](https://cla-assistant.io/flycheck/flycheck?pullRequest=1013) to state your explicit consent with Flycheck's licenses for your contribution :blush:

ibrahima:
Done!

lunaryorn:
LGTM

lunaryorn:
@cpitclaudel Would you take a look?

cpitclaudel:
LGTM :)

Issue #530, opened 2014-11-28
Error fixers


swsnr:
Add support for error fixers
# Proposed API

Add a `flycheck-try-fix-error-at-point` command, which finds a `flycheck-error` at point and invokes all “fixer” functions in a new abnormal `flycheck-fix-error-functions` hook with the `flycheck-error` object.  Each function in this hook may try to fix this error inline in the current buffer, and return non-nil if it succeeded to do so.  Otherwise it shall return `nil`.  `flycheck-try-fix-error-at-point` will invoke all functions until the first one returns non-nil.

If there are multiple errors at point, the command could either
- prompt for an error to fix,
- or signal a `user-error`.

Feedback on this API is very welcome.
# Implementing new fixers

Flycheck itself will **not** provide any fixers.  This is entirely out of Flycheck's scope, since it's too much unrelated code to maintain.

However, we will provide a `flycheck-fixers` repository in the Flycheck org which will contain a `flycheck-fixers` package with **simple** fixers.  This will mostly be maintained by the community.  More complicated fixers, particularly those that require a deep understanding of the current language, should be maintained in separate packages, though.

Inviting @purcell to this discussion.

gracjan:
Nice idea!

gracjan:
Note: error/warning fixing functionality could help haskell-mode to shed some weight and use flycheck for hlint. hlint offers fixup suggestions and hs-lint-mode makes use of that. Would be so much better if this part was inside flycheck.

lunaryorn:
@gracjan There are man features that should be in Flycheck.  This is a nice addition but it is not my top priority.  It's only available for a small subset of syntax checkers.  Most errors can't be fixed automatically. 

To be honest, if you're waiting for me to implement this you'll have to wait a long time.  On my personal backlog, there are more important things.  In particular there's a dire need for refactoring to provide better support for modes such as Ensime, Merlin or Irony to use Flycheck as a display backend for their own errors.  

I'm sorry but if you'd like to see this feature soon you'll have to implement it yourself and open a pull request.

Note, though, that my point from the original description still stands:  Flycheck itself will only provide the API for error fixers.  It'll _not_ add any fixers on its own, that's just way too much for me to maintain.  Any actual implementation for fixing errors from hlint will have to be distributed independently, probably as part of Flycheck Haskell or Haskell Mode.

gracjan:
haskell-mode would also benefit from from Flycheck error display improvements.

[...]

PR #1022, opened 2016-07-26
[WIP] Add `flycheck-try-fix-error-at-point`

ibrahima:
This interactive function attempts to fix the (first) error at point
using hooks installed into `flycheck-fix-error-functions`. Functions are
given an error object and can mutate the buffer in whatever way
necessary to fix the error, or return nil to indicate that nothing was
fixed.

Implementation of #530

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -377,6 +377,29 @@ node `(elisp)Hooks'."
   :package-version '(flycheck . "0.13")
   :risky t)
 
+(defcustom flycheck-fix-error-functions nil
+  "Functions to attempt to fix errors.
+
+Each function in this hook must accept a single argument: A
+Flycheck error to fix.  The function should, if applicable,
+attempt to fix the error, or else return nil if it cannot.
+
+All functions in this hook are called in order of appearance,
+until a function returns non-nil.  Thus, a function in this hook
+may return nil, to allow for further processing of the error, or
+any non-nil value, to indicate that the error was fully fixed
+and inhibit any further processing.
+
+Note that these functions should only ever be called
+interactively; Flycheck will not attempt to fix errors unsolicited.
+
+This variable is an abnormal hook.  See Info
+node `(elisp)Hooks'."
+  :group 'flycheck
+  :type 'hook
+  :package-version '(flycheck . "0.29")
+  :risky t)
+
 (defcustom flycheck-display-errors-delay 0.9
   "Delay in seconds before displaying errors at point.
 
@@ -4170,7 +4193,14 @@ non-nil."
     (setq flycheck-display-error-at-point-timer
           (run-at-time flycheck-display-errors-delay nil 'flycheck-display-error-at-point))))
 
-
+(defun flycheck-try-fix-error-at-point ()
+  "Attempt to fix error at point using the list of fixer functions."
+  (interactive)
+  (let ((err (car-safe (flycheck-overlay-errors-at (point)))))
+    (if err
+        (run-hook-with-args-until-success 'flycheck-fix-error-functions err)
+      (message "No errors at point to fix!"))))
+
 ;;; Functions to display errors
 (defconst flycheck-error-message-buffer "*Flycheck error messages*"
   "The name of the buffer to show long error messages in.")


flycheck.el
```diff
@@ -4170,7 +4193,14 @@ non-nil."
     (setq flycheck-display-error-at-point-timer
           (run-at-time flycheck-display-errors-delay nil 'flycheck-display-error-at-point))))
 
-
```
lunaryorn: This line shouldn't get lost.  I think it's one of the page breaks that we use these to separate sections in `flycheck.el`—try `C-x [` or `C-x ]` to navigate them and install `page-break-lines` to make them looker nicer than `^L`.

flycheck.el
```diff
@@ -377,6 +377,29 @@ node `(elisp)Hooks'."
...
+and inhibit any further processing.
+
+Note that these functions should only ever be called
+interactively; Flycheck will not attempt to fix errors unsolicited.
+
+This variable is an abnormal hook.  See Info
+node `(elisp)Hooks'."
+  :group 'flycheck
+  :type 'hook
+  :package-version '(flycheck . "0.29")
```
lunaryorn: Please drop the leading zero, we do not use dotted version numbers anymore.  It's just `29`.

flycheck.el
```diff
@@ -4170,7 +4193,14 @@ non-nil."
...
 
-

+(defun flycheck-try-fix-error-at-point ()
+  "Attempt to fix error at point using the list of fixer functions."
+  (interactive)
+  (let ((err (car-safe (flycheck-overlay-errors-at (point)))))
+    (if err
+        (run-hook-with-args-until-success 'flycheck-fix-error-functions err)
+      (message "No errors at point to fix!"))))
```
lunaryorn: I'm not sure whether it's good to just fix the first error at point.  In most cases there'll only be one error at point, for sure, but what if there are two and the first one isn't fixable whereas the second one is?  We always order overlays by error level so the user will never get a chance to fix the second error automatically.

lunaryorn:
@ibrahima Thanks!  I believe though that we've still got to refine the interface a bit, as I don't think it's sufficient to always just fix the first error at point, see my comment in the diff.

I'm also considering @chrisdone's [suggestion about the UX](https://github.com/flycheck/flycheck/issues/530#issuecomment-235224763), and maybe we should provide a more elaborate interface than what was outlined in #530. 

I'm just sketching my ideas now to collect your feedback; please don't take this as asking you to change anything.

@chrisdone's comment made me think that we should probably implement this as a two-step process:  First collect possible fixes at point, and apply them separately.  That'd allow to write different UIs for applying suggestions or fixes without changing the "fixing" functions, and it lets us avoid the problem of selecting the error at point to fix. 

So what about this interface:

1. An option `flycheck-suggestions-functions`, where each function is called with each error at point and shall return a potentially empty list of `flycheck-suggestion` objects.
2. A `flycheck-suggestion` struct with three slots: `error` (the error to which the suggestion applies), `apply-fn` (a function to apply the suggestion to the buffer) and `context` (where the suggestion function can store arbitrary stuff it might later need to apply the suggestions).
3. A `flycheck-apply-suggestion` function which takes a suggestion and invokes `apply-fn` with the error and the context as arguments.

Based on this infrastructure we could add different commands to fix errors: A `flycheck-apply-first-suggestion-at-point` or so which just applies the first suggestion at point if any, or a `flycheck-apply-suggestions` which prompts for suggestions with an interface similar to what @chrisdone has shown, or maybe a company source for suggestions at point or whatever. I think making the underlying API more flexible gives us more power on the UI side.

@ibrahima What's your opinion on this?  @chrisdone, @cpitclaudel, @purcell, what do you think?

I'm sorry that I'm so late with reconsidering the interface but this just came into my mind as I saw your pull request. 😌 

----

Related to this, if we make the interface more complex we should probably move all of this into a separate `flycheck-suggestions` library.  We could then also put suggestion functions for built-in checkers into this file and thus ship with built-in suggestions without growing `flycheck.el` beyond what's maintainable.

flycheck.el
ibrahima: oh sorry, I didn't realize, thought it was a mistake!

flycheck.el
ibrahima: ah ok, wasn't sure since other variables had the dotted version numbers.

flycheck.el
ibrahima: my thought was that it may not necessarily be the case that multiple errors can be fixed in one pass, because the mutation as a result of fixing the first error might invalidate the latter errors (at least the line/column information, for instance).

ibrahima:
> I'm sorry that I'm so late with reconsidering the interface but this just came into my mind as I saw your pull request. :relieved: 

No problem, I like the discussion and think it's important since this feature would be pretty much outside of the scope of Flycheck as it stands currently. It does make sense that this could be a separate library altogether, and I'm happy to iterate on it as we flesh out what is needed for different environments and languages. Though as my experience with elisp is still pretty basic, it's possible that I'll be out of my depth at some point but I'm happy to push this along as far as I am able to.

My concern about a more complex interface/fixing multiple issues at point is that the interface would have to ensure that fixes are composable, and I'm not sure that that's necessarily the case in general. In the case that fixes only change the line the error is on, that's fine. But for instance, the reason I initially embarked on this was to enable autocorrecting SCSS-lint's property order suggestions. It has a few built-in conventions of what order to sort CSS properties, but remembering that order is difficult for me right now since we just started using it, so I wanted to correct that within Emacs as soon as I see the error message. But such a fixer would reorder lines, and now the other linter messages are no longer valid because the line (and probably even the point) has moved. Then again, looking again at @chrisdone 's example, those fixes don't necessarily correspond to the current point/line either and some of them involve deleting lines so maybe it's possible to devise an interface such that this is possible to do cleanly.

I wonder how Intero handles applying multiple suggestions like that?

chrisdone:
@ibrahima Currently we conveniently apply them in the correct order. Modify lines. Add lines in reverse order. Delete imports in reverse order. Add language extenions at the top of the file last.

But this is the kind of thing one can achieve with markers. You set a marker for the place that the change should take place. If code moves around, it's no big deal. If code is deleted, then that suggestion is no longer applicable. We'll probably use this in Intero once the number of suggestions hits double digits, to make suggestions composable. I'd expect that changes to a region of text removes the markers in that, so that subsequent suggestions aren't available there. 

However, we're going for a batch kind of feel, where many imports from a file should be removed, or many type signatures inserted, etc. It would be awful to have to click each one manually. For Intero to use Flycheck for this task, we'd definitely need a batch apply. I can understand if you'd prefer to keep changes per flycheck message.

lunaryorn:
@ibrahima Do we need to allow to apply multiple suggestions at once?  If we don't, if we just allow to apply a single suggestions at a time in the default user interface, then suggestions don't need to compose as we'd check the buffer and recompute available suggestions before the user could apply the next one.

As for batching as @chrisdone asked the API I outlined would still allow this as you could just map over all `flycheck-current-errors`, collect all available suggestions and then call apply for all applicable suggestions at once, and thus present a more powerful UI to fix many errors at once.

But that'll be mode-specific because it implies that suggestions compose to a certain degree, and we can't guarantee that generically, so you can only batch-apply suggestions if you know that whatever tool generated the suggestions creates composable ones.

ibrahima:
I'm basically thinking the same thing as @lunaryorn , though if the API is "edit the buffer until the error is fixed" I'm not sure how suggestions can compose at all - seems like if you change the buffer, the edit operations that a subsequent suggestion would run would no longer map to the same locations in the file, so it's not clear what the right thing to do would be there. In that case, it would make more sense to have some data representation of suggestions so that flycheck or whatever thing is on top of it could know how to compose them.

lunaryorn:
I guess it depends on how invasive the suggestions are… if you apply suggestions from `(point-max)` backwards and check for overlaps then there's a good chance that you can batch-apply suggestions if the suggestion doesn't change huge chunks of code.  For instance hlint suggestions are always confined to a single top-level form, so you could easily batch-apply all hlint suggestions to different forms.

chrisdone:
It wouldn't be the end of the world to apply what suggestions are unambiguous and then tell the user "we applied the ones that are unambiguous", and then the next time a check occurs they'll see the ones that haven't been applied yet. Ambiguous means "these changes overlap".

jyp:
I think that @ibrahima 's proposal is superior. It is both simpler and much more convenient than the menu-based interface suggested by @chrisdone. Indeed: 

1. in the general case not all suggestions should be applied, so there has to be a way to select those that should be applied.

2. deciding weather to apply an autofix usually requires inspecting the context of the error. So, the workflow of going through errors one by one and applying fixes (potentially automatic) is superior to the "menu-based" workflow. Indeed, the latter requires separate inspection of the code.

3. A menu requires an additional implementation burden. This burden falls both on flycheck maintainers, but also users who may want to customise the menu interface itself (evil mode, etc).

Fuco1:
The ideal interaction model for me is ido/helm/sallet or something similar.  To make that work the api would need to be two-step: 

1. generate candidates
2. do the action of selected candidate

This aligns with the proposal of `flycheck-suggestions-functions` and friends.  I also agree with @ibrahima and @jyp that you should probably just fix one thing and then recompute the suggestions (the helm/ido interface can do that automatically for example and pop a new list of candidates immediately so the user wouldn't even notice).

If I compare this to the "traditional" IDEs you usually operate the same way: click the popup with mouse, select one of the suggestions, click on it... then repeat.

As for the UI, I would add `flycheck-read-suggestion-function` setting where the user can plug anything which would take the suggestions generated and return one of them to be applied (this is handled by flycheck with some function like `flycheck-apply-suggeston` using the above callback for selection). By having the backend separate however people can write helm sources or company or whatever and not depend on flycheck beyond the `flycheck-suggestions-functions`.

jyp:
Since progress on this issue seemed halted I've implemented a similar functionality in another package. https://github.com/jyp/attrap
Maybe it'll be helpful.

fmdkdd:
@jyp That's great!  Thank you for notifying us.  I haven't had time to look into this PR again, but I did took a quick look at `attrap`.

I see `attrap` matches against errors, and provides corrections for selected errors.  I don't think we want to go in that direction for  Flycheck itself, as that would be too much overhead (error messages change too often).  However, I know that some checkers can provide suggestions directly, e.g., for `rust` fixes are part of the compiler output.  I think we could provide a "generic" error fixer that takes a buffer range and replacement text, and then probably an API to extend that behavior.

My point is, whenever we have time to revisit this, I can see `attrap` as a potential consumer of that API.

bbatsov:
Closing — this WIP PR has been open for 9+ years and was never completed. The auto-fix niche has largely been filled by LSP code actions. If someone wants to revive this concept, please open a fresh PR. Thank you for the pioneering work on this idea!

PR #1025, opened 2016-07-30
Support latest output format of errcheck

dominikh:
errcheck now prints a colon, to more closely match the quickfix format.

We now support 3 different revisions of errcheck:

- the one that prints all indentation of the actual code line (multiple
  tabs)

- the one that separates position and code with a single tab

- the one that separates position and code with a colon and a single
  tab.


lunaryorn:
LGTM

cpitclaudel:
Same here; LGTM, thanks!

lunaryorn:
Thanks!

PR #1029, opened 2016-08-07
maint: rename compile.el -> flycheck-compile.el

dunn:
When byte-compiling flycheck.el manually, if you use `-L maint` instead
of `-l maint/compile.el`, maint/ will be placed at the front of the
load-path, overriding Emacs' build-in compile.el and causing the
following error:

```
Eager macro-expansion failure: (error "Required feature ‘compile’ was not provided")
Eager macro-expansion failure: (error "Required feature ‘compile’ was not provided")

In toplevel form:
flycheck-package.el:41:1:Error: Required feature ‘compile’ was not provided
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ flycheck-ert.elc: flycheck.elc
 flycheck-buttercup.elc: flycheck.elc
 
 $(OBJS): %.elc: %.el
-	$(EMACSBATCH) -l maint/compile.el -f flycheck/batch-byte-compile $<
+	$(EMACSBATCH) -l maint/flycheck-compile.el -f flycheck/batch-byte-compile $<
 
 doc/_static/logo.png: flycheck.svg
 ifndef HAVE_CONVERT
--- a/maint/flycheck-compile.el
+++ b/maint/flycheck-compile.el
@@ -1,4 +1,4 @@
-;;; compile.el --- Flycheck byte compiler            -*- lexical-binding: t; -*-
+;;; flycheck-compile.el --- Flycheck byte compiler            -*- lexical-binding: t; -*-
 
 ;; Copyright (C) 2016  Sebastian Wiesner and Flycheck contributors
 
@@ -73,4 +73,4 @@ Specifically set `byte-compile-error-on-warn' to t on Emacs 25."
           (kill-emacs 1)))))
   (kill-emacs 0))
 
-;;; compile.el ends here
+;;; flycheck-compile.el ends here


cpitclaudel:
Thanks Alex. This LGTM; but just to clarify, was there a reason you uses `-L maint`, or did you just use this automatically, and then got that confusing error?

dunn:
I was trying to install the HEAD version through Homebrew (https://github.com/Homebrew/homebrew-emacs), which byte-compiles packages by recursively adding `--directory <dir>`, instead of adding `-l <file>`.

lunaryorn:
@dunn LGTM but I can't say that I'm too happy about this change.
 
Please report this issue to the maintainers of that Homebrew formula; it must **not** add arbitrary files from our source tree to load path, for this reason.  I'm not sure why they did that; it's simply not safe to load arbitrary files from directories.

Even if we merged this change it could happen again.  We simply don't expect anyone to load arbitrary files from our source tree, nor will we support that.

Please understand that we do not support Homebrew for installing Flycheck, and take no responsibility for its flaws.

dunn:
>  I'm not sure why they did that; it's simply not safe to load arbitrary files from directories.

I was the one who did that 😰

The reasoning was I wanted a helper method for byte-compiling that didn't require manually specifying all the necessary directories and/or files.

> it's simply not safe to load arbitrary files from directories.

Normally I'd agree, but this seems to be the default behavior of Emacs (see `subdirs.el` and the responses to https://lists.gnu.org/archive/html/emacs-devel/2016-08/msg00020.html); also Homebrew (unlike MELPA) updates packages manually, so we're not loading and compiling things with no oversight.

dunn:
> Please understand that we do not support Homebrew for installing Flycheck, and take no responsibility for its flaws.

Of course!

lunaryorn:
@dunn Oh 😳 Please excuse for my harsh words, I'm terribly sorry that I hurt you 😞 

----

Still, please understand that I do not agree with the way the Homebrew formula works, and I do not agree that loading arbitrary files from directories is "default behaviour".

I'm aware of `subdirs.el` and `normal-top-level-add-subdirs-to-load-path` but a default Emacs installation uses these _only for `site-start`_.   But Flycheck [does _not_ support installation to `site-start`](http://www.flycheck.org/en/latest/user/installation.html#alternative-installation-methods), and we don't support our source code tree being dumped into `load-path` either:

> If you install Flycheck in any way other than our official packages you do so **at your own risk.**

We've got a set of supported installation methods, and there's nothing indicating that installing or even byte-compiling our entire source code tree would be safe, is there?

----

If you want you can think of our Git repo as a raw source code repository like that of the Linux Kernel or so.  You can't use it directly; you've got to build it first and install the resulting binaries.  

We distribute [binaries with our releases](https://github.com/flycheck/flycheck/releases) (see the `.tar` files) but they're meant for use with Emacs' built-in package manager.  Likewise MELPA builds binaries for us, using a procedure that matches ours.

You can build your own binaries but then you do good to use our build system.   You wouldn't compile the Linux kernel by globbing for `*.c` files and feeding them to GCC, would you?

Likewise you must not compile Flycheck this way.   The vast majority (e.g. test suite, support files, etc.) aren't meant to be compiled.  We've got `make compile` which compiles the files we meant to be compiled.

----

I'll merge this pull request but please, _please_ amend your Homebrew formula to use our `make compile` or—if that's not feasible—at least compile only those files that we list in `SRCS`.  

Everything else isn't safe to compile, and we can't promise that breakage like this one wouldn't occur again.

dunn:
> Please excuse for my harsh words, I'm terribly sorry that I hurt you

You didn't, don't worry!

> You wouldn't compile the Linux kernel by globbing for *.c files and feeding them to GCC, would you?

Ha, of course not, but few elisp projects are as complex as a kernel.

I'll look into updating the install tools, though.  The flycheck formula isn't currently using `make compile` to avoid the Cask dependency, since Homebrew can manage the dependencies itself.

Thank you for the feedback!

lunaryorn:
We could let you opt out of Cask in the Makefile of that's what you need.

dunn:
That'd be great!

lunaryorn:
Could you do me a favour and open a new issue for that so that I don't forget?  I'm on mobile currently…

dunn:
No problem! #1032.

Issue #978, opened 2016-05-22
Temporary file names in error messages


sergv:
Flycheck uses temporary files as input to checkers. After error is produced, it tries to replace temporary file name back to original name in order not to confuse users. However, it does not always work because some checkers report filenames in multiple places, e.g. ghc 7.10.3 with stack may yield error message like

```
Could not deduce (wa ~ w0 a)
from the context (Matrix w v,
                  ConstrainedFunctor w,
                  ElemConstraints v ~ ElemConstraints w)
  bound by the instance declaration
  at /tmp/emacs-tmp-31149btH/flycheck311494h0/Generic.hs:(218,3)-(222,29)
or from (MonadError String m, ElemConstraints v a)
  bound by the type signature for
             toDescription :: (MonadError String m, ElemConstraints v a) =>
                              NN w v n o a -> m (Description n o a)
  at /tmp/emacs-tmp-31149btH/flycheck311494h0/Generic.hs:(224,8)-(226,28)
  ‘wa’ is a rigid type variable bound by
       the type signature for convertLayer :: (v a, wa) -> m ([a], [[a]])
       at /tmp/emacs-tmp-31149btH/flycheck311494h0/Generic.hs:247:23
Relevant bindings include
  w' :: [[a]]
    (bound at /tmp/emacs-tmp-31149btH/flycheck311494h0/Generic.hs:256:11)
  w :: wa
    (bound at /tmp/emacs-tmp-31149btH/flycheck311494h0/Generic.hs:248:27)
  bias :: v a
    (bound at /tmp/emacs-tmp-31149btH/flycheck311494h0/Generic.hs:248:21)
  convertLayer :: (v a, wa) -> m ([a], [[a]])
    (bound at /tmp/emacs-tmp-31149btH/flycheck311494h0/Generic.hs:248:7)
  finalBias :: v a
    (bound at /tmp/emacs-tmp-31149btH/flycheck311494h0/Generic.hs:227:47)
  finalLayer :: (v a, w a)
    (bound at /tmp/emacs-tmp-31149btH/flycheck311494h0/Generic.hs:227:35)
  toDescription :: NN w v n o a -> m (Description n o a)
    (bound at /tmp/emacs-tmp-31149btH/flycheck311494h0/Generic.hs:227:3)
  (Some bindings suppressed; use -fmax-relevant-binds=N or -fno-max-relevant-binds)
In the first argument of ‘MC.toList’, namely ‘w’
In the expression: MC.toList w
```

[...]

PR #1030, opened 2016-08-08
Fix temporary filename in error body in addition to error header

sergv:
Some checkers report filename in multiple places in the error message, therefore it's not enough to fix only flycheck-error-filename field. If message is not fixed, the user may observe temporary filenames.

This is a proposed fix for #978.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5264,16 +5264,14 @@ function `buffer-file-name'."
       (when (seq-some (apply-partially #'flycheck-same-files-p
                                        (expand-file-name filename cwd))
                       buffer-files)
-        (let ((new-filename (buffer-file-name)))
+        (-when-let (new-filename (buffer-file-name))
           (setf (flycheck-error-filename err) new-filename
-                (flycheck-error-message err)
-                (replace-regexp-in-string
-                 (regexp-quote filename)
-                 new-filename
-                 (flycheck-error-message err)
-                 nil
-                 t ;; do literal substitution
-                 ))))))
+                (flycheck-error-message err) (replace-regexp-in-string
+                                              (regexp-quote filename)
+                                              new-filename
+                                              (flycheck-error-message err)
+                                              'fixed-case
+                                              'literal))))))
   err)
 
 
--- a/flycheck.el
+++ b/flycheck.el
@@ -5264,14 +5264,16 @@ function `buffer-file-name'."
       (when (seq-some (apply-partially #'flycheck-same-files-p
                                        (expand-file-name filename cwd))
                       buffer-files)
-        (-when-let (new-filename (buffer-file-name))
-          (setf (flycheck-error-filename err) new-filename
-                (flycheck-error-message err) (replace-regexp-in-string
-                                              (regexp-quote filename)
-                                              new-filename
-                                              (flycheck-error-message err)
-                                              'fixed-case
-                                              'literal))))))
+        (let ((new-filename (buffer-file-name)))
+          (setf (flycheck-error-filename err) new-filename)
+          (when new-filename
+            (setf (flycheck-error-message err)
+                  (replace-regexp-in-string
+                   (regexp-quote filename)
+                   new-filename
+                   (flycheck-error-message err)
+                   'fixed-case
+                   'literal)))))))
   err)
 
 
--- a/flycheck.el
+++ b/flycheck.el
@@ -5264,7 +5264,16 @@ function `buffer-file-name'."
       (when (seq-some (apply-partially #'flycheck-same-files-p
                                        (expand-file-name filename cwd))
                       buffer-files)
-        (setf (flycheck-error-filename err) (buffer-file-name)))))
+        (let ((new-filename (buffer-file-name)))
+          (setf (flycheck-error-filename err) new-filename
+                (flycheck-error-message err)
+                (replace-regexp-in-string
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Hi @sergv,

Thanks for this new contribution! However, I'm not too sure about the approach. Can you show an example of a checker that displays this issue?

Note that we're trying to move checkers to use stdin as much as possible, so maybe that issue will also tend to go away with time.

sergv:
Hi @cpitclaudel,

The example of a checker that has this issue is in the attached issue, #978. Thus, both `haskell-ghc` and `haskell-stack-ghc` checkers have this problem.

While it is nice that checkers are moved to use stdin, I think this is not possible to do for all checkers because some of the checkers are compilers which don't have capability to receive input from stdin. The `gcc` has the capability to compile right from stdin, but `ghc` does not have one https://ghc.haskell.org/trac/ghc/ticket/2427. I'm not sure whether it will be resolved any time soon - last update was three years ago and it is a low-priority feature.

However, while mainstream compilers may converge to accepting input from stdin at some point, less popular ones may never come to implementing such a feature. Thus users of non-mainstream compilers will not benefit from improved error messages, which does not seem right.

In addition, users may have already defined custom checkers which are based around files and have this problem of temporary file names appearing in error messages. Fixing the issue will benefit them, while resorting to stdin approach will force them to amend their checkers in order to avoid the issue.

flycheck.el
```diff
@@ -5264,7 +5264,16 @@ function `buffer-file-name'."
...
                       buffer-files)
-        (setf (flycheck-error-filename err) (buffer-file-name)))))
+        (let ((new-filename (buffer-file-name)))
+          (setf (flycheck-error-filename err) new-filename
+                (flycheck-error-message err)
+                (replace-regexp-in-string
+                 (regexp-quote filename)
+                 new-filename
+                 (flycheck-error-message err)
+                 nil
```
cpitclaudel: Are you sure you want `nil` here?

flycheck.el
```diff
@@ -5264,7 +5264,16 @@ function `buffer-file-name'."
...
+                 t ;; do literal substitution
```
cpitclaudel: Instead of `t` + a comment you can use a non-nil symbol, e.g. `'literally`, I think.

cpitclaudel:
Ok :) I have a slight uneasy feeling about this (it won't cover all cases: the checker might choose to print only parts of the name by abbreviating the beginning, for example). But I don't have strong arguments against it either.

I'd tend to like this better as a pre-defined error filter, such as `flycheck-increment-error-columns`, but I can't really think of a case where this wouldn't be the right thing to do, either... so maybe applying it unconditionally is right. @lunaryorn?

flycheck.el
```diff
@@ -5264,7 +5264,16 @@ function `buffer-file-name'."
       (when (seq-some (apply-partially #'flycheck-same-files-p
                                        (expand-file-name filename cwd))
                       buffer-files)
-        (setf (flycheck-error-filename err) (buffer-file-name)))))
+        (let ((new-filename (buffer-file-name)))
```
cpitclaudel: Are we sure that (buffer-file-name) is non-nil here? Otherwise the `replace-regexp-in-string` part will error out.

flycheck.el
```diff
@@ -5264,7 +5264,16 @@ function `buffer-file-name'."
...
+          (setf (flycheck-error-filename err) new-filename
+                (flycheck-error-message err)
+                (replace-regexp-in-string
```
cpitclaudel: Would this fit on the previous line? I had trouble parsing this visually at first :)

flycheck.el
sergv: No, that should be `t` indeed.

sergv:
@cpitclaudel I fixed review suggestions and checked that it still works.

lunaryorn:
@cpitclaudel  I'm unsure.  It feels very hacky, to be honest.  I don't think that we should make this a default; I guess I'd prefer if this was a separate error filter to be added only to those checkers that actually need it.  Error filtering shouldn't do too much magic out of the box, imho.

@sergv Can you refactor so that the substitution is in a separate error filter.

sergv:
@lunaryorn Moving this to error filter does sound like a cleaner solution. However, it seems to be not possible to move replacement into the error filter. Currently I replace generated file names for temporary files by buffer file name at the place where those generated file names are available. But those generated names are not be available in error filter. Error filter function receives only one argument - a list of `flycheck-error` structures, so no other information is available and it's unclear how to make list of generated file names available to the error filter. Do you have any ideas?

lunaryorn:
@sergv Oh, indeed, I forgot that we're talking about the generated temporary file names.  I think in this case we've got no other choice than to do what you've implemented.

So LGTM, I don't like the change, but it won't do no harm, and it's probably the right thing to do (i.e. remove all traces of our special temporary files before showing errors).

Thanks for the change.

cpitclaudel:
LGTM too :) Thanks for explaining things and for writing this patch!

cpitclaudel:
> Are we sure that `(buffer-file-name)` is non-nil here? Otherwise the `replace-regexp-in-string` part will error out.

This sounds almost prescient in retrospect.  There *is* a bug in this code, but it shows up when `message` is `nil`, instead of `buffer-file-name` :) I'll make a PR soon.

Issue #1035, opened 2016-08-10
Support new Rust error format


futile:
The rust compiler is switching to a new error format, see https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html. Flycheck currently only supports the old format, which is already not used anymore in rust's nightly builds.

Instead of updating to this new format, the rust compiler can also output errors as json (`rustc --error-format=json`), which might also be an option. This would also have the advantage of supporting both newer and older versions of the compiler. The json-error format seems to be pretty stable at this point, at least that's the story mentioned here: https://github.com/rust-lang/rust/pull/35401.

Actual discussion on the json-format: https://internals.rust-lang.org/t/stabilizing-json-compiler-message-output/3691

fmdkdd:
Thanks for the heads up.  I was aware that we would need to update flycheck to the new error format, but was waiting for it to stabilize.

As I understand, the new error format is now the default in nightly, but won't land in stable before 8 weeks.  But in the meantime, flycheck won't work with nightly, and then beta.

As you suggest, the JSON output seems like the best way forward.  But if the JSON format is not the same in stable and nightly, we will have to support both until the current one phases out (again, 8 weeks).

I can take a look over the week-end to see if we support the new format without too much hassle.

josephDunne:
+1 As I only use Nightly rust so it sucks to loose flycheck support

cpitclaudel:
Hi all,

Just to clarify: there's no format that will work with all these versions, right? If there is, I think it would be best to wait for the JSON format to land, and then move to that. Otherwise, JSON indeed seems like the right way to go right now.

fmdkdd:
@cpitclaudel From what I can gather, the JSON format is the same in stable and in nightly.  I've just tested it on a couple of errors, and it certainly looks the same.  What has changed is that it is now a stable option in nightly, while in stable and beta you would have to pass the extra `-Z unstable-options` argument in order to get the JSON output.

So actually, switching to the JSON output would mean keeping a single error parser for all three versions (stable/beta/nightly).  We just have to pass the extra argument when facing rustc < 1.12, and drop it for >= 1.12.  When 1.12 is stable, we can just get rid of this check in the code.

Otherwise, as the new error format is now the default in nightly, flycheck won't work for rust nightly users.

Any tips on parsing the JSON for flycheck?  I saw that the tslint checker already parses JSON, and I was going to base the code on that.

cpitclaudel:
[...]

PR #1036, opened 2016-08-13
Use JSON output for rust and rust-cargo checkers

fmdkdd:
See #1035 for the background discussion.

Essentially: this pull request restores flycheck for users of rust nightly, while (almost) preserving the functionality for rust stable users.

Almost, because some messages may have been slightly altered (see the tests update). 

I've switched the error parsers of rust and rust-cargo to parse the JSON output directly.  I've based the code on the tslint JSON parser, but the situation was alas not as simple.  

Things to watch out for: the `-Z unstable-options` might trigger an error in the future.  We can revisit it when it does.

Also, I _think_ the JSON output requires at least rust 1.7, meaning we might break some setups.  Is there a place where we should state that flycheck now requires a more recent rust version?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@
 
   - Change ``flycheck-eslint-rulesdir`` (string) to
     ``flycheck-eslint-rules-directories`` (list of strings) [GH-1016]
+  - Require rust 1.7 or newer for ``rust`` and ``rust-cargo`` [GH-1036]
 
 - New syntax checkers:
 
@@ -22,7 +23,8 @@
 
   - Add default directory for ``haskell-stack-ghc`` and ``haskell-ghc`` checkers
     [GH-1007]
-
+  - ``rust`` and ``rust-cargo`` checkers now support the new error format of
+    rust 1.12 [GH-1016]
 
 28 (Jun 05, 2016)
 =================
--- a/flycheck.el
+++ b/flycheck.el
@@ -5444,6 +5444,107 @@ about TSLint."
              (and (not (string-empty-p output))
                   (json-read-from-string output)))))
 
+(defun flycheck-parse-rust (output checker buffer)
+  "Parse rust errors from OUTPUT and return a list of `flycheck-error'.
+
+CHECKER and BUFFER denote the CHECKER that returned OUTPUT and
+the BUFFER that was checked respectively.
+
+The expected format for OUTPUT is a mix of plain text lines and
+JSON lines.  This function ignores the plain text lines and
+parses only JSON lines.  Each JSON line is expected to be a JSON
+object that corresponds to a diagnostic from the compiler.  The
+expected diagnostic format is described there:
+
+https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
+  (let* ((json-array-type 'list)
+         (json-false nil)
+         ;; Skip the plain text lines in OUTPUT, keep the JSON lines.
+         (json-lines (seq-filter (lambda (line)
+                                   (string-match-p "^{" line))
+                                 (split-string output "\n")))
+         ;; Each JSON line is a JSON object.
+         (diagnostics (seq-map #'json-read-from-string json-lines))
+         (errors))
+    ;; The diagnostic format is described in the link above.  The gist of it is
+    ;; that each diagnostic can have several causes in the source text; these
+    ;; causes are represented by spans.  The diagnostic has a message and a
+    ;; level (error, warning), while the spans have a filename, line, column,
+    ;; and an optional label.  The primary span points to the root cause of the
+    ;; error in the source text, while non-primary spans point to related
+    ;; causes.  In addition, each diagnostic can also have children diagnostics
+    ;; that are used to provide additional information through their message
+    ;; field, but do not seem to contain any spans (yet).
+    ;;
+    ;; We first iterate over diagnostics and their spans to turn every span into
+    ;; a flycheck error object, that we collect into the `errors' list.
+    (dolist (diagnostic diagnostics)
+      (let ((error-message)
+            (error-level)
+            (error-code)
+            (primary-filename)
+            (primary-line)
+            (primary-column)
+            (spans)
+            (children))
+
+        ;; Nested `let-alist' cause compilation warnings, hence we `setq' all
+        ;; these values here first to avoid nesting.
+        (let-alist diagnostic
+          (setq error-message .message
+                error-level (pcase .level
+                              (`"error" 'error)
+                              (`"warning" 'warning)
+                              (_ 'error))
+                ;; The 'code' field of the diagnostic contains the actual error
+                ;; code and an optional explanation that we ignore
+                error-code .code.code
+                spans .spans
+                children .children))
+
+        (dolist (span spans)
+          (let-alist span
+            ;; Children lack any filename/line/column information, so we use
+            ;; those from the primary span
+            (when .is_primary
+              (setq primary-filename .file_name
+                    primary-line .line_start
+                    primary-column .column_start))
+            (push
+             (flycheck-error-new-at
+              .line_start
+              .column_start
+              ;; Non-primary spans are used for notes
+              (if .is_primary error-level 'info)
+              (if .is_primary
+                  ;; Primary spans may have labels with additional information
+                  (concat error-message (when .label
+                                          (format " (%s)" .label)))
+                .label)
+              :id error-code
+              :checker checker
+              :buffer buffer
+              :filename .file_name)
+             errors)))
+
+        ;; Then we turn children messages into flycheck errors pointing to the
+        ;; location of the primary span.  According to the format, children
+        ;; may contain spans, but they do not seem to use them in practice.
+        (dolist (child children)
+          (let-alist child
+            (push
+             (flycheck-error-new-at
+              primary-line
+              primary-column
+              'info
+              .message
+              :id error-code
+              :checker checker
+              :buffer buffer
+              :filename primary-filename)
+             errors)))))
+    (nreverse errors)))
+
 
 ;;; Error parsing with regular expressions
 (defun flycheck-get-regexp (patterns)
@@ -8276,7 +8377,7 @@ Relative paths are relative to the file being checked."
 (flycheck-define-checker rust-cargo
   "A Rust syntax checker using Cargo.
 
-This syntax checker needs Rust 1.0 or newer, and Cargo with the
+This syntax checker needs Rust 1.7 or newer, and Cargo with the
 rustc command.  See URL `https://www.rust-lang.org'."
   :command ("cargo" "rustc"
             (eval (cond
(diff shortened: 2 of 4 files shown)


mkpankov:
Maybe we should add an option to work with JSON error format? Or, rather, to go back to plain text parsing that is present currently - like "compatibility mode". Sticking to interface that is *both* unstable and only available since specific version is scary IMO.

fmdkdd:
@mkpankov The JSON output is informally stable (see [this discussion]), and the [PR] talks about maintaining backward compatibility, so I would not worry on that front.

I'm also in favor of not dropping support for rust < 1.7, since rust releases may have breaking changes, and this is a legitimate reason to stay on a previous version.  However I believe flycheck maintainers would rather have only one parser, so that's why I went with JSON only in this PR.   But if that is not the case, I can add the text parser back in.  

Keep in mind though that the text parser would only work until rust 1.12, where the new text format is the default.

As one data point, the [rust poll] indicates most respondents prefer the latest stable and nightly versions.  For the moment, flycheck does not work for rust nightly users at all.

[this discussion]: https://internals.rust-lang.org/t/stabilizing-json-compiler-message-output/3691/23)
[rust poll]: https://blog.rust-lang.org/2016/06/30/State-of-Rust-Survey-2016.html
[PR]: rust-lang/rust/pull/35401

mkpankov:
Right.

I noticed PR doesn't build on Travis on snapshot Emacs, but that doesn't look like our problem to me. Looks like we'll have to wait for @lunaryorn

mkpankov:
As for PR itself: checked the branch locally, works for me with nightly and stable 1.10 :)

lunaryorn:
@fmdkdd Thank you for this change.  

I must admit, though, that I find some things in the parsing code strange.  Why do we need to preprocess the output? What is a JSON "line"?  Could you perhaps show an example of this JSON?

flycheck.el
```diff
@@ -5440,6 +5440,75 @@ about TSLint."
              (and (not (string-empty-p output))
                   (json-read-from-string output)))))
 
+(defun flycheck-parse-rust (output checker buffer)
+  "Parse rust errors from JSON OUTPUT and return a list of `flycheck-error'.
+
+CHECKER and BUFFER denote the CHECKER that returned OUTPUT and
+the BUFFER that was checked respectively."
+  (let* ((json-array-type 'list)
+         (json-false nil)
```
lunaryorn: Isn't that the default?

flycheck.el
```diff
@@ -5440,6 +5440,75 @@ about TSLint."
...
+         (errors))
+    ;; The JSON format of rust errors is described there:
+    ;; https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139
+    ;; The gist of it is that each JSON line can contain one more spans.  Each
+    ;; span points to the root cause of error in the source text (the primary
+    ;; span), or to related causes.
+    (dolist (message messages (nreverse errors))
+      (let-alist message
+        (let ((error-message .message)
+              ;; Assumption: .level is "error" or "warning"
```
lunaryorn: Please use `pcase` to explicitly turn the strings into Flycheck levels and signal an explicit error or add an explicit default if an unknown level occurs.

fmdkdd:
@lunaryorn When passing the JSON flag to rustc, each output line is a JSON object (and /not/ in a JSON array mind you).  But when passing the flag through cargo, we have a mixed output of cargo text and JSON lines from rustc.

Here is an example of what you currently get:

```
$ cargo rustc --bin chipers -- -Z no-trans -Z unstable-options --error-format=json
   Compiling chipers v0.1.0 (file:///home/fmdkdd/proj/chipers)
warning: the option `Z` is unstable and should only be used on the nightly compiler, but it is currently accepted for backwards compatibility; this will soon change, see issue #31847 for more details
warning: the option `error-format` is unstable and should only be used on the nightly compiler, but it is currently accepted for backwards compatibility; this will soon change, see issue #31847 for more details
{"message":"expected `,`, or `}`, found `pub`","code":null,"level":"error","spans":[{"file_name":"src/cpu.rs","byte_start":10872,"byte_end":10875,"line_start":30,"line_end":30,"column_start":3,"column_end":6,"is_primary":true,"text":[{"text":"  pub ram_writes: [u64; RAM_LENGTH],","highlight_start":3,"highlight_end":6}],"label":null,"suggested_replacement":null,"expansion":null}],"children":[{"message":"struct fields should be separated by commas","code":null,"level":"help","spans":[],"children":[],"rendered":null}],"rendered":null}
{"message":"aborting due to previous error","code":null,"level":"error","spans":[],"children":[],"rendered":null}
error: Could not compile `chipers`.

To learn more, run the command again with --verbose.
```

There are talks of making cargo output JSON as well in the future.  I presume we would still need to get the compiler errors out of this JSON, but the parsing code might be simplified a bit.

flycheck.el
fmdkdd: The default seems to be `:json-false` from reading the source code.

lunaryorn:
@fmdkdd Oh, that's, uhm, an interesting format :joy:  It's rather unusual to print line-wise JSON data, imho.  Would you mind to add a (lengthy?) comment that explains this format to the docstring of the parsing function?  That'd really help us to be able to maintain the parser in future.

flycheck.el
lunaryorn: Ah, interesting, I didn't know that.  Good choice, then :+1:

lunaryorn:
LGTM, but please add an explanation of the format to the docstring, as said above, and please use `pcase` to parse the error level as by my inline comment.

Other than that I think it's ready to be merged. @cpitclaudel would you like to take a look?

lunaryorn:
@mkpankov Thanks for reviewing, and for testing this pull request.

flycheck.el
fmdkdd: Right, this is probably a better idea than calling `intern` on whatever.

fmdkdd:
@lunaryorn I've added a commit with the following changes:

- Describe the expected input to the parser.
- Describe the relations between diagnostics, spans and children.
- Rename `messages` to `diagnostics`
- Use `pcase` to catch unexpected error levels.
- Replace a `concat` by a more readable `format`.

Hopefully I did not go overboard with the explanations.  The logic for this function was unfortunately less straightforward than expected. 

Thanks for the review, and indeed thanks @mkpankov for your input.

flycheck.el
```diff
@@ -5440,6 +5440,95 @@ about TSLint."
...
+    ;; level (error, warning), while the spans have a filename, line, column,
+    ;; and an optional label.  The primary span points to the root cause of the
+    ;; error in the source text, while non-primary spans point to related
+    ;; causes.  In addition, each diagnostic can also have children diagnostics
+    ;; that are used to provide additional information through their message
+    ;; field, but do not seem to contain any spans (yet).
+    ;;
+    ;; We first iterate over diagnostics and their spans to turn every span into
+    ;; a flycheck error object, that we collect into the `errors' list.
+    (dolist (diagnostic diagnostics (nreverse errors))
```
lunaryorn: Oh, good intentions :+1:, but please don't use the 3rd argument to `dolist` :blush:.   I absolutely don't like it; it's awfully confusing, particularly to folks which are not too familiar with Emacs Lisp.

Just stick to returning `(nreverse errors)` at the very end, like in all the other error parsers.

lunaryorn:
@fmdkdd Thanks a lot, particularly for these lengthy explanations.  These are really helpful!

Just two things left:

* Please don't use this fancy 3rd argument to `dolist` (see my inline comment).
* The Travis CI build fails with some compiler warnings on Emacs snapshot that I admittedly don't really understand.  Maybe you could take another look and try to get rid of these warnings on Emacs 25?  Or find a workaround?  We require that our code triggers no warnings on Emacs 25.

If you can't get your head around these warnings just tell me and I'll take a look myself this weekend :blush:

lunaryorn:
Forgot one thing: If you're done please squash all commits into a single one :blush:

flycheck.el
fmdkdd: Right!  It probably did confuse me the first time I saw it too.

fmdkdd:
@lunaryorn It appears that nesting `let-alist` calls trigger the warnings.  Unnesting definitely remove them anyway.

Once the Travis confirms this, I will squash everything for the merge.

lunaryorn:
@fmdkdd Oh, that's unfortunate.  I'm not sure whether your refactoring is the best way out of this, though.  We shouldn't use mutable variables too much, especially not to preserve a syntactic form.

I'll think about a better way to structure the code—doesn't stand against merging, though, as I can refactor later.

lunaryorn:
@fmdkdd Looks good. I've restarted the failing build, which was a random fluke.

I forgot one thing: Now that we require Rust 1.7 or newer, please mention the new version requirements in the docstrings of `rust` and `rust-cargo`, and in `doc/languages.rst` as well (you may need to rebase onto master first for the latest updates to the latter).

After that I think it's ready to go, but let's have @cpitclaudel opinion first :)

fmdkdd:
@lunaryorn I agree that the workaround is not ideal; I tried to avoid `(cdr (assq`.  I read that there is an `alist-get` in Emacs 25 that could alleviate that. 

But actually, I think it's a bug/limitation in `let-alist`.  Here is what a nested call expands to:

```
(cl-prettyexpand
 '(let-alist '((a . 1) (b . 2))
    (let-alist '((c . 3) (d . 4))
      .c)))

(let ((alist '((a . 1) (b . 2))))
  (let ((\.c (cdr (assq 'c alist))))  <-- triggers a warning for the unused binding
    (let ((alist '((c . 3) (d . 4))))
      (let ((\.c (cdr (assq 'c alist))))
        \.c))))
```

Oviously the first `let` is premature.  The correct thing would be to expand the inner `let-alist` first, then the outer one would just ignore the `\.c` and not write the first `let`.

The docstring acknowledges the fact that you can nest them, with the downside that you won't be able to access the variables of the outer one: we can see that's indeed the case because `alist` is shadowed.  This is a macro hygiene issue.

I can report this and see if the maintainers want to support nested calls at all.  This affects flycheck-rust as well.

Otherwise, I've rebased, stated that we require 1.7 in all the right places, and squashed.

flycheck.el
```diff
@@ -5440,6 +5440,107 @@ about TSLint."
...
+parses only JSON lines.  Each JSON line is expected to be a JSON
+object that corresponds to a diagnostic from the compiler.  The
+expected diagnostic format is described there:
+
+https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
+  (let* ((json-array-type 'list)
+         (json-false nil)
+         ;; Skip the plain text lines in OUTPUT, keep the JSON lines.
+         (json-lines (seq-filter (lambda (line)
+                                   (string-match-p "^\{" line))
```
cpitclaudel: I think escaping `{` is superfluous.

cpitclaudel:
Looks great, thanks (LGTM)! It's too bad that `let-alist`s can't be nested, but the commit is very readable. I made a tiny comment, but it isn't essential.

flycheck.el
fmdkdd: You are right.  I began by splitting the output using `sed` in my shell, then just used the same expression in elisp :)

lunaryorn:
@fmdkdd Thanks, LGTM. Can you add a note about this change to the `Breaking changes` section of `CHANGES.rst` before you merge :blush:?

fmdkdd:
@lunaryorn Sure.  In the same commit, or as a new one?

lunaryorn:
@fmdkdd In the same if possible :)

fmdkdd:
@lunaryorn Done.  Let me know if that works for you.

CHANGES.rst
```diff
@@ -6,6 +6,8 @@
   - Change ``flycheck-eslint-rulesdir`` (string) to
     ``flycheck-eslint-rules-directories`` (list of strings) [GH-1016]
```
lunaryorn: There's no need for this blank line.

CHANGES.rst
```diff
@@ -23,6 +25,8 @@
   - Add default directory for ``haskell-stack-ghc`` and ``haskell-ghc`` checkers
     [GH-1007]
```
lunaryorn: There's no need for this blank line.

lunaryorn:
@fmdkdd Two minor style nitpicks.  Please fix these and merge the pull request.  I think you need to rebase onto `master` first to get the Travis CI build working again: Emacs snapshot is currenlty failing, so I added Emacs pretest to the build and allowed the snapshot build to fail on `master`.

fmdkdd:
@lunaryorn Nits fixed.  Rebased and indeed CI now builds.  Should be good to go!

lunaryorn:
Please go ahead and merge 😊

cpitclaudel:
Yay! Great work :)

lunaryorn:
@fmdkdd Thanks a lot!

Issue #1021, opened 2016-07-25
Include configuration files into verification buffer


lunaryorn:
If a syntax checker uses a configuration file the presence of the configuration and its path should be included in the verification buffer.

PR #1038, opened 2016-08-16
Include config file in verification buffer if any

lunaryorn:
Closes GH-1021 which see.

LGTM, for the record.  @cpitclaudel ?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,8 @@
     [GH-1007]
   - ``rust`` and ``rust-cargo`` checkers now support the new error format of
     rust 1.12 [GH-1016]
+  - `flycheck-verify-checker` and `flycheck-verify-setup` now include
+    information about configuration files of syntax checkers [GH-1021] [GH-1038]
 
 28 (Jun 05, 2016)
 =================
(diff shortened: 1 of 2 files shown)


lunaryorn:
Oh dear, MELPA still down?

cpitclaudel:
LGTM. I guess (?) you could use a `-when-let` in there and save the `and` in `path`:

    ,@(-when-let (value (bound-and-true-p config-file-var))

But the behavior would be a bit different (it wouldn't print anything about config files when `config-file-var` has value nil)

(Unrelated: every time I see an `,@` followed by a a `list` with a single element inside a `when` I find myself wanting for a smarter version of `,@` that properly interpolates `Maybe`s)

lunaryorn:
I'm not sure; `config-file-var` is a variable poiting to a symbol, not the symbol itself, so I don't think `bound-and-true-p` would work.

LGTM (because it didn't catch the first one).

cpitclaudel:
Urgh, forgot that `bound-and-true-p` was a macro :) LGTM.

lunaryorn:
MMh, looks like LGTM.co lost track of this one.  I'll use my admin privileges to merge.

PR #1042, opened 2016-08-18
immediately show errors after explicit jumps

### LINKED ISSUE #1034: no delay for echo area message when explicitly jumping to errors
whenever i jump to an error, the echo area shows the error only after a delay. this is fine during normal editing, but when i _explicitly_ tell emacs to jump to an error, it would be useful to see the error message immediately, since the main reason to jump to an error is to investigate it. :-)

it would be great if flycheck ignored `flycheck-display-errors-delay` whenever triggered by an explicit jump to an error location using one of the flycheck "go to error" functions.

a condition like this in `(flycheck-display-error-at-point-soon)` would perhaps be a start:

```elisp
(memq this-command '(flycheck-first-error flycheck-next-error flycheck-previous-error))
```

wbolster:
immediately show any errors at point after using
flycheck-{first,next,previous}-error. fixes #1034.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3706,7 +3706,8 @@ position."
   (when (consp n)
     ;; Universal prefix argument means reset
     (setq reset t n nil))
-  (flycheck-next-error-function n reset))
+  (flycheck-next-error-function n reset)
+  (flycheck-display-error-at-point))
 
 (defun flycheck-previous-error (&optional n)
   "Visit the N-th previous error.


lunaryorn:
LGTM.

lunaryorn:
@wbolster Oh, I forgot, please amend your commit to capitalise the first letter of the commit message.

wbolster:
done.

ceterum censeo that capital letters are nonsense. :)

cpitclaudel:
Thanks! LGTM.

Issue #1032, opened 2016-08-08
Option to byte-compile without Cask


dunn:
Follow-up to https://github.com/flycheck/flycheck/pull/1029#issuecomment-238276670; it'd be nice to have a Make target that doesn't require Cask.  My usecase is Homebrew, where dependencies are already taken care of.

The path to each dependency might have to be specified in the Makefile with some sort of sensible default; this is how I did it in one case: https://github.com/dunn/homebrew-mode/blob/master/Makefile#L7-L16

lunaryorn:
@dunn Thanks for opening this issue, I'll add an option soon.  Sorry for the delay, I've been on vacation the last days.

dunn:
No worries!

lunaryorn:
@dunn See #1043.  You've got to pass `EMACSFLAGS` to setup load path explicitly via `-L`.  The makefile doesn't try and guess where dependencies might be located, as I don't think that there's a sensible default in the absence of Cask.

dunn:
Fixed by #1043, thanks again!

lunaryorn:
@dunn You're welcome.

PR #1043, opened 2016-08-19
Allow byte compilation without Cask

lunaryorn:
See #1032. Use with:

```console
$ make CASK= EMACSOPTS'<flags that setup load path accordingly>'  compile
```

@dunn Would you please try and check whether this patch suites your needs in the Homebrew formula?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,12 +29,18 @@ SELECTOR = (language $(LANGUAGE))
 endif
 
 # Internal variables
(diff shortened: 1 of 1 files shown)


dunn:
Yes! I think you meant `EMACSOPTS` rather than `EMACSFLAGS`, but this works, thanks so much.

lunaryorn:
@dunn Yup, sorry for the confusion, I've fixed the comment.

cpitclaudel:
LGTM too :)

PR #1047, opened 2016-08-21
[lua-luacheck] specify filename when reading from stdin

### LINKED ISSUE #1046: Add support for files in luacheckrc
As of #767 , luacheck are reading contents from standard input, which means following config in .luacheckrc will not work

```lua
files['*_spec.lua'] = { std = '+busted' }
```

since luacheck won't know the file name.

Anyway to workaround this?

ghprince:
Close #1046 

@lunaryorn Though I haven't written any Lisp before, I thought this gonna be a simple fix. But apparently it is not. Would you please help with the following issues?

In `:error_patterns`, after `line-start` I used `(minimal-match (zero-or-more not-newline))` instead of `(file-name)` because the latter failed to match empty file name (e.g. when buffer is not saved yet, and `source-original` return empty string.

It worked for me for my current use case, but I don't think it is a good solution (e.g. what will happen on Windows, when the file name already include `:`?)

And also, in luacheckrc, it used globbing instead of regex for matching in `files[<glob>]`
https://github.com/mpeterv/luacheck/blob/master/src/luacheck/globbing.lua#L4
http://luacheck.readthedocs.io/en/stable/config.html#per-file-and-per-path-overrides

So ideally here we should do something like:

### If `.luacheckrc` is found, 
for `--filename` we passed in the pathname **relative** to the folder which has `.luacheckrc`.

 e.g.
```
$ tree -a code
code
├── .luacheckrc
└── spec
    └── test_app_spec.lua

$ cat /Users/abc/code/spec/test_app_spec.lua | luacheck --config /Users/abc/code/.luacheckrc -filename spec/test_app_spec.lua -
```
And expect `<filename>:` at the beginning of every output line

### If `.luacheckrc` is not found or buffer-file-name is nil
do not specify `--filename`

And expect `stdin:` at the beginning of every output line

Not sure how to achieve this, can you give me some hint so that I know what the next step could be?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7446,7 +7446,7 @@ See URL `https://github.com/mpeterv/luacheck'."
             "-")
   :standard-input t
   :error-patterns
-  ((warning line-start 
+  ((warning line-start
             (optional (file-name))
             ":" line ":" column
             ": (" (id "W" (one-or-more digit)) ") "
--- a/flycheck.el
+++ b/flycheck.el
@@ -7447,12 +7447,12 @@ See URL `https://github.com/mpeterv/luacheck'."
   :standard-input t
   :error-patterns
   ((warning line-start 
-            (minimal-match (zero-or-more not-newline))
+            (optional (file-name))
             ":" line ":" column
             ": (" (id "W" (one-or-more digit)) ") "
             (message) line-end)
-   (error line-start 
-          (minimal-match (zero-or-more not-newline))
+   (error line-start
+          (optional (file-name))
           ":" line ":" column ":"
           ;; `luacheck' before 0.11.0 did not output codes for errors, hence
           ;; the ID is optional here
--- a/flycheck.el
+++ b/flycheck.el
@@ -7441,14 +7441,19 @@ See URL `https://github.com/mpeterv/luacheck'."
             "--formatter" "plain"
             "--codes"                   ; Show warning codes
             "--no-color"
+            "--filename" source-original
             ;; Read from standard input
             "-")
   :standard-input t
   :error-patterns
-  ((warning line-start "stdin:" line ":" column
+  ((warning line-start 
+            (minimal-match (zero-or-more not-newline))
+            ":" line ":" column
             ": (" (id "W" (one-or-more digit)) ") "
             (message) line-end)
-   (error line-start "stdin:" line ":" column ":"
+   (error line-start 
+          (minimal-match (zero-or-more not-newline))
+          ":" line ":" column ":"
           ;; `luacheck' before 0.11.0 did not output codes for errors, hence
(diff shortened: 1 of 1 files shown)


lunaryorn:
@ghprince Do we really need to pass the relative file name?  I'd expect that luacheck would handle this itself; other tools do as far as I know.  Could you get a comment from the luacheck maintainers?

flycheck.el
```diff
@@ -7441,14 +7441,19 @@ See URL `https://github.com/mpeterv/luacheck'."
...
             "--codes"                   ; Show warning codes
             "--no-color"
+            "--filename" source-original
             ;; Read from standard input
             "-")
   :standard-input t
   :error-patterns
-  ((warning line-start "stdin:" line ":" column
+  ((warning line-start 
+            (minimal-match (zero-or-more not-newline))
```
lunaryorn: Can you try `(optional (file-name))`?

lunaryorn:
@ghprince Does it work with `optional`?  Did you try?

mpeterv:
> @ghprince Do we really need to pass the relative file name? I'd expect that luacheck would handle this itself; other tools do as far as I know. Could you get a comment from the luacheck maintainers?

It can handle it, but `--config` shouldn't be used; instead, luacheck should be started from the directory containing the checked file. see mpeterv/luacheck#68. This way, there is also no need to manually look for the config on the side of flycheck.

lunaryorn:
@ghprince If I judge the outcome of the luacheck issue aright the path forward is to pass the absolute filename and remove the explicit config file support?

ghprince:
@mpeterv thx a lot for your comment

@lunaryorn yep, I think so. After I remove the `--config` part in command, and use `optional` in output parsing, all my use cases seem to be working 😄 
CI failed though, I'll try to fix it tonight (I am in GMT+8), I'll let you know if I feel it's too difficult for me to do.

lunaryorn:
@ghprince Thanks a lot.  But please don't mix these two changes in a pull request.  Please keep this pull request constrained to the `--filename` option, and open a *separate* PR for the removal of the configuration file option. That'd be awesome.

ghprince:
@lunaryorn it turns out there was a trailing whitespace that fails the CI, my bad. I also reverted the change about configuration file in this PR, and It it green now. Should I create another PR to remove `luacheckrc` after your merge this branch or against current master?

lunaryorn:
LGTM.

lunaryorn:
@ghprince You can just open the new pull request right away :blush:w

cpitclaudel:
LGTM; thanks!

lunaryorn:
@cpitclaudel Looks like the service didn't pick up your edit so I've force-merged the PR.  Thanks for your feedback!

PR #1057, opened 2016-08-30
let luacheck find luacheck itself

ghprince:
As discussed in #1047, `luacheck` is capable of finding its config file `.luacheckrc` itself, it also works better with `--filename` option.

Therefore, we should remove `config-file` call for `lua-luacheck`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@
 - **Breaking changes**
 
   - Flycheck now requires flake8 3.0 or newer
+  - Remove ``--config`` option in ``lua-luacheck`` in favour of ``luacheck``'s
+    own ``.luacheckrc`` detection. Therefore ``flycheck-luacheckrc`` is
+    no longer used [GH-1057]
 
 - New syntax checkers:
 
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3463,27 +3463,22 @@ Why not:
    "language/less/syntax-error.less" 'less-css-mode
    '(1 1 error "Unrecognised input" :checker less)))
 
-(flycheck-ert-def-checker-test luacheck lua syntax-error
+(flycheck-ert-def-checker-test lua-luacheck lua syntax-error
   (flycheck-ert-should-syntax-check
    "language/lua/syntax-error.lua" 'lua-mode
-   '(5 7 error "unfinished string" :id "E011" :checker luacheck)))
+   '(5 7 error "unfinished string" :id "E011" :checker lua-luacheck)))
 
-(flycheck-ert-def-checker-test luacheck lua warnings
+(flycheck-ert-def-checker-test lua-luacheck lua warnings
   (flycheck-ert-should-syntax-check
    "language/lua/warnings.lua" 'lua-mode
    '(1 1 warning "setting non-standard global variable 'global_var'"
-       :id "W111" :checker luacheck)
+       :id "W111" :checker lua-luacheck)
    '(3 16 warning "unused function 'test'"
-       :id "W211" :checker luacheck)
+       :id "W211" :checker lua-luacheck)
    '(3 21 warning "unused argument 'arg'"
-       :id "W212" :checker luacheck)
+       :id "W212" :checker lua-luacheck)
    '(4 11 warning "variable 'var2' is never set"
-       :id "W221" :checker luacheck)))
-
-(flycheck-ert-def-checker-test lua-luacheck lua no-warnings
-  (let ((flycheck-luacheckrc "luacheckrc"))
-    (flycheck-ert-should-syntax-check
-     "language/lua/warnings.lua" 'lua-mode)))
+       :id "W221" :checker lua-luacheck)))
 
 (flycheck-ert-def-checker-test lua lua nil
   (let ((flycheck-disabled-checkers '(lua-luacheck)))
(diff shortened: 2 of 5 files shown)


lunaryorn:
@ghprince Thanks a lot!  

May I ask you to remove the definition of the variable as well, and maybe add a note about the removal to `CHANGES.rst`?

CHANGES.rst
```diff
@@ -1,6 +1,11 @@
 30-cvs (in development)
 =======================
 
+- Improvements:
```
lunaryorn: Please list this as `**Breaking change**:`, as it'll break configurations that try to set this variable explicitly.

lunaryorn:
@ghprince Oh, I forgot about `doc/languages.rst` which is the documentation that appears on our website.  Please remove the documentation of the option in this document (that's what's breaking Travis currently).

ghprince:
@lunaryorn thx a lot for your advice. working on it :)  do I need to remove the section in `doc/legacy/flycheck.html`?

lunaryorn:
@ghprince Please ignore that.  It's just for reference for stuff that's not in the new manual yet, but it'll go away eventually.

ghprince:
@lunaryorn also there is a test in `test/flycheck-test.el` try to use `test/resources/language/lua/luacheckrc` as config file. After removing those, shall we add a test case to see if `luacheck` really finds the `.luacheckrc` as advertised? if yes, I wonder how... creating a file on the fly?

lunaryorn:
@ghprince Ah, sorry. I forgot about the test.  Just remove it.  With the option gone, it's redundant, and I don't think that we should try and test luacheck's behaviour.  I think we can just rely upon luacheck doing its job properly, and I hope that luacheck has tests for that 😎 

So just remove that test case and the corresponding config file at `test/resources/language/lua/luacheckrc`.

ghprince:
@lunaryorn oops, didn't see your reply... I fixed that test by copy-and-then-delete to create temporary `.luacheckrc` file. It works and passed `make LANGUAGE=lua integ` on my local machine. I also fixes other broken luacheck integration test cases. Since those broken tests failed on master too, may I ask why they are not running during CI?

lunaryorn:
@ghprince Oh, thanks for fixing these test cases.  I wished we could run them on Travis CI, and in fact we used to do that long ago but it didn't work so well.  Travis CI took ages to install all the syntax checker tools that we support, and tests would constantly break due to minor changes in the output of some tool so we'd spend a non-trivial amount of time on constantly updating the test cases.  

It got out of hand and that's where I got fed up, split the integration tests from the rest and decided to stop running the integration tests on Travis CI.  Our builds are a lot more stable now, and feedback for contributors comes much faster.

We now only use the integration tests for local testing when making changes to syntax checkers, i.e. I sometimes ask contributors to run `make LANGUAGE=foo integ` to run the integration tests for a particular language they've been making changes to.

If you've got any idea how we could run the language tests or at least a subset thereof on Travis CI or any other CI platform I'd be happy to hear it.

----

As to this PR, please rebase on master and resolve the merge conflict in `CHANGES.rst`.  Please put the breaking changes section first, before `New syntax checkers`.

test/flycheck-test.el
```diff
@@ -3463,27 +3463,29 @@ Why not:
...
 
 (flycheck-ert-def-checker-test lua-luacheck lua no-warnings
-  (let ((flycheck-luacheckrc "luacheckrc"))
-    (flycheck-ert-should-syntax-check
-     "language/lua/warnings.lua" 'lua-mode)))
+  (copy-file (flycheck-ert-resource-filename "language/lua/luacheckrc")
+             (flycheck-ert-resource-filename "language/lua/.luacheckrc"))
+  (flycheck-ert-should-syntax-check
+   "language/lua/warnings.lua" 'lua-mode)
+  (delete-file (flycheck-ert-resource-filename "language/lua/.luacheckrc")))
```
lunaryorn: That's a nice idea, but as far as I'm concerned I'd like to ask you to remove the test case entirely.  All that we're still testing here is that luacheck itself works correctly and I think that's not our concern.  I think we can rely upon luacheck to do its job correctly.

Please remove this test case and the `luacheckrc` configuration file.  Let's keep the tests confined to what's really done by Flycheck, i.e. parsing the output of Luacheck which is already covered by the other test cases for luacheck.

ghprince:
@lunaryorn Thanks a lot for your explanation! It's good to know the history. I don't have any ideas of improving the workflow now, but will definitely share with you when I come up with something.

And I have rebased this branch onto current master and fixed the conflict as you suggested.

lunaryorn:
@ghprince Thanks a bunch.  LGTM now.  Would you mind to squash your commits?

ghprince:
@lunaryorn I think github merge button and do [squash-and-merge](https://github.com/blog/2141-squash-your-commits). but anyway I just squashed this branch :)

lunaryorn:
@ghprince Yup, but just in case someone accidentally presses merge :wink: and in somecases I prefer to merge normally because Github's squash merge looses commit signatures 😞 

By the way, would you like to join Flycheck and help maintain Lua support?  We've got [no Lua team yet](http://www.flycheck.org/en/latest/community/people.html#language-teams) and could use some help here :blush:  In fact we could use any helping hand, and I'd be happy to welcome you as a contributor :blush:

cpitclaudel:
LGTM; thanks for this contribution!

ghprince:
@lunaryorn Thanks a lot for your invitation. I would like to join :)

PR #1058, opened 2016-08-31
Fix alignment bug in flycheck-compute-message-column-offset

cpitclaudel:
Introduced by 7a82b38c094a43eee2ab42608f4e3bd0bf61bf28.  The check used
to locate the message column was overly strict.

Now that message is the last column, though, I wonder if we should just do `(butlast` on the list of columns instead of this convoluted check.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3869,9 +3869,10 @@ Return a list with the contents of the table cell."
                                          (padding (plist-get props :pad-right)))
                               (cons name (+ width (or padding 1)))))
                           tabulated-list-format))
-         (before-msg (seq-take-while (lambda (fmt)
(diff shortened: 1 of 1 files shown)


lunaryorn:
LGTM, but I also like a `butlast` variant…  I don't think it's urgent to refactor but if it's easy enough please go ahead :)

I'm going to merge this now to fix the bug, but feel free to refactor.  By the way as far as I'm concerned you can push such bug fixes directly to master.

cpitclaudel:
Thanks. I'll push a `butlast` version soon.

I also noticed that the name of the checker is appearing on a single line of it's own; I'll open a PR for that part.

lunaryorn:
Thanks a lot, @cpitclaudel

PR #1063, opened 2016-09-01
Speed up elixir-dogma predicate

aaronjensen:
Running mix takes quite a bit of time so it causes stutters when it runs

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6417,13 +6417,13 @@ about dogma."
     (nreverse errors)))
 
 (defun flycheck-elixir--check-for-dogma ()
-  "Check if the `dogma' subcommand is available.
+  "Check if `dogma' is installed.
 
-Check by parsing the output of `mix help'.
+Check by looking for deps/dogma in this directory or a parent to
+handle umbrella apps.
 Used as a predicate for enabling the checker."
-  (let ((mix (flycheck-checker-executable 'elixir-dogma)))
-    (seq-find (lambda (l) (string-match-p "dogma" l))
-              (ignore-errors (process-lines mix)))))
+  (and buffer-file-name
+       (locate-dominating-file buffer-file-name "deps/dogma")))
 
 (flycheck-define-checker elixir-dogma
   "An Elixir syntax checker using the Dogma analysis tool.
--- a/doc/community/people.rst
+++ b/doc/community/people.rst
@@ -135,6 +135,7 @@ Contributors
 The following people—listed in alphabetical order—contributed substantial code
 to Flycheck:
 
+* Aaron Jensen (:gh:`aaronjensen`)
 * Alain Kalker (:gh:`ackalker`)
 * Alex Reed (:gh:`acr4`)
 * Atila Neves (:gh:`atilaneves`)


flycheck.el
```diff
@@ -6417,13 +6417,13 @@ about dogma."
...
+  "Check if `dogma' is installed and the file is saved.
 
-Check by parsing the output of `mix help'.
+Check by looking for deps/dogma.
 Used as a predicate for enabling the checker."
-  (let ((mix (flycheck-checker-executable 'elixir-dogma)))
-    (seq-find (lambda (l) (string-match-p "dogma" l))
-              (ignore-errors (process-lines mix)))))
+  (and
+   (flycheck-buffer-saved-p)
```
lunaryorn: Do we need this?  You're still using `source` in `:command` after all…

aaronjensen: ah good catch. My version of this checker used `source-original`

lunaryorn: @aaronjensen Do you use your own version?  Why's that?  I'm just being curious… if it's a better variant why not upstream it to Flycheck?

aaronjensen: I didn't know about this one until today. It's more compact and uses the flycheck format instead of json:

```
  (flycheck-define-checker elixir-dogma
    "Defines a checker for elixir with dogma"
    :command ("mix" "dogma" "--format" "flycheck" source)
    :predicate
    (lambda ()
      (file-exists-p "deps/dogma"))
    :error-patterns
    (
     (info line-start (file-name) ":" line ":" column ": " (or "C" "R" "D") ": " (optional (id (one-or-more (not (any ":")))) ": ") (message) line-end)
     (warning line-start (file-name) ":" line ":" column ": " (or "W" "E" "F") ": " (optional (id (one-or-more (not (any ":")))) ": ") (message) line-end))
    :working-directory
    (lambda (_checker)
      (locate-dominating-file buffer-file-name "mix.exs"))
    :modes (elixir-mode))
```

Let me know if I should just pull this one

lunaryorn: Thanks, doesn't look too different, though.  I prefer the JSON parsing as I've more trust in structured formats and the other differences seem minor.

I think we'll just merge this PR 👍

lunaryorn:
@flycheck/elixir can you comment?

@aaronjensen Would you like to join Flycheck's elixir team?

aaronjensen:
@lunaryorn sure!

kthelgason:
This looks fine to me, but perhaps we should keep the old function as a fallback. I think I'm right in saying that this will not work if someone has installed dogma as an archive (as they used to recommend)

aaronjensen:
@kthelgason that's true, though I'd venture to guess most people installed it in both ways. 

How would you envision the fallback working? would it be an option? We couldn't just run the old function as part of the normal predicate as it'd cause the same stutter if dogma was not installed.

kthelgason:
`(or (new-pred) (old-pred))` is fine right? I'm assuming `or` is short-circuiting in elisp :)

lunaryorn:
@kthelgason `or` does short-circuit, but as @aaronjensen said it's going to run a very slow code path on every syntax check if `dogma` isn't installed.

lunaryorn:
@kthelgason I'm not familiar with Elixir at all, could you give me a short overview how the different methods of installing dogma look like?  Maybe there's an alternative approach…

kthelgason:
Of course, that's not ideal. So it's possible to install a package as an archive, which makes it available to `mix` globally, and thus it won't be placed in a projects `deps` directory. Similar to how with `npm` one can install a package with `-g`, in which case it won't show up in `node_modules`.

However, I don't believe installing packages is very common in the Elixir community, so this is probably not much of a problem in practice. I think we can safely ignore that use case, I just wanted to point it out :smiley:

lunaryorn:
@kthelgason In that case it looks as if this PR is to help more people and harm only few if any.  I think we should just go ahead, merge this pull request and look whether someone complains 😎 

So LGTM.

kthelgason:
I agree. LGTM!

Sent from my iPhone




On Thu, Sep 1, 2016 at 4:44 PM +0200, "Sebastian Wiesner" <notifications@github.com> wrote:












@kthelgason In that case it looks as if this PR is to help more people and harm only few if any.  I think we should just go ahead, merge this pull request and look whether someone complains 😎 



So LGTM.

cpitclaudel:
LGTM, though can we confirm that there won't be default-directory issues? In a case like below, will checking `b.exs` and `c.exs` still work?

```
.
├── a.exs
├── deps
│   └── dogma
├── mix.exs
├── subdir1
│   └── b.exs
└── subdir2
    └── c.exs
```

aaronjensen:
Yes that's fine, though I do wonder if there will be issues with umbrella projects. I have to look at that.

Aaron

aaronjensen:
@cpitclaudel this now handles umbrella apps. The difference between umbrella apps and the scenario you mentioned is that there are multiple `mix.exs` files in umbrella apps (imagine a `mix.exs` in `subdir1`, for example). This would cause us not to find `deps/dogma` which we could find in your scenario because the `working-directory` takes effect for the predicate check.

aaronjensen:
hold on this

aaronjensen:
Ok, this should be good to go now. Do you want me to squash or do you use the squash and merge feature?

flycheck.el
```diff
@@ -6385,7 +6385,9 @@ This will either be the directory that contains `mix.exs' or,
 if no such file is found in the directory hierarchy, the directory
 of the current file."
   (or
-   (locate-dominating-file (buffer-file-name) "mix.exs")
+   (and
+    buffer-filename
```
kthelgason: 1: shouldn't this be `buffer-file-name` and 2: what is the purpose of this check?

aaronjensen:
Looks like an intermittent failure on travis? any way to rerun?

flycheck.el
aaronjensen: 1: oops, thanks.

2: you can't pass null to `locate-dominating-file`

flycheck.el
```diff
@@ -6385,7 +6385,9 @@ This will either be the directory that contains `mix.exs' or,
 if no such file is found in the directory hierarchy, the directory
 of the current file."
   (or
-   (locate-dominating-file (buffer-file-name) "mix.exs")
+   (and
+    buffer-file-name
+    (locate-dominating-file buffer-file-name "mix.exs"))
```
kthelgason: but `buffer-file-name` is a function, no?

cpitclaudel: The function just returns the variable :)

aaronjensen:
squashed and signed

cpitclaudel:
Thanks! I'll let @lunaryorn make the final decision on squashing, but I almost feel that these should be two separate PRs: one to fix the existing use of locate-dominating-file, and one to optimize the mix check. Thoughts?

aaronjensen:
@cpitclaudel I'm good w/ that, I'll open another

aaronjensen:
@cpitclaudel split to #1068

lunaryorn:
@aaronjensen Feel free to merge :blush:  If you'd like you can add yourself to the list of contributors in `doc/community/people.rst` first :)

aaronjensen:
pinging lgtm service...

PR #1066, opened 2016-09-01
Compute error-list-msg-offset once and for all

cpitclaudel:
This function used to be generic (it worked in any tabulated-list-mode buffer), but since it's also called out of tabulated-list-mode it makes sense IMO to specialize it to Flycheck's header format and compute the result once and for all.


cpitclaudel:
(LGTM :)

lunaryorn:
:+1: LGTM

PR #1067, opened 2016-09-01
Strip trailing newlines in error messages

cpitclaudel:
(will need rebasing after #1066; also needs a test)

This prevents the checker name to pe pushed on its own line with multiline messages that have trailing newlines.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3875,12 +3875,15 @@ Return a list with the contents of the table cell."
 (defun flycheck-flush-multiline-message (msg)
   "Prepare error message MSG for display in the error list.
 
-Prepend all lines of MSG except the first with enough space to
-ensure that they line up properly once the message is displayed."
+Strip trailing blanks and prepend all lines of MSG except the
+first with enough space to ensure that they line up properly once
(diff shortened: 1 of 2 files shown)


cpitclaudel:
Test added. LGTM.

Concrete example of the issue that existed before this: this message:

```
Some error
Blah
Bla

```

was displayed as

```
    21  26 error           Some error
                           Blah
                           Blah
(some-checker)
```

instead of 

```
    21  26 error           Some error
                           Blah
                           Blah (some-checker)
```

If preferred, we could instead display it as 


```
    21  26 error           Some error (some-checker)
                           Blah
                           Blah
```

This is nice if the last line of the message is long.  But I think it's not great to mix the checker name in the middle of the message.

cpitclaudel:
I don't think we need this, actually. The only cases in which this will pop up are when the sanitize filter isn't applied. Closing.

lunaryorn:
@cpitclaudel Yup, error messages shouldn't have trailing spaces.  If they do we can consider that as a bug in the syntax checker.

PR #1068, opened 2016-09-01
Fix elixir-dogma on non-file-backed buffers

aaronjensen:
Split from #1063

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6385,7 +6385,9 @@ This will either be the directory that contains `mix.exs' or,
 if no such file is found in the directory hierarchy, the directory
 of the current file."
   (or
-   (locate-dominating-file (buffer-file-name) "mix.exs")
+   (and
+    buffer-file-name
+    (locate-dominating-file buffer-file-name "mix.exs"))
    default-directory))
 
 (defun flycheck-elixir--parse-dogma-json (output checker buffer)


flycheck.el
```diff
@@ -6385,7 +6385,9 @@ This will either be the directory that contains `mix.exs' or,
 if no such file is found in the directory hierarchy, the directory
 of the current file."
   (or
-   (locate-dominating-file (buffer-file-name) "mix.exs")
+   (and
+    buffer-file-name
+    (locate-dominating-file buffer-file-name "mix.exs"))
```
kthelgason: I'll repeat my comment from the other PR. Am I wrong in thinking that `buffer-file-name` is a function and it's result should be passed to `locate-dominating-file`?

cpitclaudel: Partly wrong :) I replied there already.

cpitclaudel:
```
buffer-file-name is a variable defined in ‘buffer.c’.

Documentation:
Name of file visited in current buffer, or nil if not visiting a file.
This should be an absolute file name.
```

cpitclaudel:
Thanks! @kthelgason, what do you think?

flycheck.el
aaronjensen: `buffer-file-name` as a function is useful if you need the file name of a buffer other than the current. For the current you can just use the variable.

kthelgason: Oh! Thanks for clarifying :)

kthelgason:
LGTM :+1: Thanks @aaronjensen

flycheck.el
lunaryorn: I'm always using the function out of some habit that I can't explain :relaxed: but it's generally totally fine to use the variable instead.

lunaryorn:
LGTM. @aaronjensen Feel free to merge :)

cpitclaudel:
Sorry for the delay! LGTM, go ahead :)

PR #903, opened 2016-03-06
Add stylelint checker

PhilippBaschke:
These changes should enable the `css-stylelint` checker in `css-mode` and `scss-mode`.

It covers 3 different scenarios:

**1) Normal errors and warnings**

``` yaml
# .stylelintrc.yaml
rules:
  string-quotes: double
  indentation: [2, { "warn": true }]
```

``` css
/* style.css */
h1 {
  color: 'red'; /* error: single quote */
    margin: 0 auto; /* warning: wrong intentation */
}
```

stylelint output (missing error level information):

```
$ stylelint style.css
2:10    Expected double quotes (string-quotes)
3:5 Expected indentation of 2 spaces (indentation)
```

That's why I used the JSON formatter and wrote my own `:error-parser`

```
$ stylelint --formatter json style.css
[{"source":"style.css","deprecations":[],"errored":true,"warnings":[{"line":2,"column":10,"rule":"string-quotes","severity":"error","text":"Expected double quotes (string-quotes)"},{"line":3,"column":5,"rule":"indentation","severity":"warning","text":"Expected indentation of 2 spaces (indentation)"}]}]
```

**2) Syntax errors**

``` css
/* style.css */
h1 {
  color: 'red; /* syntax error: Unclosed quote */
    margin: 0 auto;
}
```

```
$ stylelint --formatter json style.css
CssSyntaxError: style.css:2:10: Unclosed quote
```

**3) Configuration errors**

``` yaml
# .stylelintrc.yaml
# should be rules instead of rles
rles:
  string-quotes: double
  indentation: [2, { "warn": true }]
```

```
$ stylelint --formatter json style.css
Error: No rules found within configuration. Have you provided a "rules" property?
```

The `:error-filter` function adds a line (line number 1) to these errors to be able to show them.

**Questions**
- What is the best way to define this checker for `css-mode` and `scss-mode`?
- What is the best way to use a combination of an `:error-parser` and some form of `:error-patterns`?
- Is it necessary to show the Syntax errors and Configuration Errors or should flycheck just fail if these happen?
- Should I rather try to create a PR at stylelint that changes their text formatter to output the error level and then create a simple checker that just uses `:error-patterns` for all 3 scenarios?

**Edit:** Connects to #785

PR #1070, opened 2016-09-04
add support for sass-lint (a pure node.js scss linter)

DEADB17:
- a current limitation is that linting is performed on the file contents
  instead of the buffer and saving is required to re-lint

The functionality was not _designed_ but put together by looking at existing code and it might be missing obvious improvements.

### DIFF
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -935,6 +935,16 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
 .. supported-language:: Sass
 
+   Flycheck checks SASS with `sass/scss-sass-lint`, falling back to `sass`.
+
+   .. syntax-checker:: sass/scss-sass-lint
+
+      Check ``sass`` and ``scss`` syntax and lint with SASS-Lint_.
+
+      .. _SASS-Lint: https://github.com/sasstools/sass-lint
+
+      .. syntax-checker-config-file:: flycheck-sass-lintrc
+
    .. syntax-checker:: sass
 
       Check syntax with the `Sass <http://sass-lang.com/>`_ compiler.
@@ -987,7 +997,8 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
 .. supported-language:: SCSS
 
-   Flycheck checks SCSS with `scss-lint`, falling back to `scss`.
+   Flycheck checks SCSS with `scss-lint`, falling back to `sass/scss-sass-lint` first and
+   `scss` if neither is available.
 
    .. syntax-checker:: scss-lint
 
@@ -1000,6 +1011,14 @@ to view the docstring of the syntax checker.  Likewise, you may use
       .. _SCSS-Lint: https://github.com/brigade/scss-lint
 
       .. syntax-checker-config-file:: flycheck-scss-lintrc
+
+   .. syntax-checker:: sass/scss-sass-lint
+
+      Check ``sass`` and ``scss`` syntax and lint with SASS-Lint_.
+
+      .. _SASS-Lint: https://github.com/sasstools/sass-lint
+
+      .. syntax-checker-config-file:: flycheck-sass-lintrc
 
    .. syntax-checker:: scss
 
--- a/flycheck.el
+++ b/flycheck.el
@@ -231,11 +231,12 @@ attention to case differences."
     ruby-jruby
     rust-cargo
     rust
-    sass
     scala
     scala-scalastyle
     scheme-chicken
     scss-lint
+    sass/scss-sass-lint
+    sass
     scss
     sh-bash
     sh-posix-dash
@@ -8541,6 +8542,24 @@ See URL `http://sass-lang.com'."
             " of " (one-or-more not-newline)
             line-end))
   :modes sass-mode)
+
+(flycheck-def-config-file-var flycheck-sass-lintrc sass/scss-sass-lint ".sass-lint.yml"
+  :safe #'stringp
+  :package-version '(flycheck . "30"))
+
+(flycheck-define-checker sass/scss-sass-lint
+  "A SASS/SCSS syntax checker using sass-Lint.
+
+See URL `https://github.com/sasstools/sass-lint'."
+  :command ("sass-lint"
+            "--verbose"
+            "--no-exit"
+            "--format" "Checkstyle"
+            (config-file "--config" flycheck-sass-lintrc)
+            source)
+  :standard-input nil
+  :error-parser flycheck-parse-checkstyle
+  :modes (sass-mode scss-mode))
 
 (flycheck-define-checker scala
   "A Scala syntax checker using the Scala compiler.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,7 @@
 - New syntax checkers:
 
   - Elixir with ``dogma`` [GH-969]
+  - sass and scss with ``sass-lint`` [GH-1070]
 
 29 (Aug 28, 2016)
 =================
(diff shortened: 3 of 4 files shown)


DEADB17:
I'll try to figure out what the errors are as well as include tests and re-submit

cpitclaudel:
Hi @DEADB17,

Thanks for your contribution! It's quite OK to leave the PR open while you work on ironing out the bugs, and we'd be happy to help, too.

Regarding your comment about having to save: have you tried `source` or `source-inplace` instead of `source-original`?

DEADB17:
Thanks will reopen it then. I didn't want to add noise with a patch that was not ready for merging as it was missing tests and possibly basic documentation
Regarding the option I did try `source` (not `source-inplace`) but the problem seems to be that the `sass-lint` bin only works on files and not std-input as far as I can tell.

cpitclaudel:
> Regarding the option I did try source (not source-inplace) but the problem seems to be that the sass-lint bin only works on files and not std-input as far as I can tell.

`std-input` is independent from `source/source-inplace`; the former indicates whether the file should be fed to the linter, whereas the latter describes a command line argument; I'd be surprised if sass-lint doesn't work with source (or at worse source-inplace, but not source-original) :)

I'd suggest trying it again; let me know if you run into any issues.

cpitclaudel:
The failing test is the documentation, btw; it'll pass once you add a documentation bit to your patch (it doesn't need to be big, btw; just a reference to the checker) :) A reference to the checker in the change log would be welcome as well.

lunaryorn:
@DEADB17 And please feel free to add yourself to the contributor list in `doc/community/people.rst` :)

DEADB17:
@cpitclaudel thanks for the suggestion regarding `source-original`, changing it to `source` fixed the issue.
Let me know if there is anything else that should be improved.
Thanks.

flycheck.el
```diff
@@ -232,6 +232,7 @@ attention to case differences."
     rust-cargo
     rust
     sass
+    sass-lint
```
lunaryorn: Shouldn't we use `sass` before `sass-lint`?  `sass` is a very basic checker that only catches some syntax errors; shouldn't we generally prefer `sass-lint` if it's available?

DEADB17: Regarding priority, since `sass-lint` can be used for both `sass` and `scss` modes but it is not as developed as `scss-lint` yet, I think the priority should be:

1. scss-lint
2. sass-lint
3. sass
4. scss

If there are no objections I can make the change later today.

lunaryorn:
@DEADB17 I wonder whether we should probably renaming the syntax checkers following our convention for C and C++, given that both `sass-lint` and `scss-lint` check both SASS and SCSS. 

How about `sass/scss-sass-lint` and `sass/scss-scss-lint`?  Or are these names as awkward as I find them? :wink:  Any better suggestions?

cpitclaudel:
I'm OK with renaming. Some part of me wants to call them `s{c|a}ss-scss-lint` and `s{c|a}ss-sass-lint` :)  But jokes aside, what about `s*ss-scss-lint` and `s*ss-sass-lint`?

lunaryorn:
@cpitclaudel Oh dear, I guess Emacs Lisp would even handle that name wouldn't it? 

I'm not too convinced of `s*ss` either… the asterisk looks kind of strange, and what if someone publishes `suss` tomorrow 😜

DEADB17:
I think the value in making a decision will be: 1. validating existing conventions and 2. defining how to handle this issue in the future. Having a convention brings predictability for both users and developers.
An example where a convention will be relevant is the *stylelint checker* (#903), which can potentially be used for any css-like syntax, like `css/scss/less-stylelint`.

Quickly skimming through the code it looks like there is only one instance of a linter used for multiple modes: `c/c++`.
The existing convention would then be:
`<mode> "/" <mode> ... "-" <linter>`
Which I think will provably work better in future cases than the `s*ss-sass-lint` alternative which is clever but specific to this linter, thus harder to use as a convention.
The one improvement I can think of  is to use a different separator for the modes and the linter:
`<mode> "/" <mode> ... ":" <linter>`, for example: `sass/scss:sass-lint`. Which would make it easier to parse if it ever needs to be automated... but honestly I'd wait until there is an actual use case.

My vote would be to use `sass/scss-sass-lint`.

cpitclaudel:
> My vote would be to use sass/scss-sass-lint.

:+1:

lunaryorn:
@DEADB17 Would you mind to implement the rename in this PR?

DEADB17:
Done.

doc/languages.rst
```diff
@@ -935,6 +935,16 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
 .. supported-language:: Sass
 
+   Flycheck checks SASS with `sass-lint`, falling back to `sass`.
```
lunaryorn: `sass-lint` -> `sass/scss-sass-lint`.

doc/languages.rst
```diff
@@ -987,7 +997,8 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
 .. supported-language:: SCSS
 
-   Flycheck checks SCSS with `scss-lint`, falling back to `scss`.
+   Flycheck checks SCSS with `scss-lint`, falling back to `sass-lint` first and
```
lunaryorn: `sass-lint` -> `sass/scss-sass-lint`.

flycheck.el
```diff
@@ -8542,6 +8543,24 @@ See URL `http://sass-lang.com'."
             line-end))
   :modes sass-mode)
 
+(flycheck-def-config-file-var flycheck-sass-lintrc sass/scss-sass-lint ".sass-lint.yml"
+  :safe #'stringp
+  :package-version '(flycheck . "0.30"))
+
+(flycheck-define-checker sass/scss-sass-lint
+  "A SCSS syntax checker using sass-Lint.
```
lunaryorn: SASS/SCSS?

DEADB17:
All set

lunaryorn:
@DEADB17 LGTM now.

flycheck.el
```diff
@@ -231,11 +231,12 @@ attention to case differences."
     ruby-jruby
     rust-cargo
     rust
-    sass
     scala
     scala-scalastyle
     scheme-chicken
     scss-lint
+    sass/scss-sass-lint
+    sass
```
cpitclaudel: I think we should keep these sorted alphabetically :)

flycheck.el
```diff
@@ -8542,6 +8543,24 @@ See URL `http://sass-lang.com'."
             line-end))
   :modes sass-mode)
 
+(flycheck-def-config-file-var flycheck-sass-lintrc sass/scss-sass-lint ".sass-lint.yml"
+  :safe #'stringp
+  :package-version '(flycheck . "0.30"))
```
cpitclaudel: The version number is 30, not 0.30

cpitclaudel:
LGTM, thanks! With a few tiny changes suggested as inline comments. Can you squash your commits?

flycheck.el
lunaryorn: > Am 11.09.2016 um 15:49 schrieb Clément Pit--Claudel <notifications@github.com>:
> 
> In flycheck.el:
> 
> >      scala
> >      scala-scalastyle
> >      scheme-chicken
> >      scss-lint
> > +    sass/scss-sass-lint
> > +    sass
> I think we should keep these sorted alphabetically :)
> 

Not possible here because scss-lint must come first.

DEADB17:
I updated the version and squashed.
Thanks for all the suggestions and maintaining this project!

lunaryorn:
@DEADB17 Merged. Many thanks for the patch, and for patiently enduring our nit-picky reviews :blush:

DEADB17:
Any time :thumbsup:

PR #1071, opened 2016-09-05
Require :mode for all syntax checkers

lunaryorn:
Follow up of #961: We've got no built-in syntax checkers without `:mode` anyway, thus making it mandatory simplifies our code.  

Should we need a more generic system I think we should rather take the route of supporting derived modes. Even a hypothetical ispell checker could probably not support all modes: There are fundamental differences between programming languages and text and likely you'll need to syntax checkers for these anyway.

This pull request is still WIP (missing some tests and doc updates), but I'm opening it anyway for discussion.

/cc @cpitclaudel

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1440,9 +1440,8 @@ have any `:modes', but a `:predicate' that returns non-nil for
 the current buffer."
   (let (checkers)
     (dolist (checker flycheck-checkers)
-      (let ((modes (flycheck-checker-get checker 'modes)))
-        (when (memq major-mode modes)
-          (push checker checkers))))
+      (when (flycheck-checker-supports-major-mode-p checker major-mode)
+        (push checker checkers)))
     (nreverse checkers)))
 
 
@@ -1743,16 +1742,25 @@ A valid checker is a symbol defined as syntax checker with
        (= (or (get checker 'flycheck-generic-checker-version) 0)
           flycheck-generic-checker-version)))
 
+(defun flycheck-checker-supports-major-mode-p (checker mode)
+  "Whether CHECKER supports the given major MODE.
+
+CHECKER is a syntax checker symbol and MODE a major mode symbol.
+Look at the `modes' property of CHECKER to determine whether
+CHECKER supports buffers in the given major MODE.
+
+Return non-nil if CHECKER supports MODE and nil otherwise."
+  (memq mode (flycheck-checker-get checker 'modes)))
+
 (defun flycheck-may-use-checker (checker)
   "Whether a generic CHECKER may be used.
 
 Return non-nil if CHECKER may be used for the current buffer, and
 nil otherwise."
-  (let ((modes (flycheck-checker-get checker 'modes))
-        (predicate (flycheck-checker-get checker 'predicate)))
+  (let ((predicate (flycheck-checker-get checker 'predicate)))
     (and (flycheck-valid-checker-p checker)
          (not (flycheck-disabled-checker-p checker))
-         (memq major-mode modes)
+         (flycheck-checker-supports-major-mode-p checker major-mode)
          (funcall predicate))))
 
 (defun flycheck-may-use-next-checker (next-checker)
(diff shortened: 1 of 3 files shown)


cpitclaudel:
I ran a quick experiment and downloaded all packages whose name contained "flycheck".  All of them had a `:modes` property, so I think this is safe :)

Some notes:

* Many extensions have a single mode, but still use the list form: `:modes (a-mode)`; other's don't use a list, though: `:modes a-mode`.

* There was at least one wrong one:

```
(flycheck-define-checker cstyle
  "A checker using cstyle.

See `https://github.com/alexmurray/cstyle/'."
  :command ("cstyle"
            (config-file "--config" flycheck-cstyle-config)
            source)
  :error-patterns ((info line-start (file-name) ":" line ":" column ":"
                         (message (minimal-match (one-or-more anything)))
                         line-end))
  :modes c-mode c++-mode)
```

(reported at https://github.com/alexmurray/flycheck-cstyle/issues/2)

flycheck.el
```diff
@@ -1598,13 +1593,10 @@ are mandatory.
      If given this syntax checker is only used in buffers whose
```
cpitclaudel: I think this should change, since the mode should always be given (I understand that the PR is still in flux :)

lunaryorn: Yup, thanks.

lunaryorn:
@cpitclaudel Thanks a lot; that's good to know.  I think the change is safe then.

Funny that folks have so many different styles :)

lunaryorn:
@cpitclaudel Finished and ready for review :blush:  For the record, LGTM.

cpitclaudel:
LGTM. IIRC buttercup has a feature to intercept function calls that could be used instead of cl-letf, but I don't think it would make a difference here :)

I like this new architecture. If a checkers want more mores, we could always introduce a different `:modes` spec, like `(:derived-from <mode>)`, in the future.

lunaryorn:
>  If a checkers want more mores, we could always introduce a different :modes spec, like (:derived-from <mode>), in the future.

That's my idea. 

Merged.

radupopescu:
Unfortunately, this merge broke the `flycheck-ycmd` checker.

lunaryorn:
@radupopescu I'm sorry but we [reserve the right to change API at any time](http://www.flycheck.org/en/latest/contributor/maintaining.html#versioning-and-releases).

radupopescu:
I was just reporting this here, too, since this checker wasn't mentioned in this thread (it's part of the `emacs-ycmd` repo, it doesn't have it's own `flycheck-*****` repo).

Anyway, I think the maintainers of `emacs-ycmd` have made a fix today.

lunaryorn:
Generally, if you're maintaining extensions to Flycheck outside of our organisation we expect you to follow our development process to be notified about breaking changes. 

However we offer to move Flycheck extensions into the Flycheck organisation—for extensions and checkers in our org we notify maintainers of breaking changes and make necessary changes to restore compatibility.

radupopescu:
Makes sense.

Are extensions part of the Flycheck organization allowed to require an external package/program to function?

lunaryorn:
@radupopescu For sure… how would Flycheck work without external programs? :wink:

radupopescu:
@lunaryorn Yes, I guess it wasn't the best question. The flycheck package would be huge if it contained the logic for checking all these languages.

Thanks!

lunaryorn:
@radupopescu Sorry, I didn't mean to offend you—it was a stupid remark of mine 😞  I hope you don't bear a grudge against me 😌 

If you've got an extension to move into Flycheck please feel free to ping us, either with an issue or on our gitter channel and we'll help you with moving the extension into our Org :blush:

radupopescu:
@lunaryorn No worries, no offense taken!

I don't personally maintain any extension, but I was trying to understand whether it would be technically feasable to include something like `flycheck-ycmd` into the `flycheck` package.

Thanks for all the information!

Cheers

PR #1072, opened 2016-09-05
Chain syntax checkers by stages

swsnr:
Improves over and superseds #961.  ~~Currently blocked by #1071~~
# TODO:
- [x] Add stages
- [x] Take conflicts into account
- [x] Declare stages for all checkers
- [x] Document stages in the list of languages
- [x] Run stages
- [x] Remove manual syntax checker selection
- [x] Remove `:next-checkers`
- [ ] Document stages in the user guide
- [ ] Stop stages at `error`s
# TODO after merging
- [ ] Update official Flycheck extensions that use `:next-checkers`.
- [ ] Update this comment with a migration guide. /cc @aaronjensen 
# Design

Instead of :next-checkers we now have three properties to control when a checker is selected:
- `:stage:` The kind of syntax checker, e.g. syntax, type, lint, style, test, which serves to order checker execution.
- `:conflicts-with:` A list of syntax checkers this syntax checker conflicts with, i.e. may not be used together with.

Unlike #961 there's no `:maximum-level` anymore. Instead the `error` level will become a hard-coded boundary for chains: If there's any `error` or more severe the chain stops otherwise it continues.

Pros & Cons are mostly the same as in #961 but the implementation is simpler.


aaronjensen:
Instead of `:conflicts-with` would it be possible to have a function that could be easily used in a predicate to tell if a checker would run before you? It's less explicit than conflicts-with but it's simpler

aaronjensen:
One other suggestion... with `stage`, what if each stage symbol correlated to a number... 10, 20, 30, 40, etc. Then if someone wanted to slot in between a stage they could just use a number instead of a symbol. I'm not sure if we want/need the extra flexibility, I just wanted to throw the idea out there.

lunaryorn:
@aaronjensen We iterated over numeric stages ("kinds" back then) in the other issues. I don't think we need those—should someone need another stage they can just file an issue—and I doubt that we'll have need for more than the current stages.

I'm not in favour of using functions for these, nor would I want to let predicates inspect the previously run syntax checkers.  I don't think that these would be simpler; on the contrary I believe functions would increase complexity.  `:conflicts-with` is as simple as it can get but a function is code that needs to be read and understood which gets the worse the more clever the conditions get.

Functions hurt static discoverability and documentation, too.  With functions we couldn't reasonably show conflicts in `C-c ! v` for instance, as we've no chance to figure out what checkers a syntax checker would conflict with without running the function—and even then the result can't be trusted as function could do crazy things like only conflict with checker A if the seconds of the current time are even :wink:

Likewise if we let predicates inspect previously run syntax checkers to make a decision we'd have no chance to compute the entire chain in advance which makes a reasonable `C-c ! v` harder and is less efficient as we'd have to recompute the next syntax at each step throughout the entire chain of checkers.

aaronjensen:
@lunaryorn I totally understand, thanks for the explanations.

lunaryorn:
@flycheck/contributors After a long hiatus I've made quite a bit of progress on last weekend, and I plan to finish the implementation this weekend.

The functionality is still there, except aborting chains if any `error` level error appears.  It'd be great if  some of you could checkout the branch and test it with different languages.

lunaryorn:
If you find any issues in the branch, please leave a comment and if possible also add a note to the TODO column of the [corresponding project](https://github.com/flycheck/flycheck/projects/1).

lunaryorn:
@flycheck/contributors It'd also be great if you could give the changeset a pull request review :blush:

doc/glossary.rst
```diff
@@ -39,6 +39,18 @@ reference`_, but we reproduce them here for convenience.
...
+      The stage denotes when a syntax checker will run.  Flycheck has a couple
+      of stages, see `flycheck-checker-stages`.  Each syntax checker has an
+      associated stage.  For instance compilers will typically sit in the
+      ``syntax`` stage, whereas semantic linters belong into the ``lint`` stage.
+
+   syntax checker chain
+      The exact sequence of syntax checkers that Flycheck runs for a single
+      syntax check.  Flycheck computes chains from syntax checkers that support
+      the current buffer, ordering them by their stages and throwing out
+      conflicting syntax checkers.
```
aaronjensen: It might be helpful to know what constitutes a conflict. Is it two at the same stage? Or something else?

flycheck.el
```diff
@@ -1590,6 +1548,15 @@ are mandatory.
      your syntax checker, because it will help users to spot
      potential setup problems.
 
+`:kind KIND'
```
aaronjensen: Should this be `:stage`? The code below uses `:stage`

doc/glossary.rst
```diff
@@ -39,6 +39,18 @@ reference`_, but we reproduce them here for convenience.
...
+
```
aaronjensen: Will there be any sort of upgrade guide both for external checker maintainers and individuals?

aaronjensen (changes requested):

doc/glossary.rst
lunaryorn: There'll be a changelog entry, and Ill update the corresponding issue with a  migration guide.

doc/glossary.rst
lunaryorn: It's across all stages. I tried to clarify that in the glossary, but the main documentation in user guide is still missing anyway, and that's where I'll try to describe the new system in detail.

aaronjensen (approved):

stsquad:
I ran into this problem with clangcheck and checkpatch where checkpatch would run first, report an error and then prevent future checkers from running. I've tweaked both (see pending patches) and the :stage keyword works fine with this pullreq branch. I'm awaiting eventual merging of this feature before I push the upstream branches.

fmdkdd:
@stsquad Unfortunately this is unlikely to merge soon due to #1177.  But if you want to pick it up, you are more than welcome :)

stsquad:
@fmdkdd I think taking over maintainership would be slightly overkill ;-) I shall wait for the new maintainers to get comfortable and keep subscribed to this should it move forward.

In the meantime how does the current tree deal with chaining? I couldn't find a clear statement in the documentation and I'm guessing a straight "conflicts with" doesn't really cut it.

fmdkdd:
@stsquad I meant pick up this current PR, mind you ;)

I'm not too familiar with this area, but glancing over the code, it seems the intent is that a `next-checker` will run only if the current one does not have errors higher than a level.  E.g., `c/c++-clang` has:
```
  :next-checkers ((warning . c/c++-cppcheck)))
```
which means that `cppcheck` will run after `clang` only if `clang` does not report warning or errors.

nickmccurdy:
Is this blocked until there's a new maintainer?

fmdkdd:
@nickmccurdy Until there's a willing soul to pick up this PR at least.  

I *might* be able to work on it next week, but no promises.

fmdkdd:
Ok, I took some time to go through it.

Like the first post implies, there is not much that remains to be done.  I've rebased on master and stopped at the first checker which returns errors (now that I think of it, the TODO stated to prevent the next *stage* from running, not the next checker, so I might have to revisit that one).

Other than that, the new system looks less flexible than what we had before, which, from looking at discussions, I believe was the intention.  The benefit is that the new system is less daunting when defining checkers.

I want to merge #1217 before pushing forward with this PR, since I want to make sure we don't break everything.  After that, it would probably be a good idea to double-check the new `:stage` and `:conflicts-with` properties of each checker.

Simplify:
Personally, I see a lot of problems with preventing next stage from running in some modes. Imagine following situation:

`main.scss`:
``` scss
@import 'variables.scss';
@import 'header';
```

`varibales.scss`:
``` scss
$header-background-color: #ccc;
```

`header.scss`:
``` scss
#header {
  background-color: $header-background-color;
}
```

When you compile `main.scss` into `main.css` everything works fine. When you are editing `header.scss` first linter will add error about undeclared `$header-background-color` variable and prevent other linters to run that can give out useful information, even if you can safely ignore that error. Solution is to import `variables.scss` in `header.scss` but that must happen in all files and may break a lot of current setups (a lot of Rails applications). Also, big problem here is going to be `bootstrap.scss`, importing that everywhere may result in slow compilation to final `css` file.

I expect similar things with JavaScript. Using `jQuery` from some CDN will report errors about undeclared `$` in your files. This one is easy to solve by defining `globals` in `eslint` config file or on top of each `js` file.

I think that there should be a setting (per major mode) to continue with next stage even if there are errors.

Simplify:
Also, before merging this we need to release new version to melpa stable, that will give people some time to change their external flycheck checkers. @cpitclaudel do you have plans to release version 31 soon? I can think of at least one external checker that can benefit from `flycheck-parse-json`.

fmdkdd:
@Simplify Thank you, that's the kind of feedback that is very helpful, since I'm flying mostly notes and past discussions on this PR.

This PR is a big change to how checker selection works, since instead of one checker (which may have `:next-checker`), you always have a checker chain that's automatically setup from valid and non-conflicting checkers in `flycheck-checkers`.  

For example for`scss`, we did not have any `:next-checker`.  So you could run only one of `scss`, `scss-lint`, `scss-stylelint` or `scss/sass-sass-lint` in one buffer.

With this PR, you could potentially run all of them in sequence.  But, the way they are currently setup, they are all in the `syntax` stage, and conflict with each other.  So in effect, only one of them can be run.  No change on that front.

In your example, you speak of multiple linters running on a SCSS file.  What specific checkers did you have in mind?

In any case, here are the possibilities I see for deciding when to stop picking up the next checker in a chain:

1. Stop at the first checker that returns an error.  
2. Stop at the last checker of the first stage which had a checker return an error.   So if a syntax-stage checker returns at least one error, all syntax-stage checkers in the chain may run, but no checkers of the later stages.
3. Don't stop; run all the checkers in the chain.

We can also, as you suggest, have it be customizable: have a toggle between behavior 2 and 3 for instance.

The only checker chain I use is for emacs-lisp, and there I think it makes sense to still run the lints even if the syntax stage failed (so, behavior 3).  Ultimately, it depends on how the tools behave and on user preference; if there are syntax errors, the linter might choke on the input and not provide any useful messages.  Or, it may be resilient and provide lints, but you might prefer to see them later. 

So I think this can go in a lot of directions.  The simplest approach that was adopted in this PR is to provide sane defaults, and then see if any concrete needs arise to make it more flexible.

I agree on keeping this PR for the next release; this will give us more time to collect feedback.

stsquad:
Would another option be to stop at the last checker for any given stage but still process all stages? Personally I'd like to know of both style and syntax errors as I go along so I don't have to do multiple passes over my code.

fmdkdd:
@stsquad It could be.  Again, it's all in the details.  If your style linter is unable to provide you with useful feedback because it cannot parse your document, then there's no point running it.  Do you have a concrete example in mind?

stsquad:
@fmdkdd well in the checkpatch case the style linter is a perl script driven by regex's. It's not overly concerned if the code is valid and compilable C but it will point out spacing issues and bracing styles.

fmdkdd:
@stsquad Then that's one point in the "let users choose whether a stage should always run" column ;)

Simplify:
> In your example, you speak of multiple linters running on a SCSS file. What specific checkers did you have in mind?

That was just example, but for example i have [colorguard](https://github.com/Simplify/flycheck-css-colorguard) that i may want to run even if there is a "error" in css/scss file.

Also, i like to use `jscs` for checking JSDoc in javascript files and `eslint` for code. So in the future we may see requests to move checkers from one stage to another. This is also bad example, because i can install `eslint` extension that improves JSDoc parsing.

> In any case, here are the possibilities I see for deciding when to stop picking up the next checker in a chain:
> 
>  1. Stop at the first checker that returns an error.
>  2. Stop at the last checker of the first stage which had a checker return an error. So if a syntax-stage checker returns at least one error, all syntax-stage checkers in the chain may run, but no checkers of the later stages.
>  3. Don't stop; run all the checkers in the chain.
> 
> We can also, as you suggest, have it be customizable: have a toggle between behavior 2 and 3 for instance.

Yeah, this is definitely something that i had in mind and should work per major mode. For example in TypeScript i use [tide](https://github.com/ananthakumaran/tide) that provides checker that uses `tsserver`. If that checker gives error, there is no point to use `tslint`.
Also, `tide` is very interesting example because you can also use it with JavaScript (any valid JavaScript is also valid TypeScript). It that case `tide` should come before any other JS checker. My understanding is that something like this Microsoft uses in latest VS, JS is parsed first with TypeScript "compiler" to detect any parse errors. So we should probably left "compiler" stage for JS checker empty for `tide` and put other JS checkers in lower priority stages.

I'm expecting a lot of requests to change stages for some checkers, especially GO, but the best is to wait and see what to do about it after this gets released.

fmdkdd:
@Simplify I see.  Based on these examples, it seems reasonable to expect users to want to tune the order of checkers.  However, the previous system did not offer this flexibility either, so we are not losing anything, at least.

We'll indeed see how much flexibility is warranted once this is usable.  Probably after #1090.

fmdkdd:
@jazen I've been meaning to go back to that one, but I need to find the time.

wcummings:
Whats the state of this? Is there anything I can do to help get this done?

fmdkdd:
@wcummings The current state is that other features and maintenance tasks have taken priority.  I wanted integration tests to run on CI before getting back to this because such a large change needed to be tested thoroughly.  That's done with #1462.

I also am a bit torn on the design choices that went into this PR.  The intent was to make the system simpler, but we somehow lose customizability in the process.  I understand the merit of giving a  sensible configuration out-of-box, but I also feel that we need to give the users the ability to define their own checker chain.  The `stages` distinction is not that clear-cut, and `conflicts-with` is debatable.  What if I /do/ want to run two syntax checkers, because each provides diagnostic the other doesn't?

Also, there are many checkers on MELPA and whatnot, and these would need to be updated to the new system.  This is not to be taken lightly.

So, from my side, I would need to review the discussions that led to this design before going forward with the implementation.

If you want to help, the best would be to provide convincing arguments for or against this PR.  Why do you want this to go forward, for instance?

lunaryorn:
Hello, I’d like to close this PR.  It’s lingering in my pull requests overview with no change since I left this project, and I’d like to clean up.  If you’re still interested in this change, I’d appreciate if someone opened a new PR instead so that this one does no longer clutter my pull requests view.

Do you mind if I close it?

fmdkdd:
@lunaryorn Sure, we can always open a new one using the same branch.

lunaryorn:
Thanks 🙏

PR #1075, opened 2016-09-09
Support debian

lunaryorn:
I'd like to update the documentation to state that we officially support the `elpa-flycheck` Debian package by @spwhitton.  

While I'm personally still in favour of Emacs' built-in package manager I feel that given @spwhitton's [awesome work on the tests](https://ci.debian.net/packages/f/flycheck/) ❤️ we can absolutely recommend the package.  In fact as far as the integration with syntax checkers is concerned it's probably even more stable than our own MELPA packages :blush:

Plus, since Debian is so kind as to run our tests for us I feel we really should do them the favour to support their package, i.e. not make any packaging changes without @spwhitton's approval and make sure that new releases seamlessly propagate to Debian. :relaxed:

I'm also adding a packaging team in this PR which lists responsibilities for Flycheck packages.  It's main purpose is so that I can just do `@flycheck/packaging` in issues that affect packaging to bring all relevant people in at once.

I'd also like to take the opportunity to say my heartfelt thanks to @spwhitton! Thanks for your labour on our all-to-flaky integration tests, and for providing us with a CI to run them.  Many many thanks, it's a pleasure to work with you! ❤️

### DIFF
--- a/doc/community/people.rst
+++ b/doc/community/people.rst
@@ -109,6 +109,18 @@ TypeScript
 
 Mention with ``@flycheck/typescript``.
 
+Packagers
+=========
+
+We would like to thank all people who package Flycheck on behalf of
+distributions and support our development efforts with their feedback, their
+patches and their testing:
+
+* Sean Whitton (:gh:`spwhitton`) and the `Debian Emacs addon team`_ (Debian
+  packages)
+
+.. _Debian Emacs addon team: https://pkg-emacsen.alioth.debian.org/
+
 Acknowledgements
 ================
 
--- a/doc/user/installation.rst
+++ b/doc/user/installation.rst
@@ -92,24 +92,35 @@ Add the following form to your init file to setup Flycheck with `use-package`_:
 Then press :kbd:`C-M-x` with point somewhere in this form to install and enable
 Flycheck for the current Emacs session.
 
-.. _flycheck-alternative-installation-methods:
+.. _flycheck-distribution-packages:
 
-Alternative installation methods
-================================
+Distribution packages
+---------------------
 
-Some users prefer to install Flycheck via other methods such as el-get, Git
-submodules, etc.
+Alternatively some distributions provide binary packages of Flycheck.  We
+officially support the following distributions:
 
-We do **not** support any of these methods, and advise against any alternative
-installation method.  We do not consider it a bug if Flycheck works when
-installed as above but not with a different installation method.
+* Debian 9 and newer: ``apt-get install elpa-flycheck flycheck-doc`` (the latter
+  for our manual).  The `Debian Emacs addon team`_ provides these packages.
+
+.. _Debian Emacs addon team: https://pkg-emacsen.alioth.debian.org/
+
+.. _flycheck-legacy-installation-methods:
+
+Legacy installation methods
+===========================
+
+Some users prefer to install Flycheck with legacy methods such as el-get, Git
+submodules, etc that were common before Emacs included a package manager.  There
+are also many 3rd party packages provided by various package managers.  We do
+neither support nor endorse any of these:
 
 .. warning::
 
    If you install Flycheck in any way other than :ref:`our official packages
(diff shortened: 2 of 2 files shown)


lunaryorn:
For the record, LGTM.

cpitclaudel:
LGTM too :) This is a very nice step forward.

lunaryorn:
@cpitclaudel Thanks.  I'd like to hear from @spwhitton whether he's ok with being added to a new Flycheck team and publicly mentioned on the `people` page before I merge, so I'll wait for his comment before merging.

spwhitton:
I really appreciate this commitment to make things easy for Debian packaging -- thank you.  I am happy to be listed as a packaging team member, but I have three commnts on the contents of this pull request.

1. The packaging team list might be taken to imply that I package flycheck on behalf of flycheck (rather than on behalf of Debian), and to imply the flycheck project (rather than Debian) has control over the contents of the flycheck package.  That doesn't bother me, because regardless I'm still the contact point, but it might bother others.  Not sure how it would be best to reword.

2. I don't think the installation instructions should refer people to Debian unstable.  As I'm sure you know, that's meant for people trying to test the next release of Debian.  Unfortunately, there is not yet any Debian stable release containing flycheck (although we are working on a backport atm).  So I suggest you write something like this "Users of Debian 9 or later or Ubuntu 16.04 or later can `apt-get install elpa-flycheck`."

3. Finally, you could add a link to the Debian Emacs addon team wiki for people who might want to get involved in maintaing our `elpa-*` namespace: https://pkg-emacsen.alioth.debian.org/  The team is quite small right now.

Thanks again.

lunaryorn:
@spwhitton 

Ad 1) I'll try to clarify the relation between us and Debian, and ping you when I've come up with something better :blush:

Ad 2) Will do.

Ad 3) I'll add a link, to the people page as well on the installation page.

spwhitton:
Great.  It would also be nice to add an `apt-get install flycheck-doc`
suggestion somewhere, since you did all that work to make the HTML docs
work offline.

-- 
Sean Whitton

lunaryorn:
@spwhitton Ah, thanks for making me aware.  I didn't notice that the docs are packaged separately.  I'll add that as well.

lunaryorn:
@spwhitton Next try :wink: Would you take another look?

spwhitton:
s/later/latter/ :)

lunaryorn:
@spwhitton I've fixed the typo, but I'm not sure about Ubuntu.  Is there a guarantee that packages automatically migrate?  Or can Ubuntu mess with the packages any time?

spwhitton:
The only only reason it wouldn't migrate is if some of flycheck's dependencies hadn't yet migrated.  Or they could replace it with a patched version which I think would prevent new versions migrating. Quite unlikely, though.  Tbh I'm not sure what it is you're worried about.  Could you describe a situation where including Ubuntu install instructions would create problems for someone?

lunaryorn:
@spwhitton I had a couple of bad experiences with downstream packaging.  I do not trust downstream distributions—Debian included—to deliver good packages anymore. 

I trust you, that's why I'm willing to support Debian, but if there's a chance—no matter how small—that the Ubuntu package deviates then I'm not willing to extend that trust to the Ubuntu package—all the less as Ubuntu has had their share of bad packaging decisions in the past (owncloud anyone?).

Now, I don't think that other packages are likely to create problems for anyone but I don't want to be left in the cold either when a broken package makes users hit our tracker and the downstream packager refuses to even acknowledge the problem.  Been there already and I don't need it again.

I'm sorry but unless I get word and promise that the automated migration doesn't stop—or at least that we're absolutely notified of it—I'm not willing to hint that we supported Ubuntu, even if the hint is just a word in the installation instructions.

spwhitton:
Fair enough, and sorry to hear about those experiences.

LGTM.

lunaryorn:
Thanks, merged.

PR #1079, opened 2016-09-09
Add configuration entry for cargo's rustc command.

c-nixon:
This new configuration entry allows arguments to be passed to cargo's
rustc subcommand, mirroring the preexisting configuration entry
for rustc itself.

This enables the user to specify cargo related flags like --features,
number of worker threads with much be passed to cargo rust before the --
which delineates the rustc subcommand's args from rustc proper's

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,11 @@
 
   - Elixir with ``dogma`` [GH-969]
 
+- New features:
+
+  - Add ``flycheck-cargo-rustc-args`` to pass multiple arguments to cargo rustc
+    subcommand [GH-1079]
+
 29 (Aug 28, 2016)
 =================
 
--- a/flycheck.el
+++ b/flycheck.el
@@ -8380,6 +8380,9 @@ See URL `http://jruby.org/'."
   :modes (enh-ruby-mode ruby-mode)
   :next-checkers ((warning . ruby-rubylint)))
 
+(flycheck-def-args-var flycheck-cargo-rustc-args (rust-cargo)
+  :package-version '(flycheck . "30"))
+
 (flycheck-def-args-var flycheck-rust-args (rust-cargo rust)
   :package-version '(flycheck . "0.24"))
 
@@ -8454,6 +8457,7 @@ rustc command.  See URL `https://www.rust-lang.org'."
                    ((string= flycheck-rust-crate-type "lib") "--lib")
                    (flycheck-rust-binary-name
                     (list "--bin" flycheck-rust-binary-name))))
+            (eval flycheck-cargo-rustc-args)
             "--" "-Z" "no-trans"
             ;; Passing the "unstable-options" flag may raise an error in the
             ;; future.  For the moment, we need it to access JSON output in all
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -903,7 +903,11 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       .. defcustom:: flycheck-rust-args
 
-         A list of additional arguments.
+         A list of additional arguments that are passed to rustc.
+
+      .. defcustom:: flycheck-cargo-rustc-args
+
+         A list of additional arguments passed to the cargo rustc subcommand
 
       .. defcustom:: flycheck-rust-check-tests
 


flycheck.el
```diff
@@ -8380,6 +8380,9 @@ See URL `http://jruby.org/'."
   :modes (enh-ruby-mode ruby-mode)
   :next-checkers ((warning . ruby-rubylint)))
 
+(flycheck-def-args-var flycheck-cargo-rustc-args (rust-cargo)
+  :package-version '(flycheck . "0.24"))
```
fmdkdd: This is supposed to indicate the flycheck version number that *introduced* this variable.  In this case, this should be "30" rather than "0.24".

fmdkdd:
Thank you for your contribution.

Out of curiosity, how do you pass this variable to flycheck in your use case?  A dir-locals file?  A file-local variable?  In your dotfile for every rust project?  Can you give an example?  Keep in mind I have never used the `flycheck-rust-args` variable, so I'm asking in order to have a better picture of our users's needs.
 
In any case, I'm in favor to merge.  Before that however, could you please make the following changes:

1. Add a line documenting this variable in `doc/languages.rst`.  Look for the documentation of `flycheck-rust-args`.  I think this is what blocks the CI; you can run the tests locally before pushing, see the [Contributor's guide](http://www.flycheck.org/en/latest/contributor/contributing.html#the-build-system).
2. Change the `package-version` argument as per my inline comment
3. Add a line in `CHANGES.rst` under the `New features` heading to signal the new variable

flycheck.el
```diff
@@ -8380,6 +8380,9 @@ See URL `http://jruby.org/'."
   :modes (enh-ruby-mode ruby-mode)
   :next-checkers ((warning . ruby-rubylint)))
 
+(flycheck-def-args-var flycheck-cargo-rustc-args (rust-cargo)
+  :package-version '(flycheck . "0.30"))
```
fmdkdd: That was quick!  But it's just `30` without the `0.`.  We have dropped the decimal since version 26.

c-nixon:
Thanks for the quick review!

I use a combination currently. By default I have dir-locals files with --features clippy and flycheck-rust-args with -Z extra-plugins=clippy in my own project directories (to avoid polluting my .rs files with #![feature(clippy)]) and have jobs limited to 2 via customize on my laptop because it's dieing and doesn't like having all 4 cores pegged for any length of time, especially on battery.

I'll fix up the numbering, add the docs and changelog once i'm back at a machine with a keyboard!

c-nixon:
I believe that's everything addressed

fmdkdd:
Great!  Good call on enhancing the documentation to `flycheck-rust-args` as well.

LGTM.

I'll leave the rest to the @flycheck/maintainers.

CHANGES.rst
```diff
@@ -8,6 +8,11 @@
     own ``.luacheckrc`` detection. Therefore ``flycheck-luacheckrc`` is
     no longer used [GH-1057]
 
+- New features:
```
lunaryorn: Thanks for updating the changelog, @c-nixon.  However, please move the "New features" item *below* the "new syntax checkers" item.  We think that new syntax checkers are more important as the extend the languages Flycheck supports.

lunaryorn:
@c-nixon I'm sorry if I'm asking a stupid question but why don't you use `flycheck-rust-args`?  As far as I can see it's already used in `rust` *and* `rust-cargo`…

fmdkdd:
@lunaryorn The `cargo rustc` command accepts two different sets of options, one set is intended for `cargo`, and the other is intended for `rustc`.  The second set must appear after the `--` separator:

```
cargo rustc [options] [--] [<opts>...]
```

So this PR adds the ability to extend the first set of option, which is specific to the `rust-cargo` checker.

lunaryorn:
Oh dear, that was a really stupid question of mine.  Thanks for enlightening me :blush:

LGTM.

lunaryorn:
@fmdkdd While we're waiting for @cpitclaudel's review I'd like to thank you for the thorough review of this pull request.  Many thanks :+1:  

Would you like to join us as a maintainer, to be able to formally approve pull requests?

@c-nixon Thanks to you, of course, for your contribution, and for going through the review!

cpitclaudel:
LGTM; thanks @c-nixon for this contribution and @fmdkdd for your review!

fmdkdd:
@c-nixon Well done! :clap:

@lunaryorn I'm humbled by the offer, but for the moment I'm afraid I cannot commit to review every PR.  There's a lot that happens here in any given week!  I tend to stick to rust-related issues because I'm not yet comfortable with the rest of flycheck.  But I'll keep on helping in any capacity I can.

lunaryorn:
@fmdkdd You're welcome, thank you very much for your contributions and your support!  😍  Taking care of Rust-related issues already helps us a lot 👍 👏

PR #1080, opened 2016-09-10
[WIP] Add a Python type checker using MyPy

cpitclaudel:
I think we'll hold this until we merge the next-checker stuff :) This is a checker for [MyPy](http://www.mypy-lang.org/). Types!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,7 @@
 - New syntax checkers:
 
   - Elixir with ``dogma`` [GH-969]
+  - Python with ``mypy`` [GH-1080]
 
 - New features:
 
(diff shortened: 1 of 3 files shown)


dakra:
Any news on this? I would like mypy support :)

cpitclaudel:
@dakra Can you try the branch and report whether it works for you? That would help :)

dakra:
I tried your changes and seems to work fine.

I had to rename a few command line arguments:
`--suppress-error-context` is now `--hide-error-context`
and instead of `--silent-imports` I always use `--follow-imports=skip` (since we never want to return errors from different files?!) and and replaced the original `--silent-imports` with `--ignore-missing-imports`.

I didn't play around with it too much but lgtm :+1:

fmdkdd:
@cpitclaudel Do you still want to wait for the `next-checker` changes?  I'd rather have that working now than in an hypothetical future ;)

fmdkdd:
Closing in favor of #1354.

PR #1093, opened 2016-09-18
Add a style guide for contributors and maintainers

lunaryorn:
I took some time to write up a style guide for Flycheck.  It should cover everything from code to documentation, tests, and even what features we consider appropriate for Flycheck.

@flycheck/maintainers @flycheck/contributors  Please take a look and tell me what you think about the rules, and whether anything is missing.

### DIFF
--- a/doc/contributor/style-guide.rst
+++ b/doc/contributor/style-guide.rst
@@ -1,0 +1,118 @@
+.. _flycheck-style-guide:
+
+=============
+ Style Guide
+=============
+
+This document describes our code style.  It tells you what to look for when
+making changes to Flycheck, or when reviewing pull requests.
+
+Features
+========
+
+Flycheck’s scope and focus is providing the infrastructure and foundations for
+on-the-fly syntax checking.  Flycheck provides the basics but deep integration
+with particular programming languages is best left to :ref:`separate packages
+<flycheck-extensions>`.
+
+Whether a feature is within the scope of Flycheck is the :ref:`maintainer’s
+<flycheck-maintainers>` judgement call.  Generally we reserve the right to
+reject any pull request for being out of scope.
+
+* Avoid a *disproportionate amount of code* for a single syntax checker or
+  language.  Look at the built-in checkers for judgement.  A syntax checker that
+  requires a lot more code than any built-in checker is likely to be rejected.
+
+* Avoid *deep integration* with a particular UI or completion framework.  Emacs’
+  standard is our standard: We will reject code that is tied to Helm or Counsel.
+
+* Likewise do not deviate from Emacs’ default behaviour too much.  Stick to
+  Emacs’ standard for key bindings, interactive functions, etc.
+
+Style
+=====
+
+.. important::
+
+   ``make check compile`` must pass on Emacs 25 or newer.  This command checks
+   for some formatting issues and compilation errors.
+
+   Run ``make format`` with Emacs 25 to automatically reformat the Emacs Lisp
+   source files.
+
+* Generally try to fit into the style of the code you see.
+
+* Indent with the default indentation rules.
+
+* Follow the :infonode:`(elisp)Programming Tips` for Emacs Lisp.
+
+* Whitespace:
+
+  * 80 characters per line.
+  * Avoid tabs and trailing spaces.
+
+* Naming:
+
+  * Prefix all variables and functions with the name of the containing library,
+    i.e. ``flycheck-`` for everything that is in :file:`flycheck.el`.
+
+  * End boolean predicates with ``-p``, i.e. ``flycheck-valid-checker-p``.
+
+* Avoid macros, and use them for syntax only.
+
+* Adhere to the :infonode:`(elisp)Key Binding Conventions`.  Particularly do not
+  define keys in Emacs’ reserved keymaps or in the :samp:`C-c {LETTER}` space
+  for user bindings.
+
+Libraries
+=========
+
+* Do **not** advise built-in or 3rd party functions and commands.
+
+* Do **not** redefine built-in or 3rd party functions, unless for compatibility,
+  but then copy the newer definition verbatim.
+
+* Do **not** use ``with-eval-after-load`` and similar functions.
+
+* Dependencies:
+
+  * Use built-in Emacs libraries freely.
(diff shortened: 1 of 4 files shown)


doc/contributor/style-guide.rst
```diff
@@ -0,0 +1,113 @@
...
+   Generally, ``make check compile`` must pass on Emacs 25 or newer.  This
+   command checks for some style errors and byte-compile warnings and errors.
+
+* Indent with the default indentation rules.
+
+* Follow the :infonode:`(elisp)Programming Tips` for Emacs Lisp.
+
+* Whitespace:
+
+  * 80 characters per line.
```
fmdkdd: We already have `indent-tabs-mode` as a buffer-local variable.  Would it make sense to enforce `fill-column` as well?

doc/contributor/style-guide.rst
```diff
@@ -0,0 +1,113 @@
...
+  * Avoid tabs and trailing spaces.
```
fmdkdd: Can we also enforce that somehow, in Emacs or in git?

doc/contributor/style-guide.rst
```diff
@@ -0,0 +1,113 @@
...
+  * Use built-in Emacs libraries freely.
+  * Introduce external dependencies with care.  Prefer built-in
+    libraries. ``dash.el`` is fine, though.
+  * Avoid dependencies on language-specific libraries.
+
+* Avoid ``cl-lib``:
+
+  * Prefer ``seq`` over ``dash`` over ``cl-lib``.  Use list functions from
+    ``cl-lib`` only as the very last resort.
+  * Prefer ``let-alist`` and ``pcase`` over ``cl-destructuring-bind``.
```
fmdkdd: Just curious: what is the reason to avoid `cl-lib`?  Is it because there's a potential to mistake Elisp for CommonLisp?

doc/contributor/style-guide.rst
fmdkdd: Shouldn't we mention the recently-introduced `make format`?

fmdkdd:
Given the care that goes into each review, it makes sense to have this guide :+1: .

Style guides are tricky to write, because they are often arbitrary.  This one at least stays factual with clear "don'ts".

One the goals of a style guide is to enforce coherence.  To me, it seems that the general advice here should be "try to mimick what's already there", be it for writing a new checker, or for writing documentation.  This is already mentionned in "avoid a disproportionate amount of code", but maybe it warrants to be stated on its own.

Also, if we can enforce coherence automatically before the code even reaches the pull request, then less work for the reviewers.  Hence my comments on enforcing `fill-column` or using `make format`.

manuel-uberti:
LGTM. :+1:

manuel-uberti:
LGTM. :+1:

doc/contributor/style-guide.rst
```diff
@@ -0,0 +1,113 @@
...
+=====
+
+.. important::
+
+   Generally, ``make check compile`` must pass on Emacs 25 or newer.  This
+   command checks for some style errors and byte-compile warnings and errors.
+
+* Indent with the default indentation rules.
+
+* Follow the :infonode:`(elisp)Programming Tips` for Emacs Lisp.
```
aaronjensen: emacs noob question, how do I find this in emacs?

doc/contributor/style-guide.rst
```diff
@@ -0,0 +1,113 @@
...
+
+Libraries
+=========
+
+* Do **not** advise built-in or 3rd party functions and commands.
+
+* Do **not** redefine built-in or 3rd party functions, unless for compatibility,
+  but then copy the newer definition verbatim.
+
+* Do **not** use ``with-eval-after-load`` and similar functions.
```
aaronjensen: Why is this?

kthelgason:
Thanks @lunaryorn for continuing to make the process of contributing to Flycheck awesome! Definitely LGTM.

doc/contributor/style-guide.rst
lunaryorn: I think that it's not a good way to try and cram all possible editing settings into buffer local variables.  There are just so many of these.  I'd rather remove `indent-tabs-mode` as well.

We have `make format` and `make check`—that's much easier imho.

doc/contributor/style-guide.rst
lunaryorn: +:+1:  Will add it

doc/contributor/style-guide.rst
lunaryorn: It's just personal preference.  If find the CL API cumbersome to use and the function names awkward, in particular in contrast to prevalent names in contemporary functional languages.  `cl-remove-if` feels really strange, but `seq-filter` is nice :)

lunaryorn:
@fmdkdd Yep, "try to fit into what you see" is a good advice, will add it.

We try and enforce much of this already:  `make check` checks for checkdoc and formatting errors, and I plan to extend it to warn about line length, and maybe other issues as well.

doc/contributor/style-guide.rst
fmdkdd: Fair enough :)

doc/contributor/style-guide.rst
lunaryorn: `M-x info-browse-manual RET elisp` (or use `C-h i` to get a list of all manuals, and then `m elisp` to open the elisp manual), and then `m Programming Tips`.

doc/contributor/style-guide.rst
lunaryorn: These are for user configuration only.  A package should not create an implicit dependency on other libraries via `with-eval-after-load`.  That's too much magic, and makes things hard to test.

If someone feels the need to do that we should either move that into an entirely separate package or add a hard `require`, depending on the circumstances.

lunaryorn:
@kthelgason Thanks for these kind words :blush:

lunaryorn:
LGTM, @cpitclaudel mind to take a look?

cpitclaudel:
LGTM.  It's a great step forward, and it's nicely written. :+1:

lunaryorn:
Merged :relaxed:  Please feel free to make adjustments

PR #1102, opened 2016-09-22
Run slim-lint inplace instead from a temp location

jtdowney:
The `slim-lint` utility makes use of a few different configuration files during its run. The main one is `.slim-lint.yml` but `.ruboconf.yml` can also be used if you have Rubocop checks enabled. Flycheck with `slim-lint` was not taking these files into account since the source was being copied to a temp location.

I don't know enough about flycheck to understand the downsides of running this inplace vs from a temp location. This just scratches my own itch of getting the correct warnings reporting while editing files.

I did look first at passing a file to `slim-lint`, which would work for `.slim-lint.yml` but not for `.ruboconf.yml`.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8896,7 +8896,7 @@ See URL `http://slim-lang.com'."
   "A Slim linter.
 
 See URL `https://github.com/sds/slim-lint'."
-  :command ("slim-lint" "--reporter=checkstyle" source)
+  :command ("slim-lint" "--reporter=checkstyle" source-inplace)
   :error-parser flycheck-parse-checkstyle
   :modes slim-mode)
 


lunaryorn:
@jtdowney In-place temporary files are known to confuse all kinds of source-watching build tools, so we'd really like to avoid them.

Can you perhaps check whether `slim-lint` allows us to pass the code to check on standard input and provides an argument to denote the file name?  Rubocop does that, for instance.

jtdowney:
It looks like it does not support stdin like you suggested. There is an option to pass to rubocop file as an environment variable. With that the two configs could be passed in and the file still copied to a temp location.

I was able to already figure out how to find/pass something as a command line flag. Is there an example already where a file path is passed as an environment variable?

lunaryorn:
@jtdowney We do not support environment variables for syntax checkers currently, and I do not think that this is the way to go.

Could you ask the slim-lint maintainers to add support for standard input and passing a file name via a command line argument?

fmdkdd:
Closing due to inactivity.  Feel free to open up a new PR/issue if this is still a problem with `slim-lint`.

Issue #1101, opened 2016-09-22
Make it possible to show longer ID


Mekk:
**The problem**

I use flycheck for python code, where it wraps pylint. In default configuration it shows error (in `*Flycheck errors*` as sth like: 
`
   87   5 warning  no-... Method could be a function (python-pylint)
`
I find fourth column (ID, `no...` in this case) problematic. It is useful – I need this identifier whenever I want to disable given message (by inserting it into ~/.pylintrc or making steering comment like `# pylint:disable=no-self-use`) or simply google for details. But it is so heavily truncated that it is unreadable.

So, please, make it somehow possible to see whole `ID` of error/warning I look at.

**Suggested solutions**

Below a few ideas (from imperfect but probably simple to implement, to more difficult):
1. For time being I simply edited `flycheck-error-list-format` in `flycheck.el`, changing id column spec to `("ID" 15 t)`. It's not perfect (lost space on checkers without identifier), but much better (whole `no-self-use` fits, and for longer messages prefix is long enough to distinguish). Unfortunately I must re-edit this variable on every flycheck upgrade.
   
    So, it would be nice, if some parts of this variable (column widths?, column widths and names?) could be customizable, or at least easily modifiable from .emacs (say by being put into simpler, non-const variable).
2. Even better, if ID colum width could be matched to the context - expanded to sth like 15-20 chars for backends which use long IDs, truncated to few chars for modes/backends which use short numerical codes, hidden for backends which do not apply them at all. This could be either configurable by backend (say, backend somehow reports "I don't use ID"/"I use short IDs"/"I use long IDs") or just calculated dynamically from actual errors (just make column wide enough to fit longest code, with some limit)

**Related suggestions**

Whenever ID is truncated, three characters (three dots) are used to signal the fact. Especially in 5 chars case this does not make sense. Unicode … (three-dot character) takes one char. Not sure whether this is possible to modify from your code level but maybe tabulated list has some arg somewhere. Or maybe ID could be truncated this way before being fed to list…

lunaryorn:
Thank you for reporting this issue.  I understand that it's a nuisance, but to me it's not important to fix to be honest: It affects only a single syntax checker, which I don't use personally.   Personally I'll not fix this anytime time soon.

While it'd be great if the error list automatically changed column widths according to the content that's probably hard to implement since there's no support for this in Tabulated List Mode.

[...]

PR #1104, opened 2016-09-23
Show untruncated error ID as help-echo in error list

cpitclaudel:
Inspired by GH-1101.  I've made the mistake of mousing over a truncated error id to see more multiple times before, so this will at least help with that.

@Mekk, what do you think?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3820,16 +3820,21 @@ the beginning of the buffer."
   "Go to the error at BUTTON."
   (flycheck-error-list-goto-error (button-start button)))
 
-(defsubst flycheck-error-list-make-cell (text &optional face)
+(defsubst flycheck-error-list-make-cell (text &optional face help-echo)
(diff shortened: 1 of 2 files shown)


lunaryorn:
@cpitclaudel That's a nice change in any way.  👍 LGTM.

But shouldn't we add a help-echo to the message column as well?  Just because we can now 😎 and it might also get truncated in some situations.

cpitclaudel:
> But shouldn't we add a help-echo to the message column as well? Just because we can now :sunglasses: and it might also get truncated in some situations.

Good point; I didn't think about long lines :) Done.

lunaryorn:
@cpitclaudel Go ahead and merge :)

cpitclaudel:
LGTM FTR :)

lunaryorn:
:blush:

lunaryorn:
😍 Thanks :+1:

Issue #1085, opened 2016-09-12
flycheck-eslint errors out with a bare setup


cowboyd:
Opened up a file with a .js extension using the javascript-eslint checker enabled. It plopped a big nasty error into my minibuffer saying that the checker might be mal-defined instead of verifying the javascript syntax as expected.

<img width="1222" alt="emacs_prelude_-_scratch_js_and_check_buffers_ _flycheck_30-cvs_documentation_" src="https://cloud.githubusercontent.com/assets/4205/18453585/93de9c44-7905-11e6-9ee2-ccf65eda6fc1.png">

It looks as though the problem is that `eslint` throws an error if you don't have at least some configuration file present, even if it's an empty one in your home directory. This seems suboptimal to me, but I don't know if this is something that I should raise upstream, or that can be worked around in flycheck.

Elsint does support a `--no-eslintrc` flag that will make the error go away:

```
$ echo "var x = 'y';" | eslint --stdin                                                                                                                                        
No ESLint configuration found.
Error: No ESLint configuration found.
    at getLocalConfig (/Users/cowboyd/.nvm/versions/node/v4.2.3/lib/node_modules/eslint/lib/config.js:162:35)
    at Config.getConfig (/Users/cowboyd/.nvm/versions/node/v4.2.3/lib/node_modules/eslint/lib/config.js:256:22)
    at processText (/Users/cowboyd/.nvm/versions/node/v4.2.3/lib/node_modules/eslint/lib/cli-engine.js:222:33)
    at CLIEngine.executeOnText (/Users/cowboyd/.nvm/versions/node/v4.2.3/lib/node_modules/eslint/lib/cli-engine.js:749:26)
    at Object.execute (/Users/cowboyd/.nvm/versions/node/v4.2.3/lib/node_modules/eslint/lib/cli.js:174:42)
    at /Users/cowboyd/.nvm/versions/node/v4.2.3/lib/node_modules/eslint/bin/eslint.js:57:36
    at ConcatStream.<anonymous> (/Users/cowboyd/.nvm/versions/node/v4.2.3/lib/node_modules/eslint/node_modules/concat-stream/index.js:36:43)
    at emitNone (events.js:72:20)
    at ConcatStream.emit (events.js:166:7)
    at finishMaybe (/Users/cowboyd/.nvm/versions/node/v4.2.3/lib/node_modules/eslint/node_modules/concat-stream/node_modules/readable-stream/lib/_stream_writable.js:475:14)
```

vs

```
[...]

PR #1105, opened 2016-09-24
ESLint support for :enabled

Simplify:
Here is `:enabled` for ESLint checker and some test fixes.

There is one issue with tests. ESLint integration tests will run on GNU Emacs 24.5. When using 25.1, both tests are skipped for some reason. In both versions of GNU Emacs `:enabled` will run and return right value, and my best guess that this is probably doe to some ERT change in 25.1.

Fixes and closes GH-1085

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@
     no longer used [GH-1057]
   - ``:modes`` is now mandatory for syntax checker definitions [GH-1071]
   - Remove jade checker [GH-951] [GH-1084]
+  - Remove ``javascript-eslintrc`` and instead rely on eslint's own configuration file
+    search [GH-1085]
 
 - New syntax checkers:
 
@@ -21,6 +23,10 @@
   - Add ``flycheck-cargo-rustc-args`` to pass multiple arguments to cargo rustc
     subcommand [GH-1079]
   - Add ``:enabled`` property to ``flycheck-define-checker`` [GH-1089]
+
+- Improvements:
+
+  - Do not use ``javascript-eslint`` if eslint cannot find a valid configuration [GH-1085]
 
 29 (Aug 28, 2016)
 =================
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -549,11 +549,14 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       Check syntax and lint with `ESLint <http://eslint.org/>`_.
 
+      .. note::
+
+         Flycheck does not use this syntax checker if eslint cannot find a valid
+         configuration file.
+
       .. defcustom:: flycheck-eslint-rules-directories
 
          A list of directories with custom rules.
-
-      .. syntax-checker-config-file:: flycheck-eslintrc
 
    .. syntax-checker:: javascript-jshint
 
--- a/flycheck.el
+++ b/flycheck.el
@@ -7489,16 +7489,11 @@ for more information about the custom directories."
   :safe #'flycheck-string-list-p
   :package-version '(flycheck . "29"))
 
-(flycheck-def-config-file-var flycheck-eslintrc javascript-eslint nil
-  :safe #'stringp
-  :package-version '(flycheck . "0.20"))
-
 (flycheck-define-checker javascript-eslint
   "A Javascript syntax and style checker using eslint.
 
 See URL `https://github.com/eslint/eslint'."
   :command ("eslint" "--format=checkstyle"
-            (config-file "--config" flycheck-eslintrc)
             (option-list "--rulesdir" flycheck-eslint-rules-directories)
             "--stdin" "--stdin-filename" source-original)
   :standard-input t
@@ -7518,6 +7513,12 @@ See URL `https://github.com/eslint/eslint'."
                                    (flycheck-error-message err))))
                           (flycheck-sanitize-errors errors))
                   errors)
+  :enabled (lambda (checker)
+             (let* ((default-directory (flycheck-compute-working-directory checker))
+                    (executable (flycheck-find-checker-executable checker))
+                    (exitcode (call-process executable nil nil nil
+                                            "--print-config" ".")))
+               (eq exitcode 0)))
   :modes (js-mode js-jsx-mode js2-mode js2-jsx-mode js3-mode)
   :next-checkers ((warning . javascript-jscs)))
 
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3321,17 +3321,16 @@ Why not:
   (let ((flycheck-disabled-checkers '(javascript-jshint)))
     (flycheck-ert-should-syntax-check
      "language/javascript/syntax-error.js" flycheck-test-javascript-modes
-     '(3 26 error "Parsing error: Unexpected token )" :checker javascript-eslint))))
+     '(3 25 error "Parsing error: Unexpected token )" :checker javascript-eslint))))
 
 (flycheck-ert-def-checker-test javascript-eslint javascript warning
   :tags '(checkstyle-xml)
-  (let ((flycheck-eslintrc "eslint.json")
-        (flycheck-disabled-checkers '(javascript-jshint javascript-jscs)))
+  (let ((flycheck-disabled-checkers '(javascript-jshint javascript-jscs)))
     (flycheck-ert-should-syntax-check
      "language/javascript/warnings.js" flycheck-test-javascript-modes
-     '(3 2 warning "Use the function form of \"use strict\"." :id "strict"
+     '(3 2 warning "Use the function form of 'use strict'." :id "strict"
          :checker javascript-eslint)
-     '(4 9 warning "\"foo\" is defined but never used" :id "no-unused-vars"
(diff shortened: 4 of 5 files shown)


CHANGES.rst
```diff
@@ -22,6 +22,11 @@
     subcommand [GH-1079]
   - Add ``:enabled`` property to ``flycheck-define-checker`` [GH-1089]
 
+- Improvements:
+
+  - When checking Javascript buffer we will disable ``eslint`` checker if that
+    checker can not find any valid ESLint configuration [GH-1085]
```
lunaryorn: I think we can shorten that to "Do not use \`\`javascript-eslint\`\` if eslint cannot find a valid configuration [GH-1085]".

doc/languages.rst
```diff
@@ -549,6 +549,10 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       Check syntax and lint with `ESLint <http://eslint.org/>`_.
 
+      .. note::
+
+         This syntax checker requires the presence of a configuration file.
```
lunaryorn: I think "Flycheck does not use this syntax checker if eslint cannot find a valid configuration file." is more precise, isn't it?

flycheck.el
```diff
@@ -1805,7 +1805,7 @@ nil otherwise."
     (if shall-enable
         (cl-pushnew checker flycheck-enabled-checkers)
       (cl-pushnew checker flycheck-disabled-checkers))
-      shall-enable))
+    shall-enable))
```
lunaryorn: Looks as if the indentation has changed here?

flycheck.el
```diff
@@ -7512,6 +7512,20 @@ See URL `https://github.com/eslint/eslint'."
                                    (flycheck-error-message err))))
                           (flycheck-sanitize-errors errors))
                   errors)
+  :enabled (lambda (checker)
+             (let* ((command (concat (flycheck-find-checker-executable checker)
+                                     (if flycheck-eslintrc
```
lunaryorn: Oh, I didn't know that we still had an option for eslint's configuration file.  I think we should remove it, shouldn't we?  After all, eslint does a much better job at finding its own configuration than Flycheck, doesn't it?

flycheck.el
```diff
@@ -7512,6 +7512,20 @@ See URL `https://github.com/eslint/eslint'."
...
+                                         (concat " --config " flycheck-eslintrc)
+                                       "")
+                                     " --print-config "
+                                     (shell-quote-argument
+                                      (flycheck-compute-working-directory
+                                       checker))
+                                     "."))
+                    (exitcode (call-process-shell-command command nil nil nil)))
```
lunaryorn: Please don't use shell commands in Flycheck.  Just use `call-process` directly; avoid the shell.

flycheck.el
```diff
@@ -7512,6 +7512,20 @@ See URL `https://github.com/eslint/eslint'."
...
+  :enabled (lambda (checker)
+             (let* ((command (concat (flycheck-find-checker-executable checker)
+                                     (if flycheck-eslintrc
+                                         (concat " --config " flycheck-eslintrc)
+                                       "")
+                                     " --print-config "
+                                     (shell-quote-argument
+                                      (flycheck-compute-working-directory
+                                       checker))
+                                     "."))
```
lunaryorn: Why this argument?  Don't we pass the working directory in the previous argument?

flycheck.el
```diff
@@ -7512,6 +7512,20 @@ See URL `https://github.com/eslint/eslint'."
...
+                    (exitcode (call-process-shell-command command nil nil nil)))
+               (if (eq exitcode 0)
+                   'true
```
lunaryorn: `true` is `t` in Emacs Lisp.  And `(if cond true false)` is just `cond` :wink:

flycheck.el
Simplify: Completely remove `flycheck-eslintrc` option?

lunaryorn: @Simplify Yes, I think we can safely do that in this pull request.  I don't think that anyone still uses that option.

flycheck.el
Simplify: For the better :)

lunaryorn: @Simplify I see, but why didn't `make check` catch that? :flushed:

lunaryorn: Scratch that, `make check` caught it, but why not on Travis CI?

lunaryorn: Oh dear, my bad: d470857af86e6927a054ecab52e7bc5f695db39c 😞 

I've fixed this formatting flaw on master, so please rebase onto `master`.  Thank you so much for making me aware!

Simplify:
Ok, I'll rebase tomorrow. I fixed other issues, so you can review it again.

Simplify:
Ok, rebased.

flycheck.el
```diff
@@ -7518,6 +7513,14 @@ See URL `https://github.com/eslint/eslint'."
                                    (flycheck-error-message err))))
                           (flycheck-sanitize-errors errors))
                   errors)
+  :enabled (lambda (checker)
+             (let* ((directory (concat
+                                (flycheck-compute-working-directory checker)
+                                "."))
```
cpitclaudel: Can you explain what the (concat … does)?

Simplify: Adds dot after directory name. Without that dot, ESLint will not look in that directory and ancestors but only ancestors. See sentence before last one in error message:

```
[1085-eslint-use-enabled]~/Projects/elisp/flycheck-upstream$ eslint --print-config test/resources/language/javascript/.
{
  "globals": {},
  "env": {},
  "rules": {
    "no-unused-vars": 1
  },
  "parserOptions": {}
}
[1085-eslint-use-enabled]~/Projects/elisp/flycheck-upstream$ eslint --print-config test/resources/language/javascript/

ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:

    eslint --init

ESLint looked for configuration files in test/resources/language and its ancestors.

If you think you already have a configuration file or if you need more help, please stop by the ESLint chat room: https://gitter.im/eslint/eslint
```

Simplify: In my original implementation I used:
`(tempfile (flycheck-save-buffer-to-temp #'flycheck-temp-file-inplace))`, however using `flycheck-compute-working-directory` should be faster.

CHANGES.rst
```diff
@@ -9,6 +9,9 @@
     no longer used [GH-1057]
   - ``:modes`` is now mandatory for syntax checker definitions [GH-1071]
   - Remove jade checker [GH-951] [GH-1084]
+  - Remove ``--config`` option in ``javascript-eslint`` in favour of
+    ``eslint``'s own config detection. Therefore ``javascript-eslintrc`` is
+    no longer used [GH-1085]
```
lunaryorn: To me that reads as if the option is still there.  How about shortening to

```
Remove ``javascript-eslintrc`` and instead rely on eslint's own configuration file search [GH-1085]
```

flycheck.el
lunaryorn: Did you try to change the working directory and just pass `.` to eslint?  IOW, like this:

```
(let* ((default-directory (flycheck-compute-working-directory checker))
          …
         (exitcode (call-process executable nil nil nil "--print-config" ".")))
   …)
```

Wouldn't that work as well?

I think that's easier to understand, and since you're computing the working directory anyway why not use it as such?

Simplify: Will that work if directory does not exist yet? When you open file in Emacs <project-dir>/components/some_component.js and components directory does not exists yet.

Simplify:
ok, ready for another review.

cpitclaudel:
LGTM :) And the code looks good, too.

This makes me wonder: now that we have `:enabled`, how does one force a refresh?  That is, suppose I add an `eslint` config file *after* opening my file. How do I force Flycheck to re-enable the checker? IOW, how do I force Flycheck to re-check `:enabled` properties? Using `M-x revert-buffer`?

lunaryorn:
@cpitclaudel That, or `C-u C-c ! x` to reenable the syntax checker.  It's disabled just as if you had used `C-c ! x`.

cpitclaudel:
@lunaryorn Neat, thanks!

lunaryorn:
@Simplify Awesome, LGTM.  Please merge :)

lunaryorn:
@Simplify Thank you so much for taking on this issue; awesome work on these two pull requests :+1: 👏

Simplify:
licence/cla — Waiting for status to be reported

lunaryorn:
@Simplify Mmh, hadn't seen that… apparently a hiccup.  Since you signed it already I'll force-merge.

Sorry for the nuisance.

lunaryorn:
@Simplify And there it is :)

Simplify:
Thanks :)

aaronjensen:
fyi this broke support for `eslint_d` because `eslint_d` did not support `--print-config`. [It does now](https://github.com/mantoni/eslint_d.js/pull/51), however, so upgrade to `eslint_d` 4.1.0 if you use it. If you don't use it, check it out! 😄

PR #1106, opened 2016-09-24
Use consistent colors in header and entries of message column

cpitclaudel:
This change highlights the word "Checker" in the header, for consistency with entries.  What do we think about this?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3793,12 +3793,22 @@ the beginning of the buffer."
     map)
   "The keymap of `flycheck-error-list-mode'.")
 
+(defun flycheck-error-list-make-last-column (message checker)
+  "Compute contents of the last error list cell.
+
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Actually, this code is circular.  Let me have a better look at whether we can easily fix this.

cpitclaudel:
Fixed. Annoyingly, I had to move the `last-column` definition before the defconst.

lunaryorn:
LGTM.  I've restarted the travis CI job, looks like a fluke.

cpitclaudel:
LGTM as well, for the record :) @lunaryorn, want to comment on the diff before I merge?

lunaryorn:
@cpitclaudel I have already :wink:  Merged.  Thanks for this change!

Issue #1107, opened 2016-09-24
[Flake8] show descriptive warning in modeline if version < 3.0


thisch:
Since you have updated the required version of flake8 to 3.0 for the python checker in the master branch, it would be nice if an error/warning is shown in the modeline if the installed flake version is < 3.0. ATM you get the `flake8: error: no such option: --stdin-display-name` error, which is hard to understand if you don't know that you need to update flake8 The version checking also applies to other flycheck checkers, of course.

lunaryorn:
@thisch I'm sorry, but as a policy we do not show such warnings for syntax checkers; version checks would cause to much maintenance effort and quickly lead to accumulated technical debt.  

Please downgrade Flycheck, or copy the flake8 definition from older releases, and please follow Flycheck's changelog for notifications about breaking changes.

Issue #1100, opened 2016-09-21
Add troubleshooting section


lunaryorn:
Sometimes checkers fail, but all our manual has to offer is [this poor section](http://www.flycheck.org/en/latest/user/syntax-checks.html#debug-syntax-checking) that I came up with.  That's too little.

Let's expand that to a proper "troubleshooting" top-level section of the user guide that mentions this command as well as `C-c ! C-c`, and gives a general guide line on how to troubleshoot Flycheck syntax checkers.

# TODO

- [x] Add troubleshooting section, documenting `C-c ! v` and `C-c ! C-c`
- [x] Suggest to update the syntax checker if it fails with unknown flags, see #1108
- [x] For OS X users, refer to exec-path-from-shell, also in the "try out section", the quickstart guide, and the installation instructions
- [x] Prominently link in the "try out section" on the front page, in the quickstart guide and after the installation instructions

fmdkdd:
Given the recent influx of issues where users had to first set up their path properly, this could definitely help.

lunaryorn:
@fmdkdd Great point, we should definitely refer to exec-path-from-shell here.

PR #1108, opened 2016-09-25
Suggest updating the checker when warning about suspicious output

cpitclaudel:
Inspired by GH-1107.  Please feel free to suggest improvements to the wording :) Hopefully this will help users who get confusing error messages when using outdated checkers.

Once we have it, we could simply include a link to the upcoming "troubleshooting" section of the manual.


lunaryorn:
@cpitclaudel LGTM, that's a nice change!  I'll also put that into the troubleshooting section in #1100.

cpitclaudel:
LGTM, FTR :)

Issue #1055, opened 2016-08-26
Use the `python-shell-interpreter` command for running checkers


benf-wspdigital:
Setting the `python-shell-interpreter` variable specifies the interpreter to use for running Python commands. This should also be used for Flycheck's checker commands.

That would make the Python checker executables:

* ${python-shell-interpreter} -m flake8
* ${python-shell-interpreter} -m pylint
* ${python-shell-interpreter} -m py_compile

That way, when a different buffer triggers a change to the `python-shell-interpreter` variable, the checkers will do the right thing for that buffer's Python code.

lunaryorn:
@benf-wspdigital Does `python-shell-interpreter` require that the program provides `-m`? IOW can we expect `python-shell-interpreter` to provide that flag?  What if it's ipython? Or bpython? Or some other interactive Python shell?

And then, can you run `pylint` and `flake8` this way? Is that behaviour documented and guaranteed by these programs?

lunaryorn:
@benf-wspdigital Did you receive my comment?

benf-wspdigital:
> Does python-shell-interpreter require that the program provides -m? IOW can we expect python-shell-interpreter to provide that flag?

I'm not sure I understand the question. If you mean, does every compliant Python interpreter program provide that option: yes.

>  What if it's ipython? Or bpython? Or some other interactive Python shell?

Those are, as you point out, designed to be interactive Python shells, not suitable for the `python-shell-interpreter` command.

> I can't answer officially for those projects's policies. But yes, they work fine that way.

lunaryorn:
@benf-wspdigital How come you think that IPython isn't suitable for `python-shell-interpreter`? Why, Python Mode even supports it officially since Emacs 24.4.  See it's commentary:

```elisp
;; Besides that only the standard CPython (2.x and 3.x) shell and
;; IPython are officially supported out of the box, the interaction
;; should support any other readline based Python shells as well
;; (e.g. Jython and PyPy have been reported to work).  You can change
;; your default interpreter and commandline arguments by setting the
;; `python-shell-interpreter' and `python-shell-interpreter-args'
;; variables.  This example enables IPython globally:

;; (setq python-shell-interpreter "ipython"
;;       python-shell-interpreter-args "-i")
```

[...]

PR #1113, opened 2016-09-27
Use python -c to invoke pylint and flake8

cpitclaudel:
**Use python -m to invoke pylint and flake8**
Part of GH-1055.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8687,6 +8687,46 @@ See URL `http://puppet-lint.com/'."
   ;; the buffer is actually linked to a file, and if it is not modified.
   :predicate flycheck-buffer-saved-p)
 
+(defun flycheck-python-find-module (checker module)
+  "Check if a Python MODULE is available.
+CHECKER's executable is assumed to be a Python REPL."
+  (-when-let* ((py (flycheck-find-checker-executable checker))
+               (script (concat "import sys; sys.path.pop(0);"
+                               (format "import %s; print(%s.__file__)"
+                                       module module))))
+    (with-temp-buffer
+      (and (eq (ignore-errors (call-process py nil t nil "-c" script)) 0)
+           (string-trim (buffer-string))))))
+
+(defun flycheck-python-needs-module-p (checker)
+  "Determines whether CHECKER needs to be invoked through Python.
+Previous versions of Flycheck called pylint and flake8 directly;
+this check ensures that we don't break existing code."
+  (not (string-match-p (rx (or "pylint" "flake8")
+                           (or "-script.pyw" ".exe" ".bat" "")
+                           eos)
+                       (flycheck-checker-executable checker))))
+
+(defun flycheck-python-verify-module (checker module)
+  "Verify that a Python MODULE is available.
+Return nil if CHECKER's executable is not a Python REPL.  This
+function's is suitable for a checker's :verify."
+  (when (flycheck-python-needs-module-p checker)
+    (let ((mod-path (flycheck-python-find-module checker module)))
+      (list (flycheck-verification-result-new
+             :label (format "`%s' module" module)
+             :message (if mod-path (format "Found at %S" mod-path) "Missing")
+             :face (if mod-path 'success '(bold error)))))))
+
+(defun flycheck-python-module-args (checker module-name)
+  "Compute arguments to pass to CHECKER's executable to run MODULE-NAME.
+Return nil if CHECKER's executable is not a Python REPL.
+Otherwise, return a list starting with -c (-m is not enough
+because it adds the current directory to Python's path)."
+  (when (flycheck-python-needs-module-p checker)
+    `("-c" ,(concat "import sys,runpy;sys.path.pop(0);"
+                    (format "runpy.run_module(%S)" module-name)))))
+
 (flycheck-def-config-file-var flycheck-flake8rc python-flake8 ".flake8rc"
   :safe #'stringp)
 
@@ -8764,7 +8804,10 @@ Update the error level of ERR according to
 
 Requires Flake8 3.0 or newer. See URL
 `https://flake8.readthedocs.io/'."
-  :command ("flake8"
+  ;; Not calling flake8 directly makes it easier to switch between different
+  ;; Python versions; see https://github.com/flycheck/flycheck/issues/1055.
+  :command ("python"
+            (eval (flycheck-python-module-args 'python-flake8 "flake8"))
             "--format=default"
             (config-file "--config" flycheck-flake8rc)
             (option "--max-complexity" flycheck-flake8-maximum-complexity nil
@@ -8782,10 +8825,13 @@ Requires Flake8 3.0 or newer. See URL
             (id (one-or-more (any alpha)) (one-or-more digit)) " "
             (message (one-or-more not-newline))
             line-end))
+  :enabled (lambda ()
+             (or (not (flycheck-python-needs-module-p 'python-flake8))
+                 (flycheck-python-find-module 'python-flake8 "flake8")))
+  :verify (lambda (_) (flycheck-python-verify-module 'python-flake8 "flake8"))
   :modes python-mode)
 
-(flycheck-def-config-file-var flycheck-pylintrc python-pylint
-                              ".pylintrc"
+(flycheck-def-config-file-var flycheck-pylintrc python-pylint ".pylintrc"
   :safe #'stringp)
 
 (flycheck-def-option-var flycheck-pylint-use-symbolic-id t python-pylint
@@ -8805,7 +8851,11 @@ This syntax checker requires Pylint 1.0 or newer.
 
 See URL `https://www.pylint.org/'."
   ;; -r n disables the scoring report
-  :command ("pylint" "-r" "n"
+  ;; Not calling pylint directly makes it easier to switch between different
+  ;; Python versions; see https://github.com/flycheck/flycheck/issues/1055.
+  :command ("python"
+            (eval (flycheck-python-module-args 'python-pylint "pylint"))
+            "-r" "n"
             "--output-format" "text"
             "--msg-template"
             (eval (if flycheck-pylint-use-symbolic-id
@@ -8830,6 +8880,10 @@ See URL `https://www.pylint.org/'."
    (info line-start (file-name) ":" line ":" column ":"
          "C:" (id (one-or-more (not (any ":")))) ":"
          (message) line-end))
+  :enabled (lambda ()
+             (or (not (flycheck-python-needs-module-p 'python-pylint))
+                 (flycheck-python-find-module 'python-pylint "pylint")))
+  :verify (lambda (_) (flycheck-python-verify-module 'python-pylint "pylint"))
   :modes python-mode)
 
 (flycheck-define-checker python-pycompile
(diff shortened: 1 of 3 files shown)


cpitclaudel:
LGTM :)

lunaryorn:
@cpitclaudel I must admit that I'm [not sure](https://github.com/flycheck/flycheck/issues/1055#issuecomment-250105591) whether I'd like to merge this pull request.  Can we discuss the problems that this is supposed to solve first, over at #1055?

lunaryorn:
@cpitclaudel Independently of the discussion in #1055 we need custom `:predicate`s for these syntax checkers now since we need to check whether the modules are actually available.

cpitclaudel:
> @cpitclaudel Independently of the discussion in #1055 we need custom :predicates for these syntax checkers now since we need to check whether the modules are actually available.

Ah, very good point. That *is* one downside of this PR. `:predicate` or `:enabled`?

lunaryorn:
@cpitclaudel I'd go for `:enabled`.  After all that's not something that'd change frequently would it?

Generally I think we should slowly move to use `:enabled` for anything that checks whether a checker is installed, as these are facts that scarcely change.  We should leave `:predicate` for checks that really change frequently, like whether the buffer is saved, or whether it contains a certain text or so.

lunaryorn:
LGTM, but I "requested changes" for the missing `:enabled` lest we forget.

lunaryorn:
@cpitclaudel Did you get a chance to work on this pull request meanwhile?

cpitclaudel:
Barely.  I'll try to get some time before the week-end

lunaryorn:
@cpitclaudel No hurry, was just curious.

fmdkdd:
@cpitclaudel Any update on this one?

cpitclaudel:
I looked into this again, but there's a problem I didn't consider at the time: backwards compatibility.  If people have custom paths for flake8 or pylint, this change will break them.

There's multiple ways to proceed.

* Accept the breakage, document it, and move on; the problem is that the fix for users won't be entirely trivial (it's tricky to load a python module that's not in your Python path), unless they use a venv in which flake8 or pylint is installed (in which case they'll just have to point Flycheck to the venv's Python instead of the venv's pylint.

* Make it possible to use either form: we add a new variable `flycheck-flake8-python-executable`, which if set takes precedence over `flycheck-flake8-executable`.  This might be tricky to implement in today's Flycheck.

* Same as above, but use a single variable for the executable, and add `-m flake8` if we think the executable is python instead of a flake8 binary.  We could either do the detection automatically (pattern-match for `python(\d(\.\d)?)?(\.exe)?`) or add a setting.  If we do that, we can also change the default value of the checker executable to python, so we keep backwards compatibility while getting the benefits of `python -m`

* Abandon this PR.  I don't like that much, because I think it's harder than it should be ATM to mix python2 and python3 sources in Flycheck.

WDYT?

dakra:
Is it possible to make `python -m flake8` the default command and if someone customizes it, it'll just use that?

cpitclaudel:
> Is it possible to make python -m flake8 the default command and if someone customizes it, it'll just use that?

Not easily (at least I think, but I'd like to be proven wrong); Flycheck has a pretty deeply-embedded notion of checker executables, where each checker has it's own binary + arguments passed to it.  `python -m flake8` is an executable + 2 arguments, so what I'm proposing to do is to change the flake8 executable to `"python"` instead of flake8, and to extend the argument list with `-m flake8`; for compatibility, I'm also suggesting to make the argument list additions conditional, so that they only happen if the executable "looks like" python, rather than flake8.

ibizaman:
Not especially pushing for this PR, but one use-case I have where calling pylint through `python -m pylint` is necessary is when having a `.pylintrc` like so:
```
[MASTER]

load-plugins=
    my_module.my_helper
```
This is the error I get in emacs:
```
Suspicious state from syntax checker python-pylint: Flycheck checker python-pylint returned non-zero exit code 1, but its output contained no errors: Using config file ./.pylintrc
  File ".virtualenvs/python2.7/bin/pylint", line 11, in <module>
    sys.exit(run_pylint())
  File ".virtualenvs/python2.7/lib/python2.7/site-packages/pylint/__init__.py", line 16, in run_pylint
    Run(sys.argv[1:])
  File ".virtualenvs/python2.7/lib/python2.7/site-packages/pylint/lint.py", line 1312, in __init__
    linter.load_plugin_modules(plugins)
  File ".virtualenvs/python2.7/lib/python2.7/site-packages/pylint/lint.py", line 494, in load_plugin_modules
    module = modutils.load_module_from_name(modname)
  File ".virtualenvs/python2.7/lib/python2.7/site-packages/astroid/modutils.py", line 190, in load_module_from_name
    return load_module_from_modpath(dotted_name.split('.'), path, use_sys)
  File ".virtualenvs/python2.7/lib/python2.7/site-packages/astroid/modutils.py", line 232, in load_module_from_modpath
    mp_file, mp_filename, mp_desc = imp.find_module(part, path)
ImportError: No module named my_module

Try installing a more recent version of python-pylint, and please open a bug report if the issue persists in the latest release.  Thanks!
```

I fixed that in another way, by adding:
```elisp
    (setq flycheck-command-wrapper-function
          (lambda (command)
            (let* ((new-prefix (replace-regexp-in-string "pylint$" "python" (car command)))
                   (new-rest (append '("-m" "pylint") (cdr command)))
                   (new-command (append (list new-prefix) new-rest)))
              new-command))))
```

I also needed to set the default-directory to be the project root instead of the file's directory because other `my_module` wasn't always found.

cpitclaudel:
Thank, that's an interesting data point.  I think your solution is similar to mine.
How does your working directory setup looks like?

fmdkdd:
> I looked into this again, but there's a problem I didn't consider at the time: backwards compatibility. If people have custom paths for flake8 or pylint, this change will break them.

I'm not sure it's worth hanging onto backward compatibility on this one.  `python-pycompile` is already using `python -m`.  If users want to preserve the previous behavior, they can always override the checker with the previous version.

In this case, there is an even simpler trick:

```emacs-lisp
(setf (flycheck-checker-get 'python-flake8 'command)
      (cddr (flycheck-checker-get 'python-flake8 'command))
```

would drop the `python -m` part of the command and restore the previous behavior.

flycheck.el
```diff
@@ -8022,6 +8040,10 @@ Requires Flake8 3.0 or newer. See URL
             (id (one-or-more (any alpha)) (one-or-more digit)) " "
             (message (one-or-more not-newline))
             line-end))
+  :enabled (lambda () (and flycheck-flake8-python-executable
```
fmdkdd: Is this a new variable that's not in this branch?  Or did you mean `flycheck-python-flake8-executable`?  The latter is already checked by Flycheck.

cpitclaudel:
I updated the PR, but there's one more issue.  From the docs of `python -m`:

```
If this option is given, the first element of sys.argv will be the full path to the module file. As with the -c option, the current directory will be added to the start of sys.path.
```

The addition of the current directory means that large projects with names that conflict with the standard library, like Sphinx, can't be checked:

```
python -m pylint -r n --output-format text --msg-template \{path\}\:\{line\}\:\{column\}\:\{C\}\:\{symbol\}\:\{msg\} --rcfile /home/clement/.emacs.d/external-config/.pylintrc /home/clement/.local/lib/python3.5/site-packages/sphinx/environment.py
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/clement/.local/lib/python2.7/site-packages/pylint/__main__.py", line 10, in <module>
    pylint.run_pylint()
  File "/home/clement/.local/lib/python2.7/site-packages/pylint/__init__.py", line 16, in run_pylint
    from pylint.lint import Run
  File "/home/clement/.local/lib/python2.7/site-packages/pylint/lint.py", line 57, in <module>
    import astroid
  File "/home/clement/.local/lib/python2.7/site-packages/astroid/__init__.py", line 54, in <module>
    from astroid.nodes import *
  File "/home/clement/.local/lib/python2.7/site-packages/astroid/nodes.py", line 54, in <module>
    from astroid.scoped_nodes import (
  File "/home/clement/.local/lib/python2.7/site-packages/astroid/scoped_nodes.py", line 25, in <module>
    import io
  File "io.py", line 12, in <module>
    from docutils.readers import standalone
  File "/usr/local/lib/python2.7/dist-packages/docutils/readers/standalone.py", line 13, in <module>
    from docutils import frontend, readers
  File "/usr/local/lib/python2.7/dist-packages/docutils/frontend.py", line 38, in <module>
    import optparse
  File "/usr/lib/python2.7/optparse.py", line 419, in <module>
    _builtin_cvt = { "int" : (_parse_int, _("integer")),
  File "/usr/lib/python2.7/gettext.py", line 584, in gettext
    return dgettext(_current_domain, message)
  File "/usr/lib/python2.7/gettext.py", line 548, in dgettext
    codeset=_localecodesets.get(domain))
  File "/usr/lib/python2.7/gettext.py", line 483, in translation
    mofiles = find(domain, localedir, languages, all=1)
  File "/usr/lib/python2.7/gettext.py", line 440, in find
    for nelang in _expand_lang(lang):
  File "/usr/lib/python2.7/gettext.py", line 133, in _expand_lang
    from locale import normalize
ImportError: cannot import name normalize
```

(this happens because Sphinx contains a "locale" module.

cpitclaudel:
Fixed by moving to `python -c`.  Let me know what you think.

flycheck.el
```diff
@@ -8687,6 +8687,45 @@ See URL `http://puppet-lint.com/'."
...
+  (-when-let* ((py (flycheck-find-checker-executable checker))
+               (script (concat "import sys; sys.path.pop(0);"
+                               (format "import %s; print(%s.__file__)"
+                                       module module))))
+    (with-temp-buffer
+      (and (eq (ignore-errors (call-process py nil t nil "-c" script)) 0)
+           (string-trim (buffer-string))))))
+
+(defun flycheck-python-needs-module-p (checker)
+  "Determines whether CHECKER needs a `-m module-name' argument.
```
fmdkdd: I guess it's not really `-m module-name`, since we end up using `-c` instead.  But the intent is the same?

cpitclaudel: Thanks, I changed the docstring.

flycheck.el
```diff
@@ -8687,6 +8687,45 @@ See URL `http://puppet-lint.com/'."
...
+Previous versions of Flycheck called pylint and flake8 directly;
+this check ensures that we don't break existing code."
+  (not (string-match-p (concat "\\(pylint\\|flake8\\)"
```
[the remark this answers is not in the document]
cpitclaudel: Sure.  Sounds like the perfect time to advertise https://github.com/cpitclaudel/easy-escape, too :)

flycheck.el
```diff
@@ -8764,7 +8803,8 @@ Update the error level of ERR according to
 
 Requires Flake8 3.0 or newer. See URL
 `https://flake8.readthedocs.io/'."
-  :command ("flake8"
```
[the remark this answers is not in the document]
cpitclaudel: Good idea

cpitclaudel:
> This can go into CHANGES.rst Improvements section.

Yup, I already added it :)

cpitclaudel:
Question: do we actually care about argument order in docstrings? I'm temper to turn checkdoc-arguments-in-order-flag off.

fmdkdd:
I don't mind about the order no.

cpitclaudel:
OK, all green.

PR #1114, opened 2016-09-28
Check that error message is non-nil before replacing temp filename

cpitclaudel:
**Check that error message is non-nil before replacing temp filename**
The issue was introduced by GH-1030.  Some command checkers may produce
nil messages (with a custom error filter to handle them), so guard
against that case before replacing filename in error message.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5332,16 +5332,12 @@ function `buffer-file-name'."
       (when (seq-some (apply-partially #'flycheck-same-files-p
                                        (expand-file-name filename cwd))
                       buffer-files)
-        (let ((new-filename (buffer-file-name)))
-          (setf (flycheck-error-filename err) new-filename)
-          (when new-filename
-            (setf (flycheck-error-message err)
-                  (replace-regexp-in-string
-                   (regexp-quote filename)
-                   new-filename
-                   (flycheck-error-message err)
-                   'fixed-case
-                   'literal)))))))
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -5328,20 +5328,16 @@ Make the file name of ERR absolute.  If the absolute file name of
 ERR is in BUFFER-FILES, replace it with the return value of the
 function `buffer-file-name'."
   (flycheck-error-with-buffer err
-    (-when-let (filename (flycheck-error-filename err))
+    (-when-let* ((filename (flycheck-error-filename err)))
```
lunaryorn: Why the star-variant?  Looks still as if there's just one binding here…

lunaryorn:
LGTM

flycheck.el
cpitclaudel: Ugh, silly me :)

cpitclaudel:
Fixed. LGTM as well, of course :)

cpitclaudel:
Hmm. I need to stop using these GitHub buttons; they loose signatures.

lunaryorn:
@cpitclaudel You mean commit signatures with "squash-merge" and "rebase" merge?  Yes, noticed that too which is why I'm doing normal merges now for my own PRs.

I opened #1118 for us to discuss this topic.  In my opinion we should probably move to normal merges.

PR #1098, opened 2016-09-20
Add :enabled property to syntax checkers

Simplify:
Closes GH-1089

PR #1116, opened 2016-09-28
Disable syntax checkers which are not installed

### LINKED ISSUE #1115: Error while checking syntax automatically: (wrong-type-argument stringp nil) for JS files
After updating to the latest version of flycheck on Melpa (flycheck-20160927.1734) I started getting this error when accessing any JS files.

Here's a screenshot of my emacs where I followed [this great article](http://www.lunaryorn.com/2015/11/29/reproduce-bugs-in-emacs-Q.html) from @lunaryorn to initially reproduce the error.

screenshot 2016-09-28 00 07 11

I ran `M-x flycheck-verify-setup` and the only thing is that stood out to me is that I don't have eslint installed.  Is this the issue?  I was able to bisect this error to this [commit](https://github.com/flycheck/flycheck/commit/ec31874a662fdf1dcd174bca99af870a20415ceb).  Using a version of flycheck before this commit does not produce this error.

Here are the rest of the details:
* Your operating system
OS X (El Capitan) v10.11.6

* Your Emacs version from `M-x emacs-version`
Emacs 25.1.1

* Your Flycheck version from `M-x flycheck-version`
30snapshot (package: 20160927.1734)

lunaryorn:
Following up to #1098 I figured that it'd be a good idea to disable syntax checkers that are not installed upfront through the new `:enabled` feature, because it's unlikely that this changes in every syntax check, so it really doesn't need to be in `:predicate`.

Additionally, `:enabled` takes no parameters anymore and now runs with the checker's default directory, both changes for consistency with `:predicate`.

As a side-effect this should fix #1115 because if missing eslint is now disabled before its own `:enabled` even runs.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1614,6 +1614,10 @@ are mandatory.
      non-nil if this syntax checker shall be used to check the
      current buffer.  Otherwise it shall return nil.
 
+     If this checker has a `:working-directory' FUNCTION is
+     called with `default-directory' bound to the checker's
+     working directory.
+
      FUNCTION is only called in matching major modes.
 
      This property is optional.
@@ -1622,17 +1626,22 @@ are mandatory.
      A function to determine whether to use the syntax checker in
      the current buffer.
 
-     This function behaves as `:predicate', except that it's only
+     This property behaves as `:predicate', except that it's only
      called the first time a syntax checker is to be used in a buffer.
 
-     If the function returns a non-nil value the checker is put in a
-     whitelist in `flycheck-enabled-checkers' to prevent further
-     invocations of `:enabled'.  Otherwise it is disabled via
-     `flycheck-disabled-checkers' to prevent any further use of it.
-
-     FUNCTION is called with CHECKER argument and shall return
+     FUNCTION is called without arguments and shall return
      non-nil if this syntax checker shall be used to check the
      current buffer.  Otherwise it shall return nil.
+
+     If FUNCTION returns a non-nil value the checker is put in a
+     whitelist in `flycheck-enabled-checkers' to prevent further
+     invocations of `:enabled'.  Otherwise it is disabled via
+     `flycheck-disabled-checkers' to prevent any further use of
+     it.
+
+     If this checker has a `:working-directory' FUNCTION is
+     called with `default-directory' bound to the checker's
+     working directory.
 
      FUNCTION is only called in matching major modes.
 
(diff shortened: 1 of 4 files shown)


lunaryorn:
I did cursory testing, and things seem to work fine, but I'd be happy if someone else could also take a look :blush:

@Simplify What do you think?

flycheck.el
```diff
@@ -7518,9 +7535,8 @@ See URL `https://github.com/eslint/eslint'."
             (flycheck-sanitize-errors errors))
     errors)
   :enabled
-  (lambda (checker)
-    (let* ((default-directory (flycheck-compute-working-directory checker))
-           (executable (flycheck-find-checker-executable checker))
+  (lambda ()
+    (let* ((executable (flycheck-find-checker-executable 'javascript-eslint))
```
cpitclaudel: Did you really mean to remove the default-directory here?  Will the call to `executable` return the same thing?

cpitclaudel:
LGTM :) I think this is a nice change. I was a bit reluctant about the scope creep of `:enabled` at first, but this does seem like the right approach.

flycheck.el
cpitclaudel: Ugh, sorry. Makes sense now.

cpitclaudel:
> Additionally, :enabled takes no parameters anymore

Neat :)

Simplify:
@lunaryorn it looks great. I was trying to think about situation when you don't want to change `default-directory` to the location of file in the buffer. Only thing that I can think of is new elixir checker that we got few weeks ago, that requires "project" directory to call `mix credo` or `mix dogma` (can't remember witch one is now in flycheck). On the other hand it does not uses `:enabled` yet and if ever required `default-directory` can be set again inside `:enabled`.
I don't have time right now to test of this works with installed `eslint`, but will have time tomorrow. This solves current crashes when `eslint` is not installed, so you don't need to wait for me, it's better than current melpa unstable package :) LGTM :)

lunaryorn:
@Simplify  Thanks for your comment, LGTM for me too then.

I guess we should refactor the Elixir syntax checker at some point to use `:enabled`, since it makes no sense to check for the project definition every single time, as it's not going to change that frequently.  But that's for another time, /cc @flycheck/elixir

Simplify:
@lunaryorn I just tested this with current Melpa unstable package with several JS projects, with or without ESLint configuration. It works. Not only that, it's great that I don't need to disable eslint checker in projects `.dir-locals.el` anymore for projects where I don't use ESLint, for example Ember projects that have JSHint integration ;)

PR #1119, opened 2016-09-28
Fix error parsing phpcs output when not using phpcs configuration defaults

ftzdomino:
Use quiet mode for phpcs to prevent errors parsing output when phpcs has been configured with show_progress enabled.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5337,16 +5337,12 @@ function `buffer-file-name'."
       (when (seq-some (apply-partially #'flycheck-same-files-p
                                        (expand-file-name filename cwd))
                       buffer-files)
-        (let ((new-filename (buffer-file-name)))
-          (setf (flycheck-error-filename err) new-filename)
-          (when new-filename
-            (setf (flycheck-error-message err)
-                  (replace-regexp-in-string
-                   (regexp-quote filename)
-                   new-filename
-                   (flycheck-error-message err)
-                   'fixed-case
-                   'literal)))))))
+        (setf (flycheck-error-filename err) buffer-file-name)
+        (when (and buffer-file-name (flycheck-error-message err))
+          (setf (flycheck-error-message err)
+                (replace-regexp-in-string
+                 (regexp-quote filename) buffer-file-name
+                 (flycheck-error-message err) 'fixed-case 'literal))))))
   err)
 
 


cpitclaudel:
Hi there,

Thanks for this contribution! It looks neat.  Let me check that I understand correctly: when show_progress is set (say, in your personal config), then phpcs will produce extra output, and flycheck chokes on that output.  The -q flag overrides your local config to silence that extra output. Is that right?

If so, then LGTM; this sounds like a neat change.  It would be nice to include the info above in the commit message.

lunaryorn:
Thank you for this change, but please also add a comment that explain the rationale for this flag to the source and write a proper commit message.

lunaryorn:
@ftzdomino Hey, did you receive our comments?

ftzdomino:
First, thank you for being so responsive with this project and sorry for the delay.  I'll try to make some time soon to rewrite the commit message into your desired format.

The issue that this addresses is that when 
`phpcs --config-set show_progress 1`
has been run, it enables show_progress on all future runs.  
So when the command
`phpcs --report=checkstyle --standard=standardname index.php`
is run, the following is output:

```
.

<?xml version="1.0" encoding="UTF-8"?>
<checkstyle version="2.6.2">
</checkstyle>
Time: 170ms; Memory: 5.5Mb
```


Adding -q restores the output to the plain XML that flycheck expects:
`<?xml version="1.0" encoding="UTF-8"?>
<checkstyle version="2.6.2">
</checkstyle>`

lunaryorn:
@ftzdomino Oh dear.  It would appear that we should really add this flag then, but could you just amend this pull request to add a comment that summarises your comment and links to this issue so that in future we can easily understand the rationale for this flag?

lunaryorn:
@ftzdomino Hey, what is the status of this?

lunaryorn:
@ftzdomino LGTM, but could you just squash the commits into a single one, please?  See our [Style Guide](http://www.flycheck.org/en/latest/contributor/style-guide.html#commits) about commits :relaxed:

ftzdomino:
Sorry for all the trouble, I'm pretty inexperienced with git/github.  I've got everything in a single commit, but for some reason it won't go into the current pull request.  Should I just submit a new pull request with this commit:
https://github.com/ftzdomino/flycheck/commit/114cec9d1744122dbc9160e48dba5dbb71f057f3

cpitclaudel:
I think you could reopen this pull request and then force-push :)

lunaryorn:
@cpitclaudel I'm not sure… the OP's change is on the proper branch, and still doesn't show up here.  I think Github messed something up.

cpitclaudel:
Indeed, thanks @ftzdomino !

PR #1128, opened 2016-10-04
Fix rust macro errors in JSON output for rust checkers

fmdkdd:
This was reported in flycheck/flycheck-rust#36.

Macro expansion errors in the JSON output for the rust checkers use a
slightly different format than other errors.  As a result, the current
parser does not report macro expansion errors in flycheck, even though
they are present in the JSON output.

This commit enhances the parser for the rust JSON output by adding the
extra logic needed to capture the macro expansion errors.

Macro expansion errors have an 'expansion' field that contains another
'span', and this 'span' can contain an 'expansion', and so on.  So
in order to collect all the spans in a diagnostic, we need not only to
do a breadth search on the list of spans, but also a depth search on the
'expansion' field.

This search is delegated to a new function
`flycheck-parse-rust-collect-spans` that recursively collects all spans
in a 'root' span object.

Spans of macro expansion errors do not have any label, so we reuse the
text from the error message itself.

I've also added two other commit to (1) fix a test (2) add a test for macro errors.


lunaryorn:
LGTM, @cpitclaudel?

lunaryorn:
@fmdkdd Thanks, particularly for the detailed commit messages and the comments.  Really helps us to understand what's going on, even in six months from now :+1:

fmdkdd:
@lunaryorn Well, I haven't found a simpler way to deal with this JSON output in code form, so commenting is the least I could do here ;)

flycheck.el
```diff
@@ -5534,6 +5534,23 @@ about TSLint."
...
                   (json-read-from-string output)))))
 
+(defun flycheck-parse-rust-collect-spans (span)
+  "Return a list of spans contained in a SPAN object."
+  (let ((spans))
+    (let-alist span
+      ;; With macro expansion errors, some spans will point to phony file names
+      ;; to indicate an error inside the std rust lib.  We skip these spans as
+      ;; they won't appear in flycheck anyway.
+      (when (not (string= .file_name "<std macros>"))
```
cpitclaudel: I think `unless` would read better than `when not` :)

fmdkdd:
@lunaryorn @cpitclaudel Should be good to go.  Want me to rebase or squash?

lunaryorn:
@fmdkdd Squash, please :)

fmdkdd:
@lunaryorn Done.  Should be good to go once the CI greenlights.

PR #1144, opened 2016-10-21
Added erlang-rebar3 checker

joedevivo:
I wrote a flycheck checker for erlang projects using rebar3, but when I tried adding it to melpa, they said it was pretty basic, so I should try including it in flycheck, so here I am :D

As per the suggestion here: https://github.com/melpa/melpa/pull/4342


lunaryorn:
@joedevivo Thank you very much.  I'll leave a first review.

flycheck.el
```diff
@@ -6920,6 +6921,25 @@ See URL `http://www.erlang.org/'."
    (error line-start (file-name) ":" line ": " (message) line-end))
   :modes erlang-mode)
 
+
+(flycheck-define-checker erlang-rebar3
+  "An Erlang syntax checker using the rebar3 build tool.
+
+See URL `https://www.rebar3.org'."
+  :command ("rebar3"
+            "compile")
```
lunaryorn: Please keep on a single line.

flycheck.el
```diff
@@ -6920,6 +6921,25 @@ See URL `http://www.erlang.org/'."
...
+  ((warning line-start
+            (file-name) ":" line ": Warning:" (message) line-end)
+   (error line-start
+          (file-name) ":" line ": " (message) line-end))
+  :modes erlang-mode
+  :working-directory flycheck-rebar3-project-root)
+
+(defun flycheck-rebar3-project-root (_checker)
+  "Return directory where =rebar.config= is located."
+  (locate-dominating-file buffer-file-name "rebar.config"))
```
lunaryorn: Does `rebar` work if `rebar.config` is missing?  If not it needs an `:enabled` function that checks whether the file is actually present.

flycheck.el
```diff
@@ -6920,6 +6921,24 @@ See URL `http://www.erlang.org/'."
    (error line-start (file-name) ":" line ": " (message) line-end))
   :modes erlang-mode)
 
+
+(flycheck-define-checker erlang-rebar3
+  "An Erlang syntax checker using the rebar3 build tool.
+
+See URL `https://www.rebar3.org'."
+  :command ("rebar3" "compile")
```
lunaryorn: I presume the syntax checker compiles the entire project? The it needs a :predicate that checks whether the current buffer is saved because in this case Flycheck won't be able to feed the actual contents of the buffer to the checker, so if the buffer is not saved the errors will not be accurate.

joedevivo:
@lunaryorn I just installed today's flycheck build and it's generating some errors:

```
Compiling file /Users/joe/.emacs.d/elpa/flycheck-20161023.643/flycheck-ert.el at Sun Oct 23 08:44:11 2016
Entering directory ‘/Users/joe/.emacs.d/elpa/flycheck-20161023.643/’

Compiling file /Users/joe/.emacs.d/elpa/flycheck-20161023.643/flycheck.el at Sun Oct 23 08:44:11 2016
flycheck.el:2610:1:Warning: Unused lexical argument ‘cwd’

In flycheck-finish-current-syntax-check:
flycheck.el:2630:20:Warning: reference to free variable ‘working-dir’

In flycheck-handle-idle-change:
flycheck.el:2748:8:Warning: function flycheck-handle-idle-change used to take
    0+ arguments, now takes 0

In end of data:
flycheck.el:9343:1:Warning: the function
    ‘flycheck-syntax-check-current-checker’ is not known to be defined.

Compiling no file at Sun Oct 23 08:44:11 2016
```

It's not related to this, but I figured you should know

lunaryorn:
@joedevivo Thanks for making me aware, but that's already been fixed in f112d0cbb8bb989fd1c0eb7bc83dad2aac44c456.   Please don't hijack discussions for unrelated topics.  Instead just open a new issue.  Thank you.

flycheck.el
```diff
@@ -6925,6 +6925,13 @@ See URL `http://www.erlang.org/'."
 (flycheck-define-checker erlang-rebar3
   "An Erlang syntax checker using the rebar3 build tool.
 
+This flychecker compiles the entire rebar3 project, which
+compensates for contextual errors that the regular erlang checker
+wouldn't be aware of. One such example is the erlang -include
+directive. Rebar3 is aware of the context for those includes,
+where the standard erlang checker is not. This checker should be
+used instead of the erlang checker for all rebar3 projects
```
lunaryorn: Thanks but I didn't mean to clarify that in the docstring, but rather write a comment on the Github issue :blush:  Please remove this rather lengthy paragraph again.

flycheck.el
```diff
@@ -6933,8 +6940,17 @@ See URL `https://www.rebar3.org'."
    (error line-start
           (file-name) ":" line ": " (message) line-end))
   :modes erlang-mode
+  :enabled (lambda () (flycheck-rebar-config-exists-p))
+  :predicate (lambda () (flycheck-buffer-saved-p))
   :working-directory flycheck-rebar3-project-root)
 
+(defun flycheck-rebar-config-exists-p ()
+  "Whether there is a rebar.config for the current buffer."
+  (let* ((executable (flycheck-find-checker-executable 'erlang-rebar3))
```
lunaryorn: Is there a reason why you check that?  Are you aware that Flycheck does that automatically?

joedevivo:
I'm happy with this pull request except for one thing. My tests are failing, and it's because the rebar3 tests aren't saving the buffer. I can't figure out how to make it do this. Any suggestion?

lunaryorn:
@joedevivo Can you show the output of the test?

And do we really need all those files?  The separate LICENSE is a no-go, in any case.

joedevivo:
@lunaryorn yes sorry, I'll take care of that and set it up so it's just the minimum rebar3 needs to recognize a project.

Here's my error output:

```
➜  flycheck git:(master) ✗ make LANGUAGE=erlang integ
cask exec emacs -Q --batch -L .  --load test/flycheck-test --load test/run.el -f flycheck-run-tests-main '(and (tag external-tool) (language erlang))'
Running tests on Emacs 25.1.1, built at 2016-09-17
Running 4 tests (2016-10-23 11:32:02-0700)
Loading erlang-skels...
Test flycheck-define-checker/erlang-rebar3/error backtrace:
  flycheck-ert-should-errors((7 nil error "head mismatch" :checker erl
  apply(flycheck-ert-should-errors (7 nil error "head mismatch" :check
  flycheck-ert-should-syntax-check("language/erlang-rebar3/src/erlang-
  (closure (python-indent-guess-indent-offset js3-mode-show-parse-erro
  ert--run-test-internal([cl-struct-ert--test-execution-info [cl-struc
  ert-run-test([cl-struct-ert-test flycheck-define-checker/erlang-reba
  ert-run-or-rerun-test([cl-struct-ert--stats (and "flycheck-" (and (t
  ert-run-tests((and "flycheck-" (and (tag external-tool) (tag languag
  ert-run-tests-batch((and "flycheck-" (and (tag external-tool) (tag l
  ert-run-tests-batch-and-exit((and "flycheck-" (and (tag external-too
  (let ((selector (car-safe (prog1 argv (setq argv (cdr argv)))))) (if
  flycheck-run-tests-batch-and-exit()
  (let ((debug-on-error t)) (flycheck-run-tests-batch-and-exit))
  (let* ((load-prefer-newer t) (source-directory (locate-dominating-fi
  flycheck-run-tests-main()
  command-line-1(("-L" "." "--load" "test/flycheck-test" "--load" "tes
  command-line()
  normal-top-level()
Test flycheck-define-checker/erlang-rebar3/error condition:
    (ert-test-failed
     ((should
       (equal expected flycheck-current-errors))
      :form
      (equal
       ([cl-struct-flycheck-error #<killed buffer> erlang-rebar3 "/Users/joe/dev/joedevivo/flycheck/test/resources/language/erlang-rebar3/src/erlang-error.erl" 7 nil "head mismatch" error nil])
       nil)
      :value nil :explanation
      (different-types
       ([cl-struct-flycheck-error #<killed buffer> erlang-rebar3 "/Users/joe/dev/joedevivo/flycheck/test/resources/language/erlang-rebar3/src/erlang-error.erl" 7 nil "head mismatch" error nil])
       nil)))
   FAILED  1/4  flycheck-define-checker/erlang-rebar3/error
Test flycheck-define-checker/erlang-rebar3/warning backtrace:
  flycheck-ert-should-errors((6 nil warning "wrong number of arguments
  apply(flycheck-ert-should-errors (6 nil warning "wrong number of arg
  flycheck-ert-should-syntax-check("language/erlang-rebar3/src/erlang-
  (closure (python-indent-guess-indent-offset js3-mode-show-parse-erro
  ert--run-test-internal([cl-struct-ert--test-execution-info [cl-struc
  ert-run-test([cl-struct-ert-test flycheck-define-checker/erlang-reba
  ert-run-or-rerun-test([cl-struct-ert--stats (and "flycheck-" (and (t
  ert-run-tests((and "flycheck-" (and (tag external-tool) (tag languag
  ert-run-tests-batch((and "flycheck-" (and (tag external-tool) (tag l
  ert-run-tests-batch-and-exit((and "flycheck-" (and (tag external-too
  (let ((selector (car-safe (prog1 argv (setq argv (cdr argv)))))) (if
  flycheck-run-tests-batch-and-exit()
  (let ((debug-on-error t)) (flycheck-run-tests-batch-and-exit))
  (let* ((load-prefer-newer t) (source-directory (locate-dominating-fi
  flycheck-run-tests-main()
  command-line-1(("-L" "." "--load" "test/flycheck-test" "--load" "tes
  command-line()
  normal-top-level()
Test flycheck-define-checker/erlang-rebar3/warning condition:
    (ert-test-failed
     ((should
       (equal expected flycheck-current-errors))
      :form
      (equal
       ([cl-struct-flycheck-error #<killed buffer> erlang-rebar3 "/Users/joe/dev/joedevivo/flycheck/test/resources/language/erlang-rebar3/src/erlang-warning.erl" 6 nil "wrong number of arguments in format call" warning nil])
       nil)
      :value nil :explanation
      (different-types
       ([cl-struct-flycheck-error #<killed buffer> erlang-rebar3 "/Users/joe/dev/joedevivo/flycheck/test/resources/language/erlang-rebar3/src/erlang-warning.erl" 6 nil "wrong number of arguments in format call" warning nil])
       nil)))
   FAILED  2/4  flycheck-define-checker/erlang-rebar3/warning
   passed  3/4  flycheck-define-checker/erlang/error
   passed  4/4  flycheck-define-checker/erlang/warning

Ran 4 tests, 2 results as expected, 2 unexpected (2016-10-23 11:32:06-0700)

2 unexpected results:
   FAILED  flycheck-define-checker/erlang-rebar3/error
   FAILED  flycheck-define-checker/erlang-rebar3/warning

make: *** [integ] Error 1
```

joedevivo:
I don't think I'll be able to act much more on this for the rest of the day, but any suggestions you have on the test front, I'll be happy to try out tonight or tomorrow, as well as finishing up the other recommendations here.

Thanks for taking the time today

lunaryorn:
Me neither, I'll take a look at the test stuff tomorrow (my time :wink:)

flycheck.el
```diff
@@ -5440,9 +5441,12 @@ ERR is in BUFFER-FILES, replace it with the return value of the
...
     (-when-let (filename (flycheck-error-filename err))
-      (when (seq-some (apply-partially #'flycheck-same-files-p
-                                       (expand-file-name filename cwd))
-                      buffer-files)
+      (when (or (seq-some (apply-partially #'flycheck-same-files-p
+                                           (expand-file-name filename cwd))
+                          buffer-files)
+                ; If buffer-files is nil, it just means we didn't use
+                ; any temp files to compile this
+                (not buffer-files))
```
joedevivo: I added this or condition to make the tests pass. Since I wasn't building with any temporary files, `buffer-files` was always coming through `nil`, so `seq-some` would always be false. This resulted in no errors making it back to the assertion. It's weird that this worked fine in real world usage.

joedevivo:
Confirming it still works in regular usage

flycheck.el
```diff
@@ -5440,9 +5441,10 @@ ERR is in BUFFER-FILES, replace it with the return value of the
 function `buffer-file-name'."
   (flycheck-error-with-buffer err
     (-when-let (filename (flycheck-error-filename err))
-      (when (seq-some (apply-partially #'flycheck-same-files-p
-                                       (expand-file-name filename cwd))
-                      buffer-files)
+      (when (or (seq-some (apply-partially #'flycheck-same-files-p
+                                           (expand-file-name filename cwd))
+                          buffer-files)
+                (not buffer-files))
```
joedevivo: I added this or condition to make the tests pass. Since I wasn't building with any temporary files, `buffer-files` was always coming through `nil`, so `seq-some` would always be false. This resulted in no errors making it back to the assertion. It's weird that this worked fine in real world usage.

test/resources/language/erlang-rebar3/rebar.lock
```diff
@@ -0,0 +1 @@
+[].
```
lunaryorn: Is this file required?

test/resources/language/erlang-rebar3/rebar.config
```diff
@@ -0,0 +1,5 @@
+{erl_opts, 
+	[
+		return_warnings
+	]}.
+{deps, []}.
```
lunaryorn: Is this file required?

flycheck.el
lunaryorn: @joedevivo I'm sorry but that's not the solution: By adding this condition you're rewriting file names of _all_ errors of syntax checkers that didn't create any temporary files.  That's going to make a mess of many other checkers.

Please remove this change again!

lunaryorn:
@joedevivo LGTM, generally, but we need to find a different way to get the tests pass.

It looks as if Flycheck doesn't get absolute file names.  Could you perhaps show the output of the rebar checker if you visit a file from the test project and then run it with `C-c ! C-c erlang-rebar3`.

Please make sure to revert the change to `flycheck-fix-error-filename` first.

test/resources/language/erlang-rebar3/rebar.lock
joedevivo: It will be generated by every build, but I can easily put it in the .gitignore

test/resources/language/erlang-rebar3/rebar.config
joedevivo: Yes! The checker determines its ability to run based on the presence of this file.

test/resources/language/erlang-rebar3/rebar.lock
lunaryorn: Yes please.  The less files we need the better.

joedevivo:
Since the `erlang-rebar3` checker looks for a `rebar.config` file in a parent directory,  I will have to move the files for the regular `erlang` checher to `erlang/erlang` and then move `erlang-rebar3` to `erlang/erlang`

joedevivo:
Here's the output from `C-c ! C-c erlang-rebar3`

```
-*- mode: compilation; default-directory: "~/dev/joedevivo/flycheck/test/resources/language/erlang/rebar3/" -*-
Compilation started at Thu Oct 27 04:27:45

rebar3 compile
===> Verifying dependencies...
===> Compiling erlang-rebar3
src/erlang-warning.erl:6: Warning: wrong number of arguments in format call

===> Compiling src/erlang-error.erl failed
src/erlang-error.erl:7: head mismatch


Compilation exited abnormally with code 1 at Thu Oct 27 04:27:45
```

but also, if it makes any difference, the output is actually colored:

screenshot 2016-10-27 04 28 47

joedevivo:
If it helps, here is the `erlang` checker run with `C-c ! C-c erlang` on the file in `erlang/erlang/error.erl`

```
-*- mode: compilation; default-directory: "~/dev/joedevivo/flycheck/test/resources/language/erlang/erlang/" -*-
Compilation started at Thu Oct 27 04:31:52

erlc -o /var/folders/bv/n_692_k15qq3b7sb2vrnd_qm0000gn/T/flycheck9772AiT -Wall /Users/joe/dev/joedevivo/flycheck/test/resources/language/erlang/erlang/error.erl
error.erl:7: head mismatch

Compilation exited abnormally with code 1 at Thu Oct 27 04:31:52
```

It's also colored output, so it seems like that's probably not an issue

lunaryorn:
@joedevivo Did you explicitly select `erlang-rebar3` in that buffer?  Normally it shouldn't get use because the plain `erlang` checker comes first…

flycheck.el
```diff
@@ -181,6 +181,7 @@ attention to case differences."
     emacs-lisp
     emacs-lisp-checkdoc
     erlang
+    erlang-rebar3
```
lunaryorn: Please move erlang before `erlang-rebar3`.

joedevivo:
The plain `erlang` checker is disabled via ` :enabled (lambda () (not (flycheck-rebar-config-exists-p)))`

flycheck.el
```diff
@@ -6918,8 +6919,40 @@ See URL `http://www.erlang.org/'."
   :error-patterns
   ((warning line-start (file-name) ":" line ": Warning:" (message) line-end)
    (error line-start (file-name) ":" line ": " (message) line-end))
+  :enabled (lambda () (not (flycheck-rebar-config-exists-p)))
```
lunaryorn: Please remove this `:enabled`… it prevents users from explicitly selecting the plain `erlang` checker in rebar projects even if they want to.  Just move `erlang-rebar3` before `erlang` and Flycheck will automatically prefer the rebar checker.

doc/languages.rst
```diff
@@ -319,6 +319,10 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
          A list of library directories.
 
+   .. syntax-checker:: erlang-rebar3
+
+      Check Erlang with the `rebar3 <https://www.rebar3.org/>`_ build tool.
```
lunaryorn: Please also add a paragraph right below `..supported-language:: erlang` that explains the order in which syntax checkers are applied, e.g. ``Flycheck checks erlang with `erlang-rebar3` in rebar projects or `erlang` otherwise.``

flycheck.el
```diff
@@ -6918,8 +6919,40 @@ See URL `http://www.erlang.org/'."
...
+   (error line-start
+          (file-name) ":" line ": " (message) line-end))
+  :modes erlang-mode
+  :enabled (lambda () (flycheck-rebar-config-exists-p))
+  :predicate (lambda () (flycheck-buffer-saved-p))
+  :working-directory flycheck-rebar3-project-root)
+
+(defun flycheck-rebar-config-exists-p ()
+  "Whether there is a rebar.config for the current buffer."
+  (locate-dominating-file buffer-file-name "rebar.config"))
```
lunaryorn: Please remove this extra function, and just use `flycheck-rebar3-project-root` in `:enabled`; you can use `&optional _checker` to be able to invoke the function without specifying a checker.  There's no need to duplicate the same code.

Please also move the function _before_ the definition of the syntax checker.

flycheck.el
lunaryorn: @joedevivo Can you please remove this paragraph?

lunaryorn:
@joedevivo Just noticed that; I've left another review because I think it's better to sort out priorities in a different way—users should still be able to fall back to the normal erlang checker if the want to, and `:enabled` would prevent that.

I'm not sure why the tests should fail.

lunaryorn:
@joedevivo Can you ping me when you've applied all the changes I asked for?  I'll checkout your PR then and try to run the tests myself… maybe I can find out what makes them fail.

joedevivo:
@lunaryorn I just pushed the changes.

test/flycheck-test.el
```diff
@@ -3044,14 +3044,26 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
...
+  (flycheck-ert-should-syntax-check
+   "language/erlang/rebar3/src/erlang-error.erl" 'erlang-mode
+   '(6 nil warning "wrong number of arguments in format call" :checker erlang-rebar3)
+   '(7 nil error "head mismatch" :checker erlang-rebar3)))
+
+(flycheck-ert-def-checker-test erlang-rebar3 erlang warning
+  (flycheck-ert-should-syntax-check
+   "language/erlang/rebar3/src/erlang-warning.erl" 'erlang-mode
+   '(6 nil warning "wrong number of arguments in format call" :checker erlang-rebar3)
+   '(7 nil error "head mismatch" :checker erlang-rebar3)))
```
lunaryorn: @joedevivo I'm aware of that but you should still see only errors for the current buffer?  If that was really the expected outcome of the test _all_ errors from _all_ files in the whole project would be highlighted in _all_ erlang buffers for the project, and that's clearly not desirable :)

joedevivo: Sorry, I wasn't sure what the expected value should be. The failures that are a problem return `nil`.

lunaryorn:
I've tested this branch and found out the syntax checker definition parses invalid filenames—the filenames contain invisible ASCII control characters, so Flycheck fails to match the parsed filenames to the actual buffer filename.

Specifically, that's the actual filename that gets parsed by the syntax checker definition: `[0msrc/erlang-warning.erl`—note the control sequence at the beginning of the filename.  I guess that's the "color reset" sequence from rebar which is printed on the next line, and thus comes after `line-start`.

That's not particularly clever behaviour of rebar3—it should not output color sequences when writing to a pipe—but maybe there's a flag to stop rebar3 from writing coloured output?

joedevivo:
I was suspicious of the coloring. I'll investigate. In the mean time, any idea why that doesn't cause a problem with the checker when I'm actually using it?

lunaryorn:
@joedevivo I don't know… you're running the tests on the terminal so maybe rebar3 somehow thinks that it's running on a TTY and enables colouring.  Note that in the `C-c ! C-c` output you've showing above (which was run in a GUI Emacs I presume) doesn't contain ASCII control sequences.

joedevivo:
Good news, the `TERM=dumb` ENV var will suppress all color and now the tests pass! It's definitely the colors. How would we go about adding that to the checker?

```
➜  flycheck git:(master) ✗ TERM=dumb make LANGUAGE=erlang integ
cask exec emacs -Q --batch -L .  --load test/flycheck-test --load test/run.el -f flycheck-run-tests-main '(and (tag external-tool) (language erlang))'
Running tests on Emacs 25.1.1, built at 2016-09-17
Running 4 tests (2016-10-27 06:58:00-0700)
Loading erlang-skels...
   passed  1/4  flycheck-define-checker/erlang-rebar3/error
   passed  2/4  flycheck-define-checker/erlang-rebar3/warning
   passed  3/4  flycheck-define-checker/erlang/error
   passed  4/4  flycheck-define-checker/erlang/warning

Ran 4 tests, 4 results as expected (2016-10-27 06:58:04-0700)
```

lunaryorn:
@joedevivo I would rather prefer to explicitly disable colored output in rebar upfront, in the syntax checker definition…

joedevivo:
@lunaryorn me too, but am I correct that the first string in the `:command` list needs to be the executable to look for on the `$PATH`? if not we could `:command ("TERM=dumb" "rebar3" "compile)`

lunaryorn:
@joedevivo Yes that's correct but that's not what I meant.  I hoped that `rebar3` had a flag for that, e.g. `rebar3 --no-color compile` or so…

joedevivo:
That's what I was hoping for as well, but even if we were to add something, it would still be risky behavior for older versions of `rebar3`, right? 

When I run `emacs` in a terminal, neither `erlang` or `erlang-rebar3` show the errors in the source code buffer, but both report the correct results with `C-c ! C-c <checker>`

lunaryorn:
> That's what I was hoping for as well, but even if we were to add something, it would still be risky behavior for older versions of rebar3, right?

We'd simply require a recent version of rebar3 then.  We've been doing that with other checkers as well.

> When I run emacs in a terminal, neither erlang or erlang-rebar3 show the errors in the source code buffer, but both report the correct results with C-c ! C-c <checker>

Not sure what you mean?  On Flycheck master `make LANGUAGE=erlang integ` succeeds if I install erlang.

joedevivo:
Yes, the make command works. I'm talking about running `emacs` and trying to check the file in the terminal version of `emacs` as opposed to the GUI version.

joedevivo:
I've been told that `rebar3` can't take global arguments, because the set of arguments are passed to each individual task.

lunaryorn:
@joedevivo That's more an issue of faces then… there are no other significant differences between GUI and terminal Emacs.

> I've been told that rebar3 can't take global arguments, because the set of arguments are passed to each individual task.

Well, `rebar3 compile --no-color` would be fine, too :wink:

joedevivo:
I found a function in the `ansi-color` package that's included with `emacs` that strips all the ANSI color codes out of the checker's output. The `LANGUAGE=erlang` tests all pass with this change

lunaryorn:
@joedevivo That is a very invasive change for a single syntax checker… I presume there's really no other way that we can prevent rebar3 from printing colours?  Did you ask the rebar3 maintainers?

joedevivo:
@lunaryorn I spoke with a maintainer, and the `TERM=dumb` solution seems to satisfy any need they have for color suppression. I figured this change to flycheck would not only solve the problem for me, but for anybody in the future that runs into the same issue with another checker that uses ANSI color

lunaryorn:
@joedevivo That is disappointing; no other checker behaves this way.  They should _at least_ check whether the output goes to a TTY in addition to looking at `$TERM`.

That said, any solution that preprocesses the output before individual syntax checkers even get the chance to look at it—which your approach does—is not an option; it's bound to create problems on the long run.

We've got two options now:

1. Add an `:environment` property to syntax checkers that allows syntax checkers to extend their `process-environment`, and add a `:environment (("TERM" . "dumb"))` to the rebar checker.
2. Add a custom `:error-parser` to rebar3 that strips ANSI colour sequences and then calls `flycheck-parse-with-patterns` with the stripped output.

2) would be simpler, but 1) would be a great addition to Flycheck generally.

We need either to be able to merge this syntax checker into Flycheck, so would you be willing to take one?  Which one would be your choice…

joedevivo:
I love the first solution, but I'm not sure I have the `elisp` chops to get it done. I might tackle it a little further down the line as I learn more. It seems like a meaty enough project to learn, and I understand the scope of the problem well.

I feel confident that I can get the second option done soon, and get this PR merged sooner, so I'm happy to take on that approach today. Are you ok with the `(require 'ansi-color)`? If so, I can turn this around pretty quickly.

lunaryorn:
@joedevivo Great, then please go ahead with option 2 👍  Please don't hesitate to ask any questions, I'll gladly help you :)

 The `(require)` is absolutely ok for now.  We can put it into the function body later to load it lazily but that gets a little ugly if one wants to avoid byte compiler warnings, so let's do that later, when the code's done, if you are ok with doing another iteration then.

joedevivo:
@lunaryorn cool. I've already pushed the change.

joedevivo:
@lunaryorn oh no, a unit test is failing:

```
1 unexpected results:
   FAILED  flycheck-checkers/should-have-either-patterns-or-parser
```

Before I go back to the drawing board, I wanted to ask you if you think it's ok that I solved this problem by having both `:error-parser` *AND* `:error-patterns`. If that's an OK fix, then maybe it's this unit test that needs to be changed.

lunaryorn:
@joedevivo You can delete the entire test; it's redundant.

flycheck.el
```diff
@@ -6920,6 +6922,27 @@ See URL `http://www.erlang.org/'."
...
+
+(defun flycheck-rebar3-project-root (&optional _checker)
+  "Return directory where =rebar.config= is located."
+  (locate-dominating-file buffer-file-name "rebar.config"))
+
+(flycheck-define-checker erlang-rebar3
+  "An Erlang syntax checker using the rebar3 build tool."
+  :command ("rebar3" "compile")
+  :error-parser
+  (lambda(output checker buffer)
```
lunaryorn: There's a space missing before the opening parenthesis :)

flycheck.el
```diff
@@ -6920,6 +6922,27 @@ See URL `http://www.erlang.org/'."
    (error line-start (file-name) ":" line ": " (message) line-end))
   :modes erlang-mode)
 
+
```
lunaryorn: This blank line is superfluous, please remove it.

flycheck.el
```diff
@@ -6924,7 +6924,16 @@ See URL `http://www.erlang.org/'."
...
   "Return directory where =rebar.config= is located."
-  (locate-dominating-file buffer-file-name "rebar.config"))
+  (let* ((rebarconfig (locate-dominating-file buffer-file-name "rebar.config"))
+         (project-dir (if rebarconfig
+                          (file-name-as-directory
+                           (substring rebarconfig
+                                      0
+                                      (string-match
+                                       "\\(/_build/[^/]+/lib\\|/_checkouts\\)"
+                                       rebarconfig)))
```
lunaryorn: @joedevivo Can you illustrate what this `substring` is supposed to do?

test/flycheck-test.el
```diff
@@ -3053,6 +3053,13 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
    "language/erlang/rebar3/src/erlang-warning.erl" 'erlang-mode
    '(6 nil warning "wrong number of arguments in format call" :checker erlang-rebar3)))
 
+(flycheck-ert-def-checker-test erlang-rebar3 erlang build
+  (flycheck-ert-should-syntax-check
+   "language/erlang/rebar3/_checkouts/dependency/src/dependency.erl" 'erlang-mode)
+  (should (not (file-exists-p
+                (flycheck-ert-resource-filename
+                 "language/erlang/rebar3/_build/default/lib/dependency/_build")))))
```
lunaryorn: What's this assertion for?  What does it test?  Why is it important this this file/directory does not exist?

flycheck.el
```diff
@@ -6920,6 +6922,35 @@ See URL `http://www.erlang.org/'."
...
+                        rebarconfig)))
+    project-dir))
+
+(flycheck-define-checker erlang-rebar3
+  "An Erlang syntax checker using the rebar3 build tool."
+  :command ("rebar3" "compile")
+  :error-parser
+  (lambda (output checker buffer)
+    (flycheck-parse-with-patterns (ansi-color-filter-apply output) checker buffer))
```
lunaryorn: Please add a little comment here that explains why we need to filter colours manually, and include a link to this pull request, so that we can easily find the discussion again when we look at this code in future.

flycheck.el
```diff
@@ -79,6 +79,7 @@
 (require 'help-mode)             ; `define-button-type'
 (require 'find-func)             ; `find-function-regexp-alist'
 (require 'json)                  ; `flycheck-parse-tslint'
+(require 'ansi-color)            ; strip ansi color from command output
```
lunaryorn: Please move this require into the body of the rebar3 `:error-parser` function.  That'll break the build due to a byte compiler warning, which we can then fix together.

flycheck.el
joedevivo: A `rebar3` project will check dependencies out into a `_build/<profile_name>/lib` folder. Your dependency will have `.erl` files, and you may want to open them in emacs to review what they're doing. If you do, `flycheck` will try and build that `rebar3` project, because it will satisfy the checker. 

Let's look at a real world example. https://github.com/joedevivo/chatterbox

Chatterbox has `src/*.erl` and we'll wan't those to work as the checker was already defined.

When I build chatterbox, it pulls dependencies into `_build/default/lib`, but also `_build/test/lib` which lets me include specific dependencies that I don't want bundled with the release, they're just there to test. If I open up any file in `_build/*/lib/*/src/*.erl`, it shouldn't try and compile that project on its own. That would wind up building `_build/<profile>/lib/<dependency>/_build/<profile>/lib/<dependency_of_the_first_depencency>` and so on. This regex will match places that rebar checks things out to, and helps us make sure we're only building the parent project.

`_checkouts` is a little different. Chatterbox is coupled to another repo I maintain: https://github.com/joedevivo/hpack. The development is done in tandem, and I can work on a master version of hpack by checking it out to `_checkouts`.

I hope this clarifies what that regex is doing here.

test/flycheck-test.el
joedevivo: Based on the comment above about the regex, this assertion is testing that when we open a buffer for a dependency, it doesn't build the dependency as if it were a standalone project. If it did, it would create that `_build` directory, so we're asserting that it doesn't.

joedevivo:
@lunaryorn I pushed up the changes, and yes, now there's a compile error.

flycheck.el
```diff
@@ -6920,6 +6920,41 @@ See URL `http://www.erlang.org/'."
...
+  :command ("rebar3" "compile")
+  :error-parser
+  (lambda (output checker buffer)
+    ;; rebar3 outputs ANSI terminal colors, which don't match up with
+    ;; :error-patterns, so we strip those color codes from the output
+    ;; here before passing it along to the default behavior. The
+    ;; relevant disucssion can be found
+    ;; https://github.com/flycheck/flycheck/pull/1144
+    (require 'ansi-color)
+    (flycheck-parse-with-patterns (ansi-color-filter-apply output) checker buffer))
```
lunaryorn: @joedevivo Thanks for the explanatory comment.

Now, to fix the compiler warning use the following as argument to `flycheck-parse-with-patterns`:

```elisp
(and (fboundp 'ansi-color-filter-apply) (ansi-color-filter-apply output))
```

The `fboundp` guard is entirely redundant, because we _know_ that `(require 'ansi-color)` brings that function in scope, but the byte compiler can't make that link.  However, the `fboundp` guard convinces it that we're only calling the function when it's defined and thus removes the warning.

Ugly, but works, and much evil is achieved 😎

test/flycheck-test.el
lunaryorn: @joedevivo Can you add a comment that explains this above this assertion?  E.g.

```
Ensure that the dependency file wasn't build as standalone project which would create a separate _build directory
```

would be sufficient (please wrap it to 80 characters though :) )

flycheck.el
```diff
@@ -6920,6 +6920,41 @@ See URL `http://www.erlang.org/'."
    (error line-start (file-name) ":" line ": " (message) line-end))
   :modes erlang-mode)
 
+(defun flycheck-rebar3-project-root (&optional _checker)
+  "Return directory where =rebar.config= is located."
+  (let* ((rebarconfig (locate-dominating-file buffer-file-name "rebar.config"))
```
lunaryorn: By the way, the name of this variable is wrong, I think.  Afaik `locate-dominating-file` returns the _directory_ containing the file, not the path to the file itself, so the variable doesn't actually contain the path to the rebar configuration file but rather the project directory.

So I guess `project-directory` is the better name here isn't it?

flycheck.el
lunaryorn: @joedevivo I see, but I'm not convinced that a regex is the right tool to use here, and I do not like string substitutions over pathnames.  That's just inherently fragile—for instance your current regex obviously breaks on windows pathnames.

How about the following algorithm instead:

1. Find the nearest `project-directory` containing `rebar.config` file with `locate-dominating-file`
2. From that file, find the nearest directory that contains a `_build` directory with `locate-dominating-file`.  If none exists return `project-directory` as the final working directory.
3. If one exists `let`-bind it to `default-directory` and then recursively call `flycheck-rebar3-project-root` again to restart the search for the top-most rebar project directory.

What do you think?

joedevivo: I like it. The issue with step 2 is that I need to look for `_build` or `_checkouts`; however,  there's an even better solution. We can just keep looking for `rebar.config`. Every `rebar3` project needs one, so a recursive call to `locate-dominating-file` would be enough. I've rust run into one snag: A recursive call to `locate-dominating-file` finds the one in `.`, so I need the recursive call to run on the parent directory, and I just can't figure out the function to give me the parent-dir, do you know how to do that? Thanks for the `elisp` lesson :D

joedevivo: I'm going to push with a solution using `(expand-file-name ".." possibly-nested-project-dir)`

lunaryorn: @joedevivo Mmh, I'm not sure whether that's not too eager: Is there no reason why anyone would ever nest two proper rebar projects under a top-level project?   Would you never use "nested" projects with rebar?

joedevivo: Chances are, if flycheck encounters this then it's a mistake on the part of the user. Something got built in a way it wasn't supposed to, or some artifact of some build survived somewhere it shouldn't have. No matter the case, the top `rebar.config` it can find is always the right answer. I think something less eager would still work 90% of the time, so if we're concerned about speed we could opt for something like that.

There's no nesting of projects with rebar. If you are doing something like, there's a way to structure your source so it will build multiple applications, but if you did that, it would still be done with a single `rebar.config` file and the applications would be bundled together into a single release artifact. 

The closest thing to what I think you're looking for would be if you had two rebar3 projects, and a third than included both of them as dependencies, but that's already covered.

Did that clear it up?

joedevivo: You're welcome! I'm happy to give back, and it's been fun learning more about elisp.

flycheck.el
```diff
@@ -6920,6 +6920,49 @@ See URL `http://www.erlang.org/'."
...
+(defun locate-rebar3-project-root (file-name)
+  "Return directory where the root =rebar.config= for the source FILE-NAME is located."
+  (let* ((possibly-nested-project-dir
+          (locate-dominating-file file-name "rebar.config"))
+         (possible-parent-dir
+          (if possibly-nested-project-dir
+              (locate-dominating-file
+               (expand-file-name ".." possibly-nested-project-dir)
+               "rebar.config")
+            nil)
```
lunaryorn: I've been thinking about whether we could simplify this code, to only call `locate-dominating-file` once.  Would that be possible?

joedevivo: Hopefully the new implementation looks good.

flycheck.el
```diff
@@ -80,7 +80,6 @@
 (require 'find-func)             ; `find-function-regexp-alist'
 (require 'json)                  ; `flycheck-parse-tslint'
 
-
```
lunaryorn: Why this change?

flycheck.el
```diff
@@ -6920,6 +6920,62 @@ See URL `http://www.erlang.org/'."
...
+non-nil of which will be our project root.
+
+Return the absolute path to the directory"
+  (if (string= file-name prev-file-name)
+      (car (seq-drop-while (lambda (x) (not x)) acc))
+    (let ((current-dir (file-name-directory file-name)))
+      (locate-rebar3-project-root (directory-file-name current-dir)
+                                  file-name
+                                  (cons (contains-rebar-config current-dir) acc))
+      )))
```
lunaryorn: Please don't let parenthesis hang on a line on their own.  Always join with the code before.

lunaryorn:
@joedevivo Ok, there are two minor stylistic issues left, but other than that I think we're ready now.

Would you like to join Flycheck as part of a new Erlang team to take care of Erlang support in Flycheck?  That'd really help us to maintain proper support for Erlang.

flycheck.el
joedevivo: accident

flycheck.el
joedevivo: done

joedevivo:
Done. Would you like me to squash the commits?

I'm happy to keep working on Erlang support as the need arises :D

lunaryorn:
@joedevivo Yup, please squash your commits.  

Thanks for joining us, I'll invite you to the team tonight, and approve the review.  Then you'll be able to merge this PR yourself 👍 

Meanwhile browse through our [Contributor's Guide](http://www.flycheck.org/en/latest/contributor/contributing.html) and our [Maintainer's Guide](http://www.flycheck.org/en/latest/contributor/maintaining.html) :blush:

joedevivo:
@lunaryorn  I squashed the commits a few days ago, but forgot to let you know. sorry about that.

joedevivo:
@lunaryorn hey, I haven't gotten the team invite, do you mind sending again?

Simplify:
@cpitclaudel Can you take look at this PR? I'll like to have it in Flycheck.

@joedevivo Sorry about delay. Are you still willing to support Erlang in Flycheck?

flycheck.el
```diff
@@ -6920,6 +6921,62 @@ See URL `http://www.erlang.org/'."
    (error line-start (file-name) ":" line ": " (message) line-end))
   :modes erlang-mode)
 
+(defun contains-rebar-config (dir-name)
+  "Return DIR-NAME if DIR-NAME/rebar.config exists, nil otherwise."
+  (if (file-exists-p (expand-file-name "rebar.config" dir-name))
+      dir-name nil))
+
+(defun locate-rebar3-project-root (file-name &optional prev-file-name acc)
```
cpitclaudel: This seems to be essentially `locate-dominating-file`, but looking for the topmost file.  Why is that needed?

flycheck.el
```diff
@@ -6920,6 +6921,62 @@ See URL `http://www.erlang.org/'."
    (error line-start (file-name) ":" line ": " (message) line-end))
   :modes erlang-mode)
 
+(defun contains-rebar-config (dir-name)
+  "Return DIR-NAME if DIR-NAME/rebar.config exists, nil otherwise."
+  (if (file-exists-p (expand-file-name "rebar.config" dir-name))
```
cpitclaudel: This should be `when` or `and` instead of `if`

flycheck.el
cpitclaudel: This seems to be essentially `locate-dominating-file`, but looking for the topmost file.  Why is that needed?

flycheck.el
```diff
@@ -6920,6 +6921,62 @@ See URL `http://www.erlang.org/'."
...
+FILE-NAME is a source file for which to find the project.
+
+PREV-FILE-NAME helps us prevent infinite looping
+
+ACC is an accumulator that keeps the list of results, the first
+non-nil of which will be our project root.
+
+Return the absolute path to the directory"
+  (if (string= file-name prev-file-name)
+      (car (seq-drop-while (lambda (x) (not x)) acc))
```
cpitclaudel: No need for seq-drop-while here; it should be enough to use `(remove nil …)`

flycheck.el
```diff
@@ -6920,6 +6921,62 @@ See URL `http://www.erlang.org/'."
...
+
+(flycheck-define-checker erlang-rebar3
+  "An Erlang syntax checker using the rebar3 build tool."
+  :command ("rebar3" "compile")
+  :error-parser
+  (lambda (output checker buffer)
+    ;; rebar3 outputs ANSI terminal colors, which don't match up with
+    ;; :error-patterns, so we strip those color codes from the output
+    ;; here before passing it along to the default behavior. The
+    ;; relevant disucssion can be found
```
cpitclaudel: `found` → `found at`

flycheck.el
```diff
@@ -6920,6 +6921,62 @@ See URL `http://www.erlang.org/'."
...
+    ;; https://github.com/flycheck/flycheck/pull/1144
+    (require 'ansi-color)
+    (flycheck-parse-with-patterns
+     (and (fboundp 'ansi-color-filter-apply) (ansi-color-filter-apply output))
```
cpitclaudel: This looks suspicious. Wouldn't `ansi-color-filter-apply` always be bound, due to the `require` above?

flycheck.el
```diff
@@ -6920,6 +6921,62 @@ See URL `http://www.erlang.org/'."
...
+     checker buffer))
+  :error-patterns
+  ((warning line-start
+            (file-name) ":" line ": Warning:" (message) line-end)
+   (error line-start
+          (file-name) ":" line ": " (message) line-end))
+  :modes erlang-mode
+  :enabled (lambda () (flycheck-rebar3-project-root))
```
cpitclaudel: No need to η-expand the `flycheck-rebar3-project-root` (you can drop the lambda).

flycheck.el
```diff
@@ -6920,6 +6921,62 @@ See URL `http://www.erlang.org/'."
...
+  :predicate (lambda () (flycheck-buffer-saved-p))
```
cpitclaudel: Same here

test/flycheck-test.el
```diff
@@ -3044,14 +3035,33 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
...
+
+(flycheck-ert-def-checker-test erlang-rebar3 erlang warning
+  (flycheck-ert-should-syntax-check
+   "language/erlang/rebar3/src/erlang-warning.erl" 'erlang-mode
+   '(6 nil warning "wrong number of arguments in format call" :checker erlang-rebar3)))
+
+(flycheck-ert-def-checker-test erlang-rebar3 erlang build
+  (flycheck-ert-should-syntax-check
+   "language/erlang/rebar3/_checkouts/dependency/src/dependency.erl" 'erlang-mode)
+  ;; Ensure that the dependency file wasn't build as standalone
```
cpitclaudel: `build` → `built`

test/resources/language/erlang/rebar3/.gitignore
```diff
@@ -0,0 +1,17 @@
...
+*.swo
+.erlang.cookie
+ebin
+log
+erl_crash.dump
+.rebar
+logs
+_build
+.idea
+rebar.lock
```
cpitclaudel: Do we really need all of these?

test/resources/language/erlang/rebar3/_checkouts/dependency/.gitignore
```diff
@@ -0,0 +1,17 @@
...
+*.swo
+.erlang.cookie
+ebin
+log
+erl_crash.dump
+.rebar
+logs
+_build
+.idea
+rebar.lock
```
cpitclaudel: Same here

cpitclaudel:
I left a few comments — there are some typos, and I'm a bit confused about some parts of the code.

flycheck.el
tsloughter: The first `rebar.config` it finds isn't necessarily the project root. This function is actually slightly wrong though. It should use `"\\(/_build/[^/]+/lib\\|/_checkouts\\|/apps\\)"`

tsloughter: To be clearer, any Erlang application in a project could have a `rebar.config` of its own, while there will also be a top level `rebar.config`. So this just makes sure that the file we are using isn't a sub-app or dependency of the project.

tsloughter:
@joedevivo want me to take this over?

joedevivo:
I was going to take a look tomorrow morning. It seems like there's not much left. :D

tsloughter:
Ok cool :), had just seen there wasn't much since 2016.

flycheck.el
joedevivo: Yes. You might want to explore how a dependency works, but that buffer might think it's the main project without this. Then it would try and fetch the dependencies dependencies, which have already been fetched by the root project.

flycheck.el
joedevivo: I don't really understand this either. @lunaryorn told me to add this [here](#discussion_r85707469)

test/resources/language/erlang/rebar3/_checkouts/dependency/.gitignore
joedevivo: Both were generated by rebar3. I reduced them to things these integration tests actually create.

flycheck.el
joedevivo: This is the only comment I left outstanding, maybe you'll have a better context for this after reading the comment I mentioned and we can go from there

flycheck.el
joedevivo: @tsloughter as implemented, this should catch all of those things. I assume that the critique here is that it might catch more than it should?

joedevivo: @cpitclaudel, @tsloughter is one of the authors of rebar3, so I'll defer to his judgement on if this needs further modification. Or, as I hope, that it's good enough for now, we can get this merged, and then come back and make small tweaks as needed.

joedevivo:
@cpitclaudel I'm happy to maintain this still

flycheck.el
```diff
@@ -6920,6 +6921,62 @@ See URL `http://www.erlang.org/'."
...
+
+(flycheck-define-checker erlang-rebar3
+  "An Erlang syntax checker using the rebar3 build tool."
+  :command ("rebar3" "compile")
+  :error-parser
+  (lambda (output checker buffer)
+    ;; rebar3 outputs ANSI terminal colors, which don't match up with
+    ;; :error-patterns, so we strip those color codes from the output
+    ;; here before passing it along to the default behavior. The
+    ;; relevant disucssion can be found at
```
cpitclaudel: disucssion → discussion

flycheck.el
```diff
@@ -6920,6 +6921,62 @@ See URL `http://www.erlang.org/'."
...
+    ;; https://github.com/flycheck/flycheck/pull/1144
+    (require 'ansi-color)
+    (flycheck-parse-with-patterns
+     (and (fboundp 'ansi-color-filter-apply) (ansi-color-filter-apply output))
```
cpitclaudel: I think the idea of this check was to silence the byte-compiler — I'd rather use with-no-warnings, or add a comment about the condition.

flycheck.el
cpitclaudel: @tsloughter Thanks, that's what I meant by `looking for the topmost file`.

flycheck.el
```diff
@@ -6920,6 +6921,62 @@ See URL `http://www.erlang.org/'."
...
+non-nil of which will be our project root.
+
+Return the absolute path to the directory"
+  (if (string= file-name prev-file-name)
+      (car (remove nil acc))
+    (let ((current-dir (file-name-directory file-name)))
+      (locate-rebar3-project-root
+       (directory-file-name current-dir)
+       file-name
+       (cons (contains-rebar-config current-dir) acc)))))
```
cpitclaudel: I think it would help with maintainability if this function was based on `locate-dominating-file`. Something like this, maybe? (untested):

```
(defun locate-topmost-dominating-file (file name)
  (let ((result nil)
        (further-up file))
    (while further-up
      (setq further-up (locate-dominating-file further-up name))
      (when further-up
        (setq result further-up)
        (setq further-up (file-name-directory (directory-file-name further-up)))))
    result))
```

cpitclaudel:
Thanks a lot. Things look great now, and this PR's discussion thread is starting to look suspiciously long :) Welcome aboard, and thanks for this contribution! Sorry for the long reviewing process.

I left a few comments — it would be great if you could address them in a separate PR.  Additionally, it would be nice to ask the rebar3 folks to update the checker to add a flag to disable color coding (or better, they could restrict color codes to TTYs).

PR #1149, opened 2016-10-29
Remove :standard-input nil from sass-lint

rski:
That is the default value for standard-input anyway


lunaryorn:
@rski I presume that `sass-lint` can't read from standard input?

rski:
@lunaryorn I couldn't find any way to do so, so probably not

lunaryorn:
Ok, LGTM then.

Issue #1154, opened 2016-11-02
Add an option to check rust project with cargo clippy


dzamlo:
`clippy` is a collection of additional lints for the rust programming language. A cargo project can be checked by running `cargo clippy`. In addition to adding more lints, `cargo clippy` will check the examples, the tests and all the binaries.

I think than an option to use it with flycheck would be a good idea.

For the implementation, running `cargo clippy` instead of `cargo rustc`, without the `--lib` or `--bin` option seems to work. I think the `--test` option should not be used with cargo clippy.

cpitclaudel:
Hi Loïc,

Do you think you could write the corresponding checker definition? It would probably share a lot with the `rustc` one, so it should be rather straightforward :)

dzamlo:
Yes, I can try. 
Do you think it should be a new checker, or an option that modify the behavior of the rust-cargo checker ?

cpitclaudel:
My vote would go to new checker, but @flycheck/rust will certainly have a more relevant opinion :)
The nice part of making it a separate checker is that it is discoverable and configurable just like all other checkers.

cpitclaudel:
On the other hand, if it's a strict superset of `rustc` in terms or errors reported, then it probably shouldn't be a separate checker (i.e., if it does `rustc` + more checks, then it could just be an option)

fmdkdd:
I've never used `clippy`, but it's a linter,  so I expect it to give linting errors only, and not compilation errors.  And the output format seems "classical" (i.e., not JSON), so it wouldn't be compatible with the current `rust` or `rust-cargo` checkers.  I think a new checker will be best :)

@dzamlo The current rust checkers might not be the best examples to base yours upon, as we parse the errors from the JSON output rather than the plain text stdout.  You can look at other linters to get an idea, or at the [previous versions](https://github.com/flycheck/flycheck/blob/00a38359bf12fb004276a878731382685a480247/flycheck.el#L8276-L8311) of the rust checkers to get an idea.

Feel free to ask questions if any arise :)

lunaryorn:
It's a separate tool, so I think we should have a separate checker.

dzamlo:
`cargo clippy` will actually give you the errors that `cargo rustc` will give you, but will give you some more warnings. It is a strict superset of the `cargo rustc` in terms of errors reported.

As far as I understand, the tool will do the same thing that `cargo rustc`, but will add some options to use the clippy lints and compile both as test and none test, will check the examples, etc.

[...]

PR #1156, opened 2016-11-03
Add the rust-cargo-clippy checker

dzamlo:
Fixes GH-1154

### DIFF
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -955,7 +955,7 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       .. note::
 
-         This syntax checker requires ``cargo clippy``, and therefore a nighlty version of rustc.
+         This syntax checker requires ``cargo clippy``, and therefore a nightly version of rustc.
 
       .. defcustom:: flycheck-cargo-clippy-args
 
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -949,6 +949,18 @@ to view the docstring of the syntax checker.  Likewise, you may use
          A list of additional library directories. Relative paths are relative
          to the buffer being checked.
 
+   .. syntax-checker:: rust-cargo-clippy
+
+      Check syntax and lint with `cargo clippy <https://github.com/Manishearth/rust-clippy>`_.
+
+      .. note::
+
+         This syntax checker requires ``cargo clippy``, and therefore a nighlty version of rustc.
+
+      .. defcustom:: flycheck-cargo-clippy-args
+
+         A list of additional arguments passed to the cargo clippy subcommand.
+
 .. supported-language:: Sass/SCSS
 
    Flycheck checks SASS with `sass/scss-sass-lint`, falling back to `sass`, and
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -959,9 +959,10 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
          This syntax checker requires ``cargo clippy``, and therefore a nightly version of rustc.
 
-      .. defcustom:: flycheck-cargo-clippy-args
+      .. defcustom:: flycheck-cargo-rustc-args
 
-         A list of additional arguments passed to the cargo clippy subcommand.
+         A list of additional arguments passed to the cargo clippy subcommand. This checker reuses
+         the variable from `rust-cargo`.
 
 .. supported-language:: Sass/SCSS
 
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -890,9 +890,11 @@ to view the docstring of the syntax checker.  Likewise, you may use
 .. supported-language:: Rust
 
    Flycheck checks Rust_ with `rust-cargo` in Cargo projects, or `rust`
-   otherwise.
+   otherwise. For Cargo projects, you can also use `rust-cargo-clippy` and get
+   the additional lints of clippy_.
 
    .. _Rust: https://www.rust-lang.org/
+   .. _clippy: https://github.com/Manishearth/rust-clippy
 
    .. syntax-checker:: rust-cargo
                        rust
--- a/flycheck.el
+++ b/flycheck.el
@@ -8727,6 +8727,8 @@ This syntax checker needs the cargo clippy subcommand. See URL `https://github.c
   :error-parser flycheck-parse-rust
   :error-explainer flycheck-rust-error-explainer
   :modes rust-mode
+  ;; cargo clippy must be run from the directory containing Cargo.toml
+  :working-directory (lambda (_) (locate-dominating-file (buffer-file-name) "Cargo.toml"))
   ;; Since we build the entire project with cargo clippy we require
   ;; that the buffer is saved.
   :predicate flycheck-buffer-saved-p
--- a/flycheck.el
+++ b/flycheck.el
@@ -8734,7 +8734,8 @@ rustc command.  See URL `https://www.rust-lang.org'."
 (flycheck-define-checker rust-cargo-clippy
   "A Rust syntax checker and linter using clippy.
 
-This syntax checker needs the cargo clippy subcommand. See URL `https://github.com/Manishearth/rust-clippy'"
+This syntax checker needs the cargo clippy subcommand. See URL
+`https://github.com/Manishearth/rust-clippy'"
   :command ("cargo" "clippy"
             (eval flycheck-cargo-rustc-args)
             "--"  "--error-format=json"
--- a/flycheck.el
+++ b/flycheck.el
@@ -8691,10 +8691,7 @@ Relative paths are relative to the file being checked."
     (flycheck-verification-result-new
      :label "Cargo.toml"
      :message (if have-cargo-toml "found" "missing")
-     :face (if have-cargo-toml 'success '(bold error))
-     )
-    )
-  )
+     :face (if have-cargo-toml 'success '(bold error)))))
 
 (flycheck-define-checker rust-cargo
   "A Rust syntax checker using Cargo.
@@ -8752,8 +8749,7 @@ This syntax checker needs the cargo clippy subcommand. See URL `https://github.c
                (flycheck-verification-result-new
                 :label "cargo clippy subcommand"
                 :message (if have-clippy "found" "missing")
-                :face (if have-clippy 'success '(bold error))
-                )))))
+                :face (if have-clippy 'success '(bold error)))))))
 
 (flycheck-define-checker rust
   "A Rust syntax checker using Rust compiler.
--- a/flycheck.el
+++ b/flycheck.el
@@ -231,6 +231,7 @@ attention to case differences."
     ruby
     ruby-jruby
     rust-cargo
+    rust-cargo-clippy
     rust
     scala
     scala-scalastyle
@@ -8608,10 +8609,13 @@ See URL `http://jruby.org/'."
   :modes (enh-ruby-mode ruby-mode)
   :next-checkers ((warning . ruby-rubylint)))
 
+(flycheck-def-args-var flycheck-cargo-clippy-args (rust-cargo-clippy)
+  :package-version '(flycheck . "31"))
+
 (flycheck-def-args-var flycheck-cargo-rustc-args (rust-cargo)
   :package-version '(flycheck . "30"))
 
-(flycheck-def-args-var flycheck-rust-args (rust-cargo rust)
+(flycheck-def-args-var flycheck-rust-args (rust-cargo rust-cargo-clippy rust)
   :package-version '(flycheck . "0.24"))
 
 (flycheck-def-option-var flycheck-rust-check-tests t (rust-cargo rust)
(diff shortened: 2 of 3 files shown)


lunaryorn:
@dzamlo Thank you very much, LGTM.  I'll take a look right away.

doc/languages.rst
```diff
@@ -949,6 +949,18 @@ to view the docstring of the syntax checker.  Likewise, you may use
          A list of additional library directories. Relative paths are relative
          to the buffer being checked.
 
+   .. syntax-checker:: rust-cargo-clippy
+
+      Check syntax and lint with `cargo clippy <https://github.com/Manishearth/rust-clippy>`_.
+
+      .. note::
+
+         This syntax checker requires ``cargo clippy``, and therefore a nighlty version of rustc.
```
lunaryorn: "nighlty" -> "nightly"

doc/languages.rst
```diff
@@ -949,6 +949,18 @@ to view the docstring of the syntax checker.  Likewise, you may use
...
+
+      .. defcustom:: flycheck-cargo-clippy-args
```
lunaryorn: Do we need this option?

doc/languages.rst
lunaryorn: Please also extend the paragraph above after `.. supported-language:: rust` to mention the new syntax checker and how it relates with the existing rust checkers.

flycheck.el
```diff
@@ -8711,6 +8715,25 @@ rustc command.  See URL `https://www.rust-lang.org'."
...
 
+(flycheck-define-checker rust-cargo-clippy
+  "A Rust syntax checker and linter using clippy.
+
+This syntax checker needs the cargo clippy subcommand. See URL `https://github.com/Manishearth/rust-clippy'"
+  :command ("cargo" "clippy"
+            (eval flycheck-cargo-clippy-args)
+            "--"  "--error-format=json"
+            (option-list "-L" flycheck-rust-library-path concat)
+            (eval flycheck-rust-args))
```
lunaryorn: Can you explain the difference between `flycheck-cargo-clippy-args` and `flycheck-rust-args` with regards to this syntax checker?

flycheck.el
```diff
@@ -8711,6 +8715,25 @@ rustc command.  See URL `https://www.rust-lang.org'."
...
+  :error-parser flycheck-parse-rust
+  :error-explainer flycheck-rust-error-explainer
+  :modes rust-mode
+  :predicate (lambda ()
+               ;; Since we build the entire project with cargo clippy we require
+               ;; that the buffer is saved.  And of course, we also need a Cargo
+               ;; file :)
+               (and (flycheck-buffer-saved-p)
+                    (locate-dominating-file (buffer-file-name) "Cargo.toml"))))
```
lunaryorn: Please move the check for `Cargo.toml` into an `:enabled` function (see `javascript-eslint` for an example of the `:enabled` field), so that we only check for the file once when the checker is used first and not every single time.

I presume the `clippy` command is not built-in to `cargo`?  If so `:enabled` also needs to check whether the subcommand is present.

doc/languages.rst
dzamlo: There is a very similar options for the `rust-cargo` checker (`flycheck-cargo-rustc-args`). I thought it would make sense to add it for this checker.

flycheck.el
lunaryorn: @dzamlo Yes.  It's just that `rust-cargo` predates the `:enabled` attribute, but nowadays these checks should always happen in `:predicate`.

lunaryorn: By the way, we'd also need a `:verify` function that informs the user about whether `Cargo.toml` was found or not, to help them figure out why a particular syntax checker isn't available in `C-c ! v`.  Take a look at `javascript-eslint` for an example.

flycheck.el
lunaryorn: Ah, I see; one's for cargo arguments, the other for clippy arguments.

fmdkdd: So we'll have three argument variables.

- `flycheck-rust-args` that are for the compiler (`rustc`)
- `flycheck-cargo-rustc-args` are arguments to the `cargo rustc` *command*, not to the compiler.
- `flycheck-cargo-clippy-args` are arguments to the `cargo clippy` command, and `cargo clippy` accepts a superset of the options of `cargo rustc`.

It's getting a bit confusing, but at least the naming is consistent ;)

fmdkdd: To be clear: I think it's the right call; just that it unfortunately be confusing for new users (it's certainly starting to confuse me!)

lunaryorn: @fmdkdd Should we perhaps merge flycheck-cargo-rustc-args and flycheck-cargo-clippy-args into a single `flycheck-cargo-args`?

fmdkdd:
LGTM.  I'm wondering how this new checker will interfere with `rust cargo` however.  If you don't have clippy installed, then clearly this new checker will be disabled.  But if you do have clippy installed, which takes precedence?  Will both run whenever the buffer is saved?  Since there's an overlap (clippy is a superset of rustc), will compilation errors be reported twice?

I know the compiler is a little finnicky, in that different flags will trigger a full recompilation of the crate.  If we do call both `cargo rustc` and `cargo clippy`, do one trigger a full recompilation before the other?

If there are any interferences, a fix would be to disable `rust cargo` when clippy is present.  A good opportunity to add this `:enabled` attribute on `rust cargo` ;)

dzamlo:
From what I understand of flychek, until you use `:next_checker` only one the first checker for the curent mode is used. You need to disable `rust-cargo` to use `rust-cargo-clippy`

I have not yet made the requested changes in the documentation.

fmdkdd:
Okay then.  I was thinking we could disable `rust-cargo` when clippy is found, so `rust-cargo-clippy` could take precendence.  But then again I noticed that I currently have `cargo-clippy` on my PATH for testing with rust nightly, but I am still using rust stable, which means that even though the executable is present, it is unusable (fails to load shared libraries).  So maybe opt-in is preferable ;)

dzamlo:
There is a multiple repetition of `(locate-dominating-file (buffer-file-name) "Cargo.toml")`. I'm not sure if I should make a function just for that because it's also very short. What do you think ?

Otherwise the code should be ready for second review.

flycheck.el
```diff
@@ -8681,6 +8685,17 @@ Relative paths are relative to the file being checked."
...
+(defun flycheck-rust--verify-cargo-toml ()
+  "Return a flycheck-verification-result indicating whether Cargo.toml was found."
+  (let ((have-cargo-toml (locate-dominating-file (buffer-file-name) "Cargo.toml")))
+    (flycheck-verification-result-new
+     :label "Cargo.toml"
+     :message (if have-cargo-toml "found" "missing")
+     :face (if have-cargo-toml 'success '(bold error))
+     )
+    )
+  )
```
lunaryorn: Please do not use hanging parenthesis.  Instead, move all the closing parenthesis up to the last code line.

I know it feels odd coming from other languages, but in Lisp it really aids readability to keep parenthesis along with the code.

flycheck.el
```diff
@@ -8704,12 +8719,41 @@ rustc command.  See URL `https://www.rust-lang.org'."
...
+                  (locate-dominating-file (buffer-file-name) "Cargo.toml")))
+  :verify (lambda (_)
+            (let ((have-clippy (funcall flycheck-executable-find "cargo-clippy")))
+              (list
+               (flycheck-rust--verify-cargo-toml)
+               (flycheck-verification-result-new
+                :label "cargo clippy subcommand"
+                :message (if have-clippy "found" "missing")
+                :face (if have-clippy 'success '(bold error))
+                )))))
```
lunaryorn: Please move these parenthesis on the line before.

lunaryorn:
@dzamlo Yes, sure, that'd be a good idea.  It reduces duplication, and—more important—helps to understand the code.  I'd call the function `flycheck-find-cargo-root` or similar…

By the way, should we perhaps merge flycheck-cargo-rustc-args and flycheck-cargo-clippy-args into a single flycheck-cargo-args?  What do you think?  /cc @fmdkdd

fmdkdd:
According to `cargo clippy --help`, it accepts two arguments: `--help` and `--features`.  Features can be used to toggle individual lints on the command line.  All the other options are the same as `cargo rustc` (since clippy will pass them to the compiler).  But `--features` is also a valid option for `cargo rustc`, so the set of valid flags is the same.

So it seems it should not cause much confusion with users if we reuse the same variable.  We can keep and reuse `flycheck-cargo-rustc-args` for clippy, since most of the flags are the same, and it's more descriptive than just `flycheck-cargo-args`.

We should however document this overloading in doc/languages.rst.

lunaryorn:
@dzamlo Can you update this pull request according to @fmdkdd's comment?

LGTM now.

dzamlo:
I'm not really sure what is the best way to reorganize languages.rst:  rust-cargo-clippy doesn't have any option on its own, it only use options that the other rust checker have. So there should be only one section for the tree checkers. On the other hand, I find the explanation of each of the rust checker become confusing when the three checkers are grouped together.

fmdkdd:
I don't think there's any harm in using the `defcustom:: flycheck-cargo-rustc-args` twice, once when describing `rust-cargo`, and once when describing the `rust-cargo-clippy` checker, and state in the second case that it reuses the variable from `rust-cargo`.

If there is a documentation check that prevents declaring it twice, we could define `flycheck-cargo-clippy-args` as an alias to `flycheck-cargo-rustc-args.

doc/languages.rst
```diff
@@ -949,6 +951,19 @@ to view the docstring of the syntax checker.  Likewise, you may use
          A list of additional library directories. Relative paths are relative
          to the buffer being checked.
 
+   .. syntax-checker:: rust-cargo-clippy
```
lunaryorn: If I understand correctly this checker is a superset of `rust-cargo`, isn't it?  

In this case please add it to the previous `.. syntax-checker::` directive with the other rust checkers, so that we don't need to document each option twice.  You can move the `.. note::` up as well and just change it to 

```
The `rust-cargo-clippy` syntax checker requires …
```

dzamlo: In terms if errors reported, this checker is a superset of `rust-cargo`.
But in term of options, only the `flycheck-rust-args` and `flycheck-cargo-rustc-args` options are used by this checker.  Most of the options of the others rust checkers doesn't apply to this checker.

lunaryorn: @dzamlo If it's a superset of `rust-cargo` and replaces `rust-cargo` in regular use, shouldn't it use at least the same options as `rust-cargo`?

flycheck.el
```diff
@@ -8704,12 +8725,40 @@ rustc command.  See URL `https://www.rust-lang.org'."
...
+  ;; Since we build the entire project with cargo rustc we require
+  ;; that the buffer is saved.
+  :predicate flycheck-buffer-saved-p
+  :enabled flycheck-rust-find-cargo-root
+  :verify (lambda(_) (list (flycheck-rust--verify-cargo-toml))))
+
+(flycheck-define-checker rust-cargo-clippy
+  "A Rust syntax checker and linter using clippy.
+
+This syntax checker needs the cargo clippy subcommand. See URL `https://github.com/Manishearth/rust-clippy'"
```
lunaryorn: Please wrap this line; it's too long and goes beyond our 80 character limit.  Just press `M-q` to automatically wrap the string :)

doc/languages.rst
fmdkdd: Agreed on `flycheck-rust-library-path`.  I had plan to phase it out of `rust-cargo`, so no need to rely on it.

As for the others, the reasoning is sound.  Thank you @dzamlo for taking the time to outline it.

fmdkdd:
Regarding the lacking error output, here is an example.  Running clippy in the terminal I get:

```
warning: this if-then-else expression returns a bool literal, #[warn(needless_bool)] on by default
   --> src/cpu.rs:248:25
    |
248 |             sprite.push(if (p & (1 << (7 - b))) > 0 { true } else { false });
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: you can reduce it to
    |             sprite.push((p & (1 << (7 - b))) > 0);
```

But currently, the rust parser for flycheck will not pick up the last code line, and you will end up with the message being only:
```
help: you can reduce it to
```
in your Emacs buffer.  That's not very helpful.

I don't want to block this PR for that, so we are fine.  But I'll be looking at enhancing the parser to catch these cases soon enough.

fmdkdd:
@dzamlo Are you still interested in this checker?  Because we'd still like to see it make it into Flycheck!

lbolla:
Any news on this checker? Can it be made a separate package, while we wait it to be merged inside flycheck?

fmdkdd:
@lbolla I think it was mostly ready.  If we do not hear from @dzamlo in the coming days, I can probably pick it up.

fmdkdd:
Closing in favor of #1385.

PR #1158, opened 2016-11-06
Set pylint output-format to text explicitly

koterpillar:
`pylintrc` can set the output format to `colorized`, which will fail to parse. Explicitly set it to `text` via the command line option.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8204,6 +8204,7 @@ This syntax checker requires Pylint 1.0 or newer.
 See URL `https://www.pylint.org/'."
   ;; -r n disables the scoring report
   :command ("pylint" "-r" "n"
+            "--output-format" "text"
             "--msg-template"
             (eval (if flycheck-pylint-use-symbolic-id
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Thanks a lot!  Is this option documented somewhere?  All I can find is https://pylint.readthedocs.io/en/latest/user_guide/output.html, and it doesn't mention `"text"` :/

It does mention JSON, however; it would be very nice to move to that, if you feel like a fun ELisp challenge :)

koterpillar:
From `pylint --help`:
  Reports:
    -f <format>, --output-format=<format>
                        Set the output format. Available formats are text,
                        parseable, colorized, msvs (visual studio) and html.

koterpillar:
I'm afraid I don't know enough ELisp to switch to `parseable`, maybe in a separate pull request later.

cpitclaudel:
Thanks for the clarification about `--help` :)

lunaryorn:
@koterpillar Thank you very much!

Issue #1060, opened 2016-09-01
Use docker in CI


ghprince:
Looking at a successful Travis CI build job of flycheck, half of the time (87s out of 160s) is spent on installing Emacs using a Makefile from [flycheck/emacs-travis](https://github.com/flycheck/emacs-travis)

Instead of compiling and installing Emacs for each build, how about create a Docker image that has all software we need for test installed, and pull it from DockerHub (maybe also cache it at Travis)? I believe it will save us a lot of time and even possible to bring integration test back to CI

lunaryorn:
I'm not so familiar with Docker as to know whether that's possible, let alone to implement it.  I've never created any docker images 😞 and I didn't even know that one could use custom docker images on Travis CI ☺️ 

I think it'd be great if we could a CI that has no setup cost, and if Docker is a good way to go, why not?  Do you think it's possible to do that on Travis CI—or is there another platform that could support this?  Could you try and implement a demo (i.e. a build with a small docker image that has just one or two languages)?

ghprince:
I did some experiment and got blocked on this issue
https://github.com/docker/docker/issues/22801

I'll try install emacs from apt-get (24.3 for 14.04 and 24.5 for 16.04) later instead of building from source.

lunaryorn:
@ghprince Ah, yes, I remember a lengthy emacs-devel thread about this issue.  I'm sorry that you hit a wall here—I guess it's impossible to build Emacs within a Docker image currently?

It'd be great to get started with apt-get—but if we are to replaced our current setup we have to achieve the same version coverage, i.e. the lowest supported version (24.3 at this point), the latest stable version (24.5), the master branch of Emacs and currently also the latest pretest.

I guess we may need a custom repository for this purpose?

cpitclaudel:
@ghprince I didn't have time to read all these discussions in detail, but things are slightly different in our case, I think.  The issue here is that build fails during what really amounts to an optimization (preloading a bunch of libraries, then saving the resulting emacs image as a separate binary — this is what emacs calls "dumping").  Couldn't we just run the undumped Emacs? It'll be slower to start, but I'm not sure it'll be that much; maybe tens of seconds at most?

lunaryorn:
@cpitclaudel Can Emacs' build system install an undumped Emacs? 

And thinking of it I'm not sure whether it's a good idea to run our tests in an Emacs build that virtually no user would ever actually run 😕

cpitclaudel:
[...]

PR #1159, opened 2016-11-09
Use docker to speed up CI

ghprince:
this is an effort to solve #1060 

It is still work in process, hacky and buggy code, please do not review until I think it is good enough.

I created the PR simply to trigger travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,26 @@
-# Use Python 3.5 for our tooling
-language: python
-python:
-  - "3.5"
 sudo: required
 services:
   - docker
 matrix:
   # Don’t wait for jobs that are allowed to fail
   fast_finish: true
+cache:
+  - directories:
+    - ~/docker/
 env:
   matrix:
-    - EMACS_VERSION=24.4
+    - EMACS_VERSION=24.3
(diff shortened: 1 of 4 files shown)


robbyoconnor:
I like this...we should ideally install dependencies for checkers and allow for all checker tests to run in CI as well..

cpitclaudel:
@robbyoconnor We use to have a testing VM, but it was too much work keeping up with all the checkers.

fmdkdd:
@robbyoconnor @cpitclaudel But any way to make it work without too much overhead would be swell ;)

I saw some thread on emacs-devel about a portable dumper branch that's supposed to fix the bug preventing Emacs to be included in a docker image.  Don't know how much helpful that would be; just mentioning.

robbyoconnor:
not running checker tests in travis negates having them

fmdkdd:
@ghprince Any update on this?  We're still very much interested in improving the CI situation, especially regarding the integration tests.

fmdkdd:
Closing due to inactivity.  See also https://github.com/flycheck/flycheck/issues/1060#issuecomment-364676605.

PR #1162, opened 2016-11-16
Avoid nil in flycheck-same-files-p

drvink:
This prevents a race condition when reloading a desktop in which Flycheck may try to check a buffer for which (buffer-file-name) returns nil.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1075,16 +1075,19 @@ to a number and return it.  Otherwise return nil."
 
 (defun flycheck-same-files-p (file-a file-b)
   "Determine whether FILE-A and FILE-B refer to the same file."
-  (let ((file-a (expand-file-name file-a))
-        (file-b (expand-file-name file-b)))
-    ;; We must resolve symbolic links here, since some syntax checker always
-    ;; output canonical file names with all symbolic links resolved.  However,
-    ;; we still do a simple path compassion first, to avoid the comparatively
-    ;; expensive file system call if possible.  See
-    ;; https://github.com/flycheck/flycheck/issues/561
-    (or (string= (directory-file-name file-a) (directory-file-name file-b))
-        (string= (directory-file-name (file-truename file-a))
-                 (directory-file-name (file-truename file-b))))))
+  (if (or (null file-a)
+          (null file-b))
+      nil
+    (let ((file-a (expand-file-name file-a))
+          (file-b (expand-file-name file-b)))
+      ;; We must resolve symbolic links here, since some syntax checker always
+      ;; output canonical file names with all symbolic links resolved.  However,
+      ;; we still do a simple path compassion first, to avoid the comparatively
+      ;; expensive file system call if possible.  See
+      ;; https://github.com/flycheck/flycheck/issues/561
+      (or (string= (directory-file-name file-a) (directory-file-name file-b))
+          (string= (directory-file-name (file-truename file-a))
+                   (directory-file-name (file-truename file-b)))))))
 
 (defvar-local flycheck-temporaries nil
   "Temporary files and directories created by Flycheck.")


lunaryorn:
@drvink Thanks!  I don't yet understand the issue, though: Could you perhaps explain the adverse effects of this race condition?  How does the user experience this issue?

drvink:
Oops, sorry. I said what it fixed but not why it needs fixing. :)

If you look at the cross references to `flycheck-same-files-p` (there are only three or four), you'll see there are places where Flycheck directly calls it with a `(buffer-file-name)` argument, and `buffer-file-name` can return `nil` (for buffers not associated with backing files), so when `flycheck-same-files-p` calls `expand-file-name` with a nil argument, an error is signaled and the resulting unhandled exception results in anything from checkers getting wedged (on the side of the process or by fouling state on the the side of the invoking elisp) to mysterious and seemingly inexplicable intermittent failures in other libraries or during Emacs startup. These would happen about every two out of ten startups with a certain restored desktop/buffer configuration. I'd give you the backtrace that allowed me to find this after sticking `(setq debug-on-error t)` in my `.emacs` and restarting until it happened if I still had it in my clipboard, but I think it's pretty clear from the code that this it's possible for this situation to occur.

Specifically, the arguments to `flycheck-same-files-p` were a valid string argument and nil, and even if you don't accept that this is a Flycheck bug, consider that Flycheck is:

* a library
* in a "dynamically typed" language
* for a highly extensible program whose design both permits and encourages doing so

...which basically means that even if some code that's criminally stupid or outright malicious can be made to coerce Flycheck to do the wrong thing, you still have to guard against it. And please don't get me wrong when I say that--I love Emacs (though not so much its implementation languages), and I'm very grateful for Flycheck!

lunaryorn:
I see, but I think the issue is rather that `flycheck-same-files-p` gets called with nil arguments. That shouldn't happen! 

Could you try and produce a backtrace that shows where the call with nil happens?  This is probably just a symptom of a deeper flaw in Flycheck, so I'd like to understand what's going on.

drvink:
OK, I'll try to get it to cough up another one--it wasn't hard previously but it would be just my luck for it to Heisenbug on me now.

drvink:
https://gist.github.com/c2665e29500c67d03e9a5e5acb56df26 Here's a backtrace.

lunaryorn:
@drvink Oh, I see 😲 Now that I look at the code I wonder how Flycheck was ever able to check buffers without files 🤔 

I'd like to take a deeper look—please give me a couple of days to check this issue. I think the fix will be different, as I'd like to avoid calling `flycheck-same-files-p` with a nil argument in the first place.

drvink:
No problem. This workaround is fine for me now, since it only ever seemed to be an issue when starting Emacs anyway.

lunaryorn:
@drvink Maybe we should inhibit Flycheck complete while desktop is restoring buffers anyway.  It doesn't seem to be reasonable to start checking buffers at that stage.

Could you try and find out whether desktop mode exposes any variable or function that we can use to check whether desktop is in the process of restoring buffers?

lunaryorn:
@drvink Maybe we should inhibit Flycheck complete while desktop is restoring buffers anyway.  It doesn't seem to be reasonable to start checking buffers at that stage.

Could you try and find out whether desktop mode exposes any variable or function that we can use to check whether desktop is in the process of restoring buffers?

drvink:
It doesn't appear to. The most useful thing available is `desktop-after-read-hook`, to get notified when desktop is done restoring a desktop.

drvink:
Oh, actually, you could use `desktop-minor-mode-handlers` in order to establish that a desktop is being restored/changed. Combine that with `desktop-after-read-hook` and you'll know when it's done too.

You'll want to check the comments at the beginning of desktop.el under "Special handling" and "Minor modes", though.

lunaryorn:
@drvink Could you perhaps summarize these comments for me?  Or maybe even sketch a solution around these features?  I'm short of time currently and can't take a look myself.

drvink:
I went ahead and tried writing the thing I proposed, but it didn't solve the problem. However, I realized that the issue isn't a race condition after all--it seems to be fairly deterministic, now that I realize what's happening.

I don't know *why* yet, and I don't think I'll have time to look into it, but each time I restart Emacs, for whatever reason, desktop doesn't give me the same buffer with which I quit--it restores them all fine, but if I simply quit and restart Emacs over and over, it cycles through them round robin (weird, but whatever). I finally noticed the correlation: the bug would happen each time the dired window happened to be the one that was displayed. And of course, dired buffers aren't associated with a file name...

I don't think I'll have time to look into this further right now (and now you might be able to reproduce it yourself too), but if you need any more information or backtraces from me, I can provide them.

drvink:
Also, while I haven't even wall clocked it and it's not relevant to this bug, if you've got nothing better to do and want to see if waiting to reenable Flycheck only after a desktop has been fully restored improves big desktop switch/startup time compared to the current behavior, this is all it takes: https://gist.github.com/6279fbf9e55abb49a0512026c0c2063e

lunaryorn:
@drvink I've fixed the cause of this issue, ensuring that Flycheck doesn't call `flycheck-same-files-p` if `buffer-file-name` is `nil`.  I don't yet understand how it could end up in this situation, but I think we're safe now.

----

Thanks for sketching a solution about desktop mode, but I'm not sure whether it's the way to go.  Ideally we'd not have care for whether Flycheck Mode is enabled as part of desktop save mode, but instead have a flag or something to check in `flycheck-must-defer-check` so that we could defer necessary syntax check until desktop restore has finished, and then only check buffers which have become visible (e.g. have a window attached).

We're already using the same approach to handle buffer reverts (see `revert-buffer-in-progress-p` in `flycheck-must-defer-check`).

I had hoped that there's something similar for desktop mode 😞

drvink:
Sorry for replying late--yeah, unfortunately, that's all there appears to be for desktop mode. I haven't had a chance yet to see whether this fixes the issue for me, but I'll just comment again if I can still reproduce it. Thanks again!

drvink:
I'm still having the issue. I think you meant for `flycheck-relevant-error-p` to be this:

```elisp
(defun flycheck-relevant-error-p (err)
  "Determine whether ERR is relevant for the current buffer.

Return t if ERR may be shown for the current buffer, or nil
otherwise."
    (let ((file-name (flycheck-error-filename err))
          (message (flycheck-error-message err)))
      (and
       ;; The error is relevant for the current buffer if it's got no file-name
       ;; and the current buffer has no file name, too, or if it refers to the
       ;; same file as the current buffer.
       (or (and (not file-name) (not buffer-file-name))
           (and buffer-file-name ; <----- HERE
                (flycheck-same-files-p file-name buffer-file-name)))
       message
       (not (string-empty-p message))
       (flycheck-error-line err)))))
```

drvink:
@lunaryorn I'm still getting this issue under the same conditions (when the dired buffer gets restored, as mentioned earlier in the thread). The code in the previous comment should fix it.

cpitclaudel:
@drvink This seems like an OK idea.  Can you make a pull request with a short summary of the issue (a test would be tricky to write, correct?) and this change? Thanks!

drvink:
@cpitclaudel Yeah, I'm not sure what a useful test for this would look like, and I assume it would be hard to really test because it relies on magic Emacs variables rather than arguments to the function. It's hard to know where to draw the line since things like this are effectively call site bugs (i.e. don't invoke the function or do something that invokes the function at the wrong time), but since we have no way to prevent that, we have to ensure our function does the right thing...which is what this patch does.

I'll make a PR in a moment.

cpitclaudel:
Thanks :) Please allow some time for review, because we're all pretty busy these days :/

legends2k:
@lunaryorn I've version `20170212.1015` from elpa I am still able to reproduce this issue.

I've `elpy` (and its dependencies including flake8) installed.  Below is the backtrace I got using `toggle-debug-on-error` when I switch `*scratch*`'s major mode to `python`.

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  expand-file-name(nil)
  flycheck-same-files-p("f:/Work/tryouts/stdin" nil)
  flycheck-relevant-error-p([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"])
  #[257 "\300!\203\207\301\207" [flycheck-relevant-error-p exclude] 3 "\n\n(fn ELT)"]([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"])
  mapcar(#[257 "\300!\203\207\301\207" [flycheck-relevant-error-p exclude] 3 "\n\n(fn ELT)"] ([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"] [cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 2 "no newline at end of file" warning "W292"]))
  #[514 "\300\"\207" [mapcar] 5 "\n\n(fn FUNCTION SEQUENCE)"](#[257 "\300!\203\207\301\207" [flycheck-relevant-error-p exclude] 3 "\n\n(fn ELT)"] ([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"] [cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 2 "no newline at end of file" warning "W292"]))
  apply(#[514 "\300\"\207" [mapcar] 5 "\n\n(fn FUNCTION SEQUENCE)"] #[257 "\300!\203\207\301\207" [flycheck-relevant-error-p exclude] 3 "\n\n(fn ELT)"] ([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"] [cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 2 "no newline at end of file" warning "W292"]) nil)
  seq-map(#[257 "\300!\203\207\301\207" [flycheck-relevant-error-p exclude] 3 "\n\n(fn ELT)"] ([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"] [cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 2 "no newline at end of file" warning "W292"]))
  seq-filter(flycheck-relevant-error-p ([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"] [cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 2 "no newline at end of file" warning "W292"]))
  flycheck-relevant-errors(([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"] [cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 2 "no newline at end of file" warning "W292"]))
  flycheck-finish-current-syntax-check(([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"] [cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 2 "no newline at end of file" warning "W292"]) "F:/Work/tryouts/")
  flycheck-report-buffer-checker-status([cl-struct-flycheck-syntax-check #<buffer *scratch*> python-flake8 #<process flycheck-python-flake8> "F:/Work/tryouts/"] finished ([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"] [cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 2 "no newline at end of file" warning "W292"]))
  apply(flycheck-report-buffer-checker-status [cl-struct-flycheck-syntax-check #<buffer *scratch*> python-flake8 #<process flycheck-python-flake8> "F:/Work/tryouts/"] (finished ([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"] [cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 2 "no newline at end of file" warning "W292"])))
  #[128 "\301\302\300#\207" [[cl-struct-flycheck-syntax-check #<buffer *scratch*> python-flake8 #<process flycheck-python-flake8> "F:/Work/tryouts/"] apply flycheck-report-buffer-checker-status] 5 "\n\n(fn &rest ARGS)"](finished ([cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 1 "undefined name 'i'" error "F821"] [cl-struct-flycheck-error #<buffer *scratch*> python-flake8 "f:/Work/tryouts/stdin" 1 2 "no newline at end of file" warning "W292"]))
  flycheck-finish-checker-process(python-flake8 1 nil "stdin:1:1: F821 undefined name 'i'\nstdin:1:2: W292 no newline at end of file\n" #[128 "\301\302\300#\207" [[cl-struct-flycheck-syntax-check #<buffer *scratch*> python-flake8 #<process flycheck-python-flake8> "F:/Work/tryouts/"] apply flycheck-report-buffer-checker-status] 5 "\n\n(fn &rest ARGS)"] "F:/Work/tryouts/")
  flycheck-handle-signal(#<process flycheck-python-flake8> "exited abnormally with code 1\n")
```

manuel-uberti:
Please keep in mind @lunaryorn does not maintain Flycheck anymore. Refer to @cpitclaudel.

marsam:
@legends2k I created #1209 to fix this flake8 error. I think this issue is a bit different.

Issue #1168, opened 2016-11-21
Registering many errors blocks emacs for a noticeable amount of time


jyp:
This is not so much a bug report, more of a question and perhaps a feature request.

My situation is that I sometimes flycheck big files, which may contain a lot of warnings (maybe a thousand). In this situation, flycheck sometimes blocks emacs for half a second or so. This happens even when the checker is run asynchronously. (I have written my own [checker](https://github.com/jyp/dante), so I am familiar with the internals. The blocking happens when the `checker` function calls its callback with the list of warnings.)

There could be two ways to solve the problem:
1. Speed up the reporting of errors by a factor of 10 or more
2. Make the reporting of errors fully asynchronous. That is, the `checker` function (that which is stored in the `:start` property) could call its callback several times. Say, in addition of 

```
‘finished’
     The syntax checker has finished with a proper error report
     for the current buffer.  DATA is the (potentially empty)
     list of ‘flycheck-error’ objects reported by the syntax
     check.

     This report finishes the current syntax check.
```

we would have

```
‘partial’
     The syntax checker has some errors to report 
     for the current buffer.  DATA is the (probably non-empty)
     list of ‘flycheck-error’ objects reported by the syntax
     check in this partial run. 

     This report does not finish the current syntax check.
```

Perhaps this would be easy to implement? Perhaps there is even a way today to report errors asynchronously? I can offer help implementing the feature.

lunaryorn:
@jyp Thank you for this issue, but with all due respect I think you're underestimating the issue. 

> Speed up the reporting of errors by a factor of 10 or more

It's not as if we kind of deliberately slowed down error reporting to make someone find that hidden potential.  Flycheck uses overlays (they are the only reasonable API for our purposes) which are known to have poor performance.  As far as I know they even have quadratic complexity, i.e. adding 1000 overlays is a hundred times slower than adding 100 overlays.  

Please go ahead and profile our code, but I do think that you'll find a tenfold speedup in there.

As for asynchronous reporting, I'm not sure whether that'd fix the issue.  We can't add overlays asychronously, so all we'd achieve is splitting a 0.5s freeze into a couple of smaller freezes.

cpitclaudel:
> As far as I know they even have quadratic complexity, i.e. adding 1000 overlays is a hundred times slower than adding 100 overlays.

[...]

Issue #476, opened 2014-09-09
Overlaying shellcheck warnings takes too long on really large bash scripts


nordlow:
One of my Bash scripts is very large which in turn causes the checker `shellcheck` to give back an overlay list that takes around 10 second for Emacs to show. It this an inevitable limitation of Emacs? If so is adding `shellcheck` to `flycheck-disabled-checkers` in the shell-mode-hook only for this buffer the best way to do it? Maybe a more general solution would be to have a buffer-size limit for `shellcheck` as it often generates a lot of warnings.

Here's what I'm using so far

``` Emacs-Lisp
(defun flycheck-shellcheck-buffer-size-check ()
  (when (> (buffer-size) 100000)
    (add-to-list 'flycheck-disabled-checkers 'sh-shellcheck)))
(add-hook 'flycheck-mode-hook 'flycheck-shellcheck-buffer-size-check t)
```

Issue #179, opened 2013-08-02
Flycheck with flake8 can make Emacs pause for several seconds


Wilfred:
I'm seeing the same behaviour as described in #142. Here's the full steps to reproduce. I'm using Emacs 24.3 and have flake8 `2.0 (pep8: 1.4.6, mccabe: 0.2.1, pyflakes: 0.7.2)`

1: I moved `~/.config/flake8` to `~/.config/not_flake8` to ensure I'm testing with default flake8 settings.

2: I started Emacs with `emacs -Q`.

3: I loaded flycheck with `(require 'package)` `(package-initialize)`

4: I opened [this file](https://gist.github.com/Wilfred/ca60fa53ad7595adaba3) with `C-x C-f`.

5: I started flycheck with `M-x flycheck-mode`.

Flycheck seems to be pausing the Emacs event loop for 5-20 seconds on my system (Intel Core i5, 12GB RAM) every time it runs. Pressing `C-n` repeatedly during that time does not do anything, but after the pause, the cursor moves the same number of times I've pressed `C-n`.

Oddly, flake8 doesn't seem to be taking much of that time. [I can run flake8 on the file in 0.2 seconds](https://gist.github.com/Wilfred/6f8cd34f835907671314).

Issue #154, opened 2013-06-14
Gradually disable Flycheck if there are too many errors


swsnr:
We need to disable on the fly syntax checking if there are too many errors in the current buffer, to avoid hangs and redisplay delays.  See #148 and #153.

We should do so gradually in four steps:
1. Decrease highlighting mode to `columns` (50)
2. Decrease highlighting mode further to `lines` (100)
3. Disable `idle-change` checking (200)
4. Disable any automatic syntax checking (300)
5. Disable Flycheck Mode completely (500)

We need to find reasonable thresholds for each of these steps.  5. should have a high threshold, to allow for explicit manual syntax checks with `C-c ! c` as long as reasonable. The numbers in parenthesis seem reasonable to me.

On each syntax check, determine the number of total errors after parsing, but _before_ highlighting.  Check each of the thresholds, and adapt settings accordingly.  At step 3 and 4., we should probably remove all current errors and interrupt the syntax check, if the threshold is exceeded and the current syntax check was started automatically.

PR #1169, opened 2016-11-22
Cache line numbers to improve performance of flycheck-add-overlay

cpitclaudel:
This is a partial fix for GH-1168.  Related issues: GH-153, GH-154, GH-179,
and GH-476.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2999,15 +2999,18 @@ ERR is a Flycheck error.  MODE may be one of the following symbols:
      Return the line region.
 
 Otherwise signal an error."
-  (pcase mode
-    (`columns (or (flycheck-error-column-region err)
-                  (flycheck-error-line-region err)))
-    (`symbols (or (flycheck-error-thing-region 'symbol err)
+  ;; Ignoring fields speeds up calls to `line-end-position' in
+  ;; `flycheck-error-column-region' and `flycheck-error-line-region'.
+  (let ((inhibit-field-text-motion t))
+    (pcase mode
+      (`columns (or (flycheck-error-column-region err)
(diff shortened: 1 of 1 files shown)


lunaryorn:
Nice!  LGTM.

lunaryorn:
I'll hopefully get to review this soon, I'm quite short of time currently.

cpitclaudel:
No hurry :) Plus, this is a pretty core part of Flycheck, so a thorough review is in order ^^

cpitclaudel:
Updated to not rely on `pcase-let`'s behavior when a pattern doesn't match

cpitclaudel:
Updated again because the byte compiler doesn't like (guard) patterns :/

manuel-uberti:
Wow, thanks for the effort.

lunaryorn:
@cpitclaudel I looked at the comment in #1168 and it seems to suggest that this PR does not contribute much, does it?  Should we still merge it, or rather focus on the other optimisations suggested by the extensive profile in #1168?

cpitclaudel:
Done :)

cpitclaudel:
I'd like to go ahead and merge this soon, as it should be harmless and the performance boost would be welcome. Objections?

cpitclaudel:
> Maybe we should point out with comments before the overlay-recenter calls that they are here for optimization purposes?

I skipped the comments because overlay-recenter doesn't rely have other purposes, but feel free to add them if you think they'd help :) Thanks for the review!

fmdkdd:
@cpitclaudel Yeah you are right, looking at the documention is probably sufficient.

PR #817, opened 2015-12-07
Don't spuriously change checkers

mkpankov:
Currently when editing the buffer, flycheck temporarily switches to `rust` checker as the buffer is not saved. It introduces misleading errors about not being able to find crates and is inconsistent overall.

PR #1171, opened 2016-11-23
Allow rust checker to run when crate root is defined

donkopotamus:
Fixes issue where the `rust` checker is disabled when `flycheck-rust-crate-root` is defined.  

A change was introduced in pull request #817 which disabled this checker if `flycheck-rust-crate-root` was defined.  This change should ensure that:

- the checker can run if `flycheck-rust-crate-root` is defined;
- the checker will not run if the file has not been saved, which should still prevent the checker switching complained about in #817 

(NOTE: I was the original contributor of the `flycheck-rust-crate-root` variable in #417, I'm just returning to using rust again!)

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8745,8 +8745,7 @@ This syntax checker needs Rust 1.7 or newer.  See URL
   :error-parser flycheck-parse-rust
   :error-explainer flycheck-rust-error-explainer
   :modes rust-mode
-  :predicate (lambda ()
-               (and (not flycheck-rust-crate-root) (flycheck-buffer-saved-p))))
+  :predicate flycheck-buffer-saved-p)
 
 (defvar flycheck-sass-scss-cache-directory nil
   "The cache directory for `sass' and `scss'.")
--- a/flycheck.el
+++ b/flycheck.el
@@ -8741,7 +8741,7 @@ This syntax checker needs Rust 1.7 or newer.  See URL
             (option-list "-L" flycheck-rust-library-path concat)
             (eval flycheck-rust-args)
             (eval (or flycheck-rust-crate-root
-                      (flycheck-substitute-argument 'source-inplace 'rust))))
+                      (flycheck-substitute-argument 'source-original 'rust))))
   :error-parser flycheck-parse-rust
   :error-explainer flycheck-rust-error-explainer
   :modes rust-mode


fmdkdd:
So you want to check rust files inside a crate with the plain `rust` checker?  Are these files that are still part of the crate, or are they totally independent but just *happen* to have a `Cargo.toml` above them?

With the proposed fix, both `rust` and `rust-cargo` would be available to run in `rust-mode`, but from what I understand, `rust-cargo` would be the selected checker by default since it appears first in the list of checkers.  Is that what you want?

Unless... there is a subtle situation where `flycheck-rust-crate-root` is set but `(locate-dominating-file (buffer-file-name) "Cargo.toml")` is nil?

donkopotamus:
Yes ... I think the ideal situation should be:

- `rust-cargo` is used if a `Cargo.toml` can be located;
- the plain `rust` checker is used if there is no `Cargo.toml`;

In the latter case the`flycheck-rust-crate-root` variable allows us to check the file while still considering the possibility that its part of a crate rather than a stand alone file.  (It was originally introduced when cargo was very immature).

For myself, I currently have historic source code that is a mix of c, c++ and rust and its still being built via `ninja`, not `cargo`, so there is no `Cargo.toml` to be found.   While I expect to soon split the build between `ninja` and `cargo`, I think it remains useful to make the `rust` checker more flexible in this regard.

### NOTE ###

(I believe pull request #817 was partly in error ... the reason the original submitter found that they kept switching between `rust-cargo` and `rust` checkers was because they would have had a `flycheck-rust-crate-root` in their file local variables when it was irrelevant to the `rust-cargo` checker in any case.  Simply removing it would have solved the problem.)

fmdkdd:
Well, the problem in #817 was more about the `or`, since then the checker was run when the buffer wasn't saved.   I think I understand the fix in #817, but I'm not sure I understand the `or` in #417 (granted, I've been using rust for only a few months), given that the value of `flycheck-rust-crate-root` is actually used in the checker, it seems strange to prevent it to run when it is set nullity.

In any case, you are not reverting #817 but removing the check to `flycheck-rust-crate-root` altogether.

Regardless, I think I understand the code better without this check, so LGTM.

donkopotamus:
@fmdkdd At the time of #417 there was no `rust-cargo` checker at all (cargo was still in its infancy).  The reason for the `or` was that if `flycheck-rust-crate-root` did **not** exist then its a standalone file and you could legitimately check it by doing a `source-inplace` save.  Conversely if it was part of a crate, then you couldn't use a temporary save.

Unfortunately, once the `rust-cargo` checker was introduced above it it caused this flipping.

fmdkdd:
Oh I see.  So you could check the buffer on the fly without actually saving it, but only for standalone files.  Then you added the functionality to check crates by providing a `crate-root`, while maintaining the ability to check individual files.  Then `rust-cargo` was added to check crates through `cargo`, and conflicts arised: a saved buffer would trigger a `rust-cargo` check but unsaved buffers could be checked by `rust` if `flycheck-rust-crate-root` was `nil`.  To fix this, #817 made sure that the `rust` checker would only run on save, but at the same time it meant you could not use the `rust` checker for crates anymore, since it wouldn't run if `flycheck-rust-crate-root` was set.

So, removing the check on `flycheck-rust-crate-root` means you can use the `rust` checker on crates again.  I think I get it now.  Thanks for taking me through it ;)

In my opinion, it would be much simpler to focus the `rust` checker for individual rust files, and to use `rust-cargo` for crates.  But, the functionality was already there, so there's no harm in restoring it.

Now, if the predicate is just `flycheck-buffer-saved-p`, couldn't we rather use `source-original` to avoid creating a temporary file?

fmdkdd:
@cpitclaudel we'll need a LGTM here.  This one is rather benign.  Not sure if it even warrants a line in the CHANGELOG.

cpitclaudel:
@fmdkdd thanks, LGTM. I'll trust your judgment on this :)

PR #1182, opened 2016-12-27
Handle errors from rustc that have no associated spans

fmdkdd:
Errors reported by rustc may not have any associated position in a file, for
instance if the error is related to the project as a whole (see flycheck/flycheck-rust#47).

Previously, we did not collect any error without spans, thus not only
misreporting errors from rustc, but also raising a suspicious checker state.

This commit enhances `flycheck-parse-rust` to collect errors without associated
spans.  By default, these errors are bound to the position (0,0) of the current
buffer.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5733,7 +5733,22 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
               :checker checker
               :buffer buffer
               :filename primary-filename)
-             errors)))))
+             errors)))
+
+        ;; If there are no spans, the error is not associated with a specific
+        ;; file but with the project as a whole.  We still need to report it to
+        ;; the user by emitting a corresponding flycheck-error object.
+        (unless spans
+          (push (flycheck-error-new-at
+                 ;; We have no specific position to attach the error to, so
+                 ;; let's use the top of the file.
+                 1 1
+                 error-level
+                 error-message
+                 :id error-code
+                 :checker checker
+                 :buffer buffer)
+                errors))))
     (nreverse errors)))
 
 


fmdkdd:
@manuel-uberti Thanks :)  Can you approve it as well if you have looked at the code?

On a related note: how do we handle errors with no associated filename or position in other checkers?  Do we pass a special phony filename?  I've left it empty and it seems to work fine, but if we have a consistent way of handling them, might as well follow it.

manuel-uberti:
@fmdkdd I'm no Elisp expert, and the code is simple enough for me to understand, that's why I said LGTM :)

Better if we wait for @cpitclaudel for a proper review and approval.

Wilfred:
For what it's worth, looks good to me too.

flycheck.el
```diff
@@ -5733,7 +5733,22 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
...
+             errors)))
+
+        ;; If there are no spans, the error is not associated with a specific
+        ;; file but with the project as a whole.  We still need to report it to
+        ;; the user by emitting a corresponding flycheck-error object.
+        (unless spans
+          (push (flycheck-error-new-at
+                 ;; We have no specific position to attach the error to, so
+                 ;; let's use the top of the file.
+                 0 0
```
cpitclaudel: Should the column be 0, or 1?

cpitclaudel:
LGTM; this looks like a reasonable approach. I could be a nice entry for a todo list to support errors that do not have a position.

One small thing: we index columns starting at 1, right?  Does the rust checker then add an offset to column numbers for all columns?

flycheck.el
fmdkdd: You're right, it should be 1 per the docs.  Though, it seems flycheck will treat a 0 as if the column was not specified and highlight the whole line in the `columns` highlighting mode.

Same goes for the line argument: 0 or 1 will highlight the same line.

If I introduce a parse error on the first character of the file, the error position reported by the compiler will be (1,1), so we can use that for consistency.

Issue #1174, opened 2016-12-05
typescript-tslint fails on deprecation warning from tslint


asmundg:
Given the following tslint.json

```
{
    "rules": {
        "no-unused-variable": true
    }
}
```

tslint produces the following output:

```
$ tsc --version
Version 2.1.1
$ tslint --version
4.0.1
$ tslint --project .
no-unused-variable is deprecated. Use the tsc compiler options --noUnusedParameters and --noUnusedLocals instead.
```

This causes typescript-tslint to fail:

```
Error from syntax checker typescript-tslint: Unrecognized keyword: "no"
```

```
Syntax checkers for buffer SignalRExecutor.ts in typescript-mode:

  typescript-tide
    - may enable:        yes
    - predicate:         t
    - Typescript server: running
    - Tide mode:         enabled

  typescript-tslint
    - may enable:         yes
    - executable:         Found at c:/Users/asgramme/AppData/Roaming/npm/tslint.cmd
    - configuration file: Found at "c:/Users/asgramme/src/LivePersonaCard/tslint.json"

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled
checkers.

--------------------

Flycheck version: 31snapshot (package: 20161025.551)
Emacs version:    25.1.1
System:           x86_64-w64-mingw32
Window system:    w32
```

Simplify:
```
$ tslint src/<filename>.ts --format=json
no-unused-variable is deprecated. Use the tsc compiler options --noUnusedParameters and --noUnusedLocals instead.

Could not find implementations for the following rules specified in the configuration:
    label-undefined
    no-constructor-vars
    no-duplicate-key
    no-unreachable
    use-strict
Try upgrading TSLint and/or ensuring that you have all necessary custom rules installed.
If TSLint was recently upgraded, you may have old rules configured which need to be cleaned up.
        
[]
```

Simplify:
ESLint all over again... We have following options:
- Ask TSLint developers not to print this warning when using `--format` option or to add warnings to JSON output (or any other non-cli format) for line 1, column 1. This is something that IMO should be done in first place. When somebody requests JSON output, JSON output should be returned.
- Ignore all output text until we get line output that starts with `[`. That is real JSON output. Basically, last line in output.
- Do nothing, developers can remove those unused settings from their `tslint.json` configuration file. Upgrading to mayor version of any tool should be manual task. However, it's not nice user experience for Flycheck users.

asmundg:
[...]

Issue #1163, opened 2016-11-16
Various test suite failures


spwhitton:
Hello,

[Various tests are failing against Debian unstable](https://ci.debian.net/packages/f/flycheck/unstable/amd64/).  I won't have time to fix this anytime soon, so filing an issue in case someone has time.

Thanks.

lunaryorn:
@spwhitton I'm sorry; I don't have time either 😞

fmdkdd:
@spwhitton Should be good?  I see green thumbs on the CI log after 30-3.

spwhitton:
@fmdkdd Thank you for your interest.  Unfortunately, the test are
passing only because I commented out large swaths of the test suite.

-- 
Sean Whitton

fmdkdd:
@spwhitton Hah.  Sorry to hear that.  I managed to fix a few of them with #1187, but that won't affect the Debian CI until the next release (right?).

spwhitton:
@fmdkdd It won't affect it even after the next release unless I
uncomment the tests.  I wasn't planning to do that because they are so
fragile.  But thank you for your work.

-- 
Sean Whitton

fmdkdd:
@spwhitton I see.  Let's keep this issue open then.  Thank you for maintaining Flycheck for Debian.

fmdkdd:
Hey @spwhitton we are now running most integration tests on our CI again, thanks to #1462.

You may be able to re-enable those integration tests on your end, should you want to do so.

spwhitton:
Hello,

On Fri, Jun 01 2018, fmdkdd wrote:

> Hey @spwhitton we are now running most integration tests on our CI
>again, thanks to #1462.
>
> You may be able to re-enable those integration tests on your end,
> should you want to do so.

Thank you for the ping, but given how often new upstream versions of
tools hit Debian unstable, I'd prefer to leave them switched off.
They're flakier for us than for you :)

-- 
Sean Whitton

fmdkdd:
>  given how often new upstream versions of tools hit Debian unstable,

Yeah, that's also why we use a stable release on our side.  I'll be closing this though, since there's not much more we can do on our side.  Feel free to re-open if you feel otherwise.

PR #1187, opened 2017-01-04
Update failing tests

fmdkdd:
I was looking at the tests failures reported in #1163, and managed to fix a few of them.

The rust failures were introduced by #1182.  I did not run the tests locally, and somehow it passed through our CI.

The haskell failures are due to small changes in GHC output.  There are now dots in the output:
```
Error.hs:4:16: error:
    • Couldn't match type ‘Bool’ with ‘[Char]’
      Expected type: String
        Actual type: Bool
    • In the first argument of ‘putStrLn’, namely ‘True’
      In the expression: putStrLn True
      In an equation for ‘foo’: foo = putStrLn True
```
I've added them to the tests as-is, but we could also trim them from the output.  Maybe @flycheck/haskell has an opinion on that.

The typescript failure was due to a deprecated rule, the same issue as #1174.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7425,7 +7425,9 @@ See URL `https://github.com/commercialhaskell/stack'."
             source)
   :error-patterns
   ((warning line-start (file-name) ":" line ":" column ":"
-            (or " " "\n    ") "Warning:" (optional "\n")
+            (or " " "\n    ") "Warning:"
+            (optional " " "[" (id (one-or-more not-newline)) "]")
+            (optional "\n")
             (message
              (one-or-more " ") (one-or-more not-newline)
              (zero-or-more "\n"
@@ -7473,7 +7475,9 @@ See URL `https://www.haskell.org/ghc/'."
             source)
   :error-patterns
   ((warning line-start (file-name) ":" line ":" column ":"
-            (or " " "\n    ") "Warning:" (optional "\n")
+            (or " " "\n    ") "Warning:"
+            (optional " " "[" (id (one-or-more not-newline)) "]")
+            (optional "\n")
             (message
              (one-or-more " ") (one-or-more not-newline)
              (zero-or-more "\n"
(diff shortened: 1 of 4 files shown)


purcell:
Ideally the Haskell patterns would match both the old and new output formats, since people will continue to use the older GHC (7.8, 7.10?) for quite some time, and Flycheck should support it. Is that the case here?

fmdkdd:
@purcell That's a good point.  The changes I've made to the error-parser are backwards-compatible, since I've only added an `optional` rx matcher for catching error IDs.

But the test output will definitely fail with previous versions of GHC.  So users won't be affected, unless they try to run the integration tests with an older GHC.

From what I gather, Flycheck makes no guarantees for backward-compatibility.  If a checker fails after updating Flycheck, we even suggest to update the external tools to a newer version.

Our integration tests are especially version-specific, since the slightest change in compiler output will make them fail.  We don't have any infrastructure in place to handle tests for multiple tool versions.  Should we?  I might open an issue if I don't find the answer clearly stated anywhere.

fmdkdd:
Thanks to flycheck/flycheck-rust#48, I noticed there is a plural form of the useless Rust error message "aborting due to previous error".  I've updated the regexp of that commit to catch these as well.

purcell:
> Our integration tests are especially version-specific, since the slightest change in compiler output will make them fail. We don't have any infrastructure in place to handle tests for multiple tool versions. Should we? I might open an issue if I don't find the answer clearly stated anywhere.

No, I wouldn't personally introduce that just for this case, but we should still do what we can to keep patterns backwards-compatible where it makes sense for core language machinery (e.g. major compiler versions). For pure add-on tools like `eslint`, it's less of a concern: people who use the latest flycheck will also want the latest eslint.

fmdkdd:
> we should still do what we can to keep patterns backwards-compatible where it makes sense

Agreed.  And *if* we had tests for major versions, they would help maintaining backward-compatibility.  That's what I'm proposing in #1190.

cpitclaudel:
Thanks! This all looks good. I replied on the separate thread.

PR #1189, opened 2017-01-05
Fix missing backquote in Tune error list example

dgchurchill:
In the "List all errors" documentation page, the "Tune error list example" seems to be missing a backquote. I think it's safe there for RST because it's in a literal code-bock. I don't know enough elisp to be sure this is correct but it didn't work without it and it works with it (I'm using Emacs 25.1.1).


cpitclaudel:
Thanks!

PR #972, opened 2016-05-17
Display errors from other files (#966)

chrisdone:
This implements the feature requested in #966 and laid out in https://github.com/flycheck/flycheck/pull/970#issuecomment-219734788.

Now:

* There's a new column in the error list, called `File`.

  ``` diff
  +  [("File" 6)
  ```
* We don't discard errors that don't match the current filename (in `flycheck-relevant-error-p`):

   ``` diff
   -       (or (not file-name) (flycheck-same-files-p file-name (buffer-file-name)))
   ```

* We only add overlays for the current buffer in `flycheck-add-overlay`:

   ``` diff
   +  (when (eq (flycheck-error-buffer err) (current-buffer))
   ```
* In `flycheck-next-error-function`, when there are no more errors in the current buffer, we jump to the first error from another file.
* The `flycheck-error-list-goto-error` function uses the same function as the previous item, both use a new function `flycheck-jump-to-error`.
* We add a new face for filenames, `flycheck-error-list-filename`, derived from `buffer-menu-buffer`.

I don't think this is an automatic merge, but it's a straight-forward one. I tried rebasing against master and it worked, but I was unable to get master to load in my Emacs due to a variety of new dependencies which have been added since I last updated flycheck. 😱  So I don't know if it actually runs on your version. Sorry!

Anyway, against `778493e` from 9 months ago it works fine. 🙊 I have to work on other things related to Intero now, but I hope this makes it upstream.

Edit (@lunaryorn): Connects to #966

PR #1191, opened 2017-01-06
Display errors from other files

TyOverby:
As a huge fan of the quickfix list in Vim, when I came to Emacs I was amazed to find flycheck, but sad to see that project-wide errors don't show up in flycheck-list-errors.  My main use-case for a flycheck-like list of errors is to make a change to a function signature, and immediately jump to all my other files where I broke things.

I found @chrisdone's PR implementing this [here](https://github.com/flycheck/flycheck/pull/972), but the PR was abandoned.

**This** PR revives #972 and addresses the code review comments.  I've also rebased and can confirm that all cask tests pass locally.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -797,6 +797,12 @@ This variable is a normal hook.  See Info node `(elisp)Hooks'."
   :group 'flycheck-faces
   :package-version '(flycheck . "0.16"))
 
+(defface flycheck-error-list-filename
+  '((t :inherit buffer-menu-buffer))
+  "Face for filenames in the error list."
+  :group 'flycheck-faces
+  :package-version '(flycheck . "0.25"))
+
 (defface flycheck-error-list-column-number
   '((t :inherit font-lock-constant-face))
   "Face for line numbers in the error list."
@@ -3130,18 +3136,11 @@ Return ERRORS, modified in-place."
 
 Return t if ERR may be shown for the current buffer, or nil
 otherwise."
-  (flycheck-error-with-buffer err
-    (let ((file-name (flycheck-error-filename err))
-          (message (flycheck-error-message err)))
-      (and
-       ;; The error is relevant for the current buffer if it's got no file-name
-       ;; and the current buffer has no file name, too, or if it refers to the
-       ;; same file as the current buffer.
-       (or (and (not file-name) (not buffer-file-name))
-           (flycheck-same-files-p file-name (buffer-file-name)))
-       message
-       (not (string-empty-p message))
-       (flycheck-error-line err)))))
+  (let ((message (flycheck-error-message err)))
+    (and
+     message
+     (not (string-empty-p message))
+     (flycheck-error-line err))))
 
 (defun flycheck-relevant-errors (errors)
   "Filter the relevant errors from ERRORS.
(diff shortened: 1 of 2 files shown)


fmdkdd:
Okay so I've looked at the code and at the previous discussions, and it seems @lunaryorn's comments still hold (see [here](https://github.com/flycheck/flycheck/pull/972#discussion-diff-71131011) and [there](https://github.com/flycheck/flycheck/pull/972#issuecomment-233302415)).

### Confusion between `:buffer` and `:filename`
Trying it out, I can see the new `File` column that indicates the file where the error originated.  But navigating does not open said file, it just jumps me in the same buffer, highlighting errors from another file in the current buffer.  I.e.: if I visit `lib.rs` and rename a public function there, I will have errors in both `lib.rs` and `main.rs`.  All these errors appear in the error list for `lib.rs`, but going down the list I am jumping in the same `lib.rs` buffer, and never going to `main.rs`.  I *see* the red squiggles for the `main.rs` errors in the `lib.rs` buffer.  That's definitely not working correctly.

I think the cause is a confusion as to what the `:buffer` field of  a `flycheck-error` is for.  Per the docstring:
> `buffer'
     The buffer that the error was reported for, as buffer object.

That is, most checkers will put `(current-buffer)` there, so the error is associated with the buffer that ran the checker.  In no way `buffer` indicates the file where the error is located; there is the `:filename` field for that. 

When I visit `lib.rs`, the checker will invoke `cargo rustc` and get back errors for the whole project (well, the selected target, but that includes multiple files at least), and in doing so will create `flycheck-error` objects that have `:buffer` set to the buffer visiting `lib.rs`, but that may have `:filename` set to `lib.rs`, `main.rs` or other files.

Currently, the PR is using `:buffer` where it should use `:filename`, when highlighting and when navigating the error list.  I think using `:filename` is a good start to make sure we are dealing with errors that are relevant to the current buffer.

Or, we can, when constructing `flycheck-error` objects, associate the error with the buffer visiting the corresponding filename.  That is, even if errors are collected when visiting `lib.rs`, we set `:buffer` to `get-file-buffer`.  But that means changing all the parsers that pass `:buffer` directly instead of using `flycheck-error-new-at`.  Using `:filename` when highlighting and navigating might be easier, but using `:buffer` to bind the error with the buffer where it should appear might be a better solution in the long run.

I'd say go with the first option (use `:filename` instead of `:buffer`) to see how it works out.  If it's too messy, we will consider other options.

### Jumping to other buffers in the error list
The error list was made to jump in the same buffer.  Jumping to another buffer, how do we deal with the errors reported by *that* buffer?  That was pointed out in [that comment](https://github.com/flycheck/flycheck/pull/972#r72588007) but never addressed.

I don't have a good solution at the moment.  But I think, for the use case you mention of fixing places after changing a function signature, that keeping the error-list as-is is fine.  Don't bother with the errors from the buffer you are visiting for now.  If I am a user and I can follow the places where my change broke something, fix the errors, and save my buffers.  After that, it will trigger another checker run, and I may have some new errors, that were already present but not reported.  But that's only a minor inconvenience.

If it breaks something, then that's bad.  But we'll see.

Invoking @cpitclaudel, because he may have valuabe insight as well.

fmdkdd:
@TyOverby Please let me know if anything is unclear, or if you want more precise instructions.  We are exploring unknown territory here, so we have to try differents solutions to see how they work out.

TyOverby:
@fmdkdd: thanks for the in-depth comments!  I'm finally able to get my local flycheck branch working with spacemacs, so I've jumped in and attempted to address some of your points.

My most recent commit [Fix filtering for overlay](https://github.com/flycheck/flycheck/pull/1191/commits/12ab6736527b237184238d3b91b9f53df85bbe5b) uses filename checking rather than buffer checking in order to decide if an error overlay should show up on a line in the file.  After running it on one of my rust projects, it appears to be working!

I can also confirm that buffer jumping is not working as it should...  I'll poke around and see if I can find a satisfactory solution.

Once again, I seriously appreciate the help.

fmdkdd:
@TyOverby I can confirm that overlays now appear to be limited to errors concerning the current file.

Your latest commit also allows one to jump to the right buffer when browsing the error list.  That's good progress!

How do you feel about the usability of this feature right now?  Testing it, I can see two issues that makes it a bit crude: 

1. Navigating to another buffer empties the error list.  If I make a change in `lib.rs`, save, browse the error list and go into `main.rs` by pressing `Enter` in the error list in order to make a corresponding change in `main.rs`, then I see the error list for that buffer instead.  As soon as I make a change, the error list is cleared.  If I want to jump to other places that needed fixing, I must first go back to the `lib.rs` buffer, navigate the error list again (where the errors for `lib.rs` have been preserved), and go to the next interesting place in `main.rs`.  Repeat for every error in buffers other than `lib.rs`.
It's a bit jarring that the error disappear, but it stems from the fact that flycheck errors are per buffer, and not shared across a project.

2. Errors in other buffers are not highlighted with overlays.  They appear to be, because the first time the error list opens a buffer, flycheck will run in that buffer as well, and there the overlays will work.  But if that file was already visited, no overlays will appear.
This, again, stems from the fact that flycheck errors are not shared.  Overlays are applied only to the current buffer, and when browsing the error list, other buffers are never instructed to apply overlay for errors that concern them, since they have no knowledge of these errors.

So, if we want to solve these issues, we need to share the errors with other flycheck buffers somehow.  I see three ways to do it:

1. One way to do that would be to have a global list of collected flycheck errors.  Collected errors would go into this global list, and each buffer would then look into this global list for overlays, error list and so on.  That's not enough, since that means we would now get *all* rust errors for a rust buffer for instance, even the errors from other projects.  And I don't think that's a good approach considering that flycheck was made to work per-buffer.  

2. Another solution would be to share the errors when browsing the error list.  If I navigate to an error that belong to `main.rs`, currently we are opening a `main.rs` buffer.  At that point, we should copy all the flycheck errors relevant for that buffer into it.  That way we share the errors with that buffer, and the overlays, and its error list, should contain the errors without re-triggering the checker there as well.

3. Or, we can assume that even if we run a checker per-file, it will return the same errors.  So, when navigating to another buffer from the error list, trigger the checker in that buffer as well, so it can collect its own errors, trigger its overlays, and populate its error list.  No need to share errors.  That's the brute force approach, and it can be costly, though in the cargo's case, we would not re-trigger a full recompilation, so it might just work.

I feel 2 and 3 can work, and 3 might be the easiest to get right, provided it's not too slow.  But 2 should also work.  Care to try?

TyOverby:
> I can confirm that overlays now appear to be limited to errors concerning the current file.

Yay!  Also, for what it's worth, your `(load-file)` workaround is no longer working for me for loading replacement packages in spacemacs.  My solution for now is symlinking `~/workspace/emacs/flycheck` to `~/.emacs.d/elpa/flycheck-20170112.1646`.  It's a terrible hack, but at least I'm no longer loading two versions of flycheck into my process...

> Your latest commit also allows one to jump to the right buffer when browsing the error list. That's good progress!

Yeah, that one was surprisingly easy.  I had a few implementations of that fix, but I think the commit that I pushed is the least invasive of the lot.

(I'll repond on the important bit later when I have more time)

fmdkdd:
@TyOverby Any update on this?  It's promising work that I would like to see continue :)

TyOverby:
Unfortunately I ran into so many other bugs in emacs / spacemacs that I gave up and moved back to neovim :(  If someone wants to continue this work, I'd be happy to help explain why I did certain things, but I'm otherwise going to abandon this PR.

fmdkdd:
That's too bad.  Thanks for update though, and happy Vim'ing :)

PR #1204, opened 2017-02-07
Add Emacs Lisp other initialization option.

QBobWatson:
This adds a simple option to the emacs-lisp checker to have flycheck evaluate an arbitrary piece of code before checking a file.  I use it to (require 'my-globals).

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6761,7 +6761,8 @@ This variable has no effect, if
 
 (defun flycheck-emacs-lisp-feature-requires (value)
   "Option VALUE filter for `flycheck-def-option-var'."
-  (mapconcat (lambda (feature) (format "(require '%s)" feature)) value " "))
+  (when value
+    (mapconcat (lambda (feature) (format "(require '%s)" feature)) value " ")))
 
 (flycheck-define-checker emacs-lisp
   "An Emacs Lisp syntax checker using the Emacs Lisp Byte compiler.
--- a/flycheck.el
+++ b/flycheck.el
@@ -6775,7 +6775,7 @@ See Info Node `(elisp)Byte Compilation'."
                     flycheck-option-emacs-lisp-package-user-dir)
             (option "--eval" flycheck-emacs-lisp-initialize-packages nil
                     flycheck-option-emacs-lisp-package-initialize)
-            (option "--eval" flycheck-emacs-lisp-other-initialization
+            (option "--eval" flycheck-emacs-lisp-other-initialization nil
                     flycheck-sexp-to-string)
             "--eval" (eval flycheck-emacs-lisp-check-form)
             "--"
--- a/flycheck.el
+++ b/flycheck.el
@@ -6752,14 +6752,16 @@ This variable has no effect, if
     (when value
       (flycheck-sexp-to-string `(setq package-user-dir ,value)))))
 
-(flycheck-def-option-var flycheck-emacs-lisp-other-initialization
+(flycheck-def-option-var flycheck-emacs-lisp-require-features
     nil emacs-lisp
-  "Other initialization to perform before byte-compiling.
-
-This can be set to any form to evaluate."
-  :type 'sexp
+  "List of features to require before checking."
+  :type '(repeat (string :tag "Feature to require"))
   :risky t
   :package-version '(flycheck . "0.14"))
+
+(defun flycheck-emacs-lisp-feature-requires (value)
+  "Option VALUE filter for `flycheck-def-option-var'."
+  (mapconcat (lambda (feature) (format "(require '%s)" feature)) value " "))
 
 (flycheck-define-checker emacs-lisp
   "An Emacs Lisp syntax checker using the Emacs Lisp Byte compiler.
@@ -6775,8 +6777,8 @@ See Info Node `(elisp)Byte Compilation'."
                     flycheck-option-emacs-lisp-package-user-dir)
             (option "--eval" flycheck-emacs-lisp-initialize-packages nil
                     flycheck-option-emacs-lisp-package-initialize)
-            (option "--eval" flycheck-emacs-lisp-other-initialization nil
-                    flycheck-sexp-to-string)
+            (option "--eval" flycheck-emacs-lisp-require-features nil
+                    flycheck-emacs-lisp-feature-requires)
             "--eval" (eval flycheck-emacs-lisp-check-form)
             "--"
             source-inplace)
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Thanks. I'm not sure how much I like the idea of running arbitrary code, though.  Can you clarify the use case? Is it for your .emacs.d? If so, you could locally advise `flycheck-option-emacs-lisp-package-initialize`, for example.

QBobWatson:
I just want to require the file that defines my load-path.  How about this?  It's just an option for extra requires.

cpitclaudel:
Understood.  I'm grateful for the contribution, but I'm just not sure how generally useful this feature would be.  Concretely, the only use is for init files, right?

QBobWatson:
Yes, that's correct. But most Emacs users will only ever open an emacs-lisp buffer when editing their config files, and some people have very extensive configuration :)

fmdkdd:
Sorry for the delay.  I can sense @cpitclaudel's hesitation for this feature here, and I, too, am on the fence.

On the one hand, the `emacs-lisp` checker was mostly made with packages in mind, not user files.  There are warnings like "add ;;; commentary section" or "missing (provide 'package)" that make no sense for init files. 

On the other hand, @QBobWatson makes a valid point: most Emacs users who edit emacs-lisp do it for their configuration, not for hacking packages.  If they are also Flycheck, users, they might want to use the `emacs-lisp` checker as it provides, besides the spurious warnings, valid feedback for incorrect syntax.

In that sense, this PR is a step in reducing the amount of spurious warnings for this use case.

In fact, I would be interesting in pursuing further.  @QBobWatson, even with this patch,  you do get `add ;;; commentary section` for your init file, don't you?  If you can update this PR to make the spurious warnings disappear, we'll glady reconsider merging it.

QBobWatson:
Actually, the `add ;;; commentary section` warnings and the like are in the `emacs-lisp-checkdoc` checker, which I have disabled for config files.  I'm not sure where `missing (provide 'package')` comes from, but I'm not getting those warnings anyway.

fmdkdd:
Oooh.  You are right, I was confused there.  So, once you add the `require`, can you actually check your whole init file without errors or spurious warnings?

QBobWatson:
That's correct -- once I add the `require` and disable `emacs-lisp-checkdoc`, I get no errors or warnings.

fmdkdd:
Then I would say it's worth it.  What say you, @cpitclaudel?

cpitclaudel:
> What say you, @cpitclaudel?

I don't know.  I don't really buy it.  @QBobWatson, your situation is pretty specific: part of your configuration seems to a be a proper, `require`-able file that your other configuration files depend on. If so, why not just `require` it in other configuration files?

> On the other hand, @QBobWatson makes a valid point: most Emacs users who edit emacs-lisp do it for their configuration, not for hacking packages. If they are also Flycheck users, they might want to use the emacs-lisp checker as it provides, besides the spurious warnings, valid feedback for incorrect syntax.

Fair enough.  But in that case do we actually need such a general setting?  Wouldn't it be enough to just add a list or extra packages?

An alternative solution would be for Flycheck to set a variable while checking ELisp files. Something like `flycheck-elisp-in-process`.  Then users could write:

```
(when (bound-and-true-p 'flycheck-elisp-in-process)
  … extra init code here …)
```

fmdkdd:
>  Wouldn't it be enough to just add a list of extra packages?

The latest commit just adds a list of features to be required.  Aren't they the same?

@QBobWatson can you give us a glimpse of your init file to see how you would use that feature, and what exact errors or warnings it allows you to get rid of?

It seems the crux of the issue is whether adding such a feature would be beneficial to many Flycheck users, or for those who have set up their init files in a way similar to yours.  And, whether there are alternative way of achieving the same result.

cpitclaudel:
> The latest commit just adds a list of features to be required. Aren't they the same?

They are, sorry :)

Thinking more about this: why isn't the following sufficient?

```
(cl-eval-when (compile)
  (require 'my-globals))
```

QBobWatson:
It was a couple years ago that I set all this up, but here's what I remember of my reasoning.

I have about 50 init files for different subsystems, which are all `load`ed by `init.el`.  Now, `init.el` requires `my-globals`, so `(require 'my-globals)` is not needed in my init files at runtime or compile time.  So the only reason to `(require 'my-globals)` in those files would be for flycheck.

Of course I could do this, but my opinion is that the emacs-lisp checker should be able to set itself up to check the file in the same context as it will be used, to the extent this is possible.  Initialization files are often not run as standalone packages.  Since the change to flycheck's code is minimal, I thought it would be worth adding.

cpitclaudel:
> Of course I could do this, but my opinion is that the emacs-lisp checker should be able to set itself up to check the file in the same context as it will be used,

Thanks for clarifying.  What confuses me is that you'll end up using this:

```
;; -*- flycheck-emacs-lisp-require-features: (my-globals) -*-
```

… instead of that:

```
(require 'my-globals)
```

Is there any reason to prefer the first one? (especially given that it'll trigger a risky-variable warning).

QBobWatson:
> Is there any reason to prefer the first one? (especially given that it'll trigger a risky-variable warning).

I have my config set up to do the `flycheck-emacs-lisp-require-features` and disable `emacs-lisp-checkdoc` automatically for all the config files in that directory.  This happens in `emacs-lisp-mode-hook`.

cpitclaudel:
I think I'll defer to @fmdkdd, but I'm not sold.  This sounds like a very specific setup, and it's not clear how useful that setting would be for others (especially since `require` is just one more line)

fmdkdd:
>  it's not clear how useful that setting would be for others

You hit the nail on the head.

I'd be happy to help you make your config neater @QBobWatson, but the workarounds suggested in this thread look sufficient for this use case to me.  The fact that it's a slightly risky option (as it tells Emacs to load arbitrary code, after all) doesn't help either. 

Unless we know for sure it's a major pain point for using the `emacs-lisp` checker, or unless you can think of another usefulness for this option, it is not enough to warrant inclusion.

QBobWatson:
> Unless we know for sure it's a major pain point for using the emacs-lisp checker, or unless you can think of another usefulness for this option, it is not enough to warrant inclusion.

I understand your point.  It's probably best if I just maintain a patch for my own use.

fmdkdd:
@QBobWatson Thank you for taking the time to discuss this feature with us.

Issue #1198, opened 2017-01-27
selecting checker with erlang-mode is too broad


massemanet:
flycheck (not unreasonably) tries to run all buffers with erlang-mode through the erlang compiler. Alas, it is quite common to have files that use erlang-mode but does not compile (e.g. header and config files).
  Adding something like this;
```:predicate (lambda () (string-suffix-p ".erl" (buffer-name)))```
  to ```flycheck-define-checker``` improves things.

cpitclaudel:
Thanks, this sounds like a good idea. I don't have much erlang experience, however, so I can't comment on the technical part.  Can you open a PR with this change, and an adequate new entry? For bonus points, pinging the original author of the elixir checker for confirmation would be awesome.

Thanks!

PR #1205, opened 2017-02-07
use erlc on .erl only

massemanet:
flycheck (not unreasonably) tries to run all buffers with erlang-mode through the erlang compiler. Alas, it is quite common to have files that use erlang-mode but does not compile (e.g. header and config files).

also see #1198

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,7 @@
     ``go-vet`` and ``go-errcheck`` as well.
   - Add a revert function to ``flycheck-verify-setup``, so hitting
     ``g`` reloads the buffer.
+  - Make sure the erlang compiler is only run on compilable files.
 
 30 (Oct 12, 2016)
 =================
--- a/flycheck.el
+++ b/flycheck.el
@@ -6968,7 +6968,7 @@ See URL `http://www.erlang.org/'."
   ((warning line-start (file-name) ":" line ": Warning:" (message) line-end)
    (error line-start (file-name) ":" line ": " (message) line-end))
   :modes erlang-mode
-  :predicate (lambda () (string-suffix-p ".erl" (buffer-file-name))))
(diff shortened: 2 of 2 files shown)


fmdkdd:
That looks harmless.  Except you lose the ability to check Erlang buffers not associated with files, and Erlang buffers that do not have files ending in `.erl`.  Though this might be a lesser inconvenience than having spurious errors for headers.

I have never used Erlang, but I find it curious that the Erlang compiler doesn't checks header files though.  Can't they have errors?  Is a separate tool used to validate header files?

`git blame` tells me @syl20bnr wrote the checker (in 2013), and @yzprofile added the library-path variable (in 2015).  They may have useful insight (or they may never this thread ;).

massemanet:
The erlang compiler can compile several different languages. It decides which kind of compilation to do based on the file name extension; a file with erlang source code must end with .erl.
  Thus, it does not make sense to have erlang code in a file that does not end with .erl. Header files can not be checked separately because they are not compilable. 
  Checking buffers without a backing file does not work because (as far as I can tell) there is no way to make flycheck create a temp file with the .erl extension.

fmdkdd:
Thanks for enlightening me.  So, currently the Erlang checker will run on header files by itself (and all files in erlang-mode), and will produce spurious errors?  Or does it just fail silently?

In either case, I agree that not running the checker is best.  In fact, you should probably use the `:enabled` checker property for that instead of `:predicate`, since `:enabled` is run once for the buffer whereas `:predicate` is run every time a syntax check is triggerred (i.e., after every text change by default).

massemanet:
Currently, it runs on all files in erlang-mode, and fails noisily on all but the .erl files.

  `:enabled` is indeed what I want. Thanks for pointing it out. I will update the PR soonish.

fmdkdd:
Thank you very much for your contribution, and happy hacking 👍

cpitclaudel:
@fmdkdd @massemanet 👍 🎉

PR #1206, opened 2017-02-07
`rust-cargo` checks tests, examples and benches targets

fmdkdd:
Previously, `rust-cargo` was unable to check integration tests and other
optional Cargo targets.  For a description of the conventional Cargo layout,
see:

http://doc.crates.io/manifest.html#the-project-layout

This commits enhances `rust-cargo` to check all files under the conventional
layout, by using `flycheck-rust-crate-type` as the target type, and
`flycheck-rust-binary-name` as the target name.

In addition, it adds :enabled and :verify properties to `rust-cargo` to ensure
these variables are correctly set, and to inform the user in case they are not.

Integration tests are added to check the support for the full conventional
layout.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -700,7 +700,7 @@ This variable is a normal hook.  See Info node `(elisp)Hooks'."
   "Functions to run if the Flycheck status changed.
 
 This hook is run whenever the status of Flycheck changes.  Each
-hook function takes the status symbol as sinlge argument, as
+hook function takes the status symbol as single argument, as
 given to `flycheck-report-status', which see.
 
 This variable is a abnormal hook.  See Info
@@ -2468,7 +2468,7 @@ buffer-local value of `flycheck-disabled-checkers'."
 
 Set `flycheck-current-syntax-check' accordingly."
   ;; Allocate the current syntax check *before* starting it.  This allows for
-  ;; synchronous checks, which call the status callback immediately in there
+  ;; synchronous checks, which call the status callback immediately in their
   ;; start function.
   (let* ((check (flycheck-syntax-check-new
                  :buffer (current-buffer)
@@ -8785,28 +8785,45 @@ if it is not modified, i.e. after it has been saved."
 (flycheck-def-option-var flycheck-rust-crate-type "lib" (rust-cargo rust)
   "The type of the Rust Crate to check.
 
-The value of this variable is a string denoting the crate type,
-for the `--crate-type' flag."
-  :type 'string
+For `rust-cargo', the value should be a string denoting the
+target type passed to Cargo.  See
+`flycheck-rust-valid-crate-type-p' for the list of allowed
+values.
+
+For `rust', the value should be a string denoting the crate type
+for the `--crate-type' flag of rustc."
+  :type '(choice (const :tag "nil (rust/rust-cargo)" nil)
+                 (const :tag "lib (rust/rust-cargo)" "lib")
+                 (const :tag "bin (rust/rust-cargo)" "bin")
+                 (const :tag "example (rust-cargo)" "example")
+                 (const :tag "test (rust-cargo)" "test")
+                 (const :tag "bench (rust-cargo)" "bench")
+                 (const :tag "rlib (rust)" "rlib")
+                 (const :tag "dylib (rust)" "dylib")
+                 (const :tag "cdylib (rust)" "cdylib")
+                 (const :tag "staticlib (rust)" "staticlib")
+                 (const :tag "metadata (rust)" "metadata"))
   :safe #'stringp
   :package-version '(flycheck . "0.20"))
 (make-variable-buffer-local 'flycheck-rust-crate-type)
 
 (flycheck-def-option-var flycheck-rust-binary-name nil rust-cargo
-  "The name of the binary to pass to `cargo rustc --bin'.
+  "The name of the binary to pass to `cargo rustc --CRATE-TYPE'.
 
 The value of this variable is a string denoting the name of the
-binary to build: Either the name of the crate, or the name of one
-of the files under `src/bin'.
+target to check: usually the name of the crate, or the name of
+one of the files under `src/bin', `tests', `examples' or
+`benches'.
 
-This variable is used only when `flycheck-rust-crate-type' is
-`bin', and is only useful for crates with multiple targets."
+This always requires a non-nil value, unless
+`flycheck-rust-crate-type' is `lib' or nil, in which case it is
+ignored."
   :type 'string
   :safe #'stringp
   :package-version '(flycheck . "28"))
 (make-variable-buffer-local 'flycheck-rust-binary-name)
 
-(flycheck-def-option-var flycheck-rust-library-path nil (rust-cargo rust)
+(flycheck-def-option-var flycheck-rust-library-path nil rust
   "A list of library directories for Rust.
 
 The value of this variable is a list of strings, where each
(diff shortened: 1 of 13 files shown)


manuel-uberti:
I let others more qualified do the review, but you're making an impressive work for Rust support in Flycheck. Great stuff.

fmdkdd:
@manuel-uberti Well, thank you!  I'm glad to help.

fmdkdd:
@cpitclaudel Any chance of reviewing this one soon?  The tests are lengthy, but they all do the smae thing.  I really need a pair of eyes on the changes to the checker, especially `:verify`.

cpitclaudel:
I'll do that right away. Sorry for the delay.

flycheck.el
```diff
@@ -8785,28 +8785,35 @@ if it is not modified, i.e. after it has been saved."
...
 
-The value of this variable is a string denoting the crate type,
-for the `--crate-type' flag."
+For `rust-cargo', the value should be a string denoting the
+target type passed to Cargo.  See
+`flycheck-rust-valid-crate-type-p' for the list of allowed
+values.
+
+For `rust', the value should be a string denoting the crate type
+for the `--crate-type' flag of rustc."
```
cpitclaudel: Any reason to make this a string rather than an explicit enumeration (with a `:choice` of symbols)?  We could `intern` old values for compatibility.

flycheck.el
```diff
@@ -8831,32 +8838,71 @@ Relative paths are relative to the file being checked."
                  (flycheck-error-message err)))
               errors))
 
+(defun flycheck-rust-valid-crate-type-p (crate-type)
+  "Whether CRATE-TYPE is a valid target type for Cargo.
+
+A valid Cargo target type is one of `lib', `bin', `example',
+`test' or `bench'."
+  (member crate-type '("lib" "bin" "example" "test" "bench")))
+
```
cpitclaudel: (then this wouldn't be needed)

flycheck.el
```diff
@@ -8831,32 +8838,71 @@ Relative paths are relative to the file being checked."
...
-  :predicate (lambda ()
-               ;; Since we build the entire project with cargo rustc we require
-               ;; that the buffer is saved.  And of course, we also need a Cargo
-               ;; file :)
-               (and (flycheck-buffer-saved-p)
-                    (locate-dominating-file (buffer-file-name) "Cargo.toml"))))
+  :predicate flycheck-buffer-saved-p
+  :enabled (lambda ()
+             (and (locate-dominating-file (buffer-file-name) "Cargo.toml")
+                  (flycheck-rust-valid-crate-type-p flycheck-rust-crate-type)
```
cpitclaudel: I'm not sure that this type check belongs here — we don't do this for other checkers, AFAIK.

flycheck.el
```diff
@@ -8831,32 +8838,71 @@ Relative paths are relative to the file being checked."
...
+                  (or (string= flycheck-rust-crate-type "lib")
+                      flycheck-rust-binary-name)))
```
cpitclaudel: Does this mean that flycheck won't run by default unless a flycheck-rust-binary-name is buffer- or dir-locally set?

flycheck.el
```diff
@@ -8831,32 +8838,71 @@ Relative paths are relative to the file being checked."
...
+  :verify (lambda (_)
+            (let* ((toml (locate-dominating-file
+                          (buffer-file-name) "Cargo.toml"))
```
cpitclaudel: I think this will throw errors when buffer-file-name is nil.

flycheck.el
```diff
@@ -8831,32 +8838,71 @@ Relative paths are relative to the file being checked."
...
+                   (crate-type flycheck-rust-crate-type)
```
cpitclaudel: Do you need this alias here?

cpitclaudel:
This looks good (I left a bunch of comments), but I'm surprised: does this mean the rust checker won't be usable anymore without an explicit flycheck-rust-binary-name?

flycheck.el
fmdkdd: You are right, and this change is too restrictive as `cargo` will build the default target if you don't pass any flag.

fmdkdd: Addendum: `cargo` will build the default target even when there are multiple targets available, but `cargo rustc` (which we use in the checker) will complain when there are multiple targets.  Still, I can change it so that we don't pass any flag when the variables are set to nil.

flycheck.el
fmdkdd: The check to `rust-valid-crate-type`?  Well, `:enabled` is rather recent, so not all checkers use it.  My thinking was that if we enable the checker with an invalid crate type, the user will get a suspicious checker state.  Since we have `:verify`, we might as well disable a checker with an invalid configuration.

Though there's an unforeseen side-effect: disabling the checker will allow for the plain `rust` checker to run on the file.  So I'm not sure anymore if that's a better situation :/

fmdkdd:
I've pushed changes to allow the checker to run without a binary name and without a crate type.   I've also added a test to catch this situation.

I've made accompanying changes to `:enabled` and `:verify`, dropped the superfluous aliases, and checker for `(buffer-file-name)`.

Remaining issues: 
1. Should we use `choice` for `flycheck-rust-crate-type`?
2. Should we disable the checker in the case of an invalid user configuration?

cpitclaudel:
Thanks!

> Should we use choice for flycheck-rust-crate-type?

I'll defer to you on this.  One last question that might orient the choice: if the two tools take different values, should the option make it possible to specify two different values too?  As in, should we have an second option that defaults to the value of the first one?

> Should we disable the checker in the case of an invalid user configuration?

I'd like to keep quick-running status checks in :verify rather than in :enabled, since they cost little and don't require reverting the buffer to redo the checks.  In the long run it would be nice to add a "re-check" button in the status buffer.

fmdkdd:
> should we have an second option that defaults to the value of the first one?

I don't think there's a realistic use case where you would want to have the rust and rust-cargo checkers running on the same file.  So, there's no need for setting two values.  For the record, I would rather use `choice` and avoid `flycheck-rust-valid-crate-type-p` for simplicity, but since we have these two sets of values, I really think leaving it as a string is best here.

> I'd like to keep quick-running status checks in :verify rather than in :enabled

My idea in adding these checks to `:enabled` was to 1) avoid doing unnecessary checks in `:predicate` and 2) avoid running the checker with a bogus configuration.

If the user sets the `flycheck-rust-crate-type` and `flycheck-rust-binary-name` incorrectly, they are greeted with a suspicious checker state message and an advice to update Cargo.  This is not really helpful, since the error lies in the user configuration of these Flycheck variables.

But, now that I think of it, we won't catch "correct" configurations that do not correspond to valid targets in the project.  So, after all, I've revised my judgement and think it might be best to pass the variables to Cargo and let the user deal with it.

Since `flycheck-rust` will take care of setting the variables correstly most of the time, this shouldn't be an issue.  And disabling the checker is a bit unpractical.

I've cut the `:enabled` checks to only include the existence of the `Cargo.toml` file.  I've left `:verify` as is.  What do you think?

cpitclaudel:
Looks good, thanks.  I'm still not entirely sure why `:choice` wouldn't be as good as string (offering the same list of choices as we currently check against), but I'm happy to trust you on this :)

fmdkdd:
1. Say we put the list `nil|lib|bin|example|test|bench` as `:choice` for `flycheck-rust-crate-type`.  Is the user able to put `rlib` in the same variable?  Not from Customize I guess, but still from `setq`?  They must be able to set values accepted by the plain `rust` checker as well.  If they can't, this is too restrictive.  If they can, then we must still check in `:verify` if the value is one that works for `rust-cargo`, so we keep `flycheck-rust-crate-type-p`.

2. Say we put the list `nil|example|test|bench|bin|lib|rlib|dylib|cdylib|staticlib|metadata` instead, merging the two sets of values.  We still have to check the subset allowed by `rust-cargo` with the predicate.

So unless I'm mistaken, having `:choice` does not mean we can get rid of the predicate.  But I guess option 2 is a bit nicer for users of Customize.

cpitclaudel:
I think that's correct :) It still requires a predicate, but option 2 is nicer for users of customize (and we can add text labels that indicate for which option which checker(s) it's valid for.

fmdkdd:
Is this what you had in mind?

```diff
-  :type 'string
+  :type '(choice (const :tag "nil (rust/rust-cargo)" nil)
+                 (const :tag "lib (rust/rust-cargo)" "lib")
+                 (const :tag "bin (rust/rust-cargo)" "bin")
+                 (const :tag "example (rust-cargo)" "example")
+                 (const :tag "test (rust-cargo)" "test")
+                 (const :tag "bench (rust-cargo)" "bench")
+                 (const :tag "rlib (rust)" "rlib")
+                 (const :tag "dylib (rust)" "dylib")
+                 (const :tag "cdylib (rust)" "cdylib")
+                 (const :tag "staticlib (rust)" "staticlib")
+                 (const :tag "metadata (rust)" "metadata"))
```

PR #1209, opened 2017-02-10
 Clean "stdin" filename from flake8 errors

marsam:
flake8 checker fails when a buffer doesn't have a `buffer-file-name`.

Also, I've update the flake8 tests, to work with flake8 v3.3.0:

```
make SELECTOR='(tag checker-python-flake8)' integ
cask exec emacs -Q --batch -L .  --load test/flycheck-test --load test/run.el -f flycheck-run-tests-main '(and (tag external-tool) (tag checker-python-flake8))'
Running tests on Emacs 26.0.50.3, built at 2017-02-09
Running 2 tests (2017-02-10 17:13:12-0500)
   passed  1/2  flycheck-define-checker/python-flake8/default
   passed  2/2  flycheck-define-checker/python-flake8/syntax-error

Ran 2 tests, 2 results as expected (2017-02-10 17:13:15-0500)
```

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8448,7 +8448,6 @@ Requires Flake8 3.0 or newer. See URL
 `https://flake8.readthedocs.io/'."
   :command ("flake8"
             "--format=default"
-            "--stdin-display-name" source-original
             (config-file "--config" flycheck-flake8rc)
             (option "--max-complexity" flycheck-flake8-maximum-complexity nil
                     flycheck-option-int)
@@ -8458,11 +8457,10 @@ Requires Flake8 3.0 or newer. See URL
   :standard-input t
   :error-filter (lambda (errors)
                   (let ((errors (flycheck-sanitize-errors errors)))
-                    (seq-do #'flycheck-flake8-fix-error-level errors)
-                    errors))
+                    (seq-map #'flycheck-flake8-fix-error-level errors)))
   :error-patterns
   ((warning line-start
-            (file-name) ":" line ":" (optional column ":") " "
+            "stdin:" line ":" (optional column ":") " "
             (id (one-or-more (any alpha)) (one-or-more digit)) " "
             (message (one-or-more not-newline))
             line-end))
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3629,7 +3629,7 @@ Why not:
   (let ((python-indent-guess-indent-offset nil))       ; Silence Python Mode!
     (flycheck-ert-should-syntax-check
      "language/python/syntax-error.py" 'python-mode
-     '(3 13 error "SyntaxError: invalid syntax" :id "E999"
+     '(3 12 error "SyntaxError: invalid syntax" :id "E999"
          :checker python-flake8))))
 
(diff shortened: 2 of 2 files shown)


fmdkdd:
I think there's another way to fix this, which is to get rid of the `--stdin-display-name`:
```diff
@@ -8283,7 +8283,6 @@ Requires Flake8 3.0 or newer. See URL
 `https://flake8.readthedocs.io/'."
   :command ("flake8"
             "--format=default"
-            "--stdin-display-name" source-original
             (config-file "--config" flycheck-flake8rc)
             (option "--max-complexity" flycheck-flake8-maximum-complexity nil
                     flycheck-option-int)
@@ -8297,7 +8296,7 @@ Requires Flake8 3.0 or newer. See URL
                     errors))
   :error-patterns
   ((warning line-start
-            (file-name) ":" line ":" (optional column ":") " "
+            "stdin" ":" line ":" (optional column ":") " "
             (id (one-or-more (any alpha)) (one-or-more digit)) " "
             (message (one-or-more not-newline))
             line-end))
```
Now all flake8 errors have `nil` as `error-filename`, which has the same effect.

marsam:
I did not notice that flag. I think if the checker can be simplified even further; the tests pass and it has not given me problems. Tell me if I should include this change in this pull request.

``` diff
@@ -8282,7 +8282,6 @@ flycheck-flake8-fix-error-level
 `https://flake8.readthedocs.io/'."
   :command ("flake8"
             "--format=default"
-            "--stdin-display-name" source-original
             (config-file "--config" flycheck-flake8rc)
             (option "--max-complexity" flycheck-flake8-maximum-complexity nil
                     flycheck-option-int)
@@ -8291,13 +8290,10 @@ flycheck-flake8-fix-error-level
             "-")
   :standard-input t
   :error-filter (lambda (errors)
-                  (let ((errors (flycheck-sanitize-errors
-                                 (flycheck-remove-error-file-names "stdin" errors))))
-                    (seq-do #'flycheck-flake8-fix-error-level errors)
-                    errors))
+                  (seq-map #'flycheck-flake8-fix-error-level errors))
   :error-patterns
   ((warning line-start
-            (file-name) ":" line ":" (optional column ":") " "
+            "stdin:" line ":" (optional column ":") " "
             (id (one-or-more (any alpha)) (one-or-more digit)) " "
             (message (one-or-more not-newline))
             line-end))
```

cpitclaudel:
Thanks.  This further change looks good, but I'm now confused as to why the checker definition originally included this.  Maybe the original implementation hoped to convey information to flake8 that way? Looking at the implementation though (https://gitlab.com/pycqa/flake8/commit/02bcbee245b9b5a0c75345ada0f1d7e70d19b3c0) flake8 doesn't use the info.

Bottom line: this change looks good, including the further simplification.  Don't you need to keep `-sanitize-errors`, though?

cpitclaudel:
@marsam Do you want to include the further patch you posted in your PR, and then we can merge it? :)

marsam:
Sorry for the late response. I've squashed the commits.

cpitclaudel:
Thanks!

Issue #1210, opened 2017-02-11
Inline error messages in buffers


fmdkdd:
So it seems that Sublime text has decent Rust syntax checking, see: [sublime-rust](https://github.com/rust-lang/sublime-rust).  Especially, errors are displayed inline, like so:

![inline errors animation](https://cloud.githubusercontent.com/assets/936006/18091147/938cb244-6ebf-11e6-9db1-b74e5bf4eebc.gif)

The way this works is by using 'phantom' lines: lines that are not buffered, and not part of the file you are editing.  This is adequate for annotations. 

I was wondering if that was possible to do with Emacs (or rather, what's the easiest way to have the same functionality).  I've looked at overlays.  It's rather easy to hack `flycheck-add-overlay` to include the message string with a face to match the error level:

![screen](https://cloud.githubusercontent.com/assets/340129/22855044/89613c22-f079-11e6-8fd1-5da4a407bfbc.png)

But I can't figure out how to add the overlay *below* the line of the error, *without* overlapping the code that's in that line.  I want the overlay to stand on its own line, like the phantoms of Sublime.

Has anyone seen similar functionality in Emacs?  Any pointers?

cpitclaudel:
Yup, I wrote a [package](https://github.com/cpitclaudel/quick-peek) that does this. It isn't on MELPA yet, but help with documenting it briefly and putting it there would be very welcome :)

Here's an animation:

![](https://raw.githubusercontent.com/cpitclaudel/company-coq/master/img/inline-docs.gif)

cpitclaudel:
(Alternatively, you can just draw inspiration from the implementation :)

fmdkdd:
Haha great!  I will try it out.

fmdkdd:
Ah yes of course, I should have added the overlay after the line.  Glad I asked in any case, as quick-peek fits the use case neatly.

Now, there multiple ways to integrate inline error messages ("phantoms") to Flycheck:
1. As another highlighting mode, `'inline`.  In addition to squiggly lines, phantoms are added as in sublime-rust.  Essentially a derivative of `flycheck-add-overlay` in `flycheck-process-error-functions`.
2. As an alternate `flycheck-display-error-function`: the phantoms appear only when point is on an error (here `quick-peek` does the job perfectly).
3. On another keybinding to show/hide phantoms for all the errors in the buffer

[...]

Issue #788, opened 2015-11-13
Only display errors after point has moved


swsnr:
Currently we start error display unconditionally in `post-command-hook`, even if the command didn't change `point`.  This leads to a slightly awkward user experience (see https://github.com/flycheck/flycheck-pos-tip/issues/7).

At some point we should track point during error display and only display errors if the command actually moved point.

lunaryorn:
@PierreR I won't work on this any time soon.  As far as I'm concerned, there are many way more important things to do in Flycheck.  Maybe @cpitclaudel takes this.

Otherwise you'll have to wait a long time for me to fix it (sorry).  A pull request would be welcome.

cpitclaudel:
I could look into this (I have a paper deadline next Friday though, so probably not before that) :) I've had a quick look just now, but it actually caused me to run into another issue; I'll post to flycheck-pos-tip.

@lunaryorn,Would it be enough to save the point in a buffer-local variable? Won't we run into trouble when multiple windows show the same buffer? (maybe we're ok with the current behaviour persisting in that case)

lunaryorn:
@cpitclaudel  I haven't yet thought about this at all.   I don't know how we could fix this.

If we used a buffer local variable, we should probably reset it when the active window changes.  Only one window can be active at a time, and we'd only need to track point position in the active window.

lunaryorn:
@cpitclaudel Do we have any news on this issue?  Should we try and fix it?

cpitclaudel:
Definitely; I've just been very busy of late.

jimmywongroo:
Is there any movement or branch up for this issue?

cpitclaudel:
Nope, I haven't touched it, and lunaryorn has transferred maintainership of this repo to @fmdkdd and me (and a large team of other people ^^).  I haven't had much time lately. On the other hand, this should be reasonably fixable, so if someone took a crack at it I'd be happy to review/shepherd a PR.

swsnr:
This issue wasn't updated in a long time, and I am no longer interested, so I'm closing this issue to clean up my issues list.

PR #1215, opened 2017-02-15
Display errors inline

fmdkdd:
**This PR is not intended for merging, just for preview and review**

Here's my implementation proposal for #1210.  There are three parts to it:

1. The `flycheck-inline` global minor mode
2. A fix to `display-error` 
3. Error groups to display related errors

The last part is mostly relevant for Rust.  `flycheck-inline` is intended to work with all checkers, and the fix is optional.

### Flycheck-inline global minor mode
Add `flycheck-inline-mode`, a global minor mode.  When activated, it changes `flycheck-display-errors-function` to `flycheck-inline-display-messages`.  Error messages under point are then displayed inline, just below the line containing the error overlays.  Here's a preview:

inline simple error

If there are 3 errors under point, they will be displayed on three lines, each error on its line.

This functionality is solely contained in `flycheck-inline.el`, which is intended as an extension to flycheck in a separate repository.  I put it in this PR to be able to review it.

The code should be straightforward.  It's mostly the same as `flycheck-pos-tip`, since the functionality is similar.

- Should this extension be hosted under the flycheck organization?  If it should, is `flycheck-inline` alright for the repository name, or is another name more appropriate?

### Fix to `display-error`
Since #1042, `flycheck-next-error` is intended to display the errors at point immediately.  However, since `-display-error-at-point-soon` is also on `post-command-hook`, invoking `flycheck-next-error` will actually trigger `-display-errors-function` *twice*.  Here is the timeline of events:
```
flycheck-next-error is called
- triggers display-error-at-point ---> errors are displayed
post-command-hook runs
- triggers display-error-at-point-soon
  - sets a timer for `display-error-at-point`
[0.9 seconds later]
- display-error-at-point is called by the timer ---> errors are displayed
```

For the default behavior of echoing messages in the minibuffer, this is unnoticeable.  For inline error messages, this means adding the overlays, then removing and adding them again after 0.9 seconds (the default delay for `-display-error-at-point-soon`).

- Is this a bug we want to fix?  Or can the `display-error` functions work around it by being (visually) idempotent?  If we want to fix it, I will open a separate PR for it.
- I've sketched a fix in the first commit here.  The idea is simple: don't call `display-error-soon` if `next-error` was the last command called.  The implementation relies on setting a flag and clearing it at the end of `post-command-hook`.  Better ideas are welcome.

### Error groups
While `flycheck-inline` can work without them, for Rust it makes the most sense to have error groups.  With errors groups, instead of displaying only the errors at point, `flycheck-inline` displays all the errors *from all the groups the errors at point belong to*.  Here is an example using groups:
inline group error

Here all the errors related to the borrowing error at point are displayed at once.  Note that the errors for the line `let zz = ...` are not displayed.

- Is this a feature we want in flycheck?  I can open a separate PR for error groups, and there we can discuss its merits and implementation, especially with regards to other checkers.

### DIFF
--- a/flycheck-inline.el
+++ b/flycheck-inline.el
@@ -103,11 +103,13 @@ Return the displayed phantom."
     (push (phantom-display msg pos) flycheck-inline--phantoms)))
 
 (defun flycheck-inline-display-messages (errors)
-  "Display ERRORS inline.
+  "Display ERRORS, and all errors from their groups, inline.
 
 ERRORS is a list of `flycheck-error' objects."
   (flycheck-inline-hide-messages)
-  (mapc #'flycheck-inline--display-phantom errors))
+  (mapc #'flycheck-inline--display-phantom
+        (seq-uniq
+         (seq-mapcat #'flycheck-errors-from-group errors))))
 
 (defun flycheck-inline-hide-messages ()
   "Hide all inline messages currently being shown."
--- a/flycheck-inline.el
+++ b/flycheck-inline.el
@@ -1,0 +1,165 @@
+;;; flycheck-inline-.el --- Display Flycheck errors inline -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2017 fmdkdd
+
+;; Author: fmdkdd
+;; URL: https://github.com/flycheck/flycheck-inline
+;; Keywords: tools, convenience
+;; Version: 0.1-cvs
+;; Package-Requires: ((emacs "25.1") (flycheck "31"))
+
+;; This file is not part of GNU Emacs.
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Provide an error display function to show Flycheck errors inline, directly
+;; below their location in the buffer.
+
+;;;; Setup
+
+;; (with-eval-after-load 'flycheck
+;;   (flycheck-inline-mode))
+
+;;; Code:
+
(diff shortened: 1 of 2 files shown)


cpitclaudel:
This looks great.

I think error groups make a lot of sense. I'd like to discuss them in a separate PR, because we definitely want them in Flycheck and they are pretty orthogonal to this :)

flycheck-pos-tip is currently in a separate repo, so it could make sense to have flycheck-inline in a separate repo too; but I'd like to hear your opinion.

cpitclaudel:
Re display-error being called twice — this does sound like a bug. Can you open a separate issue about it?

fmdkdd:
@cpitclaudel I've created two PRs for the orthogonal issues.  I'll go back to this extension when they are settled.

Regarding the repo: yes, a separate repo makes the more sense.  I was just wondering whether we'd include it in the flycheck organization, or if I should host it myself as an "unofficial" extension.

cpitclaudel:
> Regarding the repo: yes, a separate repo makes the more sense. I was just wondering whether we'd include it in the flycheck organization, or if I should host it myself as an "unofficial" extension.

Official repo, please :) It's a fine piece of software ^^

fmdkdd:
Gotcha.  I'll open up a separate repo under this organization.

fmdkdd:
The issue with `display-error` is actually related to #788.

zeusdeux:
This is awesome! Is it available in a nightly build or on melpa?
If not, how does one play with this w/o having to manually modify flycheck files?
Cheers!

fmdkdd:
@zeusdeux Thanks for your enthusiasm for this feature!  I'm afraid that if you want to play with it right now you have to use the branch of this PR and tweak your Emacs configuration.

However, now that one of the pre-requisite has been merged, I should be able to push `flycheck-inline-mode` as an extension in the organization, *soon-ish*.

terlar:
In the meantime there is this: https://github.com/stardiviner/flycheck-inline

Not sure how it compares

zeusdeux:
@fmdkdd Looking forward to it a lot tbh!
@terlar Thanks! Gonna play with it today ^_^

fmdkdd:
I've pushed the extension to its own repository under [flycheck/flycheck-inline](https://github.com/flycheck/flycheck-inline).

It might be a bit rought around the edges, so feedback and patches are welcome!

Keep in mind that until #1248 is fixed, using `flycheck-next-error` won't immediately display the error in the buffer, unlike the default behavior for the minibuffer.  You can tweak `flycheck-display-errors-delay` to mitigate that.

PR #1216, opened 2017-02-19
Treat 'note' from rustc as info, not error

Wilfred:
Given a project that is configured as a staticlib:

    [lib]
    crate-type = ["staticlib"]

rustc will output the following errors:

    $ cargo rustc -- --error-format=json
       Compiling foo v0.1.0 (file:///home/wilfred/tmp/foo)
    {"message":"link against the following native artifacts when linking against this static library","code":null,"level":"note","spans":[],"children":[],"rendered":null}
    {"message":"the order and any duplication can be significant on some platforms, and so may need to be preserved","code":null,"level":"note","spans":[],"children":[],"rendered":null}
    {"message":"library: dl","code":null,"level":"note","spans":[],"children":[],"rendered":null}
    {"message":"library: rt","code":null,"level":"note","spans":[],"children":[],"rendered":null}
    {"message":"library: pthread","code":null,"level":"note","spans":[],"children":[],"rendered":null}
    {"message":"library: gcc_s","code":null,"level":"note","spans":[],"children":[],"rendered":null}
    {"message":"library: c","code":null,"level":"note","spans":[],"children":[],"rendered":null}
    {"message":"library: m","code":null,"level":"note","spans":[],"children":[],"rendered":null}
    {"message":"library: rt","code":null,"level":"note","spans":[],"children":[],"rendered":null}
    {"message":"library: util","code":null,"level":"note","spans":[],"children":[],"rendered":null}
        Finished debug [unoptimized + debuginfo] target(s) in 0.17 secs

Previously these would be treated as errors by flycheck.

### DIFF
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3881,6 +3881,32 @@ Why not:
     (flycheck-ert-should-syntax-check
      "language/rust/lib-main/src/main.rs" 'rust-mode)))
 
+(flycheck-ert-def-checker-test rust-cargo rust notes
+  ;; We only get notes on the first build. Ensure we start from a
+  ;; clean directory each time.
+  (call-process "cargo" nil nil nil "clean" "--manifest-path"
+                (expand-file-name
+                 "language/rust/note-test/Cargo.toml"
+                 flycheck-test-resources-directory))
+
+  (let* (flycheck-rust-check-tests)
+    (flycheck-ert-should-syntax-check
+     "language/rust/note-test/src/lib.rs" 'rust-mode
+     '(1 1 info "library: util" :checker rust-cargo)
+     '(1 1 info "library: rt" :checker rust-cargo)
+     '(1 1 info "library: m" :checker rust-cargo)
+     '(1 1 info "library: c" :checker rust-cargo)
+     '(1 1 info "library: gcc_s" :checker rust-cargo)
+     '(1 1 info "library: pthread" :checker rust-cargo)
+     '(1 1 info "library: rt" :checker rust-cargo)
+     '(1 1 info "library: dl" :checker rust-cargo)
+     '(1 1 info
+         "the order and any duplication can be significant on some platforms, and so may need to be preserved"
+         :checker rust-cargo)
+     '(1 1 info
+         "link against the following native artifacts when linking against this static library"
+         :checker rust-cargo))))
+
 (flycheck-ert-def-checker-test rust rust syntax-error
   (let ((flycheck-disabled-checkers '(rust-cargo)))
     (flycheck-ert-should-syntax-check
--- a/flycheck.el
+++ b/flycheck.el
@@ -5673,6 +5673,7 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
             error-level (pcase .level
                           (`"error" 'error)
                           (`"warning" 'warning)
+                          (`"note" 'info)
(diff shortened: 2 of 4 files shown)


Wilfred:
@fmdkdd OK, I've added an integration test. It's a little awkward to get a note from the rust compiler, but the test does the right thing. Do let me know if there's any way I can make it prettier.

test/flycheck-test.el
```diff
@@ -3881,6 +3882,30 @@ Why not:
     (flycheck-ert-should-syntax-check
      "language/rust/lib-main/src/main.rs" 'rust-mode)))
 
+(flycheck-ert-def-checker-test rust-cargo rust notes
+  ;; We only get notes on the first build. Ensure we start from a
+  ;; clean directory each time.
+  (f-delete (f-join flycheck-test-resources-directory
+                    "language/rust/note-test/target")
+            t)
+  (let* ((flycheck-rust-check-tests nil))
```
fmdkdd: Wouldn't a simple `let` suffice here?

Wilfred:
@fmdkdd thanks. I've changed the let binding and moved to cargo clean. I did try `-Z parse-only` but it doesn't produce the notes we're testing for.

test/flycheck-test.el
```diff
@@ -42,6 +42,7 @@
 ;;; Requirements
 
 (require 'dash)
+(require 'f)
```
fmdkdd: We don't need that library anymore if you call `cargo clean`.

test/flycheck-test.el
```diff
@@ -3881,6 +3882,29 @@ Why not:
     (flycheck-ert-should-syntax-check
      "language/rust/lib-main/src/main.rs" 'rust-mode)))
 
+(flycheck-ert-def-checker-test rust-cargo rust notes
+  ;; We only get notes on the first build. Ensure we start from a
+  ;; clean directory each time.
+  (shell-command "cargo clean")
```
fmdkdd: Prefer `call-process` to `shell-command`, as there's no need to create a shell here.

Also, any idea of the current working directory when that command is called?  I think you can pass `--manifest-path` to make sure we are cleaning the right folder :)

Wilfred:
Updated.

test/flycheck-test.el
```diff
@@ -3881,6 +3881,33 @@ Why not:
     (flycheck-ert-should-syntax-check
      "language/rust/lib-main/src/main.rs" 'rust-mode)))
 
+(flycheck-ert-def-checker-test rust-cargo rust notes
+  ;; We only get notes on the first build. Ensure we start from a
+  ;; clean directory each time.
+  (call-process-shell-command
```
fmdkdd: Do we really need a shell?  If not, it would be faster to just do:
```
(call-process "cargo" nil nil nil "clean" "--manifest-path" (expand-file-name ...))
```

Wilfred: I initially resisted `call-process` because of all the `nil` parameters. We don't need a shell, so I'll change.

Wilfred:
Updated. Thanks for your patience :)

test/flycheck-test.el
fmdkdd: Haha, I concede that `shell-command` is less scary and a bit more readable.  I had the same reservations as you did at first.  Yet, `call-process` is the right  choice here.  After a while you don't see the `nil` (just like you don't see the parens!).

PR #1217, opened 2017-02-22
Fix (most) integration tests

fmdkdd:
I was bugged out by the fact we never actually run the whole integration test suite anywhere.  So I built a VM using Vagrant and added all the external tools.  Finding which tools and how to install them was a chore, but it all stands neatly in a single Vagrantfile.

I've fixed 31 failing tests (see other commits in PR).  2 tests are still failing and I'm not sure if that's due to upstream bugs or obsolete tests.

- `go-build/missing-package` fails because `go build` ignores the GOPATH environment variable
- `ruby-rubylint/syntax-error` fails with a backtrace from `rubylint`

Are @flycheck/go and @flycheck/ruby able to reproduce the failures above?

6 tests are still skipped:
- 2 tests due to `javascript-gjslint` being uninstalled.  As far as I can tell, `gjslint` is deprecated and unvailable from the URL in the checker docstring.  The page suggests to use lints from the Closure compiler instead.
- 2 tests due to `verilog-verilator` being uninstalled.  It's available in Arch AUR, but fails to build due to an upstream bug.
- 1 test due to `rpm-rpmlint` being uninstalled.  I found a github repo, but I'm missing the rpm python bindings.
- 1 `puppet` test which is intended to skip for puppet <= 4.0

@flycheck/javascript: should we drop the gjslint checker?  Or should we change it to use Closure compiler?  Or is everyone using eslint these days? 

If you have Vagrant and VirtualBox installed, using the VM to run the full suite is as simple as:
```
cd flycheck
vagrant up
vagrant ssh --command 'cd /vagrant; make integ'
```
**Warning**: provisioning the VM for the first time takes around 1 hour.  After that, booting should only take a few seconds.  Though since it's an Arch system, I've set it up to upgrade at boot, and that can take some time if you seldom use it.  Also, right now, it sits at 15Gb on my drive.

I'm by no means an expert in Vagrant or setting up VMs.  I just looked up enough info to make it work.  Here are some things I'm not sure about:
- Coq takes a bunch of time (and RAM!) to compile.  The website does not seem to provide Linux binaries :(
- There are 3 different versions of GHC installed (7.10, 8.01, 8.02).  One used by stack, one used by hadolint, one used by the `haskell-ghc` tests.
- Ideally Travis should run the integration tests, but due to the time it takes to install all the tools, I'm not sure it's worth it to tear it down/bring it up on each pull request.  That's why I'm pushing the Vagrantfile, so at least anyone willing can use it.

Should help with #1163.

### DIFF
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -2878,14 +2878,8 @@ The term \"1\" has type \"nat\" while it is expected to have type
   :tags '(checkstyle-xml)
   (flycheck-ert-should-syntax-check
    "language/css/syntax-error.css" 'css-mode
-   '(4 16 error "Expected LBRACE at line 4, col 16." :id "ParsingErrors"
-       :checker css-csslint)
-   '(4 16 error "Unexpected token '100%' at line 4, col 16." :id "ParsingErrors"
-       :checker css-csslint)
-   '(4 20 error "Unexpected token ';' at line 4, col 20." :id "ParsingErrors"
-       :checker css-csslint)
-   '(5 1 error "Unexpected token '}' at line 5, col 1." :id "ParsingErrors"
-       :checker css-csslint)))
+   '(4 14 error "Expected a `FUNCTION` or `IDENT` after colon at line 4, col 14."
+       :id "ParsingErrors" :checker css-csslint)))
 
 (ert-deftest flycheck-d-module-re/matches-module-name ()
   :tags '(language-d)
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3226,7 +3226,7 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
   (let ((flycheck-disabled-checkers '(haskell-ghc)))
     (flycheck-ert-should-syntax-check
      "language/haskell/Literate.lhs" 'literate-haskell-mode
-     '(6 1 warning "Top-level binding with no type signature: foo :: forall a. a"
+     '(6 1 warning "Top-level binding with no type signature: foo :: a"
          :id "-Wmissing-signatures"
          :checker haskell-stack-ghc))))
 
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3145,16 +3145,17 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
            :checker go-build)))))
 
 (flycheck-ert-def-checker-test go-build go directory-with-two-packages
-  (flycheck-ert-with-env
-      `(("GOPATH" . ,(flycheck-ert-resource-filename "checkers/go")))
-    (flycheck-ert-should-syntax-check
-     "checkers/go/src/multipkg/a.go" 'go-mode
-     `(1 nil info
-         ,(concat "can't load package: package multipkg: "
-                  "found packages a (a.go) and b (b.go) in "
-                  (flycheck-ert-resource-filename
-                   "checkers/go/src/multipkg"))
-         :checker go-build))))
+  (let ((flycheck-disabled-checkers '(go-errcheck go-unconvert)))
+    (flycheck-ert-with-env
+        `(("GOPATH" . ,(flycheck-ert-resource-filename "checkers/go")))
+      (flycheck-ert-should-syntax-check
+       "checkers/go/src/multipkg/a.go" 'go-mode
+       `(1 nil info
+           ,(concat "can't load package: package multipkg: "
+                    "found packages a (a.go) and b (b.go) in "
+                    (flycheck-ert-resource-filename
+                     "checkers/go/src/multipkg"))
+           :checker go-build)))))
 
 (flycheck-ert-def-checker-test go-test go nil
   (flycheck-ert-with-env
@@ -3178,7 +3179,7 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
   (flycheck-ert-with-env
       `(("GOPATH" . ,(flycheck-ert-resource-filename "language/go")))
     (flycheck-ert-should-syntax-check
-     "language/go/unconvert/unconvert.go" 'go-mode
+     "language/go/src/unconvert/unconvert.go" 'go-mode
      '(7 17 warning "unnecessary conversion"
          :checker go-unconvert))))
 
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -2746,13 +2746,10 @@ evaluating BODY."
 
 (flycheck-ert-def-checker-test c/c++-clang (c c++) included-file-error
   (let ((flycheck-clang-include-path '("./include"))
-        (flycheck-disabled-checkers '(c/c++-gcc))
-        (include-file (flycheck-ert-resource-filename
-                       "language/c_c++/warning.c")))
+        (flycheck-disabled-checkers '(c/c++-gcc)))
     (flycheck-ert-should-syntax-check
      "language/c_c++/in-included-file.cpp" 'c++-mode
-     `(3 nil warning ,(format "In include %s" include-file)
-         :checker c/c++-clang))))
+     `(3 nil warning "In include ./warning.c" :checker c/c++-clang))))
 
 (flycheck-ert-def-checker-test c/c++-gcc (c c++) error
   (let ((flycheck-disabled-checkers '(c/c++-clang)))
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -50,12 +50,13 @@
 (require 'flycheck-ert)
 
 ;; Make a best effort to make Coq Mode available
-(mapc (lambda (dir)
-        (add-to-list 'load-path (expand-file-name "coq/" dir)))
-      '("/usr/share/emacs/site-lisp/"
-        "/usr/local/share/emacs/site-lisp/"))
+;; (mapc (lambda (dir)
+;;         (add-to-list 'load-path (expand-file-name "coq/" dir)))
+;;       '("/usr/share/emacs/site-lisp/"
+;;         "/usr/local/share/emacs/site-lisp/"))
+(add-to-list 'load-path "/usr/share/emacs/site-lisp/")
 
-(autoload 'coq-mode "coq")
+(autoload 'coq-mode "gallina")
 
 ;; Load ESS for R-mode (its autoloads are broken)
 (require 'ess-site nil 'noerror)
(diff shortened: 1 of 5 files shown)


fmdkdd:
Note that there is only *one* change to `flycheck.el`.  Nearly all tests broke because of slight changes in error wording from updated checkers.

The only change required to *parsing* errors was for JRuby, where a semicolon was missing.  I am surprised this checker ever worked without it.

Simplify:
@fmdkdd Sorry for late reaction, I missed this PR.

- Yes, `gjslint` is deprecated and can be removed. 
- I'll check `ruby-rubylint` in next few days, probably make separate PR for that.

My guess is that `JRuby` is never used as standalone checker. Most Ruby developers use some kind of of Ruby version manager, `rvm`, `rbenv` or `chruby`, so if they have JRuby installed, it responds to `ruby` command and in that case is used in Flycheck as `ruby` linter:
```
$ ruby -v
jruby 9.1.7.0 (2.3.1) 2017-01-11 68056ae Java HotSpot(TM) 64-Bit Server VM 25.92-b14 on 1.8.0_92-b14 +jit [darwin-x86_64]
```

JRuby developers take care for compatibility with "standard" ruby, MRI (Matz's Rubu Implementation) so linting gives same output as MRI.

fmdkdd:
@Simplify Thanks for looking into this.

I'll make a separate PR for `gjslint`.

@cpitclaudel Do we want to include the `Vagrantfile` into the tree, or should I just squash all these fixes and keep the `Vagrantfile` somewhere else?

kthelgason:
This is great work, thanks for putting in this effort. One thing that bothers me a bit about this is that there tests are very dependent on the output of syntax checkers, which may change. It will get really annoying if we have to update these all the time due to syntax checkers changing. Maybe this won't be an issue in practice, or we could consider pinning the package versions in the `Vagrantfile`.

On the `Vagrantfile`, I'd vote to include it. Getting a reproduceable environment is a lot of work and this does the job well.

kthelgason:
Also, Doesn't Travis cache resources? I believe you can configure it to keep some paths between builds, so the VM setup cost should only have to be paid once. IMO the value of these tests is greatly diminished if they don't run continuously.

Simplify:
@fmdkdd feel free to remove `ruby-rubylint/syntax-error` test, `ruby-lint` just crashes on that specific file, even outside of Emacs.

About `Vagrantfile`... We had separate repo for virtual machine [here](https://github.com/flycheck/flycheck-vm/), but it looks that doesn't exists any more. I tried to use it, but box image that was used to build VM was deleted from original location. Upgrading to newer Ubuntu was pain, i gave up after 5 hours. On my question is that VM project still supported answer from Sebastian was:

> @Simplify Not really. Took too much to maintain and was too cumbersome to use.
> Currently I'm just ad-hoc installing anything I need to test with Homebrew.

That was on 2015-05-08 according to Github mail that I found. 

We can:
- Make seperate repo for VM as before. That will keep Flycheck code base cleaner in this repo.
- Add `Vagrantfile` and related scripts in this repo. That may help that more people get involved in supporting Flycheck VM, but not really sure.

fmdkdd:
> One thing that bothers me a bit about this is that there tests are very dependent on the output of syntax checkers, which may change.

I'm bothered by that as well.  When you run the tests and are greeted by a failure, you have to figure out if the checker output was updated or if there's something wrong in Flycheck.

One solution would be to decouple the two.  Instead of running `cargo` against a project and checking if the constructed `flycheck-error` objects are the ones expected, we could save `cargo` output for the project as a file, and run our tests against this saved output, without invoking `cargo`.  Then, we would have a separate test run to check if this saved output still conforms to what `cargo` outputs.  The same for all checkers.

That way, if the first run fails, there's something wrong in Flycheck, and if the second fails, the checker output was probably changed.

> Also, Doesn't Travis cache resources?

Yes, it does.  I actually sent a PR (flycheck/emacs-travis#5) to cache more things in our current CI.  It could use some eyes ;)

I agree that we should run the integration tests continuously.  However, I'm not sure if Travis will suffice here.  First, we would have to adapt the `Vagrantfile` to whatever `travis.yml` allows; second, Travis VMs are Ubuntu 12.04, so we could not test against fresh checker versions.  Also, Travis limits VM to 60 minutes, and the provisionning here is around 1 hour.  If we could bring the time below 60 minutes, then caching would help for other runs, but keep in mind that you still have to download the cached resources.

Anyone knows of an alternative service that whould do here?  One where we could set up a VM from a Vagrantfile, and have it run our tests on push?  I guess any VM hosting provider would do, but a "free for open-source projects" service could be interesting ;)

@Simplify Do you think it would be worthwhile to report the bug to `rubylint`?  Crashing on bad input is probably not their intention.

I heard about the previous VM and the reasons it got removed.  My intention with checking in the `Vagrantfile` is that other willing people would be able to run the whole suite themselves.  But, in the long run, running them continuously is probably best.  So a separate repo is fine.

Simplify:
@fmdkdd done ;)

fmdkdd:
@Simplify Great.  Let us know how that turns out.

andreas-marschke:
@fmdkdd I wonder if it would make sense to use docker instead of Vagrant to standup a dedicated environment to run the tests in. Seeing as TravisCI appears to [support Docker](https://docs.travis-ci.com/user/docker/) instead of Vagrant(?)

fmdkdd:
@andreas-marschke I'm afraid I don't know anything about Docker, or Vagrant.  I just RTFM'ed enough to make this work.

I think there's currently an issue with running Emacs inside Docker (see #1159).

But I'm interested to hear any solution that would allow us to run these tests again :)

andreas-marschke:
@fmdkdd Status Update: I've setup a new Dockerfile with most of the configuration, based off of debian unstable. I validated that debian works and starts just fine here seeing as I'm explicitly picking the `emacs25-nox` package provided by debian which does not require an X11 instance to run.

Still figuring out all of the languages and how to setup everything for each linter environment (don't know enough about R/Scala/Go but figuring out as I go). Once that's done this Docker image is uploaded to [docker hub](https://hub.docker.com/) from which travis would fetch and run the tests in the docker image. This allows us to keep a cached image that's consistent with what we need.

One caveat I can think of is that you'd have to extend the dockerfile with every new checker PR'ed if you have integration tests for it.

fmdkdd:
@andreas-marschke Setting up all these tools was a pain, yes.  Feel free to take a look at the Vagrantfile; it might be relevant for your case too.

> One caveat I can think of is that you'd have to extend the dockerfile with every new checker PR'ed if you have integration tests for it.

I'm not sure I understand what this implies.  Do you mean that `flycheck.el` is bundled in the Docker image, and we can't easily run the tests off a `flycheck.el` version fetched from master?

andreas-marschke:
@fmdkdd 
>I'm not sure I understand what this implies. Do you mean that flycheck.el is bundled in the Docker image, and we can't easily run the tests off a flycheck.el version fetched from master?

It's more that whenever we'd add a new checker with integrations tests for it, you'll have to adjust one file. It's mostly the inconvenience as a submitter of having to deal with merge conflicts should more than one person be adding a new checker

fmdkdd:
@andreas-marschke Okay, we'll see how that turns out.  Please open up a PR when you have something working 👍

andreas-marschke:
FYI: 
I have posted this public automated build setup to docker hub which I'll use to play around with still:
https://hub.docker.com/r/amarschke/flycheck-ci-docker/
It already contains a lot of what is required but still lacks packages for some of the checkers. I hope to have this running most of the tests. I'll move this into a PR here once the kinks are out and it works as intended.

You can find the Github repository referenced from the link above.

test/flycheck-test.el
```diff
@@ -50,12 +50,13 @@
...
-      '("/usr/share/emacs/site-lisp/"
-        "/usr/local/share/emacs/site-lisp/"))
+;; (mapc (lambda (dir)
+;;         (add-to-list 'load-path (expand-file-name "coq/" dir)))
+;;       '("/usr/share/emacs/site-lisp/"
+;;         "/usr/local/share/emacs/site-lisp/"))
+(add-to-list 'load-path "/usr/share/emacs/site-lisp/")
 
-(autoload 'coq-mode "coq")
+(autoload 'coq-mode "gallina")
```
cpitclaudel: These changes (the Coq ones) look fishy. Do you remember why you made them?

test/flycheck-test.el
```diff
@@ -2851,23 +2850,22 @@ evaluating BODY."
          :checker coffee-coffeelint))))
 
 (flycheck-ert-def-checker-test coq coq syntax-error
-  (skip-unless (load "coq" 'noerror 'nomessage))
+  (skip-unless (load "gallina" 'noerror 'nomessage))
```
cpitclaudel: This too looks wrong (the gallina/coq thing).

test/flycheck-test.el
fmdkdd: Sure: in the VM, `coq-mode` was provided by `gallina.el`, not `coq.el`.  The ArchLinux package is compiled straight from [the source](https://coq.inria.fr/download), and I've just checked: the latest archive has `gallina.el` providing `gallina`; no trace of `coq.el` or `(provide 'coq)`.

So this was required in order for me to run the tests inside the VM.

But I don't use Coq, so I might have missed something.  Were you able to run the tests using 8.6?

fmdkdd:
@cpitclaudel I found traces of the old VM yes.  I am definitely in favor of finding a low-burden solution to running our integration tests.

At the moment, the Vagrantfile does not seem that cumbersome to maintain once established.  But it takes too long to be able to run in the CI, so that's a downside.  At the very least, it allows anyone willing to run the the whole integration suite locally.

I'll get around to updating the PR shortly.

test/flycheck-test.el
cpitclaudel: I think the trick was that coq.el is provided by Proof General (the ELisp files in the distribution are obsolete).
Do we even need to load the mode?

cpitclaudel:
> At the moment, the Vagrantfile does not seem that cumbersome to maintain once established. But it takes too long to be able to run in the CI, so that's a downside. At the very least, it allows anyone willing to run the the whole integration suite locally.

I seem to remember Travis having an option to cache some data between builds. Could we use that?

fmdkdd:
I don't think Travis even supports running a Vagrantfile, so this is a moot point.  The Vagrantfile is a crutch for running the integration tests locally.  I included it so others would be able to replicate, but in the long run the Docker approach seems best.  Fingers crossed for @andreas-marschke :)

test/flycheck-test.el
fmdkdd: Ah, good to know :)

I think we do need to load it.  I think I remember trying without, and the test wouldn't work.  I will look at it again.

cpitclaudel: I'm actually having trouble getting ERT not to skip these two tests.  Does our test suite check the mode explicitly? I was under the impression we only ran the checker, regardless of the mode.

fmdkdd: All the integration tests need the relevant mode to run the checker.  The Coq tests are unusual as they all start with:
```
  (skip-unless (load "coq" 'noerror 'nomessage))
```

With Coq installed, and using "gallina" instead, I can run the tests.  If I remove the explicit `load` calls or `autoload`, the tests are skipped.

So, I could put ProofGeneral in the VM, but since it's not available from MELPA (any reason why?), and cloning the Git from Cask doesn't work...  I'll have to follow the PG instructions to clone and `make`, which is a bit of a hassle.

But technically, we don't need PG to run the tests, only a mode called `coq-mode` :-)

I'm okay with reverting the `gallina` changes.

fmdkdd:
@cpitclaudel I've rebased and fixed tests (again) for rust and chef-foodcritic due to updated checker output.

We still have to settle the Coq situation.  And will you want me to squash all these fixes together or leave them as is?

cpitclaudel:
I think you can leave the fixes as-is. I'm happy with your original fix for Coq, so please go ahead and merge :)

PR #1218, opened 2017-02-23
Add luacheck standards option variable

marsam:
Hi:

I've added `flycheck-luacheck-standards` option variable for luacheck checker

### DIFF
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3507,6 +3507,28 @@ Why not:
    '(4 11 warning "variable 'var2' is never set"
        :id "W221" :checker lua-luacheck)))
 
+(flycheck-ert-def-checker-test lua-luacheck lua custom-luacheckrc
+  (let ((flycheck-luacheckrc "custom.luacheckrc"))
+    (flycheck-ert-should-syntax-check
+     "language/lua/warnings.lua" 'lua-mode
+     '(1 1 warning "setting non-standard global variable 'global_var'"
+         :id "W111" :checker lua-luacheck)
+     '(3 16 warning "unused function 'test'"
+         :id "W211" :checker lua-luacheck)
+     '(4 11 warning "variable 'var2' is never set"
+         :id "W221" :checker lua-luacheck))))
+
+(flycheck-ert-def-checker-test lua-luacheck lua custom-standards
+  (let ((flycheck-luacheck-standards '("ngx_lua")))
+    (flycheck-ert-should-syntax-check
+     "language/lua/ngx_lua.warnings.lua" 'lua-mode
+     '(3 16 warning "unused function 'test'"
+         :id "W211" :checker lua-luacheck)
+     '(3 21 warning "unused argument 'arg'"
+         :id "W212" :checker lua-luacheck)
+     '(4 11 warning "variable 'var2' is never set"
+         :id "W221" :checker lua-luacheck))))
+
 (flycheck-ert-def-checker-test lua lua nil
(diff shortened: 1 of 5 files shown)


fmdkdd:
Thank you.  From the luacheck documentation, it appears you can set the standards using the configuration file also.  Any reason to single out this option?  Is `--std` more likely to be used without a configuration file? 

Also, could you add a test for these features?

marsam:
> Any reason to single out this option? Is --std more likely to be used without a configuration file?

I thinks is more convenient to use `--std` for other lua-derived modes or inside `hooks`, e.g [rockspec files](https://github.com/luarocks/luarocks/wiki/Rockspec-format):

```emacs-lisp
(define-derived-mode rockspec-mode lua-mode "RockSpec")
(add-to-list 'auto-mode-alist '("\\.rockspec\\'" . rockspec-mode))

(flycheck-add-mode 'lua-luacheck 'rockspec-mode)

(add-hook 'rockspec-mode-hook
          (lambda ()
            (setq flycheck-luacheck-standards '("rockspec"))))
```

> Also, could you add a test for these features?

Sure! but I can't find examples for `options-vars` nor `config-file-var` in the `flycheck-test.el` file. Can you point me an example?.

fmdkdd:
@marsam Ok, thanks for the example :)

You just set the variables using `let`.  This sets the variables for the duration of test.  Here is an example from the SCSS checker for a config file:
```
(flycheck-ert-def-checker-test scss-lint scss nil
  (let ((flycheck-scss-lintrc "scss-lint.yml"))
```

And here are a couple options for `rust-cargo`:
```
(flycheck-ert-def-checker-test rust-cargo rust lib-main
  (let ((flycheck-rust-crate-type "bin")
        (flycheck-rust-binary-name "lib-main"))
```

marsam:
@fmdkdd Hi, sorry for the delay. I added the tests

test/flycheck-test.el
```diff
@@ -3507,6 +3507,28 @@ Why not:
    '(4 11 warning "variable 'var2' is never set"
        :id "W221" :checker lua-luacheck)))
 
+(flycheck-ert-def-checker-test lua-luacheck lua nil
```
fmdkdd: You should replace `nil` by a test name, otherwise the two tests you added end up with the same name (`nil` is replaced by `default`), and only the second one is actually run.

marsam:
Updated, thanks for your patience

```

make SELECTOR='(tag checker-lua-luacheck)' integ
cask exec emacs -Q --batch -L .  -l maint/flycheck-compile.el -f flycheck/batch-byte-compile flycheck.el
cask exec emacs -Q --batch -L .  -l maint/flycheck-compile.el -f flycheck/batch-byte-compile flycheck-ert.el
cask exec emacs -Q --batch -L .  --load test/flycheck-test --load test/run.el -f flycheck-run-tests-main '(and (tag external-tool) (tag checker-lua-luacheck))'
Running tests on Emacs 26.0.50, built at 2017-02-22
Running 4 tests (2017-03-07 12:54:07-0500)
   passed  1/4  flycheck-define-checker/lua-luacheck/custom-luacheckrc
   passed  2/4  flycheck-define-checker/lua-luacheck/custom-standards
   passed  3/4  flycheck-define-checker/lua-luacheck/syntax-error
   passed  4/4  flycheck-define-checker/lua-luacheck/warnings

Ran 4 tests, 4 results as expected (2017-03-07 12:54:11-0500)
```

cpitclaudel:
Looks perfect :)

PR #1226, opened 2017-03-01
Enhance GCC backward compatibility

### LINKED ISSUE #1224: Make compiler flags to c/c++-gcc configurable
While I am aware that a non-ancient version of gcc is required, and I think it is reasonable, the reality of my current environment at work is that we do use an old gcc version. This means I have to comment out the "-fno-diagnostics-show-caret" flag every time flycheck.el is updated. This happens quite often as I use spacemacs and try keep my libraries up to date.

So this is not a bug report but a change request. Please make the flags defined in the `c/c++-gcc` checker configurable. I believe this to be a minor change, and it would make those of us unfortunate enough to live in the past able to use flycheck without modification to the flycheck.el file.

fmdkdd:
Closes #1224.

@Apakollaps, can you check this branch works for you?  And also report your GCC version?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6375,10 +6375,9 @@ warnings."
 (flycheck-define-checker c/c++-gcc
   "A C/C++ syntax checker using GCC.
 
-Requires GCC 4.8 or newer.  See URL `https://gcc.gnu.org/'."
+Requires GCC 4.4 or newer.  See URL `https://gcc.gnu.org/'."
   :command ("gcc"
             "-fshow-column"
-            "-fno-diagnostics-show-caret" ; Do not visually indicate the source location
             "-fno-diagnostics-show-option" ; Do not show the corresponding
                                         ; warning group
             "-iquote" (eval (flycheck-c/c++-quoted-include-directory))
--- a/flycheck.el
+++ b/flycheck.el
@@ -6378,8 +6378,6 @@ warnings."
 Requires GCC 4.4 or newer.  See URL `https://gcc.gnu.org/'."
   :command ("gcc"
             "-fshow-column"
-            "-fno-diagnostics-show-option" ; Do not show the corresponding
-                                        ; warning group
             "-iquote" (eval (flycheck-c/c++-quoted-include-directory))
             (option "-std=" flycheck-gcc-language-standard concat)
             (option-flag "-pedantic" flycheck-gcc-pedantic)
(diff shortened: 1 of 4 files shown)


Apakollaps:
@fmdkdd Looks like it works fine. Our gcc version is 4.4.7 :disappointed:.

fmdkdd:
@Apakollaps Good to know.  

@cpitclaudel is this fine?  The flags we drop enhance backward-compatibility without affecting parsing; there is just more output that isn't matched to any `:error-pattern`.  While I was at it, I captured warning groups as a `flycheck-error-id`.

flycheck.el
```diff
@@ -6381,7 +6378,8 @@ Requires GCC 4.8 or newer.  See URL `https://gcc.gnu.org/'."
    (info line-start (or "<stdin>" (file-name)) ":" line ":" column
          ": note: " (message) line-end)
    (warning line-start (or "<stdin>" (file-name)) ":" line ":" column
-            ": warning: " (message) line-end)
+            ": warning: " (message (one-or-more (not (any "\n["))))
+            (optional "[" (id (one-or-more (any "A-Z-"))) "]") line-end)
```
cpitclaudel: Are we already parsing case-insensitively? (I guess so, since it worked^^)

fmdkdd: Ha!  That did not even phase me.  But on a second reading, I'm puzzled.

I've looked into it, and we don't seem to pass anything related to case-sensitivity in Flycheck.  So I guess case-sensitivity is controlled by `case-fold-search`.

To be safe, I've changed the regexp to what other checkers do: `(one-or-more not-newline)`.

fmdkdd:
@cpitclaudel I've changed the regexp and updated the changelog.  Should be good to merge.

cpitclaudel:
Thanks. Shouldn't it be a non-greedy match though? Otherwise I won't a `]` in the message confuse us?

fmdkdd:
Hmm I don't think so.  A `]` in the message would just be part of the message.  The message is everything up to a newline or `[`, then, if there was a `[`, that's the ID.

```
(let ((str ": warning: foo ] message [-Wcpp]"))
  (string-match (rx ": warning: " (group-n 1 (one-or-more (not (any "\n["))))
                    (optional "[" (group-n 2 (one-or-more not-newline)) "]"))
                str)
  (list 
   (match-string 0 str)
   (match-string 1 str)
   (match-string 2 str)))
```
returns:
```
(": warning: foo ] message [-Wcpp]" "foo ] message " "-Wcpp")
```

cpitclaudel:
Ouch, sorry :/ For some reason I thought the warning id came first, not last. Please go ahead and merge :)

fmdkdd:
No worries :)  The warning id comes first in most checkers.

PR #1227, opened 2017-03-01
Avoid nil in flycheck-same-files-p, flycheck-relevant-error-p

drvink:


drvink:
See #1162 for context.

flycheck.el
```diff
@@ -1071,16 +1071,19 @@ to a number and return it.  Otherwise return nil."
...
-    ;; output canonical file names with all symbolic links resolved.  However,
-    ;; we still do a simple path compassion first, to avoid the comparatively
-    ;; expensive file system call if possible.  See
-    ;; https://github.com/flycheck/flycheck/issues/561
-    (or (string= (directory-file-name file-a) (directory-file-name file-b))
-        (string= (directory-file-name (file-truename file-a))
-                 (directory-file-name (file-truename file-b))))))
+  (if (or (null file-a)
+          (null file-b))
+      nil
```
cpitclaudel: This can be written as (unless (and file-a file-b) …), I think.

flycheck.el
```diff
@@ -3140,7 +3143,8 @@ otherwise."
        ;; and the current buffer has no file name, too, or if it refers to the
        ;; same file as the current buffer.
        (or (and (not file-name) (not buffer-file-name))
-           (flycheck-same-files-p file-name (buffer-file-name)))
+           (and buffer-file-name
+                (flycheck-same-files-p file-name buffer-file-name)))
```
cpitclaudel: Any reason to not make this (and buffer-file-name file-name (flycheck-same-files-p …))?

cpitclaudel:
Thanks! This makes sense, I think.  I'd like a second opinion, though.
Also, for consistency, is there any reason to spread the nil checks across both the caller and the callee? Could we either check in same-file-p and remove the check in the caller, or check both names in the callee and leave same-file-p as-is? (I think I favor the second solution)

flycheck.el
drvink: Yes. I've always disliked `unless`, though if you prefer it, I have no problem changing it.

flycheck.el
drvink: That would be fine as well.

drvink:
Oh, the intent wasn't the spread the checks across the two functions, but rather to provide additional protection for any other potential callers of `flycheck-same-files-p`.

flycheck.el
cpitclaudel: I prefer it :)

cpitclaudel:
> Oh, the intent wasn't to spread the checks across the two functions, but rather to provide additional protection for any other potential callers of flycheck-same-files-p.

Thanks for the clarification :)  We only seem to have two other uses, and I'd rather not add nil-swallowing behavior to same-files-p (it's also consistent with f-same-p from the f library), so let's go for checks in the caller.  Can you update the PR?

fmdkdd:
I agree with @cpitclaudel, `flycheck-same-files-p` is only called in 3 places.  Better to make sure not to call it with `nil` than to silently return `nil` on an invalid call.

On a related note @cpitclaudel, do you think some sort of static analysis tool for `nil`-checking could catch these bugs?

drvink:
I made the requested changes and re-pushed.

cpitclaudel:
> On a related note @cpitclaudel, do you think some sort of static analysis tool for nil-checking could catch these bugs?

Possibly? :) I don't know of much work of this sort for ELisp, though

Issue #1024, opened 2016-07-29
Remove JSCS


lunaryorn:
JSCS has reached [end of life](http://eslint.org/blog/2016/07/jscs-end-of-life), we should stop supporting it as well.

/cc @flycheck/javascript 

Edit: For reference, the [compatibility milestone](https://github.com/eslint/eslint/milestone/15)

Simplify:
On the same page:

> We have setup a milestone for tracking all of our JSCS compatibility issues. At the time of this post, we are at 20% complete and are continuing to make progress.

And:

> We recommend holding off on switching from JSCS to ESLint until most of the compatibility work has been completed (we will make an announcement when that happens).

I'll wait until that announcement. JSCS still have [27000+ downloads every day](https://www.npmjs.com/package/jscs), around same number as in April when they announced joining ESLint.

lunaryorn:
@Simplify We can wait, but I'd not like to prolong cleanup work beyond what's absolutely necessary.  Users who still need JSCS can either restore the definition, or stick to an older Flycheck release.

Simplify:
Ok, give me few days to check few things that JSCS did better than ESLint, for example checking JSDoc comments in the code.

Simplify:
We can also extract JSCS in gist.

lunaryorn:
@Simplify As I said, we can wait until eslint provides full compatibility, but as soon as that's the case, we'll remove JSCS.

Simplify:
@lunaryorn  Just a quick update: Compatibility is now at 38% and there is a draft of [JSCS migration guide](https://github.com/eslint/eslint/commit/1153955). However, there are still around 30K downloads of JSCS every day.

I'm busy next 2 weeks, but after that we can check status again. Maybe we can wait until compatibility hits 60-80% or that migration guide is done. Doe to high number of downloads, the best solution IMHO is to extract JSCS checker into GH gist, then people can load that file into Emacs if they still need JSCS support. Maybe provide link to that gist in ChangeLog.

Simplify:
JSCS is [unmaintained](https://github.com/jscs-dev/node-jscs), [merged with ESLint](https://eslint.org/blog/2016/04/welcoming-jscs-to-eslint) and [compatibility with ESLint](https://github.com/eslint/eslint/milestone/15) is 84%.

PR #1233, opened 2017-03-23
Remove javascript-gjslint checker

fmdkdd:
The linter is deprecated, as explained [there](https://developers.google.com/closure/utilities/).

However, the page also mentions that you can pass a flag to the Closure compiler to get lints.  I haven't tested it, but maybe we could adapt the checker to work with that instead.

I have no experience with either, so I'll let @flycheck/javascript chime in.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@
 - **Breaking changes**
 
   - ``rust-cargo`` now requires Rust 1.15 or newer [GH-1201]
+  - Remove javascript-gjslint checker
 
 - New syntax checkers:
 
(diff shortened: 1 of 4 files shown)


Simplify:
Great, thanks, this can be removed. 

If anyone requests support for closure compiler with flags we can add it, but so far nobody did. Currently everyone is using `eslint`, `jshint` (in Ember world, but my understanding is that they are also going to use `eslint`) and `jscs` (also deprecated, but still in use, there is ticket to remove it after `eslint` gets close to full compatibility - 70% right now - GH-1024).

fmdkdd:
Agreed: if there is a need, then we can add Closure compiler support.

PR #1238, opened 2017-04-11
Add test/.nosearch to keep test/ off the load-path

tarsius:
Adding this file ensures that the test directory does not accidentally
end up on the `load-path`.  This also helps tools used to detect the
feature provided by a package, which would otherwise claim that this
package provides `dummy-package` and some other dummy packages that
are located here.


tarsius:
Some tests failed, but it looks to my like those failures are not related to my change.

fmdkdd:
Yeah, the tests were my doing when I updated flycheck/emacs-travis.  Should be good now.

Thanks, this seems like a good idea.  But just to be sure:

How would the test directory end up on the load-path, previously?  I seem to recall seing warnings for the `dummy-package` somewhere, but now I can't find where.  What are the tools you had in mind which could be confused by this?

tarsius:
> How would the test directory end up on the load-path, previously?

That would happen if a user installed by cloning this repository and using `normal-top-level-add-to-load-path` for example. So it's not very likely that this would happen, but still possible.

fmdkdd:
Also, @cpitclaudel, it seems we have an issue with licence/cla.  According to cla-assistant/cla-assistant/issues/211, we need to change something in the repo settings.  I think you may be the only one with access :)

fmdkdd:
(Now I remember, we get warning about `dummy-package` when running `make unit`, but `.nosearch` does not change anything about them).

@tarsius I see.  Can you think of any scenario where adding this `.nosearch` would be an issue?

cpitclaudel:
@tarsius Thanks, LGTM. Sorry for the test issues (thanks @fmdkdd for your work on emacs-travis).
I fixed the CLA assistant issue.

tarsius:
> Can you think of any scenario where adding this .nosearch would be an issue?

It can be an issue when a "real library" requires a feature provided by a file in the test directory. The only reason for doing that, that I can see, is if the package comes with a command such as `foobar-run-tests`. But even then it would be better to `load`/`load-file` the test library by specifying the relative path explicitly.

fmdkdd:
@cpitclaudel Thanks :)

@tarsius Ok, thanks for taking the time to explain.  You just have to accept the CLA and then we can merge :)

And many thanks for your great work on magit by the way 👍

tarsius:
Signed.

PR #1246, opened 2017-05-03
Use standard input in nix checker

marsam:
Hi:

I've changed the nix checker to use standard input

```
make SELECTOR='(tag checker-nix)' integ
cask exec emacs -Q --batch -L .  -l maint/flycheck-compile.el -f flycheck/batch-byte-compile flycheck-buttercup.el
cask exec emacs -Q --batch -L .  --load test/flycheck-test --load test/run.el -f flycheck-run-tests-main '(and (tag external-tool) (tag checker-nix))'
Running tests on Emacs 26.0.50, built at 2017-04-19
Running 1 tests (2017-05-02 20:01:27-0500)
   passed  1/1  flycheck-define-checker/nix/default

Ran 1 tests, 1 results as expected (2017-05-02 20:01:28-0500)

```

PR #1247, opened 2017-05-03
Fix flake8 make maint check

fmdkdd:
CI was failing in the LINT build because of updated packages.

This fix is to remove the workaround for pydocstyle.

The latest flake8_docstrings requires pydocstyle 2.0.0, so the workaround
actually breaks our `make -C maint check`.

Besides, the issue with pydocstyle was fixed in 1.1.1.


fmdkdd:
@cpitclaudel This should fix the issue you encountered in #1246.

cpitclaudel:
😍

PR #1248, opened 2017-05-06
Fix calling display-error twice after next-error

fmdkdd:
So, I noticed this bug while working on #1215.  This is a separate PR including just this fix.  I'm reproducing the message to explain the error here:

Since #1042, `flycheck-next-error` is intended to display the errors at point immediately.  However, since `-display-error-at-point-soon` is also on `post-command-hook`, invoking `flycheck-next-error` will actually trigger `-display-errors-function` *twice*.  Here is the timeline of events:
```
flycheck-next-error is called
- triggers display-error-at-point ---> errors are displayed
post-command-hook runs
- triggers display-error-at-point-soon
  - sets a timer for `display-error-at-point`
[0.9 seconds later]
- display-error-at-point is called by the timer ---> errors are displayed
```

For the default behavior of echoing messages in the minibuffer, this is unnoticeable.  For inline error messages (see #1215), this means adding the overlays, then removing and adding them again after 0.9 seconds (the default delay for `-display-error-at-point-soon`).  It might affect other extensions (like [flycheck-pos-tip](https://github.com/flycheck/flycheck-pos-tip)); I have not tested.

I've sketched a fix in the first commit here.  The idea is simple: don't call `display-error-soon` if `next-error` was the last command called.  The implementation relies on setting a flag and clearing it at the end of `post-command-hook`.  Better ideas are welcome.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4391,6 +4391,8 @@ Intended for use with `next-error-function'."
       (flycheck-jump-to-error err)
     (user-error "No more Flycheck errors")))
 
+(defvar-local flycheck--display-error-caused-by-next-error nil)
+
 (defun flycheck-next-error (&optional n reset)
   "Visit the N-th error from the current point.
 
@@ -4403,7 +4405,8 @@ position."
     ;; Universal prefix argument means reset
     (setq reset t n nil))
   (flycheck-next-error-function n reset)
-  (flycheck-display-error-at-point))
+  (flycheck-display-error-at-point)
+  (setq flycheck--display-error-caused-by-next-error t))
 
 (defun flycheck-previous-error (&optional n)
   "Visit the N-th previous error.
@@ -4946,10 +4949,13 @@ non-nil."
 (defun flycheck-display-error-at-point-soon ()
   "Display the first error message at point in minibuffer delayed."
   (flycheck-cancel-error-display-error-at-point-timer)
-  (when (flycheck-overlays-at (point))
+  (when (and (flycheck-overlays-at (point))
+             (not flycheck--display-error-caused-by-next-error))
     (setq flycheck-display-error-at-point-timer
           (run-at-time flycheck-display-errors-delay nil
-                       'flycheck-display-error-at-point))))
+                       'flycheck-display-error-at-point)))
+  (setq flycheck--display-error-caused-by-next-error nil))
+
 
 
 ;;; Functions to display errors


fmdkdd:
The build is failing because of (intentional) missing docstrings.  First:

1. Is this how we want to fix the bug?
2. Is the fix sound?

@cpitclaudel Expert advice is welcome :)

cpitclaudel:
Thanks; the fix looks sound.

>  The implementation relies on setting a flag and clearing it at the end of post-command-hook. Better ideas are welcome.

Let me try one :) What about creating a function flycheck-post-command-hook that runs all functions that Flycheck currently puts on post-command-hook:

* flycheck-perform-deferred-syntax-check
* flycheck-error-list-update-source
* flycheck-error-list-highlight-errors
* flycheck-display-error-at-point-soon

There'd be an extra twist, though: that function would predicate the call to `display-error-at-point-soon` on `last-command` not being `flycheck-next-error`, or more likely on `last-command` not being a found in a list of commands for which `display-error-at-point-soon` should be skipped.

Thoughts? :)

fmdkdd:
@cpitclaudel So `last-command` is a great idea.  I tried it (or `this-command`, rather), then I remembered why I did not use it: we have no guarantee that the user called `flycheck-next-error` directly.  In fact, in Spacemacs, `spacemacs/next-error` calls `flycheck-next-error`, so `this-command` is the Spacemacs function, not the Flycheck one.

We *could* make a list like you propose, and then make a PR in Spacemacs to add their function in the list.  But that seems messier than the state variable of the current fix.

As for putting all our hooks into a function, it would help if we keep the state variable, since it would be clear that clearing it happens at the end of our hooks function.  Somehow, it would also simplify the code that attaches and removes the hooks when we enable/disable flycheck-mode.  

However, I'd be happier if we managed to not introduce extra state that is liable to cause errors and confusion in the future.  Something like "dont't call `display-error-at-point-soon` if the error is already being displayed".

fmdkdd:
See also #788.

fmdkdd:
> Do you think we could make flycheck-display-error-at-point set a flag to cancel the timer (and cancel any existing timer)?

That's already the case.  The issue was the `post-command-hook` created a timer *after* `-display-error-at-point` is called.

>  Doesn't the issue also exist if the user manually calls flycheck-display-error-at-point before the timer expires and show the error?

No, because the timer is cleared by `-display-error-at-point`.

The general fix to this issue is to have state that tells us whether the errors at point have already been displayed.  That way, we can make sure to not dislay them twice.  However, the question becomes: When do we clear that state?  When are the errors not displayed any more?  In the case of the minibuffer, we know that that any command will clear it; but that for other ways of displaying the errors, that might not be the case.

I'm still not happy about this fix, and am tempted to let extensions work around the behavior themselves.  In the case of `flycheck-inline`, it's easy to check that error overlays are active for instance.

cpitclaudel:
Revisiting this.  What about just storing the last position and buffer tick at which we called flycheck-display-errors-function, and skip the call to `flycheck-display-error-at-point` from the timer if it's redundant?  I think the patch would be fairly small, and fairly straightforward too.

fmdkdd:
If we do that and put point on the first char of an error, let the error display and `C-b C-f`, then the error won't display this time.

I've simplified the initial fix to piggy-back clearing the state variable on `display-error-at-point-soon`, so it's a little bit cleaner.  WDYT?

cpitclaudel:
> If we do that and put point on the first char of an error, let the error display and C-b C-f, then the error won't display this time.

It will if we reset the counter in display-error-at-point-too, won't it? I'm thinking of something like this: https://github.com/flycheck/flycheck/commit/338f2ed7ee89560b8a2a64947fb9daa52cd54900

> I've simplified the initial fix to piggy-back clearing the state variable on display-error-at-point-soon, so it's a little bit cleaner. WDYT?

This looks pretty small, at least :) Maybe we should add a comment pointing to the relevant issue.
Also, won't we have the same problem for other motion commands that display errors?

fmdkdd:
> It will if we reset the counter in display-error-at-point-too, won't it? I'm thinking of something like this: 338f2ed

That seems to work well, yes.  Don't remember what I did differently, but I prefer your patch!

PR #1249, opened 2017-05-06
Add the ability to group related errors

fmdkdd:
The use case is mainly for some Rust errors, like lifetime and borrowing errors,
where multiple individual errors are actually part of the same "diagnostic" for
only one issue.  Here is an example of rustc output:

```
error[E0382]: use of moved value: `x`
 --> main.rs:9:7
  |
8 |   let y = x;          // note: x's lifetime ends here
  |       - value moved here
9 |   let z = x;          // error: can no longer move x
  |       ^ value used here after move
  |
  = note: move occurs because `x` has type `NonPOD`, which does not implement the `Copy` trait
```

In the JSON output parsed by Flycheck, this one diagnostic corresponds to three
errors and one note, all of which are turned into individual flycheck-error
objects.  Consequently, the relationship between these errors is lost, and this
could confuse users not familiar with these kinds of errors.

To preserve the relationship between errors, this commit adds a `group` property
to flycheck-error objects.  Checkers should use the same `group` value for
related errors.

This commits also adds a function to get related errors,
`flycheck-errors-from-group`.  This allows extensions to display related errors
as they see fit.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2864,7 +2864,7 @@ variables of Flycheck."
                (:constructor flycheck-error-new)
                (:constructor flycheck-error-new-at (line column
                                                          &optional level message
-                                                         &key checker id
+                                                         &key checker id group
                                                          (filename (buffer-file-name))
                                                          (buffer (current-buffer)))))
   "Structure representing an error reported by a syntax checker.
@@ -2901,8 +2901,18 @@ Slots:
      The error level, as either `info', `warning' or `error'.
 
 `id' (optional)
-     An ID identifying the kind of error."
-  buffer checker filename line column message level id)
+     An ID identifying the kind of error.
+
+`group` (optional)
+     A symbol identifying the group the error belongs to.
+
+     Some tools will emit multiple errors that relate to the same
+     issue (e.g., lifetime errors in Rust).  All related errors
+     collected by a checker should have the same `group` value,
+     in order to be able to present them to the user.
+
+     See `flycheck-errors-from-group`."
+  buffer checker filename line column message level id group)
 
 (defmacro flycheck-error-with-buffer (err &rest forms)
   "Switch to the buffer of ERR and evaluate FORMS.
@@ -3159,6 +3169,18 @@ otherwise."
 Return a list of all errors that are relevant for their
 corresponding buffer."
   (seq-filter #'flycheck-relevant-error-p errors))
+
+(defun flycheck-related-errors (err)
+  "Get all the errors that are in the same group as ERR.
+
+Return a list of all errors (in `flycheck-current-errors') that
+have the same `flycheck-error-group' as ERR, including ERR
+itself."
+  (-if-let (group (flycheck-error-group err))
+      (seq-filter (lambda (e)
+                    (eq (flycheck-error-group e) group))
+                  flycheck-current-errors)
+    (list err)))
 
 
 ;;; Status reporting for the current buffer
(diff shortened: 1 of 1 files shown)


fmdkdd:
This is a pre-requisite for #1215.

@cpitclaudel I'm a bit confused by the declaration of `&optional` and `&key` arguments in the `(cl-defstruct (flycheck-error`.   `column` is marked as optional by the doc, but looks required to me.   `level` is optional but not documented as such.  `checker` is a keyword argument, so it is optional (right?), but no documented as such...

I could add a commit in this PR for cleaning this up, but I'd like your opinion.

cpitclaudel:
> To preserve the relationship between errors, this commit adds a group property
to flycheck-error objects. Checkers should use the same group value for
related errors.

Thanks. I like this approach; I find the simplicity appealing. Ideally I'd love to get an extra expert opinion (@Simplify? ^^), but otherwise I'm very much in favor of this.

cpitclaudel:
> I'm a bit confused by the declaration of &optional and &key arguments in the (cl-defstruct (flycheck-error. column is marked as optional by the doc, but looks required to me. level is optional but not documented as such. checker is a keyword argument, so it is optional (right?), but no documented as such...

I think it makes more sense if you assume that the docstring is documenting the invariants of the error data structure, and not the signature of the constructor. That being said, I wouldn't object bringing the two in closer agreement (by making the signature more strict) :)

flycheck.el
```diff
@@ -3157,6 +3167,18 @@ Return a list of all errors that are relevant for their
 corresponding buffer."
   (seq-filter #'flycheck-relevant-error-p errors))
 
+(defun flycheck-errors-from-group (err)
+  "Get all the errors that are in the same group as ERR.
```
cpitclaudel: Should this take a group rather than an error?

cpitclaudel: Or maybe it should have a different name?

flycheck.el
```diff
@@ -3157,6 +3167,18 @@ Return a list of all errors that are relevant for their
...
+
+Return a list of all errors (in `flycheck-current-errors') that
+have the same `flycheck-error-group' as ERR, including ERR
+itself."
+  (let ((group (flycheck-error-group err)))
+    (if group (seq-filter (lambda (e)
```
cpitclaudel: I think I'd use `-if-let` here :)

fmdkdd:
> I think it makes more sense if you assume that the docstring is documenting the invariants of the error data structure

Right, so the docstring stating `column (optional)` means that we can pass `nil` as second argument to the constructor?  And `checker` being required, that means we should always pass this key argument?

If so, yes it makes sense.  It's unfortunate that the signature isn't more strict, like checking that required fields passed as keywords are non-nil.  But then again, having too many unnamed arguments can be confusing when reading code.  So I'm okay leaving it as-is.

fmdkdd:
@cpitclaudel I've made the requested changes, and a small one to the docstring of `flycheck-error`.  I am too interested in other opinions on the matter, so let's see :)

fmdkdd:
@cpitclaudel Looks like it's the two of us.  Shall we go on with this approach?

fmdkdd:
Also, #1264 revealed there might be a second use-case in `flycheck-flow`, beyond the rust checker.

flycheck.el
```diff
@@ -5753,7 +5781,9 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
           :id error-code
           :checker checker
           :buffer buffer
-          :filename primary-filename)
+          :filename primary-filename
+          :group group
+          )
```
cpitclaudel: Nitpicking here, but this `)` looks lonely :)

cpitclaudel:
Yes, let's do that. I can think of many more uses, usually pertaining to related/secondary locations for errors.

PR #1251, opened 2017-05-12
Allowing shellcheck to follow files (SC1091)

seriousben:
Getting rid of Shellcheck warning SC1091 as documented in https://github.com/koalaman/shellcheck/wiki/SC1091

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9486,6 +9486,7 @@ By default, no warnings are excluded."
 
 See URL `https://github.com/koalaman/shellcheck/'."
   :command ("shellcheck"
+            "-x"
             "--format" "checkstyle"
             "--shell" (eval (symbol-name sh-shell))
             (option "--exclude" flycheck-shellcheck-excluded-warnings list


cpitclaudel:
Thanks, but don't we want this as an option, rather than always on?

CeleritasCelery:
Why would we want it as an option?

cpitclaudel:
I don't know; for the same reason, I'd expect, as shellcheck itself: why isn't it on by default there too?

CeleritasCelery:
Better put, why would we want Shellcheck enable but this feature disabled? It won't change the functionality unless they choose add a [directive](https://github.com/koalaman/shellcheck/wiki/SC1090) to point shellcheck to the nonconstanct source. I am not asking sarcastically, I am genuinely asking. I think this should just be always enabled unless someone can think of a reason not to.

cpitclaudel:
Point taken, but that's a discussion for the Shellcheck tracker: I don't see a good reason for Flycheck to diverge from Shellcheck itself in terms of defaults. Either `-x` always makes sense and Shellcheck should default to that, or it doesn't and we should have a flag.

CeleritasCelery:
Touché. Maybe I will see what they have to say about it.

seriousben:
Thanks @CeleritasCelery for asking the question https://github.com/koalaman/shellcheck/issues/902

seriousben:
@CeleritasCelery just got an answer. I think we got the 👍 from shellcheck to add the `-x` by default.

koalaman:
👍  from me.

cpitclaudel:
Thanks.  I'm fine with changing the default. Can we add an option to revert to the old behavior, though? Something like `flycheck-shellcheck-follow-includes`, maybe? We'll also want a news entry :) @CeleritasCelery , would you like to do these? Thanks a lot :)

seriousben:
I can do it if you want @CeleritasCelery

CeleritasCelery:
Since you opened this one, go ahead and make the change @seriousben

cpitclaudel:
Ping? I'd love to merge this :)

cpitclaudel:
I merged @CeleritasCelery 's PR, so we're all good :) Thanks everyone!

Issue #1259, opened 2017-05-31
haskell checkers: GHC 8 switched to lowercase `warning` and added whitespace, breaking linting


slpopejoy:
Using the `haskell-stack-ghc` checker, warnings are getting missed. Using `flycheck-compile` shows the following:

```
/Users/stuart/dev/pact/src/Pact/Compile.hs:49:1: warning: [-Wunused-imports]_
    The import of ‘Hash’ from module ‘Pact.Types.Crypto’ is redundant
```
(I added the underscore at the end of the line to highlight the whitespace that inconsistently shows up)

https://github.com/flycheck/flycheck/blob/master/flycheck.el#L7725 is explicitly looking for "Warning:", and the regex on the next line would need a " ?" after the close bracket. I would normally try using "[Ww]" (which fixes the issue for haskell mode) but rx is weird, so this is all I could figure out:

```
            (or " " "\n    ") (or "warning:" "Warning:")
            (optional " " "[" (id (one-or-more not-newline)) "] ?")
```

`haskell-ghc` needs the same change.

cpitclaudel:
Nice catch, thanks! Can you make a PR? Regarding Ww and rx, IIRC (in "Ww") should work.

PR #1261, opened 2017-06-01
recognize title- and lower-case for warnings

slpopejoy:
addresses #1259 . Tested on 7.10.x stack, 8.0.x stack, 7.10.x cabal, 8.0.x cabal.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7722,7 +7722,7 @@ See URL `https://github.com/commercialhaskell/stack'."
             source)
   :error-patterns
   ((warning line-start (file-name) ":" line ":" column ":"
-            (or " " "\n    ") "Warning:"
+            (or " " "\n    ") (in "Ww") "arning:"
             (optional " " "[" (id (one-or-more not-newline)) "]")
             (optional "\n")
             (message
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -7722,8 +7722,8 @@ See URL `https://github.com/commercialhaskell/stack'."
             source)
   :error-patterns
   ((warning line-start (file-name) ":" line ":" column ":"
-            (or " " "\n    ") "Warning:"
-            (optional " " "[" (id (one-or-more not-newline)) "]")
+            (or " " "\n    ") (in "Ww") "arning:"
+            (optional " " "[" (id (one-or-more not-newline)) "] ?")
```
cpitclaudel: Are you sure this should be " ?", rather than `(optional " ")`?

slpopejoy: you're right ... not sure how that ever caught anything. maybe i'll get rid of that change and just do the case change.

cpitclaudel: Probably weird interactions before the various `optional`. If the whitespace change is good it should stay, though :) Probably as `(optional " ")` right after that closing parenthesis.

cpitclaudel:
Thanks! Please squash and confirm whether we need that `" "`, and I'll merge :)

flycheck.el
slpopejoy: The ws change is just wrong, `flycheck-compile` is just displaying it wrong, maybe because it underlines where the carriage return is. On the cmdline, there's no extra space when I copypaste the `flycheck-compile` call. I think I was flailing when first trying to get this to work, frankly; and the bad "?" syntax proves that the change was a) broken but b) harmless. Buh-bye :)

cpitclaudel:
Thanks!

PR #1270, opened 2017-06-25
cargo-rust: remove arguments "-Z no-trans" as they are going away

lpenz:
The current code generates the following warning when used with
stable rust:
warning: the option `Z` is unstable and should only be used on the
         nightly compiler, but it is currently accepted for backwards
         compatibility; this will soon change, see issue #31847 for
         more details

See also:
- rust-lang/rust#31847
- rust-lang/rust#41751

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9110,7 +9110,7 @@ This syntax checker requires Rust 1.15 or newer.  See URL
                     flycheck-rust-binary-name))
             "--message-format=json"
             (eval flycheck-cargo-rustc-args)
-            "--" "-Z" "no-trans"
+            "--"
             ;; Passing the "--test" flag when the target is a test binary or
             ;; bench is unnecessary and triggers an error.
             (eval (when flycheck-rust-check-tests


fmdkdd:
Thank you.  I've been seeing the warning for a long time.  Is it a hard error in stable already?  Or in beta?   If no, then I don't want to get rid of `-Z no-trans` until `cargo check` can check tests (see flycheck/flycheck-rust#51).

lpenz:
It's still a warning in stable, but it exits with an error and the checker is not working as-is.
I'm at the moment redirecting flycheck-rust-cargo-executable to a shell script that remove those flags and passes the rest along to cargo - it's unusable otherwise.

fmdkdd:
Just tested.  The 1.19.0-beta.2 exits with an error, but the 1.18.0 (current stable) works fine for me.  Are you using the beta?  

In any case, we need to move to `cargo check` if we cannot use `-Zno-trans` anymore.

lpenz:
Agreed.
I was using 1.14 for compatibility reasons (app will be deployed to an older installation), but I've now tested with 1.18 and the behaviour is as you describe: there's a warning but it works. :/
I'm closing this PR, thanks.

PR #1273, opened 2017-06-27
Remove mentions to the @emacs_flycheck twitter account

fmdkdd:
Since the account doesn't exist anymore.

I am fine not having to deal with a twitter account, and don't really see the added value.  But I'm opening as a PR to let others chime in.  If you think we should have a twitter account for Flycheck, then let us know.

### DIFF
--- a/doc/_templates/about.html
+++ b/doc/_templates/about.html
@@ -30,12 +30,6 @@
 
 {% if not flycheck_offline_html %}
 <p>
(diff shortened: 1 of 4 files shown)


cpitclaudel:
I'm +1 on that :)

fmdkdd:
@cpitclaudel Wait, +1 for keeping a Twitter presence or for dropping it? :)

Issue #1269, opened 2017-06-23
Some systemd-analyze messages use a different format


joewreschnig:
Some of the messages from `systemd-analyze` don't contain line numbers and only print the file's basename. These messages appear on stderr, and the exit status is non-zero. One example is if the command for an ExecStart does not exist.

The input file (`system/foobar.service`):
```
[Unit]
Description=Example Service

[Service]
ExecStart=/home/foo/bar
```

Running `systemd-analyze`:
```
$ systemd-analyze verify system/foobar.service
foobar.service: Command /home/foo/bar is not executable: No such file or directory
```

I'm not sure how to make flycheck match an error without a full filename or line, or if such a thing is possible.

fmdkdd:
Indeed, looks like our test is failing as well.  Might be a recent change (read: last few months) to `systemd-analyze`?

Here is what `systemd-analyze` outputs for our test:
```
/tmp/flycheck4592Tnq/systemd-analyze-test.service:3: Invalid URL, ignoring: foo://bar
/tmp/flycheck4592Tnq/systemd-analyze-test.service:6: Unknown lvalue 'ExecSmart' in section 'Service'
/tmp/flycheck4592Tnq/systemd-analyze-test.service:8: Unknown section 'Dog'. Ignoring.
systemd-analyze-test.service: Service lacks both ExecStart= and ExecStop= setting. Refusing.
systemd-analyze-test.service: Failed to create systemd-analyze-test.service/start: Unit systemd-analyze-test.service is not loaded properly: Invalid argument.
```
The missing line number makes sense: it's an error at the file level, not from a specific line.  Flycheck can handle that just fine; I think we use the first line as default.

Mixing relative and absolute name seems like an oversight from `systemd-analyze` to me.  I would report it there.  If it's working as intended, then we can get always put the absolute file name ourselves with an `:error-filter` in the checker.

I've created #1275 to update the checker.  It doesn't catch the errors with the relative file name, but it should not complain about non-zero exit code.  Can you try that patch?

joewreschnig:
This is probably related to the systemd version being tested, but 232, in Debian 9 still uses the `[filename:line] message` pattern, as well as the `basename: message` pattern.

If the version you're using outputs `filename:line: message`, that's a third format that needs to be supported.

PR #1275, opened 2017-07-06
Update systemd-analyze error patterns

fmdkdd:
Closes GH-1269.


fmdkdd:
I was afraid of that, hence why I asked you to test.  Let's keep this PR open until we have an answer from systemd on the relative filename issue.

Supporting both `[filename:line]` and `filename:line` is easy enough in Flycheck anyway.

fmdkdd:
Fixed in #1624.

PR #1282, opened 2017-07-15
Pass current directory to gosimple

mpolden:
This is necessary to make gosimple aware of symbols declared in other files.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7587,7 +7587,9 @@ See URL `https://github.com/mdempsky/unconvert'."
 Requires Go 1.6 or newer. See URL `https://github.com/dominikh/go-tools'."
   :command ("gosimple"
             (option-list "-tags=" flycheck-go-build-tags concat)
-            source)
+            ;; Run in current directory to make gosimple aware of symbols
+            ;; declared in other files.
+            ".")
   :error-patterns
   ((warning line-start (file-name) ":" line ":" column ": " (message) line-end))
   :modes go-mode)
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3197,7 +3197,7 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
   (flycheck-ert-with-env
       `(("GOPATH" . ,(flycheck-ert-resource-filename "language/go")))
     (flycheck-ert-should-syntax-check
-     "language/go/src/gosimple/gosimple.go" 'go-mode
+     "language/go/src/gosimple/gosimple1.go" 'go-mode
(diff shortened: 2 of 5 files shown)


flycheck.el
```diff
@@ -7587,7 +7587,7 @@ See URL `https://github.com/mdempsky/unconvert'."
 Requires Go 1.6 or newer. See URL `https://github.com/dominikh/go-tools'."
   :command ("gosimple"
             (option-list "-tags=" flycheck-go-build-tags concat)
-            source)
+            ".")
```
fmdkdd: Please add a comment here to indicate why you are using "."

mpolden:
@fmdkdd Comment added. Didn't see an easy way to test this case. The test would have to test the absence of certain warnings (e.g. undeclared symbol), not sure if it's worth the effort.

fmdkdd:
The tests use strict equality between the expected and the actual warnings.  If `gosimple` reports extra warnings that are picked by the checker when not run on the directory, then the fact that these extra warnings are not part of the expected warnings will fail the test.

mpolden:
I see. I'll look into adding a test then, thanks.

mpolden:
@fmdkdd PTAL.

test/flycheck-test.el
```diff
@@ -3197,8 +3197,8 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
   (flycheck-ert-with-env
       `(("GOPATH" . ,(flycheck-ert-resource-filename "language/go")))
     (flycheck-ert-should-syntax-check
-     "language/go/src/gosimple/gosimple.go" 'go-mode
-     '(5 6 warning "should omit values from range; this loop is equivalent to `for range ...` (S1005)"
+     "language/go/src/gosimple/gosimple1.go" 'go-mode
+     '(4 6 warning "should omit values from range; this loop is equivalent to `for range ...` (S1005)"
```
fmdkdd: Are you sure about that `4`?  The tests expect a `5` on my machine.

mpolden: Yes, because I split it into two files, so the warning now occurs on line 4.

fmdkdd: I see that, but the `for` loop is on line 5 because you left a comment on line 4 in place of the `xs` declaration.  I ran your branch on my machine, and the test is failing.  Does it work for you?  (Integration tests are *not* tested on Travis, sadly)

mpolden: Oh right, I added the comment last 🤦‍♂️ . Fixed, test now passes locally.

fmdkdd: No worries :)  Thanks for the fix!

fmdkdd:
Thanks!  LGTM 👍

PR #1285, opened 2017-07-17
Update copyright years and maintainers for relevant files

fmdkdd:
All files that have been updated in 2017 now have an updated copyright notice.
The manual is a special case: it bears a single copyright notice at the bottom,
so this mark applies to all its pages.

Elisp packages containing the 'Maintainer' field have been updated.  Though,
another solution would be to replace them by a "see the MAINTAINERS file".

Better start using that `copyright-update` function in `before-save-hook` now.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -1,2 +1,2 @@
-Clément Pit--Claudel <hidden@example.com> (@cpitclaudel)
+Clément Pit-Claudel <hidden@example.com> (@cpitclaudel)
 fmdkdd <hidden@example.com> (@fmdkdd)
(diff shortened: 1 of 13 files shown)


fmdkdd:
@cpitclaudel Sure.  Do you wish to keep the en dash in [the manual](http://www.flycheck.org/en/latest/community/people.html#maintainers) or should it be a simple dash there as well?

cpitclaudel:
Simple dash there too :)  It was never intended to be an en-dash (just a literal double one — blame the [French government](http://www.liberation.fr/societe/2010/01/06/noms-de-famille-le-conseil-d-etat-efface-le-double-trait-d-union_602983) ^^)

fmdkdd:
Interesting!  Typography and administration make for a scary combo.

I've hunted and retired the extra dashes.

cpitclaudel:
Thanks!

Issue #1229, opened 2017-03-18
Add check-declare-function checker for emacs-lisp


darkfeline:
`check-declare-function` will ensure `declare-function` functions are actually declared in the indicated files, which the compiler does not check.  Since `declare-function` is used to silence compilation warnings for undefined functions, it would be nice to have a checker verify the `declare-function` uses.

cpitclaudel:
Nice idea :) Though I wonder why the compiler doesn't do that.  Shouldn't it?

darkfeline:
I think `declare-function` can also used to declare functions in a file that would exist at runtime, but not compile time.  It's sort of like how you only need the library headers to compile a C program, but the library itself may not be needed at compile time.

So in the general case the compiler shouldn't check it.

fmdkdd:
@darkfeline Do you want to send a PR for this?

darkfeline:
https://github.com/flycheck/flycheck/pull/1286

PR #1286, opened 2017-07-19
Add emacs-lisp-check-declare checker

darkfeline:


darkfeline:
Hm, this appears to work when I load the checker into my working Emacs, but the flycheck-test test doesn't pass, and ert is woefully opaque about why.

fmdkdd:
The output to ERT is not the most intuitive:
```
((should
       (equal expected flycheck-current-errors))
      :form
      (equal
       ([cl-struct-flycheck-error #<killed buffer> emacs-lisp-check-declare "/home/travis/build/flycheck/flycheck/test/resources/language/emacs-lisp/check-declare-errors.el" 10 1 "said ‘this-function-is-not-defined’ was defined in this-file-does-not-exist.el: file not found" warning nil nil])
       nil)
```
Inside the `:form (equal`, it expects the error you declared, but you got `nil` instead.  That means the error was either not caught by the checker, or the checker was not run at all.

I think the issue here is that your checker is never *selected* by the Flycheck run by ERT.  `emacs-lisp` is the default checker, followed by `emacs-lisp-checkdoc` (see the `:next-checker` property).

You could put your checker as a `:next-checker` for `checkdoc`.

darkfeline:
Ah, got it.  The tests pass now.

test/resources/language/emacs-lisp/check-declare-errors.el
```diff
@@ -0,0 +1,13 @@
+;;; check-declare-errors.el -- check-declare errors -*- lexical-binding: t; -*-
+
+;;; Commentary:
+
+;; Checkdoc uses the syntax table and comment syntax of the current buffer; if
```
cpitclaudel: Isn't that comment a leftover from another file?

flycheck.el
```diff
@@ -7161,12 +7162,34 @@ The checker runs `checkdoc-current-buffer'."
     (not (or (flycheck-autoloads-file-p)
              (and (buffer-file-name)
                   (member (file-name-nondirectory (buffer-file-name))
-                          '("Cask" "Carton" ".dir-locals.el")))))))
+                          '("Cask" "Carton" ".dir-locals.el"))))))
+  :next-checkers (emacs-lisp-check-declare))
```
cpitclaudel: Shouldn't this be added to the next-checkers of emacs-lisp as well?

fmdkdd:
>  shouldn't this be part of the default ELisp checker?

I agree.  It looks like we could just add the `check-declare-form` as an additional argument to the `emacs-lisp` checker, and save ourselves the trouble of calling `emacs` a third time.

fmdkdd:
(Reminder: #1229)

darkfeline:
Merged checker into emacs-lisp

darkfeline:
Looks like I need to get my hands on a build of Emacs 24, please wait warmly.

darkfeline:
Actually, it looks like I can't build Emacs 24.  I believe I'm running into the bug mentioned here: http://lists.linuxfromscratch.org/pipermail/blfs-support/2016-August/078284.html

It may be a while before I can get Emacs 24.  If someone else can fix this, feel free.  I suspect it's something like the error filter needing to be tweaked.

darkfeline:
The Emacs I used was 24.3, which flycheck doesn't support.  If the fix doesn't pass, I'll probably install a newer Ubuntu in a VM and try again.

I wonder if flycheck would benefit from shipping a Vagrantfile.

darkfeline:
Well, that was a waste of time.  As it turns out, check-declare does not print the line number on Emacs 24 and stepping through the debugger I learned that Flycheck discards errors that do not have line numbers.

I don't really feel like trying to make this work now, so I've skipped the test on Emacs 24 and documented that check-declare doesn't work on Emacs 24.

fmdkdd:
Thanks for investigating the issue.  So, currently the `check-declare` pass will be run by Flycheck on Emacs 24, but the errors won't be reported, is that it?

If the issue is missing line numbers in the output, you can check for errors that have a `nil` line and set them to 0 instead.  That way, the errors will be reported on the first line, and Flycheck will not discard them.

If it's too much a hassle to test that with Emacs 24, let me know.  I have an Emacs 24 install I can use to test this workaround.

darkfeline:
Currently on Emacs 24, `check-declare` will emit errors correctly, but they will be filtered out by `flycheck-relevant-errors`/`flycheck-relevant-error-p`.

Adding 0 line numbers to them sounds like a good solution; I'll try to do it with error-filters.

darkfeline:
Any update on this?

flycheck.el
```diff
@@ -6946,7 +6959,10 @@ See URL `https://github.com/lpil/dogma/'."
       (unwind-protect
           (byte-compile-file (car command-line-args-left))
         (mapc (lambda (f) (ignore-errors (delete-file f)))
-              flycheck-byte-compiled-files)))))
+              flycheck-byte-compiled-files))
+      (when (and (boundp 'flycheck-emacs-lisp-check-declare)
+                 flycheck-emacs-lisp-check-declare)
```
cpitclaudel: Could this be written as `bound-and-true-p`?

flycheck.el
```diff
@@ -3617,6 +3617,19 @@ IDs, that include the name of the tool."
...
+  "Set ERRORS without lines to line 0.
+
+Use as `:error-filter' for syntax checkers that output errors
+without line numbers.
+
+Return ERRORS."
+  (seq-do (lambda (err)
+            (unless (flycheck-error-line err)
+              (setf (flycheck-error-line err) 0)))
+          errors)
```
cpitclaudel: For simple iteration like this, I tend to prefer dolist over seq-do

flycheck.el
darkfeline: Done

flycheck.el
darkfeline: All of the other filters use `seq-do`, so I am inclined to keep the code consistent.

cpitclaudel:
Brilliant work, thanks!

fmdkdd:
Thank you @darkfeline for your patience, and for this nice contribution!

PR #1262, opened 2017-06-01
Improve haskell default directory detection

sergv:
I found out that `(buffer-file-name)` can return `nil` for temporary buffers, which leads to errors. Also it seemed like good idea to allow stack files like `stack-7.10.yaml` in addition to just `stack.yaml` for default directory searching purposes.

PR #1287, opened 2017-07-20
Improve default directory detection for Haskell stack projects

sergv:
As agreed in #1262, I submit improvement to detection of Haskell stack projects as a new pull request.

The brief motivation is: stack projects are expected to have `stack.yaml` file in their root directory that the `stack` tool will use up by default. However, it's reasonable to have multiple `stack-X.Y.Z.yaml` files in the root directory, each for different versions of `ghc`, the Haskell compiler. In such a case, there may not necessarily be a file named `stack.yaml` as developers will be expected to specify releval `stack-X.Y.Z.yaml` file each time they invoke the `stack` tool.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7724,8 +7724,8 @@ Otherwise return the previously used cache directory."
 (defun flycheck--locate-dominating-file-matching (file regexp)
   "Search for a file in directory hierarchy starting at FILE.
 
-Look up the directory hierarchy from START-DIR for a directory
-containing file that matches REGEXP."
+Look up the directory hierarchy from FILE for a directory
+containing a file that matches REGEXP."
   (locate-dominating-file
    file
    (lambda (dir)
--- a/flycheck.el
+++ b/flycheck.el
@@ -7721,6 +7721,20 @@ Otherwise return the previously used cache directory."
         (or flycheck-haskell-ghc-cache-directory
             (make-temp-file "flycheck-haskell-ghc-cache" 'directory))))
 
+(defun flycheck--locate-dominating-file-matching (start-dir regexp)
+  "Search for a file in directory hierarchy starting at START-DIR.
+
+Look up the directory hierarchy from START-DIR for a directory
+containing file that matches file that matches REGEXP."
+  (locate-dominating-file
+   start-dir
+   (lambda (dir)
+     (directory-files dir
+                      nil ;; use full paths
+                      regexp
+                      t ;; do not sort result
+                      ))))
+
 (defun flycheck-haskell--find-default-directory (checker)
   "Come up with a suitable default directory for Haskell to run CHECKER in.
 
@@ -7735,7 +7749,10 @@ contains a cabal file."
     (`haskell-stack-ghc
      (or
       (when (buffer-file-name)
-        (locate-dominating-file (buffer-file-name) "stack.yaml"))
+        (flycheck--locate-dominating-file-matching
+         (file-name-directory
+          (buffer-file-name))
+         "stack.*\\.yaml\\'"))
       (-when-let* ((stack (funcall flycheck-executable-find "stack"))
                    (output (ignore-errors
                              (process-lines stack "path" "--project-root")))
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -7721,6 +7721,20 @@ Otherwise return the previously used cache directory."
         (or flycheck-haskell-ghc-cache-directory
             (make-temp-file "flycheck-haskell-ghc-cache" 'directory))))
 
+(defun flycheck--locate-dominating-file-matching (start-dir regexp)
+  "Search for a file in directory hierarchy starting at START-DIR.
+
+Look up the directory hierarchy from START-DIR for a directory
+containing file that matches file that matches REGEXP."
```
fmdkdd: "containing a file that matches REGEXP" rather?

flycheck.el
```diff
@@ -7721,6 +7721,20 @@ Otherwise return the previously used cache directory."
...
+  (locate-dominating-file
+   start-dir
+   (lambda (dir)
+     (directory-files dir
+                      nil ;; use full paths
```
fmdkdd: `nil` does *not* use full paths, according to the doc I have.  Besides, I'm not sure the comments are necessary when you can lookup the docstring.

flycheck.el
```diff
@@ -7735,22 +7749,20 @@ contains a cabal file."
     (`haskell-stack-ghc
      (or
       (when (buffer-file-name)
-        (locate-dominating-file (buffer-file-name) "stack.yaml"))
+        (flycheck--locate-dominating-file-matching
+         (file-name-directory
```
fmdkdd: `(file-name-directory` is superfluous for `locate-dominating-file`, why not do the same for our new function?

flycheck.el
```diff
@@ -7735,22 +7749,20 @@ contains a cabal file."
...
+          (buffer-file-name))
+         "stack.*\\.yaml\\'"))
```
fmdkdd: Is there some official documentation for the filenames `slack` detects?  I see that `slack-nightly.yaml` is possible, so anything more restrictive than `.*` would probably be too verbose.

Also, what does the `\\'` means here?

flycheck.el
```diff
@@ -7735,22 +7749,20 @@ contains a cabal file."
...
       (-when-let* ((stack (funcall flycheck-executable-find "stack"))
                    (output (ignore-errors
                              (process-lines stack "path" "--project-root")))
                    (stack-dir (car output)))
         (and (file-directory-p stack-dir) stack-dir))))
     (_
      (when (buffer-file-name)
-       (locate-dominating-file
+       (flycheck--locate-dominating-file-matching
         (file-name-directory (buffer-file-name))
```
fmdkdd: Same here: we could pass `(buffer-file-name)` directly.

flycheck.el
sergv: Bascially, the `\\'` means the same thing as `$` here. Namely, `$` is end of line and `\\'` is end of string.

sergv:
@fmdkdd I have fixed your suggestions.

flycheck.el
fmdkdd: That's what I figured, but couldn't find the documentation on the spot.  Thanks.

flycheck.el
```diff
@@ -7721,6 +7721,16 @@ Otherwise return the previously used cache directory."
         (or flycheck-haskell-ghc-cache-directory
             (make-temp-file "flycheck-haskell-ghc-cache" 'directory))))
 
+(defun flycheck--locate-dominating-file-matching (file regexp)
+  "Search for a file in directory hierarchy starting at FILE.
+
+Look up the directory hierarchy from START-DIR for a directory
```
fmdkdd: `FILE` instead of `START-DIR`.

flycheck.el
```diff
@@ -7721,6 +7721,16 @@ Otherwise return the previously used cache directory."
...
+containing file that matches REGEXP."
```
fmdkdd: "containing **a** file"

fmdkdd:
Thanks!  @cpitclaudel do you want to take a second look?

sergv:
@cpitclaudel Yes, the change of `default-directory` second argument to `t` is on purpose in both `stack` and `cabal` cases since they all want absolute path. The path will be used as a `default-directory` later, so it's preferable to keep it absolute to avoid any ambiguity.

cpitclaudel:
Thanks!

flycheck.el
sergv: It seems this was not a good idea as this is causing following issue https://github.com/flycheck/flycheck-haskell/issues/72. What happens is that `locate-dominating-file` passes file name into lambda argument, that calls `directory-files` on it and throws an error. Should we restore previous behavior of making `flycheck--locate-dominating-file-matching` accept directory names only?

PR #1289, opened 2017-07-21
Use `cargo check` for rust-cargo checker

fmdkdd:
Now that `-Z no-trans` is not an option anymore in Rust 1.19, I had to disable the flag.

Unfortunately, that means longer wait time before getting error feedback in Flycheck, and also some unwanted side effects of actually compiling binaries.

Cargo has provided the `cargo check` command for a while now, but I've held onto using it as our main solution because I kept running into issues with it (flycheck/flycheck-rust/issues/51)... but now we have little choice. 

Here is a first implementation of `cargo check` that I made for flycheck/flycheck-rust/issues/55.  I'm using a combination of `cargo check` and `cargo test --no-run` for checking test code as well.

It has the following advantages:
- `cargo check` is the official way to get errors for a project, aside from the still-in-progress RLS.
- It supports arcane project setups that `-Zno-trans` did not.

However, there is one major hurdle right now: `cargo check` and `cargo test --no-run` report warnings only on the *first compilation*.  Subsequent calls do not report warnings (errors are always reported, because the compilation fails).  You would think that's actually not an issue, since we only call `cargo check` on file save, the file is touched, so `cargo check` will always compile and give the warnings.  But when testing, sometimes I see warnings, sometimes I don't.  This is unacceptable, and never happened with `-Zno-trans`.  

It also makes writing tests more cumbersome (I've worked around the lack of warnings by `cargo 
clean`ing the repo before each test).

Note that, without `-Zno-trans`, the *current* checkers (both `rust-cargo` and `rust`) exhibit the same behavior: warnings will only be emitted on the first successful compilation, and Flycheck will sometimes display them, but most likely will just flicker.  So at least, `cargo check` is not worst than what we have *right now*.

Also, we cannot pass arguments to `rustc` anymore.  The variable `flycheck-cargo-rustc-args` is of no use anymore, and `flycheck-rustc-args` will only be used by the plain `rust` checker.  That might break some setups.

TL;DR: For Rust 1.19 and up `cargo check` is the better solution, but overall it's worst than what we had since 1.15, unless we find a solution to the disappearing warnings.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@
 - **Breaking changes**
 
   - Remove javascript-jscs checker
+  - ``rust-cargo`` now requires Rust 1.17 or newer [GH-1289]
+  - Rename ``flycheck-cargo-rustc-args`` to ``flycheck-cargo-check-args``
+    [GH-1289]
+  - ``rust-cargo`` does not use the variable ``flycheck-rust-args`` anymore
+    [GH-1289]
 
 - New syntax checkers:
 
@@ -25,6 +30,7 @@
   - Flycheck will execute ``rubocop`` from directory where ``Gemfile`` is
     located. If ``Gemfile`` does not exist, old behaviour of running command
     from directory where ``.rubocop.yml`` is found will be used [GH-1368]
+  - ``rust-cargo`` now uses ``cargo check`` and ``cargo test`` [GH-1289]
 
 31 (Oct 07, 2017)
 =================
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -1019,12 +1019,12 @@ to view the docstring of the syntax checker.  Likewise, you may use
                        rust
 
       Check syntax and types with the Rust_ compiler.  In a Cargo_ project the
-      compiler is invoked through ``cargo rustc`` to take Cargo dependencies
+      compiler is invoked through ``cargo check`` to take Cargo dependencies
       into account.
 
       .. note::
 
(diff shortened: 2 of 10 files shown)


fmdkdd:
Update: the issue of disappearing warning is reported: rust-lang/cargo/issues/3624.

jjwest:
Is there any update on the current status of this PR?

fmdkdd:
@jjwest  The upstream issue is not going to be resolved any time soon.  I don't think we want to workaround it and cache the warnings ourselves in Flycheck.  So, the situation has not evolved much, I'm afraid.

fmdkdd:
I've updated this PR.  We can merge it after #1381 (as it's based on it).

`cargo check` still suffer from emitting warnings only on the first run, but the issue won't be resolved upstream in the foreseeable future, as there are no caching facilities in `cargo` proper.

However, `cargo check` is also quite a bit faster for most use cases.  And we are just losing *some* warnings, not any errors.  So I now believe the tradeoff is a good one to make.

The other downside is the loss of `rustc` args.  I didn't use them before, but hopefully it won't break anyone's setup.  If it does, users still have the option of defining the old `rust-cargo` checker in their config.

fmdkdd:
Oh, and this new PR is based on #1381, so better merge that before.

fmdkdd:
Rebased and ready to review.

cc @cpitclaudel

fmdkdd:
No worries, I'd rather wait for reviews than not have reviews at all 🎉

PR #1290, opened 2017-07-22
Add go-megacheck

mpolden:
I was initially going to add a checker for `staticcheck`, but then I found [megacheck](https://github.com/dominikh/go-tools/tree/master/cmd/megacheck), which runs both the `gosimple`, `staticcheck` and `unused` tools.

This replaces the existing gosimple checker, as it's no longer needed.


flycheck.el
```diff
@@ -7429,6 +7429,17 @@ This option requires Go 1.6 or newer."
...
 
+(flycheck-def-option-var flycheck-go-megacheck-disabled-checkers nil go-megacheck
+  "A list of checkers to disable when running `megacheck'.
+
+The value of this variable is a list of strings, where each
+string is a checker to be disabled. Valid checkers are `simple',
+`staticcheck' and `unused'. When nil, all checkers will be
+enabled. "
+  :type '(choice (repeat :tag "Disabled checkers"
+                         (string :tag "Checker name")))
```
fmdkdd: I think you can use `const` instead of `string` here to not allow arbitrary strings, but only the names of the checkers that make sense.

mpolden:
> `go` seems to have a lot of external tools. Good for them!

Yes, Go and the community seem to follow the UNIX philosophy closely, e.g "Do one thing and it well".

> You'll want to change the line in `CHANGES.rst` as well.

Done.

> I think you can use const instead of string here to not allow arbitrary strings, but only the names of the checkers that make sense.

Done, not that familiar with customize as I rarely use it. Changed to `set` and `const`. Verified expected behavior by customizing the variable.

PTAL. Love the rigorous testing and reviewing in this project by the way! 👍

fmdkdd:
LGTM.

> Love the rigorous testing and reviewing in this project by the way!

Thanks!  I guess that's a consequence of having most users pulling from `master` via MELPA to get Flycheck.  You really don't want to merge stuff without taking a good look at it ;)

@cpitclaudel Want to double check?

cpitclaudel:
LGTM. Just curious though — shouldn't we have both simplecheck and megacheck? Or do we expect everyone to just install megacheck?

dominikh:
Using megacheck will be a lot more efficient than running gosimple, staticcheck and unused separately, so it could be worth only supporting megacheck and pushing users in that direction.

cpitclaudel:
Ok, works for me :)

cpitclaudel:
Thanks @mpolden! Congrats on this nice contribution.

PR #1292, opened 2017-07-23
Make `flycheck--locate-dominating-file-matching' operate on directories

### LINKED ISSUE #1291: Opening directory: not a directory error
When working with Haskell files I'm getting this:

```
Error while checking syntax automatically: (file-error "Opening directory" "Not a directory" "/home/mark/projects/new-tests/ih-masks/src/Main.hs") [2 times]
Opening directory: Not a directory, /home/mark/projects/new-tests/ih-masks/src/Main.hs
```

If you can't reproduce, I'll bite the bullet and try to mess with `emacs -Q` to reproduce (which is a huge pain because my Emacs is heavily customized and I'm accustomed to it).

sergv:
This is necessary because it uses `locate-dominating-file' which, if
acting on a regular file will pass that regular file name into it's
second lambda argument. But we're passing lambda that expects only
directories.

Fixes #1291 and https://github.com/flycheck/flycheck-haskell/issues/72.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7738,13 +7738,13 @@ Otherwise return the previously used cache directory."
         (or flycheck-haskell-ghc-cache-directory
             (make-temp-file "flycheck-haskell-ghc-cache" 'directory))))
 
-(defun flycheck--locate-dominating-file-matching (file regexp)
-  "Search for a file in directory hierarchy starting at FILE.
+(defun flycheck--locate-dominating-file-matching (directory regexp)
+  "Search for a file in directory hierarchy starting at DIRECTORY.
 
-Look up the directory hierarchy from FILE for a directory
+Look up the directory hierarchy from DIRECTORY for a directory
 containing a file that matches REGEXP."
   (locate-dominating-file
-   file
+   directory
    (lambda (dir)
      (directory-files dir t regexp t))))
 
@@ -7763,7 +7763,7 @@ contains a cabal file."
      (or
       (when (buffer-file-name)
         (flycheck--locate-dominating-file-matching
-         (buffer-file-name)
+         (file-name-directory (buffer-file-name))
          "stack.*\\.yaml\\'"))
       (-when-let* ((stack (funcall flycheck-executable-find "stack"))
                    (output (ignore-errors
@@ -7773,7 +7773,7 @@ contains a cabal file."
     (_
      (when (buffer-file-name)
        (flycheck--locate-dominating-file-matching
-        (buffer-file-name)
+        (file-name-directory (buffer-file-name))
         ".+\\.cabal\\'")))))
 
 (flycheck-define-checker haskell-stack-ghc


fmdkdd:
@sergv  Sorry for the suggestion of passing `file` directly to `flycheck--locate-dominating-file-matching`.  The behavior of `locate-dominating-file` is surprising, though.

I think I'd still like like it better if `flycheck--locate-dominating-file-matching` took a `file`, and we use `file-name-directory` on it before passing it to `locate-domainting-file` (what @cpitclaudel suggested in #1291).

cpitclaudel:
(related: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=27798)

sergv:
So, to the API of `flycheck--locate-dominating-file-matching` should be that it accepts either file or directory as its first argument?

flycheck.el
```diff
@@ -7744,7 +7744,12 @@ Otherwise return the previously used cache directory."
 Look up the directory hierarchy from FILE for a directory
 containing a file that matches REGEXP."
   (locate-dominating-file
-   file
+   ;; We must not pass regular file here because
+   ;; `locate-dominating-file' will just pass it into it's second
+   ;; lambdaargument . But our lambda expects only directories.
+   (if (file-directory-p file)
```
fmdkdd: You don't need this test; you can just always call `(file-name-directory file)` even if `file` is really a directory, because in the end they are just strings and the function will always return a directory.

fmdkdd:
@sergv Yes.  It's rather common in (E)Lisp to have union arguments.

flycheck.el
sergv: Yes, but if I pass `"/usr/bin"` there then calling `file-name-diroctory` on it will produce `"/usr/"` directory, thus unconditionally stripping `bin` while I may have meant to consider too.

fmdkdd: That's a fair point.  It does make the function more robust, even though we shouldn't run into any issue considering we always pass a filename at present.

flycheck.el
```diff
@@ -7744,7 +7744,12 @@ Otherwise return the previously used cache directory."
 Look up the directory hierarchy from FILE for a directory
 containing a file that matches REGEXP."
   (locate-dominating-file
-   file
+   ;; We must not pass regular file here because
+   ;; `locate-dominating-file' will just pass it into it's second
```
fmdkdd: "it's" -> "its"

flycheck.el
```diff
@@ -7744,7 +7744,12 @@ Otherwise return the previously used cache directory."
...
+   ;; lambdaargument . But our lambda expects only directories.
```
fmdkdd: "lambdaargument" -> "lambda argument", and there's an extra space before the dot "."

flycheck.el
```diff
@@ -7745,8 +7745,8 @@ Look up the directory hierarchy from FILE for a directory
 containing a file that matches REGEXP."
   (locate-dominating-file
    ;; We must not pass regular file here because
-   ;; `locate-dominating-file' will just pass it into it's second
-   ;; lambdaargument . But our lambda expects only directories.
+   ;; `locate-dominating-file' will just pass it into its second
+   ;; lambda argument . But our lambda expects only directories.
```
fmdkdd: "argument ." -> "argument."

sergv: That's embarrassing... Thanks for bearing with me!

flycheck.el
```diff
@@ -7744,7 +7744,12 @@ Otherwise return the previously used cache directory."
 Look up the directory hierarchy from FILE for a directory
 containing a file that matches REGEXP."
   (locate-dominating-file
-   file
+   ;; We must not pass regular file here because
+   ;; `locate-dominating-file' will just pass it into its second
+   ;; lambda argument . But our lambda expects only directories.
+   (if (file-directory-p file)
+       file
+     (file-name-directory file))
```
cpitclaudel: Do you need the `if` here? (Isn't `file-name-directory` idempotent?)

fmdkdd: > Yes, but if I pass "/usr/bin" there then calling file-name-diroctory on it will produce "/usr/" directory, thus unconditionally stripping bin while I may have meant to consider it too.

sergv: It's not idempotent - it either strips trailing slash or last path element, `/usr/bin/ => /usr/bin`, `/usr/bin => /usr/`.

cpitclaudel: Which version of Emacs are you on? 

```
ELISP> (file-name-directory "/usr/bin/")
"/usr/bin/"
ELISP> (file-name-directory "/usr/bin")
"/usr/"
```

fmdkdd: @cpitclaudel The issue here is that when passing `/usr/bin`, the search would start at `/usr` and not `/usr/bin`.   His argument is that can be surprising, since `/usr/bin` happens to be a directory.

So, it boils down to making a hard distinction between files and directories.

sergv: I'm on version `25.2`.

I still don't understand all the reasons why shouldn't `flycheck--locate-dominating-file-matching` be restricted to requiring directories? This way callers can decide whether they have a regular file path at hand or a directory path and call `file-name-directory` appropriately instead of trying to make `flycheck--locate-dominating-file-matching` as generic as possible.

cpitclaudel: @fmdkdd ok; thanks for the explanation. @sergv I'm fine with that too, as long as we document the behavior (@fmdkdd, WDYT?)

cpitclaudel: Actually, let's please do this.  Eli said on debbugs that passing a file is never valid (it should always be a directory)

fmdkdd: >  I still don't understand all the reasons why shouldn't flycheck--locate-dominating-file-matching be restricted to requiring directories?

Coherence with how `locate-dominating-file` was already used, mostly.

But if we actually have been using `locate-dominating-file` wrong all this time, then there is no reason to held onto it.  (Note: we should probably hunt these down in the code now)

Looks like you had the right idea all along @sergv  👍  Sorry for the back-and-forth.

Please restore the `DIRECTORY` as first argument to the new flycheck function (we don't need the comment anymore).

Please also squash you commits as one before merging.

cpitclaudel: > But if we actually have been using locate-dominating-file wrong all this time, then there is no reason to held onto it. (Note: we should probably hunt these down in the code now)

Well, we can find solace in knowing that we're note alone :) https://debbugs.gnu.org/cgi/bugreport.cgi?bug=27798#11

@sergv Thanks for your efforts and patience

sergv:
Okay, back to square one :)

fmdkdd:
Thank you for your patience @sergv.  Happy to have you aboard!

Now we can fix this bug.

PR #1299, opened 2017-08-05
Quote checkdoc variables in form passed to subprocess

juergenhoetzel:
Prevents invalid function evaluation of list variables like
`checkdoc-symbol-words'.

Fixes error when setting 

```elisp
(setq checkdoc-symbol-words '("top-level")) 
```

which results in:
_Suspicious state from syntax checker emacs-lisp-checkdoc: Flycheck checker emacs-lisp-checkdoc returned non-zero exit code 255\
, but its output contained no errors: Invalid function: "top-level"_

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7182,7 +7182,7 @@ See Info Node `(elisp)Byte Compilation'."
 
 Variables are taken from `flycheck-emacs-lisp-checkdoc-variables'."
   `(progn
-     ,@(seq-map (lambda (opt) `(setq-default ,opt ,(symbol-value opt)))
+     ,@(seq-map (lambda (opt) `(setq-default ,opt ',(symbol-value opt)))
                 (seq-filter #'boundp flycheck-emacs-lisp-checkdoc-variables))))
 
 (flycheck-define-checker emacs-lisp-checkdoc


flycheck.el
```diff
@@ -7182,7 +7182,7 @@ See Info Node `(elisp)Byte Compilation'."
 
 Variables are taken from `flycheck-emacs-lisp-checkdoc-variables'."
   `(progn
-     ,@(seq-map (lambda (opt) `(setq-default ,opt ,(symbol-value opt)))
+     ,@(seq-map (lambda (opt) `(setq-default ,opt ,(custom-quote (symbol-value opt))))
```
cpitclaudel: Would a simple quote be enough here? Could it hurt?

juergenhoetzel: `quote` doesn't eval it's argument. Also we don't want to quote keywords/strings and so on...

cpitclaudel: I meant putting the quote outside of the `,(…)`.
I don't think quoting keywords/string is an issue, is it?

juergenhoetzel: I see.

But it _will hurt_ for example:
```elisp
(if 'nil false t)    
```
I suspect this is the reason why `custom-quote` exists.

cpitclaudel: Sorry to turn this into a long thread :/ I'm not sure what your example shows...

```
ELISP> (if 'nil false t)
t
ELISP> (equal (if 'nil false t) (if nil false t))
t
``

juergenhoetzel:
> Sorry to turn this into a long thread :/ I'm not sure what your example shows...

It's my fault: I confounded things. You are right: just amended the commit.

fmdkdd:
@cpitclaudel Is this good to merge now?

cpitclaudel:
Yup, thanks for the reminder :)

PR #1304, opened 2017-08-20
Add proselint checker

marsam:
Hi:

This is a rebase of #939

Minor changes with respect to #939:

+ Use standard input
+ Added `gfm-mode` to `:modes`
+ Replaced `json-read-from-string` with `flycheck-parse-json`

Related:
+ https://github.com/melpa/melpa/pull/4938

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,7 @@
   - Ruby with ``reek`` [GH-1244]
   - Go with ``megacheck`` [GH-1290]
   - LLVM IR with ``llc`` [GH-1302]
+  - Text prose with ``proselint`` [GH-1304]
 
 - New features:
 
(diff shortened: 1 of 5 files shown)


purcell:
I've looked through this, and it looks like a no-brainer, but would a more active maintainer care to click "merge" if they agree please?

cpitclaudel:
@purcell Thanks for the review. @marsam Thanks for your patience and your excellent work!

cpitclaudel:
Any reason to use ``(pcase (`"error" …) …)`` rather than ``(pcase ("error" …) …)`` btw? We use the former in multiple places (and it's valid according to the docs), but I hadn't paid attention to it before and I always use the latter.

fmdkdd:
@cpitclaudel I never quite managed to understand the docstring of `pcase` on that point to figure out if they were really equivalent, so I always go with the backtick, for coherency with the existing code.

cpitclaudel:
They expand to the same thing :) (as an example, here's what our use of pcase in the checkstyle parser expands to:)

```
(cond ((equal .severity '"error")
       'error)
      ((equal .severity '"warning")
       'warning)
      ((equal .severity '"info")
       'info)
      (t 'error))
```

fmdkdd:
@cpitclaudel Good to know (but still confusing to have two ways to state the same thing) :)

PR #1305, opened 2017-08-21
Make sure that idle timer handlers run in the right buffer

cpitclaudel:
This needs careful review, but at least I have a reasonably simple repro. This problem is the Flycheck side of https://github.com/FStarLang/fstar-mode.el/issues/73 , and it causes Flycheck to repeatedly re-check the same buffer, in a loop.

There's an easy way to see a limited form of this issue; I'll explain how to get the loop after that:

* Create two buffers with Flycheck enabled (say, in Python mode)
* Make a change in one and immediately change to the other (in less `flycheck-idle-change` seconds)
* Observe that Flycheck runs in the current buffer, not in the one that was modified.

This yields a loop when the error parser of a checker creates a buffer in which Flycheck can run:

1. Start a generic check
2. In e.g. the error parser, create a temporary buffer to process the checker's output
3. Put that temporary buffer in a mode that can run Flycheck
4. Make modifications in the temporary buffer

In that scenario, changes in the temporary buffer will cause an idle timer to be registered, and when the idle timer fires Flycheck will run a new check in the *current* buffer (not in the one where the change happened).  This run will cause a new timer to be scheduled, etc.

The fix is to make sure that the timer runs in the right buffer.  WDYT?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2744,12 +2744,14 @@ buffer."
           (flycheck-buffer-automatically 'new-line 'force-deferred)
         (setq flycheck-idle-change-timer
               (run-at-time flycheck-idle-change-delay nil
-                           #'flycheck-handle-idle-change))))))
+                           #'flycheck-handle-idle-change (current-buffer)))))))
 
-(defun flycheck-handle-idle-change ()
-  "Handle an expired idle time since the last change."
-  (flycheck-clear-idle-change-timer)
-  (flycheck-buffer-automatically 'idle-change))
+(defun flycheck-handle-idle-change (buffer)
+  "Handle an expired idle timer in BUFFER since the last change."
+  (when (buffer-live-p buffer)
+    (with-current-buffer buffer
+      (flycheck-clear-idle-change-timer)
+      (flycheck-buffer-automatically 'idle-change))))
 
 (defun flycheck-handle-save ()
   "Handle a save of the buffer."


fmdkdd:
Running the check on idle in another buffer is indeed incorrect.  Thanks for catching that one :)

The patch looks good.  One way to view the problem is that `run-at-time` is global and not somehow restricted to the buffer that called it.  Passing the buffer works around that.

> Put that temporary buffer in a mode that can run Flycheck

For this particular case, ensuring that `flycheck-mode` is disabled in the temporary buffer would have worked as well, right?  Or at least, ensuring it never checks automatically?

In fact, should Flycheck even be enabled in temporary buffers by default?  (Though that's a separate issue)

fmdkdd:
Ah yes, we have `flycheck-may-enable-mode` that already skips ephemeral buffers, but not temporaries.  Well, temp buffers created by `with-temp-buffer` are named `*temp*` by default.  

Since these are not user-facing, it makes sense to not turn on Flycheck automatically for these.  WDYT?

cpitclaudel:
Disabling Flycheck in temp buffers would be good, but that wouldn't have been enough to hide that bug :) In my case I actually reuse an ephemeral buffer (with a space at the beginning of the name) across syntax checks.

Looking at `flycheck-may-enable-mode`, you'd think I shouldn't have seen Flycheck running. And indeed, it's true that in that case `flycheck-mode-on-safe` leaves Flycheck off. But I also have an explicit call to `(flycheck-mode)` in my `prog-mode-hook`, which bypasses `flycheck-may-enable-mode` entirely and does enable Flycheck (that `may-enable-mode` check only runs when the mode is automatically enabled through `flycheck-global-mode`).

Note also that in some cases enabling flycheck-mode is not enough to cause an actual syntax check to run: once `flycheck-mode` is enabled, `flycheck-may-check-automatically` still needs to return non-nil before an automatic syntax check runs. Since `flycheck-may-check-automatically` calls `flycheck-ephemeral-buffer-p` too (just like `flycheck-may-enable-mode`), with this PR Flycheck never actually ends up running a checker in the ephemeral buffer and we don't get into a loop.

I wonder whether flycheck-may-check-automatically should have more conditions in common with `flycheck-may-enable-mode`…

fmdkdd:
> Disabling Flycheck in temp buffers would be good, but that wouldn't have been enough to hide that bug :)

Oh, okay.  I wasn't suggesting that we substitute that for your patch in any case.  Passing the buffer that created the timer is the correct thing to do.  I was merely wondering if it was fruitful to let `global-flycheck-mode` to enable itself on temporary buffers (though you'd need to change the buffer mode to something other than Fundamental?).

> I wonder whether flycheck-may-check-automatically should have more conditions in common with flycheck-may-enable-mode…

Well, this is only useful if you called `flycheck-mode` yourself in buffers that would have been skipped by `may-enable-mode`.  In that case, you probably want flycheck to run, no?

cpitclaudel:
> I was merely wondering if it was fruitful to let global-flycheck-mode to enable itself on temporary buffers (though you'd need to change the buffer mode to something other than Fundamental?).

I think you made a good point. But looking at it again, I also think flycheck is disabled in those too. The buffer name isn't `*temp*`, it's ` *temp*`, so we'll classify these as ephemeral and we won't automatically enable ourselves in them.

>>  I wonder whether flycheck-may-check-automatically should have more conditions in common with flycheck-may-enable-mode…

> Well, this is only useful if you called flycheck-mode yourself in buffers that would have been skipped by may-enable-mode. In that case, you probably want flycheck to run, no?

Not necessarily. As a concrete example, org-mode uses fontification buffers in which it activates apropriate major modes to fontify code snippets.  If our users have something like `(add-hook 'prog-mode-hook 'flycheck-mode)` in their configuration (and [many people](https://github.com/search?q=add-hook+%27prog-mode-hook+%27flycheck-mode&type=Code&utf8=%E2%9C%93) do), then flycheck will get enabled in these buffers (it won't automatically run a syntax check, though, because of the ephemeral nature of the buffer).

I hope I'm making sense :)

tchajed:
I think you meant that the name of the temp buffer is `" *temp*"` (apparently trimming the leading space was done in [accordance with CommonMark](http://spec.commonmark.org/0.28/#code-spans)).

fmdkdd:
Ah, indeed, temporary buffers have a leading space, so they count as ephemeral (thankfully), and are already skipped by Flycheck.

> flycheck will get enabled in these buffers (it won't automatically run a syntax check, though, because of the ephemeral nature of the buffer).

So, your point is that there are common cases where `flycheck-mode` is unintendedly turned on, bypassing `flycheck-may-enable-mode`?  Granted.

Which checks from `flycheck-may-enable-mode` do you want to reuse?
```
Flycheck mode is not enabled for

- the minibuffer,
- `fundamental-mode'
- major modes whose `mode-class' property is `special',
- ephemeral buffers (see `flycheck-ephemeral-buffer-p'),
- encrypted buffers (see `flycheck-encrypted-buffer-p'),
- remote files (see `file-remote-p'),
- and major modes excluded by `flycheck-global-modes'.
```

cpitclaudel:
Thanks for pulling out the list. Let's see :)

- the minibuffer,
I don't think we need this one; people should just not enable Flycheck there

- `fundamental-mode'
Same here

- major modes whose `mode-class' property is `special',
Same here

- ephemeral buffers (see `flycheck-ephemeral-buffer-p'),
We already check for that in the idle change handler

- encrypted buffers (see `flycheck-encrypted-buffer-p'),
I think we could enable this one. It's bad if Flycheck causes an unencrypted copy of a file to be saved somewhere.

- remote files (see `file-remote-p'),
That one is tricky. I don't think we support tramp explicitly, but some generic checkers support it just fine.

- and major modes excluded by `flycheck-global-modes'.
The docs of this variable say "Only has effect when variable ‘global-flycheck-mode’ is non-nil.", so we probably shouldn't use it there.

Thinking more about this: part of the issue is that these checks can't be bypassed, so it might not make sense to make them too stringent. There are modes that work just fine with Flycheck in remote buffers, for example.

Given this, I'm inclined to backtrack and leave things as is. OK with you, @fmdkdd?
And I'll merge this soon, if you give it the go-ahead :)

fmdkdd:
> Given this, I'm inclined to backtrack and leave things as is. OK with you, @fmdkdd?

I agree.

> And I'll merge this soon, if you give it the go-ahead :)

LGTM!

articuluxe:
FYI.  This throws an error when auto-complete is installed, as auto-complete advices flycheck-handle-idle-change

cpitclaudel:
Thanks for the heads-up. Sorry for the trouble that this caused / is causing. Did you report the error on the autocomplete side? Can you give us more details, like a backtrace? I'll look at autocomplete (but I've never used it). I wonder why it advises Flycheck functions :/

cpitclaudel:
The originating commit seems to be this, 4 years ago:
https://github.com/auto-complete/auto-complete/commit/1edc7a9d46cf05b2fa65e37ca015a67780c060a3
Not sure yet what it does / why it's there.

articuluxe:
It seems that if auto-complete is in the middle of completion, it reschedules flycheck-handle-idle-change to not activate until after completion has ended (polling until ac-completing is nil).

I believe the intention is to avoid spurious errors from syntax checkers originating from the contents of the completion dialog itself, which obviously would fail syntax checks.

The specific error is due to 'wrong-number-of-arguments.  I can try to get a good stack trace later, it is all byte-compiled right now.

cpitclaudel:
Thanks, that's very useful. I think the autocomplete dialog is an overlay, so Flycheck wouldn't check that, but it nonetheless would make sense to delay the checks until the user is finished typing their complete-able symbol.

The wrong-number-of-arguments error makes sense, too.
The fix on the autocomplete side is easy, and it should be backwards-compatible (just take an extra `buffer` argument, only make sure it's optional and use `(current-buffer)` if it isn't present (for compatibility with older `flycheck`s). How likely is it that we could get such a patch merged quickly? I see the latest auto-complete commit is 7 months old :/

We could also fix it on our side, at the cost of some ugliness. Basically instead of passing the buffer argument to that function directly, we'd call a lambda wrapping the function in a with-current-buffer.  Something like this:

```diff
(run-at-time flycheck-idle-change-delay nil
-            #'flycheck-handle-idle-change (current-buffer)))))))
+            (let ((cb (current-buffer))) 
+              (lambda () 
+                (when (buffer-live-p cb)
+                  (with-current-buffer cb (flycheck-handle-idle-change)))))
```

(and of course we'd remove the buffer argument added by this PR)

I'd tend to be -1 on this and +1 on patching auto-complete, but happy to discuss and sad to have broken things :/

Long term, we should offer an API for this, instead of getting advised. ELisp has pretty strict guidelines about not having advice in packages:

> It is […] a bad idea for one Lisp package to advise a function in another Lisp package
> https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html

fmdkdd:
> Long term, we should offer an API for this, instead of getting advised. ELisp has pretty strict guidelines about not having advice in packages:

+1 on that.  In fact, instead of adding an advice, the auto-complete authors should just have pinged us about adding an API for holding the idle change checks.  Using an advice *is* effectively creating an ad-hoc API on, but without us knowing, thus is likely to break.

I'm also in favor of patching auto-complete rather than adding a workaround in Flycheck.  That said, it looks like there is a [significant number of people](https://github.com/search?p=1&q=flycheck-handle-idle-change&type=Code&utf8=%E2%9C%93) either using auto-complete or overriding `flycheck-handle-idle-change`.

cpitclaudel:
Indeed, most of them seem to be coming from https://github.com/magnars/.emacs.d :/
Ouch.
Maybe we should go with the lambda option then? I can prepare a PR.

cpitclaudel:
CC @magnars (Hi!), since I expect it'll be hard to get all users of his config to update (harder than one commit in auto-complete). @magnars, the TL;DR is that this PR changed the signature of a function that your config overrides, which will break things for all users of your config, as far as we can tell (This wouldn't really have been a problem if your override had taken an `&rest` argument). Am I wrong in assuming that your config doesn't auto-update, and thus that fixing things there won't immediately fix things for all of your users?

We'd like it better if package and config authors didn't advise or override functions from Flycheck, so maybe you can give us an idea of the API you were missing? Thanks! :)

magnars:
It's been 4 years, so thankfully the override stated its reason:

> "This is an overwritten version of the original flycheck-handle-idle-change, which removes the forced deferred. Timers should only trigger inbetween commands in a single threaded system and the forced deferred makes errors never show up before you execute another command."

From what I can tell, my override is now equivalent to the version in flycheck itself - so it seems the Flycheck authors agreed with me at some point.

I'll remove my override, but you're right: People who have been using my config won't get the fix. This is certainly suboptimal. If you want to be helpful to these poor souls, maybe you could check the arity of the function and issue a warning?

Sorry about the inconvenience.

cpitclaudel:
> From what I can tell, my override is now equivalent to the version in flycheck itself - so it seems the Flycheck authors agreed with me at some point.

Super.

> I'll remove my override, but you're right: People who have been using my config won't get the fix. This is certainly suboptimal. If you want to be helpful to these poor souls, maybe you could check the arity of the function and issue a warning?

Makes sense. I don't think I know a better way to do an arity check than to [try to call the function](https://stackoverflow.com/questions/17693187/elisp-get-function-arity), but I think there's another OK workaround here. @fmdkdd I'll make a PR soon.

> Sorry about the inconvenience.
No worries. Thanks for the quick response!

fmdkdd:
@cpitclaudel Good.  The workaround looks like the best course of action due to the number of configs that implicitly depend on that function, unfortunately.

Issue #1297, opened 2017-08-01
Option to run Flycheck after buffer switch


raxod502:
I'd like to have Flycheck run after a short delay (specifically, after `flycheck-idle-change-delay`) whenever I switch buffers.

The current behavior is that syntax checks occur only after modifications to the buffer, or after a save, as per the default setting of `flycheck-check-syntax-automatically`.

However, it often happens that I change some configuration externally, or I make modifications in another file. Then, when I switch back to an old buffer that is supposed to now have errors in it, Flycheck does not run another check. I have to manually edit the buffer or save it. My current workflow is to switch to a buffer, insert a character, delete it, and save—after that, Flycheck will show the errors. And I feel really dumb every time I do that.

I'd propose for a new value `buffer-change` to be added to `flycheck-check-syntax-automatically`, which would cause a syntax check to be run after `flycheck-idle-change-delay` (or an analogous variable), whenever the user switched buffers.

fmdkdd:
Thanks for the idea.  I like it!

The only downside I'm seeing is that with this new setting, a syntax check would begin every time you switch to this buffer, even if the buffer hasn't changed, right?

But what if your configuration file hasn't changed either?  You don't need to run the syntax check, but Flycheck cannot know because it has no knowledge of the dependencies between your buffers.

It looks like the correct solution here is to make Flycheck *project-aware*, which is out of scope for now.

However, your suggestion is a nice compromise to that.  We don't *need* to know the dependencies, we just run the syntax check regardless.  

I'd be wary about making this new option part of the default value for `flycheck-check-syntax-automatically`, but otherwise I'm okay offering it as a choice.

raxod502:
I agree with all the things you have said. I might see if I can make a pull request (but that depends on how complicated it turns out to be).

fmdkdd:
@raxod502 Sure, that would be lovely.  You can make an incomplete PR and we can answer any questions you might have in there directly.

raxod502:
I have created pull request #1308 to implement this feature.

PR #1308, opened 2017-08-26
Allow running syntax check after buffer switch

raxod502:
This pull request addresses #1297. It does not yet include documentation and tests, since I would first like to ask if there are any technical problems with the patch.

In particular, we may not want to introduce a dependency on `switch-buffer-functions`, and instead implement that functionality ourselves (it is only about 20 lines of code).


raxod502:
Unfortunately, Cask does not work for me (see https://github.com/cask/cask/issues/393), so I'll likely be pushing a number of broken commits and using Travis CI to run the tests. Sorry about that.

raxod502:
Also, this project is absolutely wonderful to contribute to on account of the detailed documentation and automated testing frameworks. Thank you!

fmdkdd:
Thank you for pushing forward with this!  👍 

I don't have time to review it thoroughly right now, but I've taken a brief look:

1. About `switch-buffer-functions`.  Indeed, we might not want to use that, besides adding a dependency, I don't like that it runs on `post-command-hook`.  It seems you can hook into `buffer-list-update-hook` and you would get mostly the same effect without being so wasteful.

2. In the docstring, you mention:
> If the buffer is switched again during this time, Flycheck will wait again
I don't think that's the case currently.  The way I read it, if you switch to a buffer in flycheck mode,`flycheck-handle-buffer-switch` will start a timer.  If you then switch to another non-flycheck buffer before the timer ends, the timer is not cleared, and the check will start in the previous buffer after the delay runs out.

If you keep switching between different flycheck buffers (say, go to buffer A, then B, then C, then D, etc), you will start a timer in each one, and never clear them.  Which *might* be useful for your use case, as a quick way to "refresh" the checks in the buffer that are part of the same project.

But maybe we also want the behavior of not checking the buffer if we switch to another one before the check has happened.  We add an option for both behaviors.

I might be able to review this in depth on Monday.  Until then, maybe @cpitclaudel can take a look at it.

raxod502:
> It seems you can hook into `buffer-list-update-hook` and you would get mostly the same effect without being so wasteful.

I had no idea this hook existed, despite concerted searching for something like that. Might be good to merge that into `switch-buffer-functions` upstream, although I also agree that it probably makes sense to just implement the logic here in Flycheck.

> If you then switch to another non-flycheck buffer before the timer ends, the timer is not cleared, and the check will start in the previous buffer after the delay runs out.

Whoops. I forgot that the timer is buffer-local.

> Which *might* be useful for your use case, as a quick way to "refresh" the checks in the buffer that are part of the same project.

I wouldn't want this, but I suppose some people might, indeed.

raxod502:
OK. I've finally got time to look at this again, and I fixed:

1. merge conflicts
2. the fact that the timer was buffer-local (it's global now, and independent of the idle change timer)
3. the dependency on `switch-buffer-functions` (we use `buffer-list-update-hook` instead now, but see also https://github.com/10sr/switch-buffer-functions-el/issues/1)

raxod502:
Let me know what to do about the tests. The manual says that new ERT tests are not to be added, and that existing tests should be moved to Buttercup. Unfortunately I may not have the expertise to port the existing tests relating to this functionality.

flycheck.el
```diff
@@ -2762,6 +2783,36 @@ because some users override that function, as described in URL
...
+      (setq flycheck-buffer-switch-timer
+            (run-at-time flycheck-buffer-switch-delay nil
+                         #'flycheck-handle-idle-buffer-switch
+                         (current-buffer))))))
+
+(defun flycheck-handle-idle-buffer-switch (buffer)
+  "Handle an expired idle timer in BUFFER since the last buffer switch."
+  (when (buffer-live-p buffer)
+    (with-current-buffer buffer
+      (flycheck-clear-idle-change-timer)
```
fmdkdd: Is this for preventing an automatic check from `idle-change` to be triggered?  It seems to me the interaction of these two timers warrants some discussion. 

What happens if after switching to a buffer, I edit some text, then wait?  With both `idle-change` and `buffer-switch`, surely I should get a check, preferably only one, on `idle-change`, after the `idle-change-delay` has expired.  This means that a change should also clear the `buffer-switch-timer`.

fmdkdd:
@raxod502 Thanks!  I've taken another look at it.  I'm afraid I have more questions than exact fix suggestions at the moment; this change is slightly trickier than expected.

fmdkdd:
>  Unfortunately I may not have the expertise to port the existing tests relating to this functionality.

We have tests in place for `idle-change` in `flycheck-test.el`, but they are for ERT.  I will take a look at how to do the same thing in buttercup.

raxod502:
Indeed, handling this multiple timer stuff is highly nontrivial. When I get some time, I'll write down a list of all the possible use cases, and what we would want to happen in each of them, and make sure the implementation gets all of them right.

raxod502:
I'm back with an updated pull request! (It took long enough, sorry...)

So this version, I believe, addresses all of our concerns.

* _If you switch between several buffers quickly, should syntax checks be run in all of them?_ Both ways are supported, according to the value of `flycheck-buffer-switch-check-intermediate-buffers`.
* _Logic for detecting buffer switches was wrong._ It's fixed now, and we still don't have a dependency on `switch-buffer-functions`.
* _What if you switch to a buffer and then edit some text?_ The check is delayed until `flycheck-idle-change-delay` after the edit.
* _What if you edit some text, switch to another buffer, then switch back?_ The check in the first buffer is delayed until `flycheck-idle-buffer-switch-delay` after the second buffer switch. The check in the second buffer may or may not happen depending on the value of `flycheck-buffer-switch-check-intermediate-buffers`.
    * _What if `flycheck-buffer-switch-check-intermediate-buffers` is nil, and the user switches to a third buffer before the check in the first buffer can be run?_ It's still run, although the timing is determined by `flycheck-idle-buffer-switch-delay` rather than `flycheck-idle-change-delay` as you might expect. However, if `idle-buffer-switch` is not contained in `flycheck-check-syntax-automatically`, then this doesn't apply: the timing is still determined by `flycheck-idle-change-delay` as before.

What do you think, @fmdkdd ?

raxod502:
OK, apparently `buffer-list-update-hook` doesn't always run when you switch windows. I changed it back to `post-command-hook`. Unfortunate, but I don't think there's any way around it.

raxod502:
Also, unfortunately the current/previous buffer bookkeeping doesn't work if you are switching between Flycheck and non-Flycheck buffers. I think the only way to handle this correctly is to put `flycheck-handle-buffer-switch` on `post-command-hook`, unfortunately.

This might be an argument for a dependency on `switch-buffer-functions`, actually, since then multiple packages can detect buffer switches without each placing a separate function on `post-command-hook`.

raxod502:
I'll leave it to you to decide what should be done about testing.

fmdkdd:
Thanks for getting back on this!  I will take a look ASAP.

raxod502:
Ping :)

flycheck.el
```diff
@@ -2792,6 +2837,62 @@ because some users override that function, as described in URL
...
+            (run-at-time flycheck-idle-buffer-switch-delay nil
+                         #'flycheck-handle-idle-buffer-switch
+                         (current-buffer))))))
+
+(defun flycheck-handle-idle-buffer-switch (buffer)
+  "Handle an expired idle timer in BUFFER since the last buffer switch."
+  (when (buffer-live-p buffer)
+    (with-current-buffer buffer
+      (flycheck-clear-idle-buffer-switch-timer)
+      (when (or flycheck-buffer-switch-check-intermediate-buffers
```
fmdkdd: I'm not sure I understand this condition.  If `-check-intermediate-buffers` is non-nil, then all idle-buffer-switch should trigger a check?

I don't get the `(equal buffer (current-buffer))` however.  Isn't that always the case since we are in `with-current-buffer buffer`?

flycheck.el
```diff
@@ -2792,6 +2837,62 @@ because some users override that function, as described in URL
...
+                flycheck--force-checking-intermediate-buffers
+                (equal buffer (current-buffer)))
+        ;; Setting this variable to non-nil guarantees exactly one
+        ;; syntax check.  After that, we can remove the effects.
+        (setq flycheck--force-checking-intermediate-buffers nil)
```
fmdkdd: Could you elaborate on the role of this variable?  I'm not sure to follow what it does.

flycheck.el
```diff
@@ -2792,6 +2837,62 @@ because some users override that function, as described in URL
   (flycheck-clear-idle-change-timer)
   (flycheck-buffer-automatically 'idle-change))
 
+(defvar-local flycheck-idle-buffer-switch-timer nil
+  "Timer to mark the idle time since switching to a new buffer.")
+
+(defvar-local flycheck--force-checking-intermediate-buffers nil
```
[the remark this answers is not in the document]
raxod502: I have done this, along with `flycheck-idle-change-timer` as well, in the new `flycheck-empty-variables` function which is called from `flycheck-teardown`. I also had `flycheck-teardown` call `flycheck-clear-idle-change-timer` and `flycheck-clear-idle-buffer-switch-timer`.

flycheck.el
```diff
@@ -2792,6 +2837,62 @@ because some users override that function, as described in URL
...
+Ignore `flycheck-buffer-switch-check-intermediate-buffers' in
+that case.")
+
+(defun flycheck-clear-idle-buffer-switch-timer ()
+  "Clear the idle buffer switch timer."
+  (when flycheck-idle-buffer-switch-timer
+    (cancel-timer flycheck-idle-buffer-switch-timer)
+    (setq flycheck-idle-buffer-switch-timer nil)))
+
+(defvar flycheck--last-buffer (current-buffer)
```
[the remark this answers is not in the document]
raxod502: It's a global variable. Do you mean setting it to nil? If so, I could put that in `flycheck-global-teardown`, where I also removed the function on `buffer-list-update-hook`.

flycheck.el
```diff
@@ -2400,7 +2438,9 @@ buffer manually.
                                                next-error-function
                                              :unset))
     (when flycheck-standard-error-navigation
-      (setq next-error-function #'flycheck-next-error-function)))
+      (setq next-error-function #'flycheck-next-error-function))
+
+    (add-hook 'post-command-hook #'flycheck-handle-buffer-switch))
```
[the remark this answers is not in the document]
raxod502: Done!

raxod502:
I think the behavior I observed of `buffer-list-update-hook` was due to only adding the function to the hook locally. If we add it globally, then all buffer switches are correctly detected as far as I can tell. This is explained in a comment where I add the hook.

flycheck.el
raxod502: Yes, if `flycheck-buffer-switch-check-intermediate-buffers` is non-nil, then `flycheck-idle-buffer-switch-delay` seconds after you switch to a buffer, Flycheck will run a syntax check, regardless of whether the user is still in that buffer.

The `(equal buffer (current-buffer))` was a mistake; I didn't realize that `(current-buffer)` was inside a `(with-current-buffer ...)`. That has now been fixed.

flycheck.el
raxod502: Basically, the point of that variable is to ensure that the interaction between the two timers never results in a situation where the user edits a buffer, but a syntax check does not result.

That could happen in the following situation:

1. I edit buffer A. That starts `flycheck-idle-change-timer` in buffer A.
2. I switch to buffer B. That starts `flycheck-idle-buffer-switch-timer` in buffer B.
3. I switch back to buffer A. That starts `flycheck-idle-buffer-switch-timer` also in buffer A. We only want to trigger one syntax check, so that also cancels `flycheck-idle-change-timer` in buffer A.
4. Now suppose I switch back to buffer B again. That resets `flycheck-idle-buffer-switch-timer` in buffer B.
5. When `flycheck-idle-buffer-switch-timer` expires in buffer B, a syntax check is performed.
6. When `flycheck-idle-buffer-switch-timer` expires in buffer A, however, a syntax check is only performed if `flycheck-buffer-switch-check-intermediate-buffers` is non-nil. Otherwise, nothing happens. Thus, we have forgotten about the fact that the user edited buffer A.

To avoid this, we set the variable `flycheck--force-checking-intermediate-buffers` after canceling `flycheck-idle-change-timer` in step 3. Then in step 6, the fact that the variable is non-nil forces a syntax check to be performed, regardless of the value of `flycheck-buffer-switch-check-intermediate-buffers`.

raxod502:
After making the fixes you suggested, the basic functionality still seems to work in my tests. Let me know if there's more I can do :)

fmdkdd:
@raxod502 Thanks!  I will come back to this soon enough.  We may have to wait for #1462 to fix our CI.

raxod502:
@fmdkdd Rebased to fix conflicts. Anything I can do to help this along?

fmdkdd:
> Anything I can do to help this along?

Pinging me is a good step ;)  I will do my best to review it this week.

flycheck.el
```diff
@@ -2787,6 +2842,11 @@ buffer."
     (when flycheck-mode
       ;; The buffer was changed, thus clear the idle timer
       (flycheck-clear-idle-change-timer)
+      ;; If we can guarantee that this timer is actually going to
+      ;; result in a syntax check, we can cancel any other
+      ;; pre-existing timers that would cause syntax checks.
+      (when (memq 'idle-change flycheck-check-syntax-automatically)
+        (flycheck-clear-idle-buffer-switch-timer))
```
fmdkdd: Instead of having two separate timers that cancel each other, can't we just keep one timer variable?  That way, we would just have to check whether the timer is set, cancel it if it is, and set its new value.  We wouldn't have to peek into `flycheck-check-syntax-automatically`.

Maybe name it `flycheck-idle-trigger-timer`.

raxod502: @fmdkdd Are you okay with the user not being able to customize the delay for checking after an edit separately from the delay for checking after a buffer switch?

fmdkdd: We keep the two delays, but we use only one `timer` variable for the results of `run-at-time`.  This is just a simpler way to ensure only one "delayed action" can exist at a time.  Do you see what I mean?

raxod502: I think that could work. I'll look into it.

raxod502:
Whoops, I did not know that Magit renamed the remote branch as well when you rename a local branch.

raxod502:
@fmdkdd The new version is **much** better. Not only is using a single timer simpler but the new implementation makes it trivial to add more idle trigger events.

flycheck.el
```diff
@@ -2785,28 +2870,47 @@ buffer."
...
+  ;; the temp buffer!), and that would trigger spurious syntax checks.
+  ;; It seems that reading (car (buffer-list)) gets us the correct
+  ;; current buffer in all important real-life situations (although it
+  ;; doesn't necessarily catch uses of `set-buffer').
+  (with-current-buffer (car (buffer-list))
+    (unless (equal flycheck--last-buffer (current-buffer))
+      (setq flycheck--last-buffer (current-buffer))
+      (when flycheck-mode
+        (flycheck--clear-idle-trigger-timer)
+        (cl-pushnew 'idle-buffer-switch flycheck--idle-trigger-conditions)
```
fmdkdd: Can you elaborate on why we accumulate the conditions here?  It seems this is necessary because we have only one function `flycheck--handle-idle-trigger` now.  Is that it?

raxod502:
I think that your simplified code misbehaves in the following case:

1. I have `flycheck-buffer-switch-check-intermediate-buffers` set to nil
2. I make a change in buffer A
3. I quickly switch to buffer B before the syntax check happens
4. The syntax check in buffer A is canceled, when it shouldn't be!

flycheck.el
raxod502: Basically, it is so that I can handle the case that I mention in my comment below—an idle change followed by a buffer switch, when `flycheck-buffer-switch-check-intermediate-buffers` is nil. We want to be able to say "cancel any syntax check that was due to a buffer switch, but don't cancel any syntax check that was due to an edit". If I use a buffer-local list of conditions, it is easy to handle that:

1. When there is an edit, add `idle-change` to the list
2. When there is a buffer switch, add `idle-buffer-switch` to the list
3. When it comes time to perform the syntax check, remove `idle-buffer-switch` from the list unless the user is in the same buffer as they were in when the syntax check was scheduled

I think that handling this correctly will always require the use of an additional variable. In my previous attempt, that variable was `flycheck--force-checking-intermediate-buffers`.

flycheck.el
```diff
@@ -2785,28 +2870,47 @@ buffer."
...
+  ;; `current-buffer' is bogus in one of those triggers (the one just
+  ;; after the temp buffer is killed).  If we rely on the bogus value,
+  ;; Flycheck will think that the user is switching back and forth
+  ;; between different buffers during the `with-temp-buffer' call
+  ;; (note: two different normal buffers, not the current buffer and
+  ;; the temp buffer!), and that would trigger spurious syntax checks.
+  ;; It seems that reading (car (buffer-list)) gets us the correct
+  ;; current buffer in all important real-life situations (although it
+  ;; doesn't necessarily catch uses of `set-buffer').
+  (with-current-buffer (car (buffer-list))
```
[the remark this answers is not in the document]
raxod502: Well, that's what I thought. Unfortunately, as I mentioned in the comment above, 

```elisp
;; Switching buffers here is weird, but unfortunately necessary.  It
;; turns out that `with-temp-buffer' triggers
;; `buffer-list-update-hook' twice, and the value of
;; `current-buffer' is bogus in one of those triggers (the one just
;; after the temp buffer is killed).  If we rely on the bogus value,
;; Flycheck will think that the user is switching back and forth
;; between different buffers during the `with-temp-buffer' call
;; (note: two different normal buffers, not the current buffer and
;; the temp buffer!), and that would trigger spurious syntax checks.
```

That is to say, the value of `(car (buffer-list))` is actually a different value than `(current-buffer)` in some situations, and `(car (buffer-list))` is actually the value we want to use if we don't want to trigger spurious syntax checks.

(In my case, I had an idle timer in my own Emacs configuration that updated my mode line with the current Git status. This used `with-temp-buffer` twice to collect the Git output, and without this `with-current-buffer`, Flycheck was getting four buffer switches between Flycheck-enabled buffers (n.b.: _not_ one Flycheck-enabled buffer and the temp buffer, as you might have expected—as I said, the value of `(current-buffer)` is totally bogus) every time that idle timer went off.)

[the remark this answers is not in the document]
fmdkdd: I see, it seems our issue is that we want to detect buffer switches made by the user interactively, and (mostly) ignore buffer switches made programmatically.

What if we checked if `flycheck-mode` is enabled for the current buffer first in our `buffer-list-update-hook` function?  `flycheck-mode` should only be enabled in interactive buffers, so we would care only about switchs to *other* `flycheck-mode` buffers, and disregard any temporary buffers.

That is:

```emacs-lisp
(defun flycheck-handle-buffer-switch ()
  (when flycheck-mode
    (unless (equal current flycheck--last-buffer)
     ...
```

Does it gets rid of the spurious checks in your setup?

flycheck.el
fmdkdd: > "cancel any syntax check that was due to a buffer switch, but don't cancel any syntax check that was due to an edit"

Ok, that's indeed a case I had not considered.  Thanks for pointing it out.

> I think that handling this correctly will always require the use of an additional variable.

I concur.  Although, for this particular case, is it necessary that we keep track of all the "causes" for triggers?  I.e., since there is only one active timer, just checking if the last timer was triggered by a buffer-switch is enough to be able to cancel it if `flycheck-buffer-swithc-check-intermediate-buffers` is nil.

> When it comes time to perform the syntax check, remove idle-buffer-switch from the list unless the user is in the same buffer as they were in when the syntax check was scheduled

Shouldn't we always clear the list of causes when we actually run the check instead?  And what purpose does this condition serve?  It doesn't seem necessary to handle the case spelt out above.

raxod502: >  just checking if the last timer was triggered by a buffer-switch is enough to be able to cancel it if `flycheck-buffer-switch-check-intermediate-buffers` is nil.

I don't think that's the case, and I have a more complex example to demonstrate why. Suppose the user makes an edit in the buffer. Then the last timer was triggered by an edit. But now suppose that, before the timer triggers, the user quickly switches from that buffer to a different one, then back. We have to reset the timer, and now the last timer was triggered by a buffer-switch. Finally, suppose that before *that* timer expires, the user again switches to another buffer. We notice that the last timer was triggered by a buffer-switch, and cancel it. But now no timer will ever trigger in that buffer: we have forgotten about the edit.

I think you *have* to keep track of each condition separately. A toggle, or a single condition name, is simply not enough.

> Shouldn't we always clear the list of causes when we actually run the check instead? 

Indeed, we actually do that:

<https://github.com/flycheck/flycheck/pull/1308/files#diff-374cdd9563b826c712b1dc6a373a0bc1R2825>

You're probably wondering why we *also* do it in `flycheck--handle-idle-trigger`. That's because we want to clear out the list of conditions even if `flycheck-check-syntax-automatically` decides not to perform a syntax check. Otherwise the list will just accumulate over time.

> It doesn't seem necessary to handle the case spelt out above.

Does my more complex example in the first paragraph answer this question?

fmdkdd: > But now no timer will ever trigger in that buffer: we have forgotten about the edit.

Right!  And if we keep track of both causes, we would not cancel the last timer, which would eventually run a check for `idle-change`.  That justifies the variable.

Although it's arguable whether the added complexity is worth it.  If your check is not triggered after this frantic buffer switching, you would just switch back to the original buffer to trigger an `idle-buffer-switch` check.

I'll take another look at the PR with these clarifications in mind.  Thank you again for spelling things out.

However, now I'm wondering what happens if you do *not* have `idle-buffer-switch` in `flycheck-check-syntax-automatically`.  We do not check for the value of that variable, so the exact same events will unfold, right?  We will reset the timer and delay the `idle-change` by `flycheck-idle-buffer-switch-delay`.  Although we will eventually run the check, it will have been delayed unexpectedly.  This seems to contradict the docstring:
```
This variable has no effect if `idle-buffer-switch' is not
contained in `flycheck-check-syntax-automatically'
```
We can probably add a guard that looks for `idle-buffer-switch` to behave as that docstring states.  I'd rather not go back to the two timers implementation.

raxod502: > We can probably add a guard

Agreed; added.

flycheck.el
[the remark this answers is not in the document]
fmdkdd: Darn, I had a comment here I forgot to send.  Here it is.

Thank you for providing an example.

Indeed there is something counter-intuitive going on with `current-buffer`.  I wonder if this comment in `with-temp-buffer` is related:
> ;; FIXME: kill-buffer can change current-buffer in some odd cases.

In any case, even if it's an Emacs bug, we cannot wait for a fix to land.  So, as a workaround, `(car (buffer-list))` seems okay (i.e., it should always return what we want).  However, `(window-buffer)` might be clearer?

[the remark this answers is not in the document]
raxod502: In my testing, `(window-buffer)` seems fine. I changed it.

fmdkdd:
Since we have been iterating a lot on the functionality, I'm gathering all the information in one post about what the PR should do:

1. Whenever the user switches to a flycheck-mode buffer, that buffer should run a check after a customizable delay.
2. That behavior only happens when "buffer-switch" is in `flycheck-check-syntax-automatically`.
3. If the user starts typing in the buffer right after switching to a buffer, the check should be rescheduled (since the check results would be obsolete).
4. If the user was typing (hence scheduling a check) in a buffer, and the user switches to another buffer and back, the check should *not* be rescheduled.

   (In practice, it may be simpler to always reschedule checks; it is correct but will delay the check unnecessarily).

5. If the user rapidly switches to multiple flycheck-mode buffers, checks should be scheduled for all buffers.  This behavior can be toggled off, in which case only the final flycheck-mode buffer will run a check.

The following table describes the interactions between idle-change and idle-buffer-switch.  In all cases, the point is that **only one check** should run.

|                                   | idle-change    | idle-buffer-switch | both             |
|-----------------------------------|----------------|--------------------|------------------|
| Switch to flycheck buffer         | -              | schedule check     | schedule check   |
| Switch, then type                 | schedule check | schedule check     | reschedule check |
| Type, then switch away            | schedule check | -                  | schedule check   |
| Type, switch away and back        | schedule check | schedule check     | reschedule check |
| Type, switch away, back, and away | schedule check | schedule check*    | reschedule check |
| Rapidly switch multiple times     | -              | schedule checks*   | schedule checks* |

(*) Only if flycheck-check-intermediate-buffers is non-nil.

- **Schedule check**: create a timer that will call flycheck-buffer-automatically
     after a customizable delay.
- **Reschedule check**: cancel any pending check timer and create another one
     using the customizable delay for the latest trigger.

Another table for the interactions between mode-enabled and idle-buffer-switch. Again, only one check should run:

|                          | mode-enabled | idle-buffer-switch | both        |
|--------------------------|--------------|--------------------|-------------|
| Open new flycheck buffer | defer check  | schedule check     | defer check |
| Enable flycheck-mode     | defer check  | -                  | defer check |

- **Defer check**: force a deferred check (which will be called on post-command-hook).

So basically, idle-buffer-switch should have no effect on toggling the minor mode (since whether a check happens at that time is already governed by mode-enabled).

fmdkdd:
@raxod502 The implementation currently ticks all the marks of my previous post, except in two places.  

The first is point 4 above, where do the practical thing and reschedule the check.  It's not optimal, but it's not worth obsessing over (the delays are small, and can be customized anyway).

The second is a situation we haven't considered yet, and that is the interaction between `mode-enabled` and `idle-buffer-switch`.

When I have these two conditions enabled, and I have `global-flycheck-mode` turned on, if I visit a file for which I had no previous buffer, and this file turns on flycheck-mode automatically, then I will get two syntax checks: one for `mode-enabled`, and one for `idle-buffer-switch`, after the `idle-buffer-switch-delay`.

I would be happier if we could avoid that second redundant check, but on the other hand I'm not sure how to fix it reliably.  So I'm fine leaving it as a FIXME, unless you have another idea.

I've added tests for the intermediate buffer checking.  Could you try to pass the CI before I do a final proofread?

raxod502:
> I would be happier if we could avoid that second redundant check

Can't we just refrain from triggering a syntax check in `flycheck--handle-idle-trigger` if `flycheck-deferred-check-p` returns non-nil?

> Could you try to pass the CI before I do a final proofread?

Yes, I will work on that now.

raxod502:
Oh, never mind—I see that what I suggested won't work. I'll think about it more.

raxod502:
I have a few ideas for further minor improvements, so give me a bit before doing your final proofread.

fmdkdd:
> I have a few ideas for further minor improvements, so give me a bit before doing your final proofread.

Sure, let me know when you think you are done :)

raxod502:
So there was the minibuffer thing, and I also looked at the double check caused by `mode-enabled`. I still don't understand 100% what's going on there. It seems like the trouble is that the deferred syntax check triggered by `mode-enabled` is somehow run before `buffer-switch-update-hook` even gets called.

But the symptom is not all that bad and you could always remove `mode-enabled` from the list if you were really concerned about the performance impact, so I agree with you that it can be fixed later.

With that, I think I'm done for now.

raxod502:
Ping! :)

flycheck.el
```diff
@@ -2913,7 +3024,8 @@ variables of Flycheck."
   (dolist (buffer (buffer-list))
     (with-current-buffer buffer
       (when flycheck-mode
-        (flycheck-teardown)))))
+        (flycheck-teardown))))
+  (remove-hook 'buffer-list-update-hook #'flycheck-handle-buffer-switch))
```
fmdkdd: So we add the hook anytime `flycheck-mode` is activated, but the only time we remove it is when Emacs exits?

Obviously, we should remove the hook when we kill the last flycheck buffer, right?  @cpitclaudel is there an idiomatic way of cleaning up in that situation, or do we have to iterate through the buffer-list or keep track of the number of flycheck-mode buffers?

raxod502:
@fmdkdd Fixed your comments, thanks (except for the question of how to remove the `buffer-list-update-hook`—I'm not really sure what the best approach is). And sorry again for the delay.

fmdkdd:
@raxod502 Thanks!  LGTM.

I'd like @cpitclaudel to take a look as well, especially about removing that hook.

cpitclaudel:
Thanks all for working on this. I'll take a quick look this week-end

flycheck.el
```diff
@@ -2414,7 +2452,14 @@ buffer manually.
...
     (when flycheck-standard-error-navigation
-      (setq next-error-function #'flycheck-next-error-function)))
+      (setq next-error-function #'flycheck-next-error-function))
+
+    ;; This hook must be added globally since otherwise we cannot
+    ;; detect a change from a buffer where Flycheck is enabled to a
+    ;; buffer where Flycheck is not enabled, and therefore cannot
+    ;; notice that there has been any change when the user switches
+    ;; back to the buffer where Flycheck is enabled.
+    (add-hook 'buffer-list-update-hook #'flycheck-handle-buffer-switch))
```
cpitclaudel: The hook is never removed, right? If not, we should make sure that flycheck is enabled as the first thing that flycheck-handle-buffer-switch does

flycheck.el
cpitclaudel: We have to iterate, AFAIK.  Just keeping a count would probably be OK, but it's more brittle than needed; just going over the list of buffers should be good.
There's a cheap function `buffer-local-value`, that we can use to check many buffers quickly.

flycheck.el
```diff
@@ -2725,54 +2770,95 @@ current syntax check."
...
+  (when flycheck--idle-trigger-timer
+    (cancel-timer flycheck--idle-trigger-timer)
+    (setq flycheck--idle-trigger-timer nil)))
 
-(defun flycheck-clear-idle-change-timer ()
-  "Clear the idle change timer."
-  (when flycheck-idle-change-timer
-    (cancel-timer flycheck-idle-change-timer)
-    (setq flycheck-idle-change-timer nil)))
+(defun flycheck--handle-idle-trigger (buffer)
```
cpitclaudel: I'm not sure whether we can do this: many people seem to overwrite this function in their config. https://github.com/search?q=flycheck-clear-idle-change-timer&type=Code

We should try to understand why they do that, and make sure we're not breaking their config.

flycheck.el
```diff
@@ -2725,54 +2770,95 @@ current syntax check."
...
+`idle-buffer-switch' in it, at least until the idle timer
+expires.")
 
 (defun flycheck-buffer-automatically (&optional condition force-deferred)
   "Automatically check syntax at CONDITION.
 
 Syntax is not checked if `flycheck-may-check-automatically'
-returns nil for CONDITION.
+returns nil for CONDITION.  (CONDITION may be a single condition
+or a list of them.)
```
cpitclaudel: Wouldn't it be simpler to force `condition` to always be a list? Does any external code depend on accepting non-lists?

flycheck.el
```diff
@@ -558,13 +562,14 @@ If nil, never check syntax automatically.  In this case, use
...
               (const :tag "After the buffer was changed and idle" idle-change)
+              (const :tag "After switching to buffer" idle-buffer-switch)
               (const :tag "After a new line was inserted" new-line)
               (const :tag "After `flycheck-mode' was enabled" mode-enabled))
   :package-version '(flycheck . "0.12")
   :safe #'flycheck-symbol-list-p)
 
 (defcustom flycheck-idle-change-delay 0.5
-  "How many seconds to wait before checking syntax automatically.
+  "How long to wait after a change before checking syntax automatically.
```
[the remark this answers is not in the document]
raxod502: It does, yes, but I'll reword it to include the "number of seconds" part without being too long.

flycheck.el
```diff
@@ -558,13 +562,14 @@ If nil, never check syntax automatically.  In this case, use
   :group 'flycheck
   :type '(set (const :tag "After the buffer was saved" save)
               (const :tag "After the buffer was changed and idle" idle-change)
+              (const :tag "After switching to buffer" idle-buffer-switch)
```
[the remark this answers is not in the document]
raxod502: "After switching the current buffer"?

flycheck.el
raxod502: We can't do that. At the very least, we have to record that the user switched buffers; otherwise, it's not detected when the user switches from a Flycheck buffer to a non-Flycheck buffer and then back.

It would be better to make sure the hook is removed when it is no longer necessary, I think.

flycheck.el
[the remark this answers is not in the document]
cpitclaudel: Thanks ^^

flycheck.el
cpitclaudel: OK, I'm onboard with that :)

flycheck.el
cpitclaudel: Ah, thanks, I remember.  There were two issues here, I think: @magnars' config, and the advice added by auto-complete.   Could we define an alias so that the advice still works?

flycheck.el
cpitclaudel: `-` vs `--`: mostly historical reasons, I think; this is a pretty recent convention AFAIK.
Most of the uses I could find are from the same snippet that we discussed above.  @fmdkdd , WDYT?

cpitclaudel:
> I just started the fall semester at college

Good luck :)

flycheck.el
fmdkdd: +1 for the `defalias`.  Seems a good way to maintain backward compatibility while keeping the internal code clean.

flycheck.el
fmdkdd: > - vs --: mostly historical reasons, I think; this is a pretty recent convention AFAIK.

I'm not against starting using it for new code (as in this PR).  My main contention is that it's just that: a convention.  We have no way to refrain people from monkey-patching our "private API" functions.  We can just signal that "you are on your own if we remove it in the future".

Regarding the `condition` parameter: accepting single conditions and lists of condition is fine with me.  The code is not that complex.

flycheck.el
raxod502: Sure, I'll add an alias.

flycheck.el
raxod502: Single conditions and lists of conditions are fine from a maintainability perspective, I think. However, what _is_ a bit confusing is that the `condition` parameter is optional, so passing an "empty list" is the same as omitting the argument (which has a different effect than passing an empty list would be expected to have). How about making the argument non-optional, and adding a special value (some symbol) that can be passed to recover the current behavior for the optional argument being omitted? (This would allow me to remove at least one other place in the code where I intentionally refrain from calling `flycheck-buffer-automatically` when the condition list is empty.)

raxod502:
Well, I was not expecting for this to sit for so long. I am sorry.

However, I've now addressed all the comments above, except for the request to add a backwards compatibility alias for `flycheck-handle-idle-change`. I don't think this is necessary. This is because in fact you can define advice just fine for functions that don't exist (this of course has no effect), so an alias shouldn't be needed to make sure the `auto-complete` advice continues to work.

The code for the alias would look like this:

```elisp
(defalias 'flycheck-handle-idle-change #'flycheck--handle-idle-trigger
  "Backwards-compatibility alias for `flycheck--handle-idle-trigger'.")
(make-obsolete
 'flycheck-handle-idle-change
 "it is an internal function and should not be used in your code.")
```

fmdkdd:
@raxod502 Thanks!  We'll be delighted to finally merge this PR.  You'll need to rebase on master for the CI to pass I think (ess is now incompatible with 24.5, so we have a special Cask.24 file that disables it).

@cpitclaudel Do you have time to check the latest version addresses your comments?

flycheck.el
```diff
@@ -2742,54 +2788,105 @@ current syntax check."
...
   (flycheck-clear)
-  (flycheck-cancel-error-display-error-at-point-timer))
+  (flycheck-cancel-error-display-error-at-point-timer)
+  (flycheck--clear-idle-trigger-timer)
+  (flycheck--empty-variables)
+  (unless (or ignore-global
+              (->> (buffer-list)
+                   (seq-remove (apply-partially #'equal (current-buffer)))
+                   (seq-some (apply-partially #'buffer-local-value
+                                              'flycheck-mode))))
```
cpitclaudel: This looks a bit convoluted, and it traverses the buffer list twice. Would this do the same?

```
(seq-some (lambda (buf) (and (not (equal buf (current-buffer)))
                        (buffer-local-value 'flycheck-mode buf)))
          (buffer-list))
```

raxod502:
@cpitclaudel Changed (I used to write Clojure, where the multi-step approach would be more idiomatic).

@fmdkdd Rebase already done!

fmdkdd:
@raxod502 Ah sorry, but our `all-tools` image we use for running the integration test was rebuilt yesterday, so we have one more failing test that I just fixed on master.  You'll need one last rebase to grab that one.

While you are at it, please squash all your commits.  Then we can finally land this!

raxod502:
Here you go. I can squash the two remaining commits together if you want, but this way you get credit for writing the tests.

fmdkdd:
@raxod502 I didn't mind, but thanks!

Now we can finally merge this! :tada:  :balloon: 

Thank you for coming back to this PR and seeing it through the end. :clap:

PR #1314, opened 2017-08-30
Remove google analytics

fmdkdd:
We don't even have access to it, and there's no point in tracking readers of the manual.

Now actually, it seems this is not enough to remove Analytics from the hosted version at readthedocs.  I've built this branch here: http://www.flycheck.org/en/remove-google-analytics/

Analytics are [still included](http://www.flycheck.org/en/remove-google-analytics/_static/readthedocs-dynamic-include.js), only with a *different* analytics code (maybe one for readthedocs directly?)

It seems this is on purpose: rtfd/readthedocs.org/issues/1221.

I've added a second commit that prevents the injected script from loading, so we never hit Google's servers.  I couldn't find any terms of service for readthedocs, so I guess we are not violating any!

### DIFF
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -1,0 +1,8 @@
+{% extends "!layout.html" %}
+
+{% block extrahead %}
+<script type="text/javascript">
+  /* No-op to prevent injected analytics from running */
+  document.createElement = function() {}
+</script>
+{% endblock %}
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,3 +1,4 @@
+# Copyright (C) 2017 Flycheck contributors
 # Copyright (C) 2016 Sebastian Wiesner and Flycheck contributors
 
 # This file is not part of GNU Emacs.
@@ -120,9 +121,6 @@ html_theme_options = {
     'github_type': 'star',
     'github_banner': True,
     'travis_button': False,
-    # Google Analytics ID for our documentation.  On ReadTheDocs it's set via
-    # the Admin interface so we'll skip it here.
-    'analytics_id': 'UA-71100672-2' if not ON_RTD else None,
 }
 html_sidebars = {
     '**': [
@@ -274,7 +272,6 @@ def build_offline_html(app):
         app.info('Building offline documentation without external resources!')
         app.builder.theme_options['github_banner'] = 'false'
         app.builder.theme_options['github_button'] = 'false'
-        app.builder.theme_options['analytics_id'] = None
 
 
 def add_offline_to_context(app, _pagename, _templatename, context, _doctree):


fmdkdd:
Hmm, the build is failing probably because of [this change] in buttercup :(

I'll try to pin the buttercup version we pull first, because it's the right thing to do anyway.

fmdkdd:
Update: looks like one does not simply pin version using Cask.  The [documentation](http://cask.readthedocs.io/en/latest/guide/dsl.html#dependencies) shows you can fetch a specific commit from git, but doing the following:
```emacs-lisp
(depends-on "buttercup"                ; BDD test framework for Emacs
             :git "http://github.com/jorgenschaefer/emacs-buttercup.git"
             :ref "e6f97d0ee4da1d061c2ff5f3e3984a30956a9fbc") ; 1.8 release
```
and then trying `make specs` results in:
```
cask exec buttercup -L . --pattern '.*' test/specs
cask exec: error: Failed to execute buttercup -L . --pattern .* test/specs: [Errno 2] No such file or directory
Did you run cask install?
Makefile:123: recipe for target 'specs' failed
make: *** [specs] Error 1
```
Might be a bug in `Cask` or `package-build`.

The other option is of course to update our custom buttercup matchers to deal with the closures introduced in the latest master.   But I don't like that our CI can fail for changes in dependencies; especially for an update in our test runnner.  I'd rather have pinned versions for reproducibility.

WDYT @cpitclaudel?

fmdkdd:
Filled cask/cask/issues/395 in the meantime.

cpitclaudel:
> Now actually, it seems this is not enough to remove Analytics from the hosted version at readthedocs. 

I find this a bit surprising¸ given that there's an advanced settings page that lets us specify an ID, and our is blank.

fmdkdd:
@cpitclaudel Yeah, I also blanked that one; it was filled with same analytics ID we had in the `conf.py` previously.  What it did was to add *another* analytics ID to the `_gaq` array in the JS file included by readthedocs (the "User code"):
```
var _gaq = _gaq || [];
_gaq.push(['_setAccount', 'UA-17997319-1']);
_gaq.push(['_trackPageview']);


// User Analytics Code
_gaq.push(['user._setAccount', 'UA-71100672-2']);
_gaq.push(['user._trackPageview']);
// End User Analytics Code
```
So blanking the field only removes the second code, but not the first one used by readthedocs on all projects.

lunaryorn:
@fmdkdd I'm sorry, seems I forgot to remove the tracking code.  It's been invalid since a long time anywayy, I've long stopped tracking access to the manual.  I don't even have an Analytics account anymore.

However, readthedocs tracks all access to documentation hosted by the platform to its own GA property regardless of GA settings in their panel.  Can't find the reference in their docs in a cursory glance, but I'm sure I read that somewhere.  I don't think they let you opt out, tho.

fmdkdd:
>  I'm sorry, seems I forgot to remove the tracking code.

No worries :)  It's a good thing you pinged us about readthedocs; it's what got me looking around in the first place.

fmdkdd:
@cpitclaudel The CI has been fixed now.  I think we can go ahead on that one, unless you have objections (are you still on mobile?).

cpitclaudel:
Thanks; sorry for the delay!

Issue #1175, opened 2016-12-07
eslint: config errorneous instead of missing 


jethrokuan:
I was facing some troubles enabling `eslint`, and `flycheck-verify-setup` told me my config file was "missing".

After some searching, I found that the config file was found with something along the lines of `eslint --print-config .`. This led me to realize that I had yet to install a npm package `eslint-plugin-import` which it depended on. In truth, the config file wasn't missing but was errorneous, and the "missing" message was somewhat misleading.

Would be good if the error thrown by `eslint` could be parsed and figured out.

cpitclaudel:
I think parsing the error message is beyond our scope.  However, it's true that it would be nice if you got a different message in that case.

One way to solve this would be to ask eslint to implement a new flag, --locate-config.

jethrokuan:
Sounds perfect, I'll create an issue over on eslint.

Simplify:
Hi, sorry about late reaction, too much work and too many obligations this week.

Yes, that message is misleading, should be changed to "Config is missing or incorrect" or something like that. When trying to enable ESlint for specific file/buffer we run `eslint --print-config .` in directory where current js file is located. If there is problem with configuration, ESlint will print message with explanation what is wrong with config file and return non-zero exit code. If exit code is not 0, we disable ESlint checker for that file/buffer. At this time [we do nothing](https://github.com/flycheck/flycheck/blob/master/flycheck.el#L7628) with actual output (we are not using output in `call-process`, one of the `nil`'s) of `--print-config .` if there is a problem with ESlint configuration, that message is several lines of text. We can use it to create Flycheck error/warning message or print that text in standard Emacs `Messages` buffer. First option to get error/warning popup may be annoying for some users, second option is almost invisible to users. Maybe combining second option with message in mini-buffer/echo-area "ESlint disabled, please check Message buffer" is better option.

Request for `--locate-config` option in ESlint is not very helpful, it does not provide anything of value that we don't have right now. I think you can close eslint/eslint#7719

Example of `eslint --print-config .` output when something goes wrong:
```
$ eslint --print-config .

Oops! Something went wrong! :(

ESLint couldn't find the plugin "eslint-plugin-react". This can happen for a couple different reasons:

1. If ESLint is installed globally, then make sure eslint-plugin-react is also installed globally. A globally-installed ESLint cannot find a locally-installed plugin.

2. If ESLint is installed locally, then it's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

[...]

PR #1317, opened 2017-09-02
Improve eslint config verification message

juergenhoetzel:
I had a hard time to find to root cause (syntax error in `.eslintrc` file) of the disabled `flycheck-eslint`, because the flycheck pointed out a `missing` config file.

Current behavior when config is valid:

```
javascript-eslint
-   may enable: yes
-   executable: Found at /usr/bin/eslint
-   config file: found
```
(the config is not only found but also valid!)

Current behavior when config is invalid (missing might be misleading)
```
javascript-eslint
-   may enable: Automatically disabled!
-   executable: Found at /usr/bin/eslint
-   config file: missing
```

Because `eslint` only writes to `stderr` if parsing of the config goes wrong it would be nice to use this information in the verification message:

New behavior when config is invalid

```
javascript-eslint (disabled)
-   may enable: Automatically disabled!
-   executable: Found at /usr/bin/eslint
-   config file: Cannot read config file: /home/juergen/shared/js/bot1/.eslintrc.js
```

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8062,7 +8062,7 @@ for more information about the custom directories."
   :package-version '(flycheck . "29"))
 
 (defun flycheck-eslint-config-exists-p ()
-  "Whether there is an eslint config for the current buffer."
+  "Whether there is a valid eslint config for the current buffer."
   (let* ((executable (flycheck-find-checker-executable 'javascript-eslint))
          (exitcode (and executable (call-process executable nil nil nil
                                                  "--print-config" "."))))
@@ -8104,7 +8104,7 @@ See URL `http://eslint.org/'."
       (list
        (flycheck-verification-result-new
         :label "config file"
-        :message (if have-config "found" "missing")
+        :message (if have-config "found" "missing or incorrect")
         :face (if have-config 'success '(bold error)))))))
 
 (defun flycheck-parse-jscs (output checker buffer)


flycheck.el
```diff
@@ -8060,12 +8060,14 @@ for more information about the custom directories."
...
-  "Whether there is an eslint config for the current buffer."
-  (let* ((executable (flycheck-find-checker-executable 'javascript-eslint))
-         (exitcode (and executable (call-process executable nil nil nil
-                                                 "--print-config" "."))))
-    (eq exitcode 0)))
+(defun flycheck-eslint-config-error ()
+  "String of eslint config error for the current buffer."
+  (let ((executable (flycheck-find-checker-executable 'javascript-eslint))
+        (stderr-file (flycheck-temp-file-system "eslint-config-errors")))
+    (call-process executable nil `(nil ,stderr-file) nil "--print-config" ".")
```
cpitclaudel: Don't you still need to check for the executable? Won't this raise an error if it can't be found?

flycheck.el
```diff
@@ -8060,12 +8060,14 @@ for more information about the custom directories."
...
+    (with-temp-buffer
+      (insert-file-contents stderr-file)
+      (buffer-string))))
```
cpitclaudel: I'm not a big fan of returning an empty string to indicate success, and I'd rather not depend too strongly on eslint writing to stderr. Could we keep the return code check, and return a cons of `status . error-message`?

flycheck.el
juergenhoetzel: Good point! i broke that!

flycheck.el
juergenhoetzel: Yes this is indeed suboptimal. I fill amend this PR.

Simplify: Few possible pointers:
- If config is missing `eslint --print-config .` will not print in `stderr` but `stdout` and exit with exit code 1.
- I'm not sure how long we can relay on this behavior from `eslint`, it looks that every mayor version break everything. Maybe we just need to check if output from command is JSON and use better but more generic message "Config file: missing or incorrect".
- This implemented right will close GH-1175
- This may slow even more opening of any JS file, so we may add to the documentation how to disable this check for users who don't want it, as described in GH-1129.

juergenhoetzel:
> 
> If config is missing eslint --print-config . will not print in stderr but stdout and exit with exit code 1.

I can't confirm this. I get:
```
juergen@samson:~/testjs → eslint --print-config . >/dev/null 

...
```

> I'm not sure how long we can relay on this behavior from eslint, it looks that every mayor version break everything. Maybe we just need to check if output from command is JSON and use better but more generic message "Config file: missing or incorrect".
 
:+1: After consideration of all possible issues of the different-versions/compatibility issues I also prefer to use just the generic message "Config file: missing or incorrect".

juergenhoetzel:
Aa @Simplify  already pointed out: It is not feasible to parse/interpret all possible external checkers/versions error output. I  amended  01d4214c51e73f0f393ed29cf3112e14dfd70593 to just generalize the verification error message.

flycheck.el
```diff
@@ -8104,7 +8104,7 @@ See URL `http://eslint.org/'."
       (list
        (flycheck-verification-result-new
         :label "config file"
-        :message (if have-config "found" "missing")
+        :message (if have-config "Found" "Missing or incorrect")
```
cpitclaudel: I think we usually use lowercase. Can we do that here too? And maybe we can discuss changing the case as a separate change? (I see that we have a few other inconsistencies, but I'd rather not add more.)

juergenhoetzel: I see. Amended e8f99fd

cpitclaudel:
Super, thanks a lot!

PR #1320, opened 2017-09-04
Update flycheck-buttercup matchers to latest buttercup API

fmdkdd:
Buttercup has updated the `expect` macro to wrap its arguments in closures [1],
so we now have to `funcall` the arguments passed to `buttercup-define-matcher`
in order to obtain their actual value.

[1]: https://github.com/jorgenschaefer/emacs-buttercup/commit/929a904debd4cf5846bbcb7e3c9ee21f7636c5bb

### DIFF
--- a/flycheck-buttercup.el
+++ b/flycheck-buttercup.el
@@ -55,12 +55,17 @@
 ;;; Data matchers
 
 (buttercup-define-matcher :to-be-empty-string (s)
-  (if (equal s "")
-      (cons t (format "Expected %S not be an empty string" s))
-    (cons nil (format "Expected %S to be an empty string" s))))
+  (let ((s (funcall s)))
+    (if (equal s "")
+        (cons t (format "Expected %S not be an empty string" s))
+      (cons nil (format "Expected %S to be an empty string" s)))))
 
 (buttercup-define-matcher :to-match-with-group (re s index match)
-  (let* ((matches? (string-match re s))
+  (let* ((re (funcall re))
+         (s (funcall s))
+         (index (funcall index))
+         (match (funcall match))
+         (matches? (string-match re s))
          (result (and matches? (match-string index s))))
     (if (and matches? (equal result match))
         (cons t (format "Expected %S not to match %S with %S in group %s"
@@ -76,7 +81,7 @@
 ;;; Emacs feature matchers
 
 (buttercup-define-matcher :to-be-live (buffer)
-  (let ((buffer (get-buffer buffer)))
+  (let ((buffer (get-buffer (funcall buffer))))
     (if (buffer-live-p buffer)
         (cons t (format "Expected %S not to be a live buffer, but it is"
                         buffer))
@@ -84,7 +89,7 @@
                         buffer)))))
 
 (buttercup-define-matcher :to-be-visible (buffer)
-  (let ((buffer (get-buffer buffer)))
+  (let ((buffer (get-buffer (funcall buffer))))
     (cond
      ((and buffer (get-buffer-window buffer))
       (cons t (format "Expected %S not to be a visible buffer, but it is"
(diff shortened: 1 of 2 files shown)


fmdkdd:
Since I can't manage to pin the buttercup version, the remaining option is to update our matchers.

There is one test that still fails on my machine:
```
Encrypted buffers flycheck-encrypted-buffer-p recognizes an encrypted buffer

  (unwind-protect (let ((temp-buffer (generate-new-buffer " *temp*"))) (save...
  (let ((temp-buffer (generate-new-buffer " *temp*"))) (save-current-buffer ...
  (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn (flyc...
  (unwind-protect (progn (flycheck/encrypt-string-to-file "Hello world" pass...
  (progn (flycheck/encrypt-string-to-file "Hello world" passphrase file-name...
  (let ((inhibit-message t)) (insert-file-contents file-name (quote visit)))
  (insert-file-contents "/tmp/flycheck-encrypted-file8630TNK.txt.gpg" visit)
  (epa-file-handler insert-file-contents "/tmp/flycheck-encrypted-file8630TN...
  (apply epa-file-insert-file-contents ("/tmp/flycheck-encrypted-file8630TNK...
  (epa-file-insert-file-contents "/tmp/flycheck-encrypted-file8630TNK.txt.gp...
  (signal file-error ("Opening input file" "Decryption failed" ""))
error: (file-error "Opening input file" "Decryption failed" "")
```
 But it is skipped on the CI, so it might have been broken for a while.

@cpitclaudel Can you double check that?  Adding `funcall` to our custom matcher is straightforward, but now that `expect` suspends its arguments, we have to dance around in `test-help` when passing `(match-string 1)` for instance.  I'm surprised it's the only place where such problems arise.

cpitclaudel:
Thanks for working on this. I don't understand the buttercup change; in particular, I don't understand what this bit of code that quotes the args does:

```
(mapcar (lambda (expr) `(lambda () (quote ,expr) ,expr)) args)
```

...?!

What's the quoted expression for?

More on topic: I'm not sure the quoting of `match-string 1` works.  Doesn't it just never evaluate the expect form?

I'm not sure how we're supposed to do this, except maybe to inline everything *into* the expect.

fmdkdd:
> What's the quoted expression for?

Looking at the commit, the idea is to save the expression passed to the matcher in order to be able to give more informative errors.  The quoted expression is useful when `expr` contains a macro, to preserve the expression prior to macro-expansion.  See the docstring to `buttercup--enclosed-expr`:
```emacs-lisp
(defun buttercup--enclosed-expr (x)
  "Given a zero-arg closure, return its unevaluated expression.

The closure MUST have one of the following forms:

\(closure (ENVLIST) () EXPR)
\(closure (ENVLIST) () (quote EXPR) EXPR)

and the return value will be EXPR, unevaluated. The latter form
is useful if EXPR is a macro call, in which case the `quote'
ensures access to the un-expanded form."
```

>  I'm not sure the quoting of match-string 1 works. Doesn't it just never evaluate the expect form?

Indeed!  Thanks for catching that.  I should have made sure the test could fail :(

The issue is that most matchers now modify `match-data` (see ` buttercup-define-matcher-for-unary-function`), so we have to save that ourselves.  I've updated the PR.  Though, I wonder if buttercup should use `save-match-data`.

I still don't know why the GPG test fails though.

cpitclaudel:
Very nice, thanks a lot for the explanation.

The fix looks good, though maybe you should test whether the match succeeded before calling `match-string`?

The GPG test can fail if you don't have a properly set up agent to prompt you for the passphrase. Could that be it?

fmdkdd:
> The GPG test can fail if you don't have a properly set up agent to prompt you for the passphrase. Could that be it?

That's probably it, as I don't get the prompt for a passphrase :)

> The fix looks good, though maybe you should test whether the match succeeded before calling match-string?

Hmm, `match-string` cannot fail right?  If we get garbage if the `looking-at` failed, then since we `expect` the result of `looking-at` before using the result of `match-string`, this should be fine.

fmdkdd:
@cpitclaudel If it's alright with you, can we merge this soon?  It's preventing us to review other PRs.

I didn't change `match-string` due to the reasoning in my previous comment, and also because I didn't want to make the code more convoluted that it already is.  If you have a nice suggestion, I'm all ears.

cpitclaudel:
Sorry, missed your previous reply. OK to merge this :)

Re match-string, I think it can fail if it points to out-of-range positions. Not entirely sure though, and for tests it doesn't matter much.  A simple workaround might be to move the match string part of the let to be after the expect. Would that work?

cpitclaudel:
Can you do the actual merge ? I'm on mobile ATM and I don't trust my GitHub app too much.

fmdkdd:
> A simple workaround might be to move the match string part of the let to be after the expect. Would that work?

No, that's the rub: `expect` modifies the match data, so we have to save it first.  I've guarded the `match-string` call with `and` on the match result, so that should be good.

cpitclaudel:
> No, that's the rub: expect modifies the match data, so we have to save it first. I've guarded the match-string call with and on the match result, so that should be good.

Duh. Sorry :/

fmdkdd:
Dont' worry; that's why four eyes are better than two ;)

PR #1321, opened 2017-09-04
Work around problems introduced by GH-1305

cpitclaudel:
Haven't had much time to test this out yet.
Let's look at it after we fix the test suite (thanks for the hard work @fmdkdd!)

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2745,14 +2745,22 @@ buffer."
           (flycheck-buffer-automatically 'new-line 'force-deferred)
         (setq flycheck-idle-change-timer
               (run-at-time flycheck-idle-change-delay nil
-                           #'flycheck-handle-idle-change (current-buffer)))))))
+                           #'flycheck--handle-idle-change-in-buffer
+                           (current-buffer)))))))
 
-(defun flycheck-handle-idle-change (buffer)
-  "Handle an expired idle timer in BUFFER since the last change."
+(defun flycheck--handle-idle-change-in-buffer (buffer)
+  "Handle an expired idle timer in BUFFER since the last change.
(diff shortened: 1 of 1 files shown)


fmdkdd:
You'll need to rebase this on master to fix the CI.

flycheck.el
```diff
@@ -2745,14 +2745,21 @@ buffer."
           (flycheck-buffer-automatically 'new-line 'force-deferred)
         (setq flycheck-idle-change-timer
               (run-at-time flycheck-idle-change-delay nil
-                           #'flycheck-handle-idle-change (current-buffer)))))))
+                           #'flycheck--handle-idle-change (current-buffer)))))))
 
-(defun flycheck-handle-idle-change (buffer)
-  "Handle an expired idle timer in BUFFER since the last change."
+(defun flycheck--handle-idle-change (buffer)
```
fmdkdd: LGTM, except that I would like the wrapper to have a more discerning name than just an added dash.  Maybe add `-internal` or `-in-buffer` to the end?

cpitclaudel: Good point :) Updated.

fmdkdd:
@cpitclaudel LGTM.  Though I'm wondering if we shouldn't add a test to catch regressions on this one.  It looks like we already have a couple of tests related to idle-change in `flycheck-test.el` that could serve as a base.

cpitclaudel:
Possibly. I'm between trips ATM, so I have very little time. I'd vote for merging this now since the bug is already impacting people, and opening an issue for the tests.  I can tackle the issue in 2 weeks.

fmdkdd:
@cpitclaudel Good for me.  See #1324.

PR #1325, opened 2017-09-13
Make Haml-Syntax-Checker compatible with haml ~> 5.0.4

TeaMoe:
Using HAML from stdin is broken if the file has a ruby syntax error, but no haml error.
And Ruby-Syntax errors are reported in a different format, so i had to extend the error_patterns.


fmdkdd:
> Using HAML from stdin is broken if the file has a ruby syntax error, but no haml error.

Could that be an upstream bug?

TeaMoe:
> Could that be an upstream bug?
Yes, i would think so.

fmdkdd:
> Yes, i would think so.

Looks like it has [already been reported](https://github.com/haml/haml/issues/957).  Let's wait to see if they intend to fix it before working around it in Flycheck.

In any case, thank you for promptly suggesting a patch (including a test, even!).

TeaMoe:
Hi, I've updated my commit to work with haml v5.0.4, which includes a fix for --stdin.

PR #1340, opened 2017-09-28
Typo fixes for flycheck-versus-flymake.rst

msherry:
Small typo fixes.

### DIFF
--- a/doc/user/flycheck-versus-flymake.rst
+++ b/doc/user/flycheck-versus-flymake.rst
@@ -124,7 +124,7 @@ contributor agreement.
 **Flycheck** is not part of GNU Emacs, and is **unlikely to ever be** (see
 `issue 801`_).  However, it is free software as well, and publicly developed on
 the well-known code hosting platform :gh:`Github <flycheck/flycheck>`.
-Contributing to Flycheck does not require a copyright assignments.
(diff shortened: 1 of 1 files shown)


fmdkdd:
Thanks for noticing these.

Actually, I guess this:
> Contributing to Flycheck does not require a copyright assignment.

is not true anymore, since we make all contributors accept the contributor license agreement.  

cc @cpitclaudel

cpitclaudel:
The part about copyright assignments is still true, actually : we require a CLA but not an assignment

fmdkdd:
Oh, so I'm confused about the difference between the two situations. 

Is it that, in the assignment case, the FSF is the sole copyright owner of all the code, but with a CLA, the individuals retain their copyright and agree to waive parts of it for the situations specified by the license?

cpitclaudel:
Essentially, yes. With a CLA we as maintainers have no more rights or ownership over the code than other contributors: we essentially have the freedoms afforded by the GPL, but nothing more; each individual author retains sole copyright of their code.  With an FSF-style assignment, ownership of the code goes to the FSF.

fmdkdd:
Okay, I think I got it.  I've reworded the paragraph in c114b38 to explicitly mention the CLA.

PR #1342, opened 2017-10-01
Fixes #1332

### LINKED ISSUE #1332: Disable flycheck-display-error-at-point by timer
How can I disable `flycheck` behaviour  automatically display errors at point by timeout?
Just want to see it with `C-c ! h` on demand.

Pitometsu:
- Add custom flycheck-automatically-display-error-at-point
- Add fun flycheck-automatically-display-error-at-point.

# Thank you for `flycheck`!

According to https://github.com/flycheck/flycheck/issues/1332#issuecomment-332380258

@fmdkdd, please review. It works for me, but not sure, is I miss something important.
Any suggestions are welcome. 👍


fmdkdd:
@Pitometsu Thank you for pushing forward with this!  Please be patient until one of us has the time to review the PR :)

Pitometsu:
However, hooks  are here yet. They removing only on `flycheck-automatically-display-error-at-point` switching, but keeps when it customized, and `flycheck-mode` just load.

fmdkdd:
@Pitometsu I've taken a look.  I didn't know about the `:set` on `defcustom`, but I see your point: this doesn't get called when `flycheck-mode` loads.  Maybe you could extract the lambda as a `defun`, put the defun in the `set`, and also call it when `flycheck-mode` is enabled?

Otherwise, the easiest way to disable the calls to `display-error` would be to use the `flycheck-automatically-display-error-at-point` in the hooks.   That way, the hooks would still run, but then they the actual call to `display-error-at-point` would not take place while the variable is `t`.

Pitometsu:
@fmdkdd looks like fixed hooks now, please, review and check.

flycheck.el
```diff
@@ -2357,7 +2399,12 @@ buffer manually.
    (flycheck-mode
     (flycheck-clear)
 
-    (pcase-dolist (`(,hook . ,fn) flycheck-hooks-alist)
+    (pcase-dolist
+        (`(,hook . ,fn)
+         (if flycheck-automatically-display-error-at-point
+             flycheck-hooks-alist
```
fmdkdd: Shouldn't this conversion be reversed?  If we `flycheck-automatically-...` is `t`, then we want to bind all the hooks, right?

Pitometsu: yep, you right.

Pitometsu:
@fmdkdd could you review it now, please?

fmdkdd:
@Pitometsu Sorry, I'm short on time at the moment, and I want to get this a thorough look.  I should have more time in december.

In the meantime, I saw in #1332 that you mentioned the performance issue of `display-next-error` was fixed for you, so it did not come from Flycheck itself?  If performance is not the issue anymore, then we can make this PR much simpler by letting the hooks in place.

Pitometsu:
@fmdkdd current implementation done right and clear. So in my opinion it would be better to give it a try. Please feel free to leave suggestions/improvement here, if you have any.

flycheck.el
```diff
@@ -934,6 +934,25 @@ currently listed."
   :risky t
   :package-version '(flycheck . "0.20"))
 
+(defcustom flycheck-automatically-display-error-at-point t
```
fmdkdd: Please rename this variable to `flycheck-enable-display-error-at-point`.

flycheck.el
```diff
@@ -934,6 +934,25 @@ currently listed."
...
+
+If t, use `flycheck-display-error-at-point-timer' to show error under point
+after commands, and `flycheck-idle-change-timer' after changes.
+
+Otherwise disable all unnecessary timers to suppress messages."
+  :group 'flycheck
+  :type 'boolean
+  :risky t
+  :package-version '(flycheck . "0.31")
+  :set
```
fmdkdd: Please remove the whole `:set`, we will just use the value when activating/deactivating the minor mode instead.

flycheck.el
```diff
@@ -934,6 +934,25 @@ currently listed."
...
+  "Automatically display display error at point when available.
+
+If t, use `flycheck-display-error-at-point-timer' to show error under point
+after commands, and `flycheck-idle-change-timer' after changes.
+
+Otherwise disable all unnecessary timers to suppress messages."
+  :group 'flycheck
+  :type 'boolean
+  :risky t
+  :package-version '(flycheck . "0.31")
```
fmdkdd: The version string should just be "32".

flycheck.el
```diff
@@ -1354,6 +1373,21 @@ FILE-NAME is nil, return `default-directory'."
           (file-name-directory file-name)
         (expand-file-name default-directory)))))
 
+(defun flycheck-automatically-display-error-at-point-enable ()
```
fmdkdd: We don't need these two functions after all.

flycheck.el
```diff
@@ -2300,7 +2334,6 @@ Slots:
   '(
     ;; Handle events that may start automatic syntax checks
     (after-save-hook        . flycheck-handle-save)
-    (after-change-functions . flycheck-handle-change)
```
fmdkdd: I don't think we need to move this to the second hooks list.  This just triggers a syntax check on idle, if they are activated, but the displaying is controlled by `flycheck-finish-current-syntax-check` anyway.

flycheck.el
```diff
@@ -2639,7 +2689,7 @@ WORKING-DIR."
         (flycheck-error-list-refresh)
         (run-hooks 'flycheck-after-syntax-check-hook)
         (when (eq (current-buffer) (window-buffer))
-          (flycheck-display-error-at-point))
+          (flycheck-automatically-display-error-at-point))
```
fmdkdd: I would rather add a check for `flycheck-enable-display-error-at-point` in the `when`, and get rid of the new function.

flycheck.el
```diff
@@ -4310,6 +4360,11 @@ non-nil."
     (cancel-timer flycheck-display-error-at-point-timer)
     (setq flycheck-display-error-at-point-timer nil)))
 
+(defun flycheck-automatically-display-error-at-point ()
```
fmdkdd: I don't think we really need this function, since if the `defcustom` is nil, the hooks will never be called, and the only other place we call display-error, we can guard it directly (see above).

flycheck.el
```diff
@@ -4326,7 +4381,8 @@ non-nil."
   (flycheck-cancel-error-display-error-at-point-timer)
   (when (flycheck-overlays-at (point))
     (setq flycheck-display-error-at-point-timer
-          (run-at-time flycheck-display-errors-delay nil 'flycheck-display-error-at-point))))
+          (run-at-time flycheck-display-errors-delay nil
+                       'flycheck-automatically-display-error-at-point))))
```
fmdkdd: Same thing here, we can leave this unchanged, since the hooks shouldn't be run when the `defcustom` is nil.

cpitclaudel:
My understanding is that this PR was created to fix a performance problem, but it turns out that this problem wasn't caused by Flycheck itself.  Is that correct? If so, maybe we should just close this?

Pitometsu:
No, this's about UX and unwanted popups in echo area.

cpitclaudel:
Oh, wait, I think I understand now.  The problem is the automatic call that happens in -finish-current-syntax-check, right?

cpitclaudel:
Got it, sorry for the confusion.  I'm confused by the implementation.  Why is it not enough to just prevent the calls to flycheck-display-error-at-point?

cpitclaudel:
Gentle ping @Pitometsu ; I'd love to get that issue fixed, but I want to understand the consequences of the change and the logic of the implementation first.

fmdkdd:
@cpitclaudel My recollection of the issue is that the initial demand was clear: Flycheck should provide an option to disable the automatic error display in the echo area.  This is easily achieved by guarding the calls, as you suggest.

But then, if we do not need to call display-errors, we do not need the hook in post-command-hook either.  So the PR went with for this more complete, but more tricky to get right, solution.

If the performance of the post-command hook is a non-issue, then we should just go with the former solution.

cpitclaudel:
> If the performance of the post-command hook is a non-issue, then we should just go with the former solution.

I this it's tiny, so I'd feel more comfortable with the smaller and more trustworthy change.

fmdkdd:
@Pitometsu Do you still want to make the PR with the simpler solution that does not remove the hooks?  If you don't have time, it's fine, I can take care of it.  The idea is still sound and we'd still like to have it in Flycheck.

bbatsov:
Closing due to staleness (8+ years) and unsigned CLA. The related issue #2108 (which requested similar functionality) has also been closed. If this feature is still desired, please open a fresh PR. Thank you!

PR #1344, opened 2017-10-05
flycheck-mode-line-status-text: distinguish no-checker and interrupted

stepnem:


fmdkdd:
That's not a bad idea, I guess. I'm not even sure what `:interrupt` is useful for, to be honest.  Do you have a use-case?

stepnem:
> That's not a bad idea, I guess. I'm not even sure what `:interrupt` is
> useful for, to be honest. Do you have a use-case?

I don't, and I'm not sure either (other than what's stated in the doc
string of flycheck-define-generic-checker), I just checked it _can_ be
interrupted e.g. calling flycheck-stop. It might seem bike-shedding; I
simply didn't like the idea of not being able to distinguish between no
checker being available and a check being interrupted, even if the
latter might not happen often or at all.

fmdkdd:
> I simply didn't like the idea of not being able to distinguish between no checker being available and a check being interrupted

I agree that it's strange these two would share a symbol.  Reusing the bang `!` for the `errored` status would have seemed more sensible to me.  But having different symbols is even better.

Thank you for the patch.

cpitclaudel:
Thanks for the patch @stepnem. I use `interrupted` commonly; the problem is that Flycheck has it's own idea of when a check should run, and often checkers based on persistent subprocesses can't satisfy it immediately (it might be busy with another task, for example); in these cases, `interrupted` is useful (and doesn't indicate an error condition).

fmdkdd:
@cpitclaudel Ah, I see.   You provide an `:interrupt` function that checks if the subprocess is busy, in order to abort the check?  If that's the case, couldn't guarding the check with `:predicate` also work?

cpitclaudel:
> If that's the case, couldn't guarding the check with :predicate also work?

Indeed, yes; I have that test in the `predicate` as well (I guess I got scared about potential race conditions?).

Flychecking is low-priority in that example, so it could be preempted by other tasks and then interrupted.  But admittedly that doesn't happen currently.

Issue #1230, opened 2017-03-19
flycheck-buffer-saved-p causes an error on any buffer modification


Wilfred:
If `flycheck-checker` is set to `rust-cargo` (e.g. with `flycheck-select-checker`), every modification to the file produces an error in the minibuffer:

```
Error while checking syntax automatically: (error "Flycheck cannot use rust-cargo in this buffer, type M-x flycheck-verify-setup for more details")
```

This is because `flycheck-get-checker-for-buffer` expects `(flycheck-may-use-checker flycheck-checker)` to be `t`. If `flycheck-checker` isn't set, it works fine.

`:predicate` for `rust-cargo` is:

```emacs-lisp
(and (flycheck-buffer-saved-p)
                    (locate-dominating-file (buffer-file-name) "Cargo.toml"))
```

so this returns `nil` whenever the file is modified.

I think that `flycheck-get-checker-for-buffer` should either fall back to `flycheck-checkers`, or silently do nothing.

fmdkdd:
Wow.  It seems we haven't tested for `flycheck-checker` in a long time since,`rust-cargo` is not the only checker to have `flycheck-buffer-saved-p` as predicate, and no one caught this one until now.

I agree that if you manually set a checker, it should still obey the `:predicate`, and in this case do nothing if the predicate is nil.

@cpitclaudel ?

cpitclaudel:
Yup, I agree.  The predicate is a strict validity condition, so it makes no sense to ignore it;  but on the other hand, there's no point in warning repeatedly.

fmdkdd:
Somewhat related: #1072 currently removes `flycheck-checker`, so it would make this bug obsolete.

PR #1347, opened 2017-10-19
Don't raise an error an explicitly selected checker cannot be used

fmdkdd:
Closes GH-1230, which see.

We don't raise an error when no checker can be found anyway.  And we have
`flycheck-verify-setup` to diagnose misconfiguration.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2390,10 +2390,8 @@ otherwise search for the best checker from `flycheck-checkers'.
 Return checker if there is a checker for the current buffer, or
 nil otherwise."
   (if flycheck-checker
-      (if (flycheck-may-use-checker flycheck-checker)
-          flycheck-checker
-        (error "Flycheck cannot use %s in this buffer, type M-x flycheck-verify-setup for more details"
-               flycheck-checker))
+      (when (flycheck-may-use-checker flycheck-checker)
+        flycheck-checker)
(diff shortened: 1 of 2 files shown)


cpitclaudel:
This looks good. Could we move the error to the interactive version flycheck-select-checker? This way interactively picking an unusable checker would still error out.

fmdkdd:
We already have a check in `flycheck-select-checker`:

```emacs-lisp
  (unless (or (not checker) (flycheck-may-use-checker checker))
      (flycheck-verify-checker checker)
      (user-error "Can't use syntax checker %S in this buffer" checker))
```

If I try to select a checker for another mode, that raises a message and the verify window.

However, it's a bit buggy with `buffer-saved-p` in `:predicate`: if the buffer is not saved, and I do `M-x flycheck-select-checker rust RET`, then it raises the error (because the predicate is false), but since `flycheck-verify-checker` saves the current buffer, now the predicate is true, and all is green.

cpitclaudel:
Ah, I see. Shouldn't we use the `enabled` check in there, rather than the predicate?

fmdkdd:
Probably?  My feeling is that when `:enabled` was introduced, it took on part of the role that `:predicate` already had.   So some of these functions still rely on `:predicate` whereas they should have switched to `:enabled`.  But maybe all checkers have not been updated to use `:enabled` for one-time checks.

I'll have to look into it.

Issue #1298, opened 2017-08-02
Flycheck throws an error with coffeelint and control characters


AdamNiederer:
To reproduce: Press `C-q C-c` in any coffeescript buffer with flycheck and coffeelint enabled. Your coffeescript buffer will look like this:
```
^C
```

Flycheck will throw this error:

```
Suspicious state from syntax checker coffee-coffeelint: Flycheck checker coffee-coffeelint returned non-zero exit code 1, but its output contained no errors: <?xml version="1.0" encoding="utf-8"?>
<checkstyle version="4.3">
<file name="stdin">
<error line="1"
    severity="error"
    message="[stdin]:1:1: error: unexpected 

^; context: "
    source="coffeelint"/>
</file>
</checkstyle>

Try installing a more recent version of coffee-coffeelint, and please open a bug report if the issue persists in the latest release.  Thanks!
```
Versioning & Configuration:

```
Syntax checkers for buffer kek.coffee in coffee-mode:

  coffee
    - may enable: yes
    - executable: Found at /usr/bin/coffee

  coffee-coffeelint
    - may enable:         yes
    - executable:         Found at /usr/bin/coffeelint
    - configuration file: Found at "/home/redacted/.coffeelint.json"

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled checkers.

--------------------

Flycheck version: 31snapshot (package: 20170721.911)
Emacs version:    25.2.1
System:           x86_64-unknown-linux-gnu
Window system:    x
```

```bash
$ coffeelint --version
1.16.0```

fmdkdd:
Thanks for the report.

Looks like `flycheck-parse-checkstyle` cannot parse this correctly.  I'm not sure the XML output by `coffeelint` is correct though: are attribute values allowed to span multiple lines?

fmdkdd:
Setting `flycheck-xml-parser` to `flycheck-parse-xml-region` instead of `libxml-parse-xml-region` fixes it.

So it looks like the bug is with libxml which does not like some escape characters.

@cpitclaudel, should this go to the Emacs bug tracker?

cpitclaudel:
Can you provide a copy of the input that doesn't parse? Calling libxml-parse-xml-region works fine for me :/ (though it drops the newlines)

fmdkdd:
[...]

Issue #1330, opened 2017-09-22
Suspicious state from syntax checker javascript-eslint


Grox-Ni:
Hello,
 
Flycheck version: 31snapshot (package: 20170911.127)
Emacs version:    25.3.1(I am using Spacemacs with React layer)
System:           x86_64-apple-darwin16.4.0
Window system:    ns

I am opening a js file, and the flycheck is not working properly.

Error message is:

```

Suspicious state from syntax checker javascript-eslint: Flycheck checker javascript-eslint returned non-zero exit code 1, but its output contained no errors: can't resolve reference #/definitions/basicConfig from id #
can't resolve reference #/definitions/basicConfigOrBoolean from id #
can't resolve reference #/definitions/basicConfigOrBoolean from id #
<?xml version="1.0" encoding="utf-8"?><checkstyle version="4.3"><file name="/Users/grox/LintCode/LintCode/react/src/apps/problem/containers/ProblemListPage.js"><error line="19" column="26" severity="error" message="Multiple spaces found before &apos;style&apos;. (no-multi-spaces)" source="eslint.rules.no-multi-spaces" /></file></checkstyle>

Try installing a more recent version of javascript-eslint, and please open a bug report if the issue persists in the latest release.  Thanks!


```
I used yarn global to install node packages(like eslint) and have upgraded them all.

fmdkdd:
It seems to me this is the same issue as #1323.  If so, the error really is from the react plugin for `eslint`, and a fix has already been issue from their side, as far as I understand.

fmdkdd:
@Grox-Ni Have you tried updating `eslint-plugin-react` to 7.2 or later?  This should fix your issue.

n3dst4:
I am getting the `Suspicious state from syntax checker` error on a minimal project with no eslint plugins. Seems like eslint always sets exit code 1 if any problems are found. I'm using `add-node-modules-path` from MELPA to get local `node_modules/.bin` on PATH.

Flycheck: 20171019.83
add-node-modules-path: 20170914.191
Emacs: GNU Emacs 25.3.1 (x86_64-w64-mingw32) of 2017-09-2

I've pushed my minimal repro repo here: https://github.com/n3dst4/eslint-exit-codes

I'm new to emacs and thus also flycheck, so if there's any more info I can provide or anything else I can do to help, let me know.

fmdkdd:
@n3dst4 Thanks for the repro.  The relevant part is:
```
[...]

PR #1349, opened 2017-10-25
Rewrite flycheck-parse-xml-region (use xml-parse-region if libxml fails)

cpitclaudel:
Closes GH-1298, GH-1330.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5525,15 +5525,26 @@ Wrapper around `xml-parse-region' which transforms the return
 value of this function into one compatible to
 `libxml-parse-xml-region' by simply returning the first element
 from the node list."
-  (car (xml-parse-region beg end)))
+  (ignore-errors (car (xml-parse-region beg end))))
 
-(defvar flycheck-xml-parser
-  (if (fboundp 'libxml-parse-xml-region)
-      'libxml-parse-xml-region 'flycheck-parse-xml-region)
-  "Parse an xml string from a region.
+(defun flycheck-parse-xml-region-with-fallback (beg end)
+  "Parse the xml region between BEG and END.
 
-Use libxml if Emacs is built with libxml support.  Otherwise fall
-back to `xml-parse-region', via `flycheck-parse-xml-region'.")
+Try parsing with libxml first; if that fails, revert to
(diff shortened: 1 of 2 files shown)


flycheck.el
```diff
@@ -5521,19 +5521,25 @@ function `buffer-file-name'."
...
+
+The output of `xml-parse-region' is made compatible with that of
 `libxml-parse-xml-region' by simply returning the first element
-from the node list."
-  (car (xml-parse-region beg end)))
+of the node list."
+  ;; FIXME use `libxml-available-p' when it gets implemented.
+  (or (and (fboundp 'libxml-parse-xml-region)
+           (libxml-parse-xml-region beg end))
+      (ignore-errors (car (xml-parse-region beg end)))))
```
fmdkdd: Is this wise to silence errors here?

cpitclaudel: Do you mean here vs somewhere else, or silencing vs not silencing? If the latter, I don't think we have much of a choice: that's how libxml behaves (it just returns nil when it fails to parse), which we then complain about (*unexpected output …*).  Otherwise we'd throw an error, and we don't currently do that.

flycheck.el
```diff
@@ -5521,19 +5521,25 @@ function `buffer-file-name'."
...
+XML (see URL `https://github.com/flycheck/flycheck/issues/1298'),
+or on Windows by a missing libxml DLL with a libxml-enabled Emacs
+build (see URL `https://github.com/flycheck/flycheck/issues/1330').
+
+The output of `xml-parse-region' is made compatible with that of
 `libxml-parse-xml-region' by simply returning the first element
-from the node list."
-  (car (xml-parse-region beg end)))
+of the node list."
+  ;; FIXME use `libxml-available-p' when it gets implemented.
```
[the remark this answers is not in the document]
cpitclaudel: Correct

cpitclaudel:
> Although, I wonder if we shouldn't make a new function instead of modifying flycheck-xml-parse-region.

Indeed, absolutely.

flycheck.el
fmdkdd: Okay, so now at least it's coherent with `libxml`.

Issue #1348, opened 2017-10-25
Add option for --force-exclusion to ruby-rubocop checker?


mmcnl:
As discussed in https://github.com/bbatsov/rubocop/issues/893#issuecomment-95600333 when running rubocop from a linter such as flycheck, it is impossible to ignore a particular file or directory from the usual rubocop.yml config file. To address this, the rubocop project added the `--force-exclusion` command line option (https://github.com/bbatsov/rubocop/pull/922). Would it be possible to add `--force-exclusion` to flycheck's default invocation of rubocop, or to add a flycheck var to enable that option? 
Thanks so much for your work on this great tool!

fmdkdd:
> Would it be possible to add --force-exclusion to flycheck's default invocation of rubocop, or to add a flycheck var to enable that option?

Looking at the discussion in https://github.com/bbatsov/rubocop/issues/893#issuecomment-95600333, I get that `--force-exclusion` will tell `rubocop` to ignore the files that are specified in your configuration.  Is that it?

If you don't want to check a specific file, why run Flycheck on it then?  Do you have a specific use case that requires this option?

mmcnl:
Thanks for the reply, @fmdkdd . My use case (which I believe is common) is to be able to have the rubocop checker on my Rails-based project without being swamped with error highlighting when I open the `db/schema.rb` file.  (This file is continually auto-generated and thus it's impossible for me to fix the style offenses that get re-created there). This is the use case (for linters like sublime-linter or atom-linter) for which the rubocop team added the `--force-exclusion` option.

cpitclaudel:
> This is the use case (for linters like sublime-linter or atom-linter) for which the rubocop team added the --force-exclusion option.

Maybe I'm missing something, but this sounds odd.  Why do we need that option to achieve this? Isn't it simpler to just ask Flycheck not to lint these files, using whichever predicate you'd use to turn that option on?

IOW, what does it gain us to run rubocop on some file `a.rb` with the `--force-exclusion a.rb` flag? Why is that better than just disabling Flycheck for `a.rb`?

Maybe it'd help if you can show how you'd set this hypothetical new variable :) Let's assume we add `flycheck-rubocop-force-exclusion`, taking a list of files.  How would you use it?

cpitclaudel:
Oh, wait, I think I understand now.  `--force-exclusion` doesn't take a file name, right?
[...]

PR #1351, opened 2017-10-26
Add --force-exclusion flag to rubocop command

Simplify:
Fixes and closes GH-1348

Tested manually, works as expected. Didn't wrote any integration test, feels more as testing rubocop output than testing actual integration with flycheck.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,7 @@
 
   - Add ``flycheck-cppcheck-suppressions-file`` to pass a suppressions
     file to cppcheck [GH-1329]
+  - Add ``--force-exclusion`` flag to ``rubocop`` command [GH-1348]
 
 31 (Oct 07, 2017)
 =================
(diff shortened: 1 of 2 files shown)


Simplify:
I don't expect that anyone in the future will request removal or configuration option for this. Default rubocop configuration have following excludes:
- 'node_modules/**/*'
- 'vendor/**/*'

First one is for javascript files and does not make much sense to me. Second one is used to unpack specific versions of ruby gems (sometimes all needed gems on production environment, depends on deployment strategy) and usually you do not edit those.

On Rails you may need to add more files in there. Auto generated files like `db/schema.rb` will contain strings with double quotes and rubocop does not like that. Almost all lines in that file contain a string so get a warning for almost every line. You can't fix that because every time you add new migration and migrate database, new schema.rb is generated. On large projects that file is huge. That explains why people add that kind of files in excludes and you don't want to get 100+ errors when you just need to see database structure in project.

PR #1354, opened 2017-10-31
Add mypy as a flycheck checker

Wilfred:
This is based on https://github.com/lbolla/emacs-flycheck-mypy and
adds support for warnings too.

This was originally proposed by Lunaryorn: https://github.com/lbolla/emacs-flycheck-mypy/issues/1 and the original author is happy with us doing this.

Mypy is becoming more common in the Python world, it doesn't overlap with pylint or flake8, so I think it's a great fit for flycheck.

Let me know if I've missed anything or you'd like any changes.

### DIFF
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -900,6 +900,23 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       .. syntax-checker-config-file:: flycheck-flake8rc
 
+   .. syntax-checker:: python-mypy
+
+      Type check python with  `mypy <http://www.mypy-lang.org/>`_.
+
+      .. note::
+
+         This syntax checker requires mypy 0.580 or newer.
+
+      .. syntax-checker-config-file:: flycheck-python-mypy-ini
+
+      .. defcustom:: flycheck-python-mypy-cache-dir
+
+         Directory used to write ``.mypy_cache`` directories.
+
+         Set to ``null-device`` to disable writing cache directories
+         entirely.
+
    .. syntax-checker:: python-pylint
 
       Check syntax and lint with `Pylint <https://pylint.org/>`_.
--- a/flycheck.el
+++ b/flycheck.el
@@ -225,6 +225,7 @@ attention to case differences."
     python-flake8
     python-pylint
     python-pycompile
+    python-mypy
     r-lintr
     racket
     rpm-rpmlint
@@ -8949,6 +8950,35 @@ See URL `https://docs.python.org/3.4/library/py_compile.html'."
           "', ('" (file-name (one-or-more (not (any "'")))) "', "
           line ", " column ", " (one-or-more not-newline) line-end))
   :modes python-mode)
+
+(flycheck-def-config-file-var flycheck-python-mypy-ini python-mypy
+                              "mypy.ini"
+  :safe #'stringp)
+
+(flycheck-def-option-var flycheck-python-mypy-cache-dir nil python-mypy
+  "Directory used to write .mypy_cache directories."
+  :safe #'stringp
+  :type '(choice
+          (const :tag "Write to the working directory" nil)
+          (const :tag "Never write .mypy_cache directories" null-device)
+          (string :tag "Path"))
+  :package-version '(flycheck . "32"))
+
+(flycheck-define-checker python-mypy
+  "Mypy syntax and type checker.  Requires mypy>=0.580.
+
+See URL `http://mypy-lang.org/'."
+  :command ("mypy"
+            (config-file "--config-file" flycheck-python-mypy-ini)
+            (option "--cache-dir" flycheck-python-mypy-cache-dir)
+            source-original)
+  :error-patterns
+  ((error line-start (file-name) ":" line ": error:" (message) line-end)
+   (warning line-start (file-name) ":" line ": warning:" (message) line-end))
+  :modes python-mode
+  ;; Ensure the file is saved, to work around https://github.com/python/mypy/issues/4746.
+  :predicate flycheck-buffer-saved-p
+  :next-checkers '(t . python-flake8))
 
 (flycheck-def-option-var flycheck-lintr-caching t r-lintr
   "Whether to enable caching in lintr.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,7 @@
   - MarkdownLint CLI with ``markdownlint`` [GH-1366]
   - Rust with ``rust-clippy`` [GH-1385]
   - VHDL with ``ghdl`` [GH-1160]
+  - mypy with ``python-mypy`` [GH-1354]
 
 - New features:
 
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3742,27 +3742,29 @@ Why not:
        :checker puppet-lint)))
 
 (flycheck-ert-def-checker-test python-flake8 python syntax-error
-  (let ((python-indent-guess-indent-offset nil))       ; Silence Python Mode!
+  (let ((python-indent-guess-indent-offset nil) ; Silence Python Mode!
+        (flycheck-python-flake8-executable "python3"))
     (flycheck-ert-should-syntax-check
      "language/python/syntax-error.py" 'python-mode
      '(3 12 error "SyntaxError: invalid syntax" :id "E999"
          :checker python-flake8))))
 
 (flycheck-ert-def-checker-test python-flake8 python nil
-  (flycheck-ert-should-syntax-check
-   "language/python/test.py" 'python-mode
-   '(5 1 warning "'.antigravit' imported but unused" :id "F401"
-       :checker python-flake8)
-   '(7 1 warning "expected 2 blank lines, found 1" :id "E302"
-       :checker python-flake8)
-   '(12 29 warning "unexpected spaces around keyword / parameter equals"
-        :id "E251" :checker python-flake8)
-   '(12 31 warning "unexpected spaces around keyword / parameter equals"
-        :id "E251" :checker python-flake8)
-   '(21 1 warning "expected 2 blank lines after class or function definition, found 1"
-        :id "E305" :checker python-flake8)
-   '(22 1 error "undefined name 'antigravity'" :id "F821"
-        :checker python-flake8)))
+  (let ((flycheck-python-flake8-executable "python3"))
+    (flycheck-ert-should-syntax-check
+     "language/python/test.py" 'python-mode
+     '(5 1 warning "'.antigravit' imported but unused" :id "F401"
+         :checker python-flake8)
+     '(7 1 warning "expected 2 blank lines, found 1" :id "E302"
+         :checker python-flake8)
+     '(12 29 warning "unexpected spaces around keyword / parameter equals"
+          :id "E251" :checker python-flake8)
+     '(12 31 warning "unexpected spaces around keyword / parameter equals"
+          :id "E251" :checker python-flake8)
+     '(21 1 warning "expected 2 blank lines after class or function definition, found 1"
+          :id "E305" :checker python-flake8)
+     '(22 1 error "undefined name 'antigravity'" :id "F821"
+          :checker python-flake8))))
 
 (flycheck-ert-def-checker-test python-mypy python nil
   (let ((flycheck-disabled-checkers '(python-flake8))
@@ -3775,14 +3777,16 @@ Why not:
 
 (flycheck-ert-def-checker-test python-pylint python syntax-error
   (let ((flycheck-disabled-checkers '(python-flake8))
-        (python-indent-guess-indent-offset nil)) ; Silence Python Mode
+        (python-indent-guess-indent-offset nil) ; Silence Python Mode
+        (flycheck-python-pylint-executable "python3"))
     (flycheck-ert-should-syntax-check
      "language/python/syntax-error.py" 'python-mode
      '(3 1 error "invalid syntax (<string>, line 3)"
          :id "syntax-error" :checker python-pylint))))
 
 (flycheck-ert-def-checker-test python-pylint python nil
-  (let ((flycheck-disabled-checkers '(python-flake8)))
+  (let ((flycheck-disabled-checkers '(python-flake8))
+        (flycheck-python-pylint-executable "python3"))
     (flycheck-ert-should-syntax-check
      "language/python/test.py" 'python-mode
      '(1 1 info "Missing module docstring" :id "missing-docstring" :checker python-pylint)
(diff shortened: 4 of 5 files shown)


flycheck.el
```diff
@@ -8788,6 +8789,26 @@ See URL `https://docs.python.org/3.4/library/py_compile.html'."
           line ", " column ", " (one-or-more not-newline) line-end))
   :modes python-mode)
 
+(flycheck-def-args-var flycheck-python-mypy-args python-mypy)
```
cpitclaudel: We usually avoid generic `-args` variables, in favor of specific variables for each relevant flag; I think it's be better to do this here too.

flycheck.el
```diff
@@ -8788,6 +8789,26 @@ See URL `https://docs.python.org/3.4/library/py_compile.html'."
...
+
+(flycheck-define-checker python-mypy
+  "Mypy syntax and type checker.  Requires mypy>=0.3.1.
+
+Customize `flycheck-python-mypy-args` to add specific args to default
+executable.
+
+E.g. when processing Python 2 files, add \"--py2\".
```
cpitclaudel: For example, this could controled by a `flycheck-mypy-python-version` variable.

flycheck.el
```diff
@@ -8788,6 +8789,26 @@ See URL `https://docs.python.org/3.4/library/py_compile.html'."
...
+
+See URL `http://mypy-lang.org/'."
+
+  :command ("mypy"
+            (eval flycheck-python-mypy-args)
+            source-original)
```
cpitclaudel: mypy supports stdin, so we should use that.

flycheck.el
Wilfred: Which do you think are the most relevant?

```
$ mypy --help
usage: mypy [-h] [-v] [-V] [--python-version x.y] [--platform PLATFORM] [-2]
            [--ignore-missing-imports]
            [--follow-imports {normal,silent,skip,error}]
            [--disallow-any {unimported, expr, unannotated, decorated, explicit, generics}]
            [--disallow-untyped-calls] [--disallow-untyped-defs]
            [--disallow-incomplete-defs] [--check-untyped-defs]
            [--disallow-subclassing-any] [--warn-incomplete-stub]
            [--disallow-untyped-decorators] [--warn-redundant-casts]
            [--no-warn-no-return] [--warn-return-any] [--warn-unused-ignores]
            [--warn-unused-configs] [--show-error-context]
            [--no-implicit-optional] [-i] [--quick-and-dirty]
            [--cache-dir DIR] [--skip-version-check] [--strict-optional]
            [--strict-optional-whitelist [GLOB [GLOB ...]]]
            [--junit-xml JUNIT_XML] [--pdb] [--show-traceback] [--stats]
            [--inferstats] [--custom-typing MODULE]
            [--custom-typeshed-dir DIR] [--scripts-are-modules]
            [--config-file CONFIG_FILE] [--show-column-numbers]
            [--find-occurrences CLASS.MEMBER] [--strict]
            [--shadow-file SOURCE_FILE SHADOW_FILE] [--any-exprs-report DIR]
            [--cobertura-xml-report DIR] [--html-report DIR]
            [--linecount-report DIR] [--linecoverage-report DIR]
            [--memory-xml-report DIR] [--old-html-report DIR]
            [--txt-report DIR] [--xml-report DIR] [--xslt-html-report DIR]
            [--xslt-txt-report DIR] [-m MODULE] [-c PROGRAM_TEXT] [-p PACKAGE]
            [files [files ...]]
```

I like to run with `--py2 --strict --disallow-any=generics` for new scripts, or `--py2 --check-untyped-defs` when I'm gradually porting a codebase (to check calls to annotated functions even when the caller isn't yet annotated).

`--strict` includes quite a few arguments -- do we offer the individual arguments too?

flycheck.el
Wilfred: How do you feel about using `source-inplace` here instead? That way mypy can follow imports too.

Wilfred:
@cpitclaudel thanks for sharing your version!

I tend to use mypy with flake8 afterwards, so what should I put as `:next-checkers`? I see you've used `python-pylint`, but that doesn't seem to have `:next-checkers` configured.

dakra:
I use https://github.com/flycheck/flycheck/pull/1080 since 10 month now. Like in the comment there, I had to rename some arguments that got changed in newer mypy versions but otherwise works like a charm.

So I'm +1 for merging this here or #1080 or any mypy checker really ;)

flycheck.el
cpitclaudel: My usual approach is to add whatever the original writer of the checker finds useful, and accept pull requests for more as needed :)

flycheck.el
cpitclaudel: mypy can follow imports if you give it --shadow-file, and this way we don't need to pollute the current folder with `flycheck_`-prefixed files

cpitclaudel:
> I tend to use mypy with flake8 afterwards, so what should I put as :next-checkers? I see you've used python-pylint, but that doesn't seem to have :next-checkers configured.

It's an annoying issue, indeed, which is what stalled my previous PR for a mypy checker.
I'm OK with leaving flake8 in there.

dakra:
@Wilfred bump
I would like a mypy checker :)

Wilfred:
@cpitclaudel could you review this again? I think this is in much better shape now.

We now have just one configurable option, the path to a mypy.ini configuration file. This allows you to configure mypy in any way, which I find really convenient. If you have different mypy.ini files you can use .dir-locals.el.

This checker works pretty well for buffers that are backed by files, but I don't know how to handle other buffers. I want to do something like this:

```emacs-lisp
(if source-original
    ;; If this file is backed by a buffer, tell mypy where the
    ;; original file was but shadow it with a temporary file of the
    ;; current buffer contents.
    (concat "--shadow-file" source-original source source-original)
  ;; Just run mypy on a temporary file containing the buffer contents.
  source)
```

but I'm not sure how to express this with `eval` or the `:command` DSL.

Wilfred:
Added docs and changed the checker to use `(config-file ...)` which I hadn't spotted before.

fmdkdd:
> This checker works pretty well for buffers that are backed by files, but I don't know how to handle other buffers.

If `--shadow-file` is indeed checking the temporary created by `source`, then it should work seemlessly for buffers not backed by files.  Is this not the case?

Wilfred:
@fmdkdd `--shadow-file` allows you to say "use this temporary file as if it was this other file on disk". There is no 'other file on disk' for buffers that aren't backed by files.

fmdkdd:
> There is no 'other file on disk' for buffers that aren't backed by files.

Indeed, but why does `mypy` care about the actual file, since we already gave it the content it should check?  My question is, if we pass `foo.py` as a phony file path, with a real temporary file given to `--shadow-file`, does `mypy` complain?

cpitclaudel:
@fmdkdd I think it's a mypy bug:

```
$ cd /tmp/
$ mkdir python
$ cd python/
$ echo '"""test"""' > a.py
$ mypy --shadow-file b.py a.py a.py
mypy: can't read file 'a.py': No such file or directory
```

cpitclaudel:
I opened https://github.com/python/mypy/issues/4746

fmdkdd:
Great!  Keep us posted.

cpitclaudel:
Edit: acknowledged as an issue, but I don't have time to fix it myself, so let's just add a check to ensure that the buffer is saved.

We should also pass a flag to prevent all these `.mypy` cache directories from being created.

Then I'd be in favor of merging.  I think it shouldn't be enabled for all files by default, though, since it's pretty slow.  Or maybe it should just run last?

fmdkdd:
> Edit: acknowledged as an issue, but I don't have time to fix it myself, so let's just add a check to ensure that the buffer is saved.

Agreed.  We should add a comment stating that checking on saved buffers is a workaround for the mypy bug, and add a link to it.

> We should also pass a flag to prevent all these .mypy cache directories from being created.
Then I'd be in favor of merging. I think it shouldn't be enabled for all files by default, though, since it's pretty slow. Or maybe it should just run last?

Aren't these two related?  If you leave the cache, then potentially it can be reused for later calls?

In any case, if we put `python-mypy` after `python-pycompile`, then we force users to opt-in by selecting `mypy` in their own config (since `pycompile` should always be enabled if `python` is available).  I'd rather have that than a checker disabled by default.

fmdkdd:
> We could add an option.

Sure.

@Wilfred Are you still free to work on this PR?  The following changes need to be made:

1. Ensure that the buffer is saved in `:predicate` to workaround python/mypy#4746 (and add a comment to that effect with a link to the bug)
2. Move `python-mypy` in `flycheck-checkers` after `python-pycompile`
3. Add a user option to disable creating `.mypy-cache` folders

Wilfred:
@fmdkdd thanks, I had some time today and your explicit steps were very helpful :)

I've made the changes you requested, and rebased on latest master.

flycheck.el
```diff
@@ -8950,6 +8951,35 @@ See URL `https://docs.python.org/3.4/library/py_compile.html'."
...
+  :safe #'stringp)
+
+(flycheck-def-option-var flycheck-python-mypy-disable-incremental nil python-mypy
+  "Disable incremental type checking in mypy.
+
+When non-nil, mypy will run more slowly, but it will not create
+.mypy_cache files."
+  :safe #'booleanp
+  :type 'boolean
+  :package-version '(flycheck . "0.16"))
```
fmdkdd: The version number should be `32`, not `0.16`.

doc/languages.rst
```diff
@@ -900,6 +900,17 @@ to view the docstring of the syntax checker.  Likewise, you may use
...
+   .. syntax-checker:: python-mypy
+
+      Type check python with  `mypy <http://www.mypy-lang.org/>`_.
+
+      .. syntax-checker-config-file:: flycheck-python-mypy-ini
+
+      .. defcustom:: flycheck-python-mypy-disable-incremental
+
+         Switches off incremental type checking. This is slower, but
+         prevents mypy from writing ``.mypy_cache`` files.
```
cpitclaudel: I'd suggest to say `creating .mypy_cache folder` instead

cpitclaudel: I also wonder if the setting should be called `enable`, and be off by default.  These folders are really annoying. Opinions?

cpitclaudel: Or maybe there's a way to make mypy put all these folders in one place on the file system, and we should let users customize that?

flycheck.el
Wilfred: Fixed.

doc/languages.rst
Wilfred: I've updated the doc as you suggested.

I did think about having an enable flag, but this is consistent with the mypy behaviour, which enables caching by default. Personally I don't mind the .mypy_cache directory being created, and I find mypy rather slow otherwise.

fmdkdd:
Thanks!  One last thing: please add a line to `CHANGES.rst` under `New syntax checkers`.

It would also be nice to add a test for the checker in `flycheck-test.el`.

Wilfred:
I've updated `CHANGES.rst`.

I'm not sure how to add a test. Looking at e.g. `(flycheck-ert-def-checker-test python-flake8 python nil` in flycheck-test.el then it looks like the test depends on flake8 being installed.

I can see `main/requirements.txt` is installing flake8, but that doesn't look related. If I look at another test like `(flycheck-ert-def-checker-test puppet-parser puppet parser-error-puppet-4` in flycheck-test.el then it seems to be shelling out to `puppet`, but I can't see puppet being installed either.

fmdkdd:
@Wilfred right, so `make integ` can be run locally and it runs on our CI.  For both, the command will run only the tests of checkers for which there is an executable present.  The tests do not install the executable.

So, if you have `mypy` installed, you can add a test, and it should run locally with `make integ`.  Then, it will be skipped on CI because there `mypy` is not installed.  For that, we need to add `mypy` to the docker image we use for running the tests, over at [flycheck/docker-tools](https://github.com/flycheck/docker-tools).

I've just added `mypy` there https://github.com/flycheck/docker-tools/commit/f32347c4d56afa9c272577cb6311215277974186, so that should be good.

For writing the test, you probably want to set the `flycheck-disabled-checkers` variable (as in the `python-plint` tests), so that Flycheck picks `python-mypy` for the buffer being tested.

Wilfred:
@fmdkdd that's really helpful, thanks. I've had a crack at writing a test and I think it's correct now.

I also noticed an issue with `--disable-incremental` not actually doing what I expected, so I've added a new configuration variable to disable writing `.mypy_cache` properly.

Finally, I've corrected the docs regarding the minimum mypy version required.

fmdkdd:
@Wilfred Thanks!  That's real nice.  Looks like your test is passing on the CI, but now we some checkers are freaking out because of we are using Python3 instead of Python2.

I'll look into that and restart the build if necessary.

fmdkdd:
Ok so the python-pylint failures are due to `pylint` not being found when using `python3`.   Looks like a consequence of #1113.  In the Docker image:

```
$ pylint
```
invokes `pylint` correctly, but what we actually run by default is:
```
$ python -c 'import sys,runpy;sys.path.pop(0); runpy.run_module("pylint")'
  File "<string>", line 1, in <module>
  File "/usr/lib/python2.7/runpy.py", line 182, in run_module
    mod_name, loader, code, fname = _get_module_details(mod_name)
  File "/usr/lib/python2.7/runpy.py", line 107, in _get_module_details
    raise error(format(e))
ImportError: No module named pylint
```

Changing `python` to `python3` in the checker fixes it, but we probably don't want to do that.

Now, my understanding is that as it stands, python checkers need to be configured in order to work properly.  This was not case before #1113 if installing tools with `pip`.

Of course, one solution for the CI is to use python2 for all the Python tools, and use python3 for mypy.  But using python3 for all seemed simpler.

@cpitclaudel What is your take on this?

cpitclaudel:
@fmdkdd it should be enough to let-bind flycheck-python-pylint-executable to `python3` around the test — does that work?

fmdkdd:
> it should be enough to let-bind flycheck-python-pylint-executable to python3 around the test — does that work?

Yes it works.  And so does let-binding it to `pylint` more directly.  But what I was worried about was that users who had a similar setup would have to jump through the same hoops.  I thought the point of #1113 was to make the switch between `pylint for python2` and `pylint for python3` easier, and now I'm unclear as to how it eases that.

cpitclaudel:
>  But what I was worried about was that users who had a similar setup would have to jump through the same hoops. I thought the point of #1113 was to make the switch between pylint for python2 and pylint for python3 easier, and now I'm unclear as to how it eases that.

Correct (more precisely, the point of #1113 was to make it easy to use the pylint matching your version of python.  If you're using python3, you set the both python-shell-executable and flycheck-python-pylint-executable to `python3`; for python 2, you set both of them to `python2`.

(Did I get the question right?)

Wilfred:
Is there anything else you need from me? :)

fmdkdd:
@Wilfred I don't think so.  I just need to fix the python integration tests on CI and we can merge.

On a related note, has anyone taken a look at https://github.com/facebook/pyre-check/, how it compares to mypy, and whether we could integrate it with Flycheck?

PR #1357, opened 2017-11-07
Add coala checker

Grox-Ni:
This is a checker of coala(https://coala.io/#/home) the original plugin repo is https://github.com/coala/coala-emacs

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,7 @@
 - New syntax checkers:
 
   - Jsonnet with ``jsonnet`` [GH-1345]
+  - Coala with ``coala`` [GH-1357]
 
 - New features:
 
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -893,6 +893,10 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       Check syntax with Python's byte compiler (see :py:mod:`py_compile`).
 
+   .. syntax-checker:: coala
+
+      Check python with customized bears with `coala <https://coala.io>`_.
+
 .. supported-language:: R
 
    .. syntax-checker:: r-lintr
--- a/flycheck.el
+++ b/flycheck.el
@@ -168,6 +168,7 @@ attention to case differences."
     c/c++-cppcheck
     cfengine
     chef-foodcritic
+    coala
     coffee
     coffee-coffeelint
     coq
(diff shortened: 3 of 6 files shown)


fmdkdd:
Thanks for the contribution.

If I understand correctly, Coala is a checker that runs a plethora of other checkers, most of which can already be run by Flycheck already.  Is there a good reason to use Coala for, i.e., pep8, when we already have flake8?  Are there any checkers specific to Coala?  Or is the advantage in using a single configuration file?

Also, the way the checker is defined to be enabled in all the `:modes`, and the fact that it's high in the list of checkers, means that it will be the **default** checker for most of the major programming modes.  I.e., if I try to run the full `make integ` suite of tests, most of them will fail because `coala` is now the default checker.  I don't think we want that.  If we include coala in flycheck, the checker should probably be at the end of the list, and then users can opt-in to make it the default in their setup.  

WDYT @cpitclaudel ?

cpitclaudel:
Agreed :)

Grox-Ni:
@fmdkdd The good thing about coala is it only uses one configuration file and runs out of emacs. The bears in coala are similar to checkers in flycheck,  I don't think we want coala to be the default one either. Could you tell me how to do that? 

Also, it seems we have to add coala checker's description to all the mode it accepts(now I only put it under python), is there a way we can do better than putting it manually?😛

fmdkdd:
>  I don't think we want coala to be the default one either. Could you tell me how to do that?

Putting `coala` at the end of the `flycheck-checkers` list should be enough to make it available in all modes, but never overriding the default ones.

However, I'm still not convinced by the value of adding `coala` to flycheck.  I understand that a single configuration file is convenient, but if most of the checkers that Coala uses are already included in flycheck, and if you have to install these checkers anyway, the benefit is thin.

Unless, that is, Coala has useful lints that are not already provided by other checkers in Flycheck.  Otherwise, I'm inclined to let Coala live as the extension `coala-emacs` which already exists (and mention it in the list of extensions).

Grox-Ni:
In fact, coala has many checkers(called bears) that flycheck currently is not supporting, for example, swift, gitcommit, matlab and writing-related tools. (see https://github.com/coala/coala-bears/tree/master/bears) 

An extension is good except it's difficult for beginners who want to try it(they have to manually install it)

What do you think?😉

cpitclaudel:
> An extension is good except it's difficult for beginners who want to try it(they have to manually install it)

But don't they have to manually install Flycheck anyway? And, since Flycheck is a dependency of emacs-coala, they can just install Coala directly, right?

fmdkdd:
New lints we don't support certainly go into the "pros" column.

What I'm uneasy about:
1. The overlap with linters that are already included in Flycheck.  I see that `coala` can call `tslint`, `eslint`, etc.  Though there is a precedent for overlapping checkers (especially with the JavaScript ones).   Having multiple ways to run the same checks is confusing for users.
2. Usually, we put linters in the `:next-checker` list of syntax checkers.  Since `coala` mainly has lints, it should go into the `:next-checker` property of quite a handful of checkers.  Without this, `coala` will never run if a syntax checker is enabled for the language.  E.g., for python, you will always at least have `python` installed, and we have a checker for that without any `:next-checker` property, so `coala` would never run by default.
So that leaves a choice of either adding `coala` to the `:next-checker` property of all relevant checkers, or to let the user manually select `coala`.  One is a hassle for us (and gets in the way of #1072), and the other is a hassle for users.
(I'm assuming users want to run syntax checkers before getting lints from coala.  If this is not a common usage for coala, then ignore this point)
3.  Coala supports a bunch of modes, but we don't actually know if calling `coala` will produce lints for the current file, since it depends on what "bears" are actually included in the configuration, right?  It's not an issue if coala reports nothing for the file, but we should make sure it doesn't *complain* about not having any bears activated for the current file.

Grox-Ni:
That's convincing.😉 
I agree it is better to leave it as an extension, as it's easier to handle the behavior if users want to use coala.

fmdkdd:
> I agree it is better to leave it as an extension, as it's easier to handle the behavior if users want to use coala.

Well, I don't see anything in the extension right now that makes coala play nice with other checkers.  So either users of this extension stick to the `coala` checker exclusively, or they jump through hoops to make it work with other checkers for a given language.

My second point was more that our current way of chaining checkers via `:next-checker` was not a good fit for tools like coala.  We'll certainly keep this use case in mind for #1072.

That is to say, I believe it's best to keep `coala-emacs` as an extension, but only because we don't have the facilities in Flycheck right now to handle it well.

I would be curious to hear more about how coala is used in conjuction with other checkers.

fmdkdd:
Closing as the discussion has gravitated towards not merging.

PR #1360, opened 2017-11-17
Add flycheck-eslint-args var

gpittarelli:
It can be useful to set custom eslint arguments, for example if I want to see stricter lint rules while editing than what the project currently enforces in `.eslintrc`. `--quiet` or `--maxwarnings` also seem potentially useful. 

Example value of `flycheck-eslint-arg` :
```
("--rule" "no-unused-vars: [2]")
```

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8083,6 +8083,9 @@ See URL `http://www.jshint.com'."
   :modes (js-mode js2-mode js3-mode rjsx-mode)
   :next-checkers ((warning . javascript-jscs)))
 
+(flycheck-def-args-var flycheck-eslint-args javascript-eslint
+  :package-version '(flycheck . "32"))
+
 (flycheck-def-option-var flycheck-eslint-rules-directories nil javascript-eslint
   "A list of directories with custom rules for ESLint.
 
@@ -8109,6 +8112,7 @@ for more information about the custom directories."
 See URL `http://eslint.org/'."
   :command ("eslint" "--format=checkstyle"
             (option-list "--rulesdir" flycheck-eslint-rules-directories)
+            (eval flycheck-eslint-args)
             "--stdin" "--stdin-filename" source-original)
   :standard-input t
   :error-parser flycheck-parse-checkstyle
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -624,6 +624,10 @@ to view the docstring of the syntax checker.  Likewise, you may use
          this syntax checker if eslint cannot find a valid configuration file
          for the current buffer.
 
+      .. defcustom:: flycheck-eslint-args
+
+         A list of additional arguments that are passed to eslint.
+
       .. defcustom:: flycheck-eslint-rules-directories
 
          A list of directories with custom rules.


fmdkdd:
We tend to shy away from including catch-all arguments variables, mainly for security reasons.  I believe the attack scenario here would be a directory or file local variable from foreign code setting the variable value to some malicious content, and Flycheck would `eval` the content automatically when you load the file.  

@cpitclaudel, is this the main reason?

Here for instance, can't you use the already available `--rulesdir` option?

cpitclaudel:
> mainly for security reasons.

It's certainly one thing to be careful about, but I'm not sure how it ranks vs clean API design.  Our general approach has been to favor configuration files, possibly project-local ones (because they are not Flycheck-specific), and then to fall back to specific options, one per flag (they work nicely with Emacs' customization interface, and they help maintain the invariant that tweaking Flycheck setting shouldn't be able to break a checker).

fmdkdd:
> and then to fall back to specific options, one per flag (they work nicely with Emacs' customization interface, and they help maintain the invariant that tweaking Flycheck setting shouldn't be able to break a checker).

On the other hand, it's more work for us to state out explicitly which arguments Flycheck can pass to a checker, and a more friction for users when the option they want to pass is not already in Flycheck.  But I'm okay sleeping in that bed.  Thanks for stating the reasoning out.

That invariant is surely worth it, and it gives me ideas about fuzzing customization variables for testing it...

Back to this PR: @gpittarelli, we would have to consider the merit of each argument (`--quiet` `--maxwarnings` and `--rule`) separately, and add a separte customization variable for each.  Can you suggest concrete use case where these would be useful?

cpitclaudel:
> On the other hand, it's more work for us to state out explicitly which arguments Flycheck can pass to a checker

Definitely (hence my preference for checker-specific config files ^^)

> and more friction for users when the option they want to pass is not already in Flycheck

But not too too much. For example, here, it's enough to write this:

```elisp
(setf (flycheck-checker-get 'javascript-eslint 'command)
      '("eslint" "--format=checkstyle"
        (option-list "--rulesdir" flycheck-eslint-rules-directories)
        "--rule" "no-unused-vars: [2]" ;; ← this line changed
        "--stdin" "--stdin-filename" source-original))
```

It's definitely extra friction, but not that much :) The trickiest part is figuring out the template:

```elisp
(setf (flycheck-checker-get '<checker-name> 'command)
      '(<command-line>))
```

I'd be OK to document this, though I'd warn against it, too.

fmdkdd:
> I'd be OK to document this, though I'd warn against it, too.

Well, the obvious downside of this kind of patching is that you have to keep that in sync whenever you update Flycheck, or you risk running into weird issues.

And, if other users have the same use case as you, then they have to use the same work around.  If you submit a PR instead, then everyone can benefit ;)

flycheck.el
```diff
@@ -8083,6 +8083,9 @@ See URL `http://www.jshint.com'."
   :modes (js-mode js2-mode js3-mode rjsx-mode)
   :next-checkers ((warning . javascript-jscs)))
 
+(flycheck-def-args-var flycheck-eslint-args javascript-eslint
+  :package-version '(flycheck . "0.32"))
```
fmdkdd: The version string should be `"32"` not `"0.32"`.

cpitclaudel:
@fmdkdd Yup, I'm OK with it :)  Thanks for the discussion!

fmdkdd:
@gpittarelli Great!  Thank you.  One last thing: Please squash your commit before we can merge.

viniciussbs:
Hi, guys! Any update on this? With this var I can pass `--ext .js --ext .graphql` to also [lint GraphQL files](https://github.com/apollographql/eslint-plugin-graphql).

cpitclaudel:
Woops. Thanks for the reminder!

PR #1361, opened 2017-11-20
Add cwl checker

tom-tan:
This request adds a syntax checker for [Common Workflow Language](http://www.commonwl.org) (CWL) that describes analysis workflows and tools.
Its syntax is defined as [Semantic Annotations for Linked Avro Data](http://www.commonwl.org/v1.0/SchemaSalad.html) (SALAD).
I added a syntax checker for CWL using `schema-salad-tool` ([link to pip repository](https://pypi.python.org/pypi/schema-salad)), which is a validator of SALAD format.

To test the CWL checker, I copied the schema files for CWL from [the repository of schema salad](https://github.com/common-workflow-language/schema_salad/tree/master/schema_salad/tests/test_schema), which are licensed under Apache-2.
I guess it is harmless to include these files to this repository for this purpose because [Apache-2 and GPL-3 are compatible](https://en.wikipedia.org/wiki/Apache_License#GPL_compatibility).


test/flycheck-test.el
```diff
@@ -2873,6 +2873,13 @@ The term \"1\" has type \"nat\" while it is expected to have type \"bool\"." :ch
    '(4 14 error "Expected a `FUNCTION` or `IDENT` after colon at line 4, col 14."
        :id "ParsingErrors" :checker css-csslint)))
 
+(flycheck-ert-def-checker-test cwl cwl syntax-error
+  (let ((flycheck-cwl-schema-path "schema/CommonWorkflowLanguage.yml"))
+    (flycheck-ert-should-syntax-check
+     "language/cwl/cwl.cwl" 'cwl-mode
+     '(6 5 error "the `inputBinding` field is not valid because value is a str, expected null or CommandLineBinding"
```
fmdkdd: Seeing as there is only one expected error, do we really need to include all these files for testing?  In other words, can you build a minimal test that triggers the same error?

tom-tan:
Unfortunately, we can not reduce files in `schema/` because they are referred by `schema/CommonWorkflowLanguage.yml`.

IMO it is difficult to remove other schema files.
If we modify the schema to remove the dependency, the test checks the message for *modified* CWL and it is not what we want to test.

tom-tan:
ping.
Is it ready to merge or should we need a workaround not to use the schemas?

Vagrantfile
```diff
@@ -171,6 +172,9 @@ Vagrant.configure("2") do |config|
                                                   typescript \
                                                   js-yaml
 
+    echo "Installing packages with pip..."
+    pip install schema-salad
```
fmdkdd: This needs a `--user` flag.

fmdkdd:
@tom-tan I just needed to find some time to run the test locally.  Thank you for adding one, by the way.  This is a good PR all around.

Could you also add a line to `CHANGES.rst` under `New syntax checkers`?  And squash your commits when you are done, then we can merge.

tom-tan:
Done!

fmdkdd:
Great!  Nice work 👏 , and sorry for the delay.

tom-tan:
Thank you for merging!

PR #1363, opened 2017-11-22
Fix ESLint working directory

vxcamiloxv:
Some plugins like [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import) not works property because the working directory is not set, if you run `node_modules/.bin/eslint --format\=checkstyle --stdin --stdin-filename /user/path/project/myTest.js < /user/path/project/myTest.js` outside of project show a lot of errors (using `eslint-plugin-import`) but in root project works fine.

This PR fixes that.

[See Also](https://github.com/benmosher/eslint-plugin-import#sublimelinter-eslint) for more details

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8164,6 +8164,23 @@ See URL `https://eslint.org' for more information about ESLint."
           (let-alist (caar (flycheck-parse-json output))
             .messages)))
 
+(defun flycheck-eslint--find-working-directory (_checker)
+  "Look for a working directory to run CHECKER in.
+
+This will either be the directory that contains `.eslintrc' by
+supports configuration files formats or `.eslintignore', if no
+such file is found in the directory hierarchy, searches
+`node_modules' directory to detect root of project."
+  (let* ((regex-config (concat "\\`\\.eslintrc"
+                               "\\(\\.\\(js\\|ya?ml\\|json\\)\\)?\\'")))
+    (when buffer-file-name
+      (or (locate-dominating-file buffer-file-name "node_modules")
+          (locate-dominating-file buffer-file-name ".eslintignore")
+          (locate-dominating-file
+           (file-name-directory buffer-file-name)
+           (lambda (directory)
+             (> (length (directory-files directory nil regex-config t)) 0)))))))
+
 (flycheck-define-checker javascript-eslint
   "A Javascript syntax and style checker using eslint.
 
@@ -8175,6 +8192,7 @@ See URL `http://eslint.org/'."
   :error-parser flycheck-parse-eslint
   :enabled (lambda () (flycheck-eslint-config-exists-p))
   :modes (js-mode js-jsx-mode js2-mode js2-jsx-mode js3-mode rjsx-mode)
+  :working-directory flycheck-eslint--find-working-directory
   :verify
   (lambda (_)
     (let* ((default-directory


flycheck.el
```diff
@@ -8103,6 +8103,25 @@ for more information about the custom directories."
                                                  "--print-config" "."))))
     (eq exitcode 0)))
 
+(defun flycheck-eslint--find-working-directory (_checker)
+  "Look for a working directory to run CHECKER in.
+This will either be the directory that contains `.eslintrc' in some of
+supports configuration files formats, if no such file is found in the directory
```
cpitclaudel: Did some text get omitted?

flycheck.el
```diff
@@ -8103,6 +8103,25 @@ for more information about the custom directories."
...
+hierarchy, searches `node_modules' directory to detect root of project."
+  (let* ((config-regexp (concat "\\`"
+                                (regexp-quote ".eslintrc")
```
cpitclaudel: I'd just inline this (`\.eslintrc`)

flycheck.el
```diff
@@ -8103,6 +8103,25 @@ for more information about the custom directories."
...
+                                "\\(\\.\\(js\\|ya?ml\\|json\\)\\)?\\'")))
```
cpitclaudel: …but regex-opt might be appropriate here

flycheck.el
```diff
@@ -8103,6 +8103,25 @@ for more information about the custom directories."
...
+    (or
+     (and
+      buffer-file-name
+      (or (locate-dominating-file
+           default-directory
+           (lambda (directory)
+             (> (length (directory-files directory nil config-regexp 0)))))
+          (locate-dominating-file buffer-file-name "node_modules")
+          (file-name-directory buffer-file-name)))
```
cpitclaudel: I'm not sure this line is needed — isn't default-directory set to that by default?

flycheck.el
```diff
@@ -8103,6 +8103,25 @@ for more information about the custom directories."
...
+supports configuration files formats, if no such file is found in the directory
+hierarchy, searches `node_modules' directory to detect root of project."
+  (let* ((config-regexp (concat "\\`"
+                                (regexp-quote ".eslintrc")
+                                "\\(\\.\\(js\\|ya?ml\\|json\\)\\)?\\'")))
+    (or
+     (and
+      buffer-file-name
+      (or (locate-dominating-file
+           default-directory
```
cpitclaudel: I'd use buffer-file-name here for consistency

flycheck.el
vxcamiloxv: No really, no for eslint

cpitclaudel: I think it's always the case that opening a file in Emacs sets default directory to the containing folder, so I wouldn't worry about removing this line :)

flycheck.el
```diff
@@ -8103,6 +8103,24 @@ for more information about the custom directories."
...
+    (or
+     (locate-dominating-file
+      default-directory
+      (lambda (directory)
+        (> (length (directory-files directory nil regex-opt t)) 0)))
+     (and
+      buffer-file-name
+      (or (locate-dominating-file buffer-file-name "node_modules")
+          (file-name-directory buffer-file-name)))
+     default-directory)))
```
cpitclaudel: Sorry, that's not what I has in mind. I was thinking of something like this:

```elisp
(when buffer-file-name
  (or (locate-dominating-file
       buffer-file-name
       (lambda (directory)
         (> (length (directory-files directory nil regex-opt t)) 0)))
      (locate-dominating-file
       buffer-file-name
       "node_modules")))
```

(you don't need default-directory, since that function can return `nil` to indicate that)

flycheck.el
```diff
@@ -8103,6 +8103,24 @@ for more information about the custom directories."
                                                  "--print-config" "."))))
     (eq exitcode 0)))
 
+(defun flycheck-eslint--find-working-directory (_checker)
+  "Look for a working directory to run CHECKER in.
+This will either be the directory that contains `.eslintrc' by supports
+configuration files formats, if no such file is found in the directory
+hierarchy, searches `node_modules' directory to detect root of project."
+  (let* ((regex-opt (concat "\\`\\.eslintrc"
```
cpitclaudel: ^^ I meant to say that you could use the *function* regex-opt :)  But it's fine like this (just rename the variable)

vxcamiloxv: Ho sorry :D

flycheck.el
vxcamiloxv: 1) `(locate-dominating-file buffer-file-name (lambda (directory)...` not works because require a directory, should be `(locate-dominating-file (file-name-directory buffer-file-name) (lambda (directory)...`

2) That is the problem, `default-directory` is not being established as working directory, in other parts of the code, the function to set `working-directory` returns a `default-directory` https://github.com/flycheck/flycheck/blob/master/flycheck.el#L6902

cpitclaudel: 1. Thanks.
2. I may be missing something, but I don't understand this :/

vxcamiloxv: After some test `default-directory` is not necessary

Simplify:
@cpitclaudel I'm ok with this PR, unless you have more requests for code changes. 

@vxcamiloxv can you squash your 6 commits into single one? We prefer one commit per PR.

vxcamiloxv:
@Simplify @cpitclaudel Done.

fmdkdd:
@vxcamiloxv Thanks.  @cpitclaudel Can you take another look?

flycheck.el
```diff
@@ -8152,6 +8152,21 @@ See URL `https://eslint.org' for more information about ESLint."
           (let-alist (caar (flycheck-parse-json output))
             .messages)))
 
+(defun flycheck-eslint--find-working-directory (_checker)
+  "Look for a working directory to run CHECKER in.
+This will either be the directory that contains `.eslintrc' by supports
+configuration files formats or `.eslintignore', if no such file is found in
+the directory hierarchy, searches `node_modules' directory to detect root of project."
```
cpitclaudel: Please rewrap :)

vxcamiloxv: I don't understand

vxcamiloxv: All docstrings looks similar for me, you need a space between first and second line?

Simplify: Yes, and try not to make documentation lines longer than 60 chars.

fmdkdd: I think `M-x fill-paragraph` with the cursor inside the docstring should do it, right?

Simplify: Was that also `M-q`?

fmdkdd: @Simplify On a stock Emacs, yes. But I favor suggesting calling commands directly, as not everyone has the same keybindings, and so that can be confusing ;)

vxcamiloxv: Done

flycheck.el
```diff
@@ -8164,6 +8164,22 @@ See URL `https://eslint.org' for more information about ESLint."
           (let-alist (caar (flycheck-parse-json output))
             .messages)))
 
+(defun flycheck-eslint--find-working-directory (_checker)
+  "Look for a working directory to run CHECKER in.
```
fmdkdd: You forgot to add a newline after the first sentence.  After that, all good!

vxcamiloxv: done.

fmdkdd:
@vxcamiloxv Thank you for sticking to this!  👏

cpitclaudel:
Indeed, thanks a lot!

jmpridgen:
Please forgive my lack of elisp training, however shouldn't the search for .eslintrc come first in the or statement? This function is causing my working-directory to default to the project root. I changed my function to look like this and now it is working correctly for my application.

```
(defun flycheck-eslint--find-working-directory (_checker)
  "Look for a working directory to run CHECKER in.

This will either be the directory that contains `.eslintrc' by
supports configuration files formats or `.eslintignore', if no
such file is found in the directory hierarchy, searches
`node_modules' directory to detect root of project."
  (let* ((regex-config (concat "\\`\\.eslintrc"
                               "\\(\\.\\(js\\|ya?ml\\|json\\)\\)?\\'")))
    (when buffer-file-name
      (or (locate-dominating-file
           (file-name-directory buffer-file-name)
           (lambda (directory)
             (> (length (directory-files directory nil regex-config t)) 0)))
          (locate-dominating-file buffer-file-name ".eslintignore")
          (locate-dominating-file buffer-file-name "node_modules")
          ))))

```

Simplify:
@jmpridgen what is exact problem that you are facing when your working directory is the same as root of project? Do you have multiple `.eslintignore` files in sub-directories? That is only thing that I can think about that may cause issues.

We already had config detection for eslint and we remove it almost 2 years ago. it's not reliable enough, for example user can store eslint configuration inside `package.json`. Best tool to detect eslint config file is eslint itself. Eslint will always pickup config file closest to the file Flycheck wants to check.

There are several reasons that working directory is same as project root: 
- Most people will want that working-directory is root of the project in order to easily include `node_modules/.bin` to the PATH. 
- Having working-directory same as project root helps with easier CI integration later. 
- Most JS frameworks come with CLI nowdays that work from project root. 
- In most other languages your working-directory will be project root.

fmdkdd:
Regardless of whether we should look for `.eslintrc` first, the docstring seems to imply that we do (and is missing some words).  We should at least fix that.

vxcamiloxv:
For eslint the really important is look fist when lives de dependencies, the working directory if you have multiples .eslintrc eslint look automatic look for them and use it. https://eslint.org/docs/user-guide/configuring#relative-glob-patterns

vxcamiloxv:
You can put the structures projects like this:
```
.eslintrc // Global configuration
-- server/
--- .eslintrc // extend global configuration
-- client/ 
--- .eslintrc // extend global configuration 
```

and in your **server/client** `.eslint`

```javascript
{
    "extends": [
        "../.eslintrc" // Global configuration
    ],
    "rules": { // Additional rules
        "eqeqeq": "warn" 
    }
}
```

or adds a `.eslintignore`
```
.eslintignore // project root
-- server/
--- .eslintrc // configuration for server
-- client/ 
--- .eslintrc // configuration for client
```

jmpridgen:
I will try this later today

jmpridgen:
@vxcamiloxv 
That worked. I now have a base .eslintrc in my project root and two more that extend that file inside a server and client folder. Sorry for the delay and thanks for your help.

Issue #1378, opened 2017-12-26
Flycheck doens't work with latest nightly Rust compiler


kngwyu:
In Cargo project, flycheck always report no error if using latest compiler (nightly 2017-12-25).
I checked that with a bit older compiler(toolchain nightly-2017-12-24-x86_64-unknown-linux-gnu),
flycheck works properly.
However,  the output of `cargo rustc --message-format=json ` to stdout of both of two compilers are exactly same, so I have no idea why this bag happens(sorry).

To test latest compiler,  
`rustup toolchain install nightly`or if youv'e installed, just`rustup update` and `rustup default nightly`.
To test a bit older one,
`rustup toolchain install nightly-2017-12-24`
`rustup default nightly-2017-12-24`

fmdkdd:
Okay, thanks for the heads up.  I will try to look into it.

fmdkdd:
Can reproduce with:

```
nightly-x86_64-unknown-linux-gnu unchanged - rustc 1.24.0-nightly (cddc4a62d 2017-12-26)
```

kngwyu:
Thanks.
I will try to explore the cause of this problem depending on my environment.

fmdkdd:
What I found on my end for now is that there does not seem to be any significant changes between the stable and nightly version in `cargo rustc`.  However, when looking at the errors Flycheck parses, there *does* seem to be a difference between the two versions in file names.  Ultimately, the nightly version will produce file names that do not match the buffer filename.

We should probably switch to get absolute filenames from the JSON output anyway.

fmdkdd:
What I've witnessed is that the filenames produced by nightly are relative to the Cargo.toml file, and previously they were not.

One way to fix it is to add the `working-directory` property:

```
  :working-directory (lambda (_)
                       (-when-let (file (buffer-file-name))
                         (locate-dominating-file file "Cargo.toml")))
```

With this added, nightly works fine.

kngwyu:
Adding `working-directory` property worked!
Many thanks.

cpitclaudel:
Do we need the null-check? Or do we only call working-directory when we have a `buffer-file-name`?

fmdkdd:
@cpitclaudel The null check is probably redundant with `:enabled`, but I wasn't sure.

Concerning this bug, I'll look into making a proper PR once I've found the corresponding change in nightly.

ArulselvanMadhavan:
[...]

PR #1381, opened 2017-12-31
Fix rust cargo nightly

fmdkdd:
See #1378.

The fix should be backwards-compatible (I've tested with stable and nightly).

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1,6 +1,6 @@
 ;;; flycheck.el --- On-the-fly syntax checking -*- lexical-binding: t; -*-
 
-;; Copyright (C) 2017 Flycheck contributors
+;; Copyright (C) 2017-2018 Flycheck contributors
 ;; Copyright (C) 2012-2016 Sebastian Wiesner and Flycheck contributors
 ;; Copyright (C) 2013, 2014 Free Software Foundation, Inc.
 ;;
@@ -1709,7 +1709,7 @@ are mandatory.
      nil to fall back to the `default-directory' of the current
      buffer.
 
-     This property is optional.  If omitted invoke `:start'
+     This property is optional.  If omitted, invoke `:start'
      from the `default-directory' of the buffer being checked.
 
 Signal an error, if any property has an invalid value."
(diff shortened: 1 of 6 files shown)


flycheck.el
```diff
@@ -9329,6 +9329,8 @@ This syntax checker requires Rust 1.15 or newer.  See URL
   :enabled (lambda ()
              (-when-let (file (buffer-file-name))
```
cpitclaudel: Just noticing this now: I think it'd be simpler to use `buffer-file-name` (the variable) instead of the function.  As in:

```
(when buffer-file-name 
  (locate-dominating-file buffer-file-name "Cargo.toml"))
```
  
Or even this:

```
(and buffer-file-name 
     (locate-dominating-file buffer-file-name "Cargo.toml"))
```

flycheck.el
```diff
@@ -9329,6 +9329,8 @@ This syntax checker requires Rust 1.15 or newer.  See URL
...
                (locate-dominating-file file "Cargo.toml")))
+  :working-directory (lambda (_)
+                       (locate-dominating-file (buffer-file-name) "Cargo.toml"))
```
cpitclaudel: Can we assume that we have a buffer-file-name when `:working-directory` is set? If so, should we document this? (presumably in `flycheck-define-generic-checker` we'd say that the function is only run if the `:predicate` and `:enabled` functions ran.)

fmdkdd: I don't think we can assume that, no.  I just tested it, and it can definitely fail if the buffer has no file name.

In fact, looking at `flycheck-define-generic-checker`, I see that `working-directory` is used as the default directory for the `:enabled` and `:predicate` predicates, so `:working-directory` must guard `buffer-file-name`, otherwise the `enabled` check may fail.

flycheck.el
fmdkdd: Should we change to the variable elsewhere in the code?  There are 80 occurrences.  Not only in checkers, but in the core code as well.

flycheck.el
```diff
@@ -9326,39 +9326,46 @@ This syntax checker requires Rust 1.15 or newer.  See URL
...
+  (lambda ()
+    (and buffer-file-name
+         (locate-dominating-file buffer-file-name "Cargo.toml")))
+  :working-directory
+  (lambda (_)
+    (and buffer-file-name
+         (locate-dominating-file buffer-file-name "Cargo.toml")))
+  :verify
+  (lambda (_)
+    (and buffer-file-name
```
cpitclaudel: This top-level and doesn't look right to me: it will cause `C-c ! v` to fail if there's no buffer-file-name.  Is that what we want? (Maybe it's an uncommon enough case that we're OK with it?)

flycheck.el
cpitclaudel: I wonder.  I'm not sure why there's a function and a variable, except for the convenience of passing the function around.

I'm all for dropping things like `(when-let (fname (buffer-file-name)) …)`; I don't think it's worth polluting the history in cases where it's just cosmetic.

flycheck.el
cpitclaudel: OK, yay for code review then ^^

flycheck.el
fmdkdd: It won't fail `verify-setup`; there just won't be any verification message:

```
  rust-cargo
    - may enable: Automatically disabled!
    - predicate:  nil
    - executable: Found at /usr/sbin/cargo
```

We could also emit a `verification-result` indicating that the checker cannot work without a filename, but I don't think it's common enough to be worth the trouble.

flycheck.el
fmdkdd: I found a few places with this pattern, but I don't think using the variable would improve the situation:

In `flycheck-haskell--find-default-directory`:
```
     (or
      (when (buffer-file-name)
        (flycheck--locate-dominating-file-matching
         "stack.*\\.yaml\\'"))
      (-when-let* ((stack (funcall flycheck-executable-find "stack"))
    ...
```

In `flycheck-locate-sphink-source-directory`:
```
  (-when-let* ((filename (buffer-file-name))
               (dir (locate-dominating-file filename "conf.py")))
    (expand-file-name dir)))
```

In `flycheck-locate-config-file-ancestor-directories`:
```
(-when-let* ((basefile (buffer-file-name))
               (directory (locate-dominating-file basefile filename)))
    (expand-file-name filename directory)))
```

However, we can get rid of the `sphinx-source-directory` function by using the last one instead.

flycheck.el
fmdkdd: @cpitclaudel Is this good for you, or should I add a verification-result?

cpitclaudel: Woops, sorry for missing the original message (and for the confusion!). Yes, it's great :)

PR #1385, opened 2018-01-15
Add cargo clippy

fmdkdd:
This adds `clippy` as another checker for rust.  I've picked up #1156, where most of the heavy-lifting had already been done by @dzamlo.

I've tweaked the rust parser to handle a corner case with suggested replacement by Clippy.

Adding a third checker on the same `syntax-checker::` line in `doc/languages.rst` broke the tests, as the regexp did not handle this case.  I've extended it for 3 lines, but... I'm sure there is a more definitive solution waiting to be found ;)

The branch is based on #1381 and #1289, so we should probably merge these before this one.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -238,6 +238,7 @@ attention to case differences."
     ruby-jruby
     rust-cargo
     rust
+    rust-clippy
     scala
     scala-scalastyle
     scheme-chicken
@@ -4711,7 +4712,7 @@ the syntax checker definition, if the variable is nil."
         (flycheck-checker-default-executable checker))))
 
 (defun flycheck-find-checker-executable (checker)
-  "Get the full path of the executbale of CHECKER.
+  "Get the full path of the executable of CHECKER.
 
 Return the full absolute path to the executable of CHECKER, or
 nil if the executable does not exist."
@@ -5786,7 +5787,7 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
     ;; Turn each span into a flycheck error
     (dolist (span spans)
       (let-alist span
-        ;; Children lack any filename/line/column information, so we use
+        ;; Children may not have filename/line/column information, so we use
         ;; those from the primary span
         (when .is_primary
           (setq primary-filename .file_name
@@ -5813,16 +5814,27 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
          errors)))
 
     ;; Then we turn children messages into flycheck errors pointing to the
-    ;; location of the primary span.  According to the format, children
-    ;; may contain spans, but they do not seem to use them in practice.
+    ;; location of the primary span.
     (dolist (child children)
       (let-alist child
         (push
          (flycheck-error-new-at
-          primary-line
-          primary-column
+          ;; Use the line/column from the first span if there is one, or
+          ;; fallback to the line/column information from the primary span of
+          ;; the diagnostic.
+          (or (cdr (assq 'line_start (car .spans)))
+              primary-line)
+          (or (cdr (assq 'column_start (car .spans)))
+              primary-column)
           'info
-          .message
+          ;; Messages from `cargo clippy' may suggest replacement code.  In
+          ;; these cases, the `message' field itself is an unhelpful `try' or
+          ;; `change this to'.  We add the `suggested_replacement' field in
+          ;; these cases.
+          (-if-let (replacement
+                    (cdr (assq 'suggested_replacement (car .spans))))
+              (format "%s: `%s`" .message replacement)
+            .message)
           :id error-code
           :checker checker
           :buffer buffer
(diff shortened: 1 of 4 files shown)


test/specs/test-documentation.el
```diff
@@ -40,12 +41,15 @@
...
+          (cl-pushnew (intern match) matches))
+        (-when-let (match (match-string 3))
           (cl-pushnew (intern match) matches))))
     matches))
 
 (defconst flycheck/checker-re
   (rx bol "   .. syntax-checker:: " (group (1+ nonl)) "\n"
-      (? "                       " (group (1+ nonl)) eol)))
+      (? "                       " (group (1+ nonl)) "\n")
+      (? "                       " (group (1+ nonl))) eol))
```
cpitclaudel: As an alternative, would replacing the original `?` with a `*` work?

fmdkdd: The regexp would still match, but the trouble is in extracting the result with `match-string`.  The group in the start `*` will contain only the /last/ match (here, `rust-clippy`).

cpitclaudel: Ah, yes, of course. Then a single `?` with a `+` *inside* the capturing group? The calling code would have to then further split the match (string-split on spaces)

cpitclaudel: I think this is just fine as it is, though — we can deal with this problem later, if/when we need an extra addition

fmdkdd:
Rebased on master, now that both previous PR have been merged.

flycheck.el
```diff
@@ -5813,16 +5814,30 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
...
     (dolist (child children)
       (let-alist child
         (push
          (flycheck-error-new-at
-          primary-line
-          primary-column
+          ;; Use the line/column from the first span if there is one, or
+          ;; fallback to the line/column information from the primary span of
+          ;; the diagnostic.
+          (or (and (car .spans)
```
cpitclaudel: I wonder if this `and` is needed: if `(car .spans)` is `nil`, then `assq` will return nil as well.
Hence maybe `(let-alist (car .spans) .column_start)` works?

flycheck.el
```diff
@@ -5813,16 +5814,30 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
...
+                   (cdr (assq 'line_start (car .spans))))
+              primary-line)
+          (or (and (car .spans)
+                   (cdr (assq 'column_start (car .spans))))
```
cpitclaudel: Same question/suggestion here

flycheck.el
```diff
@@ -5813,16 +5814,30 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
...
+              primary-column)
           'info
-          .message
+          ;; Messages from `cargo clippy` may suggest replacement code.  In
+          ;; these cases, the `message` field itself is an unhelpful `try` or
+          ;; `change this to`.  We add the `suggested_replacement` field in
+          ;; these cases.
+          (-if-let* ((first-span (car .spans))
+                     (replacement
+                      (cdr (assq 'suggested_replacement first-span))))
```
cpitclaudel: and here :)

flycheck.el
```diff
@@ -5813,16 +5814,30 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
...
+              (format "%s: `%s`" .message replacement)
```
cpitclaudel: It might be better to use `` `…' `` instead of `` `…` `` here?

flycheck.el
```diff
@@ -9375,6 +9390,24 @@ Relative paths are relative to the file being checked."
                  (flycheck-error-message err)))
               errors))
 
+(defun flycheck-rust-manifest-directory ()
+  "Return the neaerst directory holding the Cargo manifest.
```
cpitclaudel: typo `neaerst`

flycheck.el
```diff
@@ -9466,6 +9492,36 @@ This syntax checker needs Rust 1.7 or newer.  See URL
   :modes rust-mode
   :predicate flycheck-buffer-saved-p)
 
+(flycheck-define-checker rust-clippy
+  "A Rust syntax checker using clippy.
+
+See URL `https://github.com/rust-lang-nursery/rust-clippy'."
+  :command ("cargo" "\+nightly" "clippy" "--message-format=json")
```
cpitclaudel: That `\` looks suspicious

flycheck.el
fmdkdd: We can't nest `let-alist` without running into byte-compile warnings.  But we can certainly get rid of the `and` since `nil` is safe.

After programming in stricter languages than Elisp, you tend to put guards defensively ;)

flycheck.el
fmdkdd: I believe `rustc` uses the Markdown convention, so it might be more coherent to follow that in the displayed message.  It's just text, so it doesn't really matter.  Unless someone tries to prettify that down the road.

flycheck.el
fmdkdd: It does!  But the command didn't work without it.

cpitclaudel: Huh.  That *is* strange.  AFAICT this `\` doesn't do anything: it escapes a non-special character, which is the same as not escaping it (I mean that `(string= "\+test" "+test")` evaluates to `t`).  Am I missing something?

flycheck.el
cpitclaudel: That's what I'm afraid about (prettification).  I think popups do get prettified, though arguably it's a bug on the Emacs side.

flycheck.el
cpitclaudel: ^^

cpitclaudel: Which warnings do you get? I don't get any for the following in Emacs 24.5, nor on master:

```elisp
;;; test --- test
;;; Commentary:
;;; Code:

(require 'let-alist)

(defun ~/access (x)
  "X."
  (let-alist x
    (let-alist (car .x) .xx)))

(~/access '((x . (((xx . 1) (yy . 2)) ((xx . 4) (yy . 5))))))

(provide 'test)
;;; test.el ends here
```

flycheck.el
fmdkdd: Hmm.  I could not reproduce the issue on this machine, so I removed the `\`.  Hopefully it was a fluke.

PR #1388, opened 2018-01-19
Testing Emacs pretest on Travis

fmdkdd:
So I was trying to test Flycheck under the [latest pretest](https://lists.gnu.org/archive/html/info-gnu-emacs/2017-10/msg00008.html) on Travis.

Turns out, we are hitting a bug in Cask: cask/cask/issues/373.  This is also why the snapshots tests have been failing for a while on Travis.

The thing is, Cask is looking increasingly unmaintained.   I'll open a separate issue to discuss our options and leave this PR for when this is resolved.


cpitclaudel:
IIUC, the issue that we're running into is that somehow signature verification for packages fails on Travis. Is that right?

fmdkdd:
Yes, the signature fails, but only on Emacs 26.

cpitclaudel:
https://emacs.stackexchange.com/questions/37805/m-x-list-package-fails-with-error-certificate-could-not-be-verified

fmdkdd:
That thread does not include a solution, right?

I've tried using HTTP only, but somehow, in `cask install`, something is still pulling from HTTPS source.

cpitclaudel:
No, but it gave rise to a thread on Emacs devel (that's how I found it).
The emacs-devel thread mentions that upgrading gnutls on the CI machine works, but it means we need to build it from source.

fmdkdd:
After much hassle, I'm happy to report we are building fine on Emacs 26.0.91 (and snapshot, but they should be pretty close).

The short story is that using a fresher GnuTLS works, but there is another showstopper with Cask.  
The long story is in the commit message.

fmdkdd:
@cpitclaudel Can we merge this?  I'm not quite happy with having to build GnuTLS, but unless Travis updates its images, there's not much else we can do as I understand it.

fmdkdd:
Ah, hold on, it seems the certificate issue has been fixed.  I'm testing that.

fmdkdd:
Okay so the certificate issue might have been fixed, but we still need a recent GnuTLS.

I've pushed that change to flycheck/emacs-travis actually, since it makes more sense to do that there.

It should be good to take a look now.

test/flycheck-test.el
```diff
@@ -3059,11 +3059,16 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
...
+      ((version< emacs-version "25")
+       '(0 nil warning "`this-function-is-not-declared' was defined in this-file-does-not-exist.el: file not found"
+           :checker emacs-lisp))
+      ((version< emacs-version "26")
        '(9 1 warning "`this-function-is-not-declared' was defined in this-file-does-not-exist.el: file not found"
-           :checker emacs-lisp)))))
+           :checker emacs-lisp))
+      (t
+       '(9 nil warning "`this-function-is-not-declared' was defined in this-file-does-not-exist.el: file not found"
+           :checker emacs-lisp))))))
```
cpitclaudel: 26 has less information that 25.1?

fmdkdd: Looks like it.  I was surprised by that as well.  Maybe an oversight in a larger rewrite of checkdoc?

fmdkdd:
> Thanks a lot for working on this.

It's especially dreadful work, waiting for builds to finish just to see if a minor change makes a difference.  But it's so useful!

PR #1160, opened 2016-11-13
Add a VHDL checker using GHDL

dzamlo:

PR #1399, opened 2018-02-10
Add a VHDL checker using GHDL

fmdkdd:
Rebased #1160.  This was already reviewed.  I've just updated the version numbers.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -262,6 +262,7 @@ attention to case differences."
     texinfo
     typescript-tslint
     verilog-verilator
+    vhdl-ghdl
     xml-xmlstarlet
     xml-xmllint
     yaml-jsyaml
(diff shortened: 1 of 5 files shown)


cpitclaudel:
Super. Should we just merge?

fmdkdd:
@cpitclaudel Sure.  But if you want to take a second look, that's fine by me ;)

flycheck.el
```diff
@@ -10042,6 +10043,43 @@ See URL `https://www.veripool.org/wiki/verilator'."
...
+
+(flycheck-define-checker vhdl-ghdl
+  "A VHDL syntax checker using GHDL.
+
+See URL `http://ghdl.free.fr/'."
+  :command ("ghdl"
+            "-s" ; only do the syntax checking
+            (option "--std=" flycheck-ghdl-language-standard concat)
+            (option "--workdir=" flycheck-ghdl-workdir concat)
+            source)
```
cpitclaudel: We could consider asking for stdin support at https://github.com/ghdl/ghdl.

fmdkdd: I'm not sure how that fits for their use case.  The `-s` command can take multiple files, so they probably would have to do a special case for checking a single file.

Issue #527, opened 2014-11-26
Start and end position of an error


katspaugh:
Hi Sebastian,

I'm very glad you've added the ability to define checkers as generic functions. I've made [a checker for tern-lint](https://gist.github.com/katspaugh/7d646074e547cf59dd6c) based on your flycheck-ocaml example.

The thing is, tern-lint provides error locations as start and end positions, since it uses a JavaScript parser. So I wonder if passing these positions to flycheck could allow it to highlight symbols more accurately then `bounds-of-thing-at-point` (which I assume to be the only way in flycheck to highlight symbols)?

To illustrate, here's how flycheck-tern-lint displays errors in a test file:

![Imgur](http://i.imgur.com/FCES1BQ.png)

For comparison, here's what I get when [using start and end positions directly](https://github.com/katspaugh/tern-lint.el/blob/master/tern-lint.el#L40):

![Imgur](http://i.imgur.com/JsrmcHY.png)

jbaum98:
What's the status of this? I'm working on adding support for Java and I currently have javac giving me file positions for the start and end of the error, and I'd like to put that into flycheck for the most helpful highlighting.

Is this in progress? If not, is there somebody who can give me some direction on what needs to be done? I'd be happy to work on this myself.

cpitclaudel:
Hi @jbaum98!

> What's the status of this? 

I don't think any of us is working on this ATM.  lunaryorn stepped down as maintainer (@fmdkdd and I replaced him).  But! We're still very interested in getting this feature.  It's just that no-one has found the time.

> I'm working on adding support for Java

That would be great!

>is there somebody who can give me some direction on what needs to be done? I'd be happy to work on this myself.

That would be awesome.  It shouldn't be too too hard, but it needs care.  The idea I think is to define a range structure and use that instead of the line/column pairs that we currently have in errors.  Then the code that determines overlay bounds would be moved up to set proper end bounds in cases where the checker returns a single point instead of a range, and the overlay-creation code would be simplified to just use the bounds present in the code.

This would be great to have.  Feel free to open a PR early for code review/discussion as you work on this! :) We'll be very happy to help.

jbaum98:
I've started work on this a little. What should happen to the current highlighting modes? It seems like if we have start and end positions it is reasonable to use as much specificity as the tool give us.

For example:
* If the tool give us just a line (or range of lines), we highlight the whole line(s).
* If the tool gives us a line and a column, we highlight that single character?
[...]

Issue #774, opened 2015-10-29
Provide Flycheck Annotation Mode


swsnr:
We should extract the functionality to display and list errors into a separate library and mode, to allow package which have their own error checking (Ensime, Merlin, Irony, etc.) to still use Flycheck for displaying their errors.

Initially, we'd implement `flycheck-annotation-mode`, which lets packages add annotations to the current buffer, features a list of annotations, navigation and annotation message display.  Essentially it'd provide everything related to the `flycheck-error` struct, under the new name `flycheck-annotation`.  For some time we should include this mode in addition to the `flycheck-error` code in Flycheck, to let Ensime and similar packages test the feature.

Ultimately Flycheck itself would use Flycheck Annotation Mode.

This breaks backwards compatibility, so it's a necessary prerequisite for Flycheck 1.0.  Will also address #540 and #527.  We'll use [buttercup](https://github.com/jorgenschaefer/emacs-buttercup) for the tests of this mode.

lunaryorn:
Inviting @fommil and @sarcasm

fommil:
:+1:

fommil:
@aemoncannon @edani

edani:
That'd be great!

Sarcasm:
I'm sure an annotation framework would be useful to Emacs, I could find some uses for it, I think this kind of modularity would be nice as other projects can benefit from it, I sometimes I feel frustrated by the UI options offered by Emacs, flycheck has a nice UI so I'm all for it.

TBH, for irony what I would need from flycheck, is to have more opportunities to update the errors.
What would work is to have a polling mechanism/periodic refresh, flycheck could ask me every now and then "do you want to update the errors?". Irony can know that when the header file has changed, the source file may need to be rechecked for example, I would just like flycheck to ask me.

In addition to polling, I would love to be able tell flycheck directly, "hey, I have some fresh stuff, please update the error list when you see fit (soon/now)". This happens for example when a completion popup from company is on, I want to display in a specific manner.

I think this could be 2 new values in `flycheck-check-syntax-automatically` and for irony syntax checking that would be enough.

lunaryorn:
@Sarcasm I'd rather like to get rid of the indirection through Flycheck Mode and custom syntax checkers.  That's a very baroque API for just showing errors.

[...]

PR #1400, opened 2018-02-12
Add ability to set end column and line on error

jbaum98:
Addresses #527. Adds the ability to specify an error using buffer offsets.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1224,17 +1224,43 @@ spliced into the resulting list."
               (t (error "Invalid result type for option: %S" result)))))))
     (apply #'append (seq-map prepend items))))
 
+(defun flycheck-line-column-at (pos)
+  "Move to POS and return the current line and column numbers."
+  (goto-char pos)
+  (cons (line-number-at-pos)
+        ;; `current-column' measures character widths, not just offsets
+        (- (point) (point-at-bol))))
+
+(defmacro flycheck-with-buffer (buf &rest forms)
+  "Switch to buffer BUF and evaluate FORMS.
+
+If BUF is not live, FORMS are not evaluated."
+  (declare (indent 1) (debug t))
+  `(when (buffer-live-p ,buf)
+     (with-current-buffer ,buf
+       ,@forms)))
+
+(defmacro flycheck-with-widened-buffer (buf &rest forms)
+  "Switch to buffer BUF, widen, and evaluate FORMS.
+
+If BUF is not live, FORMS are not evaluated.  Position and
+restrictions are restored after running FORMS."
+  (declare (indent 1) (debug t))
+  `(flycheck-with-buffer ,buf
+     (save-excursion
+       (save-restriction
+         (widen)
+         ,@forms))))
+
 (defun flycheck-find-in-buffer (pattern)
   "Find PATTERN in the current buffer.
 
 Return the result of the first matching group of PATTERN, or nil,
 if PATTERN did not match."
-  (save-restriction
-    (widen)
-    (save-excursion
-      (goto-char (point-min))
-      (when (re-search-forward pattern nil 'no-error)
-        (match-string-no-properties 1)))))
+  (flycheck-with-widened-buffer (current-buffer)
+    (goto-char (point-min))
+    (when (re-search-forward pattern nil 'no-error)
+      (match-string-no-properties 1))))
 
 (defun flycheck-buffer-empty-p (&optional buffer)
   "Whether a BUFFER is empty.
@@ -2921,15 +2947,79 @@ variables of Flycheck."
 
 
 ;;; Errors from syntax checks
-(cl-defstruct (flycheck-error
-               (:constructor flycheck-error-new)
-               (:constructor flycheck-error-new-at
-                             (line column
-                                   &optional level message
-                                   &key checker id group
-                                   (filename (buffer-file-name))
-                                   (buffer (current-buffer)))))
+(cl-defstruct (flycheck-coords
+               (:constructor flycheck-coords-new
+                             (start-line start-column end-line end-column)))
+  "Structure representing error locations using line and column numbers.
+
+For compatibility with external tools and unlike Emacs
+itself (e.g. in Compile Mode) Flycheck uses _1-based_
+columns: The first character on a line is column 1.
+
+Occasionally some tools try to proactively adapt to Emacs
+and emit 0-based columns automatically.  In these cases, the
+columns must be adjusted for Flycheck, see
+`flycheck-increment-error-columns'.
+
+Slots:
+
+`start-line'
+    The line on which the error begins.  Nil is the same as 0.
+
+`start-column'
+    The column at which the error begins.  If nil, the error is assumed
+    to cover the entire line.
+
+`end-line'
+    The line on which the error ends.  If nil, this is assumed to be
+    equal to `start-line'.
+
+`end-column'
+    The column at which the error ends.  If nil, this is inferred by
+    later code according to `flycheck-highlighting-mode'."
+  start-line start-column end-line end-column)
+
+(cl-defstruct
+    (flycheck-error
+     (:constructor nil)
+     (:constructor
+      flycheck-error-new
+      (&key
+       line column ;; Backwards compatibility
+       end-line end-column buffer checker filename message level id group
+       &aux
+       (-region nil)
+       (-coordinates (flycheck-coords-new line column end-line end-column))))
+     (:constructor ;; Legacy constructor
+      flycheck-error-new-at
+      (line
+       column
+       &optional level message
+       &key checker id group end-line end-column
+       (filename (buffer-file-name)) (buffer (current-buffer))
+       &aux (-region nil)
+       (-coordinates (flycheck-coords-new line column end-line end-column))))
+     (:constructor
+      flycheck-error-new-with-coords
+      (line
+       column
+       end-line
+       end-column
+       &optional level message
+       &key checker id group
+       (filename (buffer-file-name)) (buffer (current-buffer))
+       &aux (-region nil)
+       (-coordinates (flycheck-coords-new line column end-line end-column))))
+     (:constructor
+      flycheck-error-new-with-region
+      (start-pos
+       end-pos
+       &optional level message
+       &key checker id group
+       (filename (buffer-file-name)) (buffer (current-buffer))
+       &aux (-region (cons start-pos end-pos)) (-coordinates nil))))
   "Structure representing an error reported by a syntax checker.
+
 Slots:
 
 `buffer'
@@ -2941,20 +3031,20 @@ Slots:
 `filename'
      The file name the error refers to, as string.
 
-`line'
-     The line number the error refers to, as number.
+`-coordinates'
+     The boundaries of this error as a `flycheck-coords' record.
+     This field may be nil or partially populated; use the
+     `flycheck-error-coordinates' wrapper function to access a
+     consistent version of it, or use the individual accessor functions
+     `flycheck-error-start-line', `flycheck-error-start-column',
+     `flycheck-error-end-line', `flycheck-error-end-column' (all of these are
+     `setf'-able, as long as `-region' is unset).
 
-`column' (optional)
-     The column number the error refers to, as number.
-
-     For compatibility with external tools and unlike Emacs
-     itself (e.g. in Compile Mode) Flycheck uses _1-based_
-     columns: The first character on a line is column 1.
-
-     Occasionally some tools try to proactively adapt to Emacs
-     and emit 0-based columns automatically.  In these cases, the
-     columns must be adjusted for Flycheck, see
-     `flycheck-increment-error-columns'.
+`-region'
+     The boundaries of this error as a cons of two points (beginning and
+     end of this error).  This field may be nil; use the
+     `flycheck-error-region' wrapper function to access a consistent
+     version of it.  If set, this takes precedence over `-coordinates'.
 
 `message' (optional)
      The error message as a string, if any.
@@ -2965,7 +3055,7 @@ Slots:
 `id' (optional)
      An ID identifying the kind of error.
 
-`group` (optional)
+`group' (optional)
      A symbol identifying the group the error belongs to.
 
      Some tools will emit multiple errors that relate to the same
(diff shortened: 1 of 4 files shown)


jbaum98:
Another thing to add would be special regexes to make it easy to use these when defining a generic checker. Not sure how it's already implemented for line and column right now.

cpitclaudel:
Thanks, this looks very exciting :) I'm hoping to have time to review this within the week, but it may have to wait until next weekend. Hopefully @fmdkdd or another team member can beat me to it ^^

cpitclaudel:
Ouch. I said before the end of the week two months ago.
Will do today. Thanks for the ping @fmdkdd 
Generally, feel free to ping me much earlier @jbaum98  :)

flycheck.el
```diff
@@ -3035,19 +3122,26 @@ ERR is a Flycheck error.  MODE may be one of the following symbols:
...
+                 (pcase mode
+                   (`columns (or (flycheck-error-column-region err)
+                                 (flycheck-error-line-region err)))
+                   (`symbols (or (flycheck-error-thing-region 'symbol err)
+                                 (flycheck-error-region-for-mode err 'columns)))
+                   (`sexps (or (flycheck-error-thing-region 'sexp err)
+                               (flycheck-error-region-for-mode err 'columns)))
+                   (`lines (flycheck-error-line-region err))
+                   (_ (error "Invalid mode %S" mode))))))
+        (setf (flycheck-error-region err) region))))
```
cpitclaudel: This is a bit of a change in behavior — it's legit to cache the region when we have a all 4 fields (beginning and end of line and column), but it cases like symbols or sexp, we don't currently cache the result.

So maybe we could make flycheck-error-exact-region set the `region` field (and return it), and there would be no more need for the final `setf`.

flycheck.el
```diff
@@ -5969,7 +6075,12 @@ otherwise."
          :checker checker
          :filename (if (or (null filename) (string-empty-p filename))
                        (buffer-file-name)
-                     filename))))))
+                     filename)
+         :end-line (flycheck-string-to-number-safe end-line)
+         :end-column (flycheck-string-to-number-safe end-column)
+         :region (when (or region-start region-end)
+                   (cons (flycheck-string-to-number-safe region-start)
+                         (flycheck-string-to-number-safe region-end))))))))
```
cpitclaudel: Should we add region-start and region-end patterns if we don't have use cases for them yet? I wouldn't expect many CLI tools that only provide a text-based interface to provide regions.

flycheck.el
```diff
@@ -2930,7 +2955,35 @@ Slots:
...
+            (goto-char (car (flycheck-error-region err)))
+            (setf (flycheck-error-start-line err) (line-number-at-pos))
+            (setf (flycheck-error-start-column err) (current-column))
+            (goto-char (cdr (flycheck-error-region err)))
+            (setf (flycheck-error-end-line err) (line-number-at-pos))
+            (setf (flycheck-error-end-column err) (current-column)))))))
+
+(defun flycheck-error-line (err)
+  (flycheck-error-calc-line-col-from-region err)
+  (flycheck-error-start-line err))
```
cpitclaudel: Shouldn't we have accessors of this style for all 4 fields?

jbaum98:
Thanks for the comments, I'll try and respond this week! I would say that the reason I lazily compute everything is because that's what we discussed. My instinct is to precompute everything on error construction because it simplifies the logic by ruling out invalid states, but you thought from an efficiency standpoint the lazy approach was better.

cpitclaudel:
>  I would say that the reason I lazily compute everything is because that's what we discussed.

Yeah, I expected that :)  Confusingly, I can't find traces of that previous conversation here on Github, probably because it was attached to old commits? I'm not rejecting that approach, just trying to refine it :)

> but you thought from an efficiency standpoint the lazy approach was better.

Right, I think I remember.  Reading through the code, I also think there's one thing that (might) force us to be lazy, namely `flycheck-highlighting-mode`.  Changing that variable and redisplaying should cause overlay bounds to be recomputed.  The real question is, do we ever redisplay without fetching a fresh set of errors from the checker? If not, then that point is moot. I think we likely don't and even if we do it's not clear that it matters very much.

Another factor is lazy overlay display (only adding overlays to the visible portion of the buffer).  We don't do it currently; we might do it in the future.  My natural tendency is to ignore that issue entirely and design the current patch for what we currently have.

So, as the code is currently, lazy-computing lines and columns *will* save time for checkers that return a region, but lazy-computing the region won't save time in any case, because we always need a region for overlays.

Amusingly, though, *caching* the region after computing it also doesn't buy us anything, because we only need it once, to create an overlay.  Conversely, caching lines and columns is worth it, because we can display the error list multiple times.

I also think your concern about consistency (ruling out invalid states) is valid.
  
This suggests to me that the right way to think about this is that there are two types of locations: regions, and line-column pairs.  We need to convert from line-column pairs to regions to display overlays, and we need to convert from regions to lc pairs to display the error list.  Additionally, lc pairs can be partial (a checker can give us only a beginning line and a beginning column, and no end-line and end-column, or even just a line).

So, what about the following: let's give errors two fields, coordinates (two line-column pairs) and region.  We then maintain two invariants: `region` is never nil (it's populated at construction time, possibly eagerly).  `coordinates` is all-or-nothing: either full (all 4 fields non-nil) or nil. And we add a single accessor for `coordinates` that computes the coordinates from region if needed, and caches the result.  

A nice property is that for checkers that return partial coordinates, computing the region will also require filling the missing coordinate bits, so really we'll only hit the lazy case when the checker returned a region.

I think this is a very small change, and I think it'll make things clearer (and the invariant is fairly easy to state, too).  What do you think @jbaum98 ?

jbaum98:
I really like that idea, I think that makes a lot of sense. I'll make the change and the other details sometime this week. Thanks for the thoughtful feedback!

jbaum98:
It seems like right now there are actually two different uses for the error region. The first is to add the overlay and is exposed through `flycheck-error-region-for-mode`. The second is to go to the error and is exposed through `flycheck-error-pos`. The difference is that `flycheck-error-region-for-mode` respects `flycheck-error-highlighting-mode` and `flycheck-error-pos` does not.

What we discussed above includes always precomputing the error region, which means that `flycheck-error-region-for-mode` would just lookup the precomputed region in the error, and we would take into account the highlighting mode at the time we construct the error. To me, it seems like `flycheck-error-pos` should also refer to this precomputed region, but that would involve a change in behavior in the case where we don't have an exact region and use the highlighting mode because that would cause the `flycheck-error-pos` region to use the highlighting mode as well.

It seems like the old behavior is kind of strange to begin with, so I think we should unify the two and just have a single region we use for overlaying and jumping. Is there a reason we should worry about this change in behavior though?

flycheck.el
```diff
@@ -5793,11 +5887,13 @@ https://github.com/rust-lang/rust/blob/master/src/libsyntax/json.rs#L67-L139"
...
                 primary-line .line_start
-                primary-column .column_start))
+                primary-column .column_start
+                primary-line-end .line_end
+                primary-col-end .column_end))
         (push
          (flycheck-error-new-at
           .line_start
-          .column_start
+          (- .column_start 1)
```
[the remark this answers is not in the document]
jbaum98: I think this was an existing bug that nobody caught because the highlighting mode meant it still highlighted the correct region.

flycheck.el
jbaum98: I only added this to make sure I wasn't breaking the "public" API of the error struct.

jbaum98:
I think I've got it working. I didn't update the documentation because I wasn't sure where to put what, so let me know about that. I also took away the regexes and rust checker update, but we can add that after.

flycheck.el
```diff
@@ -2878,13 +2878,39 @@ variables of Flycheck."
...
+  start-line end-line start-col end-col)
+
+(cl-defstruct (flycheck-region
+               (:constructor flycheck-region-new (start-pos end-pos)))
+  "Structure representing the location of an error as byte-offsets from the start of a file."
+  start-pos end-pos)
+
+(cl-defun flycheck-region-from-cons ((start-pos . end-pos))
+  "Create a region from a cons cell of `(START-POS . END-POS)`."
+  (flycheck-region-new start-pos end-pos))
```
cpitclaudel: I think the usual representation of ranges in Emacs is a simple cons.  Unless we need something fancier, it might be better to stick with that?

flycheck.el
```diff
@@ -2878,13 +2878,39 @@ variables of Flycheck."
 
 
 ;;; Errors from syntax checks
+(cl-defstruct (flycheck-coords
+               (:constructor flycheck-coords-new (start-line end-line start-col end-col)))
+  "Structure representing the location of an error as lines and columns."
+  start-line end-line start-col end-col)
```
cpitclaudel: I'd call this flycheck-error-coords :)

flycheck.el
```diff
@@ -2930,31 +2956,144 @@ Slots:
...
-  buffer checker filename line column message level id group)
+  buffer checker filename coords region message level id group)
+
+; Compute and cache coords if nil
+(advice-add 'flycheck-error-coords :after-until
+            (lambda (err)
+              (let ((coords (flycheck-region-to-coords (flycheck-error-region err)
+                                                       (flycheck-error-buffer err))))
+                (setf (flycheck-error-coords err) coords)
+                coords)))
```
cpitclaudel: Advice is typically reserved for user code, or desperate situations; let's find a different way to do this.

flycheck.el
```diff
@@ -2930,31 +2956,144 @@ Slots:
...
+(defun flycheck-error-column (err)
+  "Return the column of an errror ERR.
+
+Defined for compatibility."
+  (flycheck-coords-start-col (flycheck-error-coords err)))
+
+(gv-define-simple-setter flycheck-error-column
+                         (lambda (err x)
+                           (setf (flycheck-coords-start-col (flycheck-error-coords err)) x)
+                           (setf (flycheck-coords-end-col   (flycheck-error-coords err)) x)))
```
cpitclaudel: We should update the error region here, right?

flycheck.el
```diff
@@ -2930,31 +2956,144 @@ Slots:
...
+          (goto-char (flycheck-region-end-pos region))
+          (let ((end-line (line-number-at-pos))
+                (end-col  (current-column)))
+            (flycheck-coords-new start-line end-line start-col end-col)))))))
+
+(defun flycheck-sparse-coords-to-region (start-line start-col end-line end-col buf)
+  "Convert coordinates to a full set of coordinates and a set of buffer-offsets.
+
+Returns a cons-cell of `(FULL-COORDS . REGION)` where FULL-COORDS
+is a set of coordinates where non are nil, and REGION is a
```
cpitclaudel: Typo: `none`

flycheck.el
jbaum98: Will this make everything work with error filters, because the way they work is by modifying errors through `setf`? Is this maybe a drawback to our new practice of precomputing the region; ie. when we change the line and column we need to recompute?

jbaum98: Also, should we be exposing an API for modifying `start-line` and `end-line` etc that will also update the region? Right now the only way for them to do this is to modify it within the coords, which won't update the error region. Should there be `flycheck-error-start-line` getters and setters to create the illusion that the `start-line` and others are stored directly on the error?

cpitclaudel:
Hey @jbaum98,

I've finally managed to set aside enough time to look at this seriously, including writing new tests and adjusting existing ones, and updating docstrings.

There were a few things left to solve, which we had discussed:

- Aggressively computing regions meant doing redundant work (many checkers apply a filter that changes error columns after parsing), and handling partly constructed errors (some checkers apply a filter to add error lines when they are missing).

- Storing computed regions was not entirely correct, since the actual region often depends on the highlighting mode.

- We wanted to expose a nicer API than having to fetch the coordinate object and then reading from it.

- We hadn't updated the existing tests and made sure that they still passed.

I've pushed two commits on top of your branch to address these points (it felt only fair that I should do that work, given how long it's taken me to get back to this); see https://github.com/flycheck/flycheck/tree/1400-error-ranges .  I have not tested it much beyond ensuring that the unit tests pass.

Would you like to review these two patches? I could open a separate PR with your commits plus mine, or more simply you could pull these two commits into your branch and review them here.

(Btw, I changed flycheck-error-coords back to flycheck-coords; the former name made everything too verbose. Sorry for asking you to change it previously).

jbaum98:
Great! I'll pull them and review them here when I get the chance. Probably sometime this week after Monday.

cpitclaudel:
Thanks a lot.

jbaum98:
@cpitclaudel So sorry that it took so long. I looked at your changes and they look great. I squashed to make the rebase easier and rebased from master.

cpitclaudel:
Thanks! I have one more small fix to push, and then we'll ask @fmdkdd for a final proofread :) I should have time this weekend or early next week.

fmdkdd:
>  then we'll ask @fmdkdd for a final proofread 

Ahead of that, it would be good to fix the errors reported by the CI (it seems there are formatting errors at the moment).

jbaum98:
@fmdkdd @cpitclaudel What's the status of this? It looks the CI is all good. Is there anything else we need to do before merge?

cpitclaudel:
I didn't forget about this :) Sorry for not posting more updates.

I've been running this as my main Flycheck since our last discussion, but I did run into a subtle issue that I didn't expect. The issue is that this change breaks packages that depend on Flycheck, like flycheck-ocaml and others, until they are recompiled — but package.el doesn't recompile dependent packages when installing an updated dependency.

I described the issue at https://lists.gnu.org/archive/html/emacs-devel/2018-07/msg00436.html .  I haven't found a fully satisfactory answer yet :/

jbaum98:
Ah, I understand. Do we want to consider wrapping all the field accessors in functions so that this doesn't happen again in the future, assuming we find a way to deal with this breaking change once?

fmdkdd:
So, when we added the `group` property to the `flyckeck-error` struct, we broke the bytecode compatibility.  Thankfully, it did not change the order of other fields (it was added last), so it went mostly unnoticed.

How chaotic would it be if we merged the PR now?  Are we talking major breakage of many packages?  Is there a way to approximate how many packages would be affected?  (Using i.e., Github search?)

In any case, the fix is to, at worst, recompile all packages?  We could provide a small Elisp snippet that does that.  Recompiling doesn't take much time, and doesn't have other downsides.

If we can avoid any breakage, then all for the better.  But the fix has to be reasonable, and shouldn't delay this PR into oblivion.

cpitclaudel:
> So, when we added the group property to the flyckeck-error struct, we broke the bytecode compatibility.

:fearful: indeed.
 
> Thankfully, it did not change the order of other fields (it was added last), so it went mostly unnoticed.

Right, adding at the end is a neat trick, as long as we don't have external packages passing errors to us, in which we expect a group to appear.  This trick won't work with positions, though.

> How chaotic would it be if we merged the PR now? Are we talking major breakage of many packages? Is there a way to approximate how many packages would be affected? (Using i.e., Github search?)

Chaotic enough that I noticed it — the package that broke for me was merlin.  My own dafny and fstar packages would break, too.  Any package that accesses an error position would break too.  And any package that passes us errors with positions, which is any external checker that uses a subprocess. 

> In any case, the fix is to, at worst, recompile all packages? We could provide a small Elisp snippet that does that. Recompiling doesn't take much time, and doesn't have other downsides.

Indeed.  The issue is cleanly detecting the issue.

> If we can avoid any breakage, then all for the better. But the fix has to be reasonable, and shouldn't delay this PR into oblivion

Agreed (sorry @jbaum98, it's indeed taking forever!).  My current thinking is that we may be able to add code at the end of Flycheck.el that recompiles packages that depend on it if needed.

But that will only work with package.el.  With other package managers, people will have to trigger a recompile themselves, so we'll likely want to have some way of detecting the problem.

cpitclaudel:
I agree.  But part of me was hoping for a simple fix, hance the discussion on emacs-devel first.

jbaum98:
Could this be a good opportunity to do something related to #774 also? If we're already committed to a  sort of breaking change?

cpitclaudel:
Indeed, probably… but I'd prefer shipping your change soon instead of packaging it into something a lot larger.

Fuco1:
There's a cool function `display-warning` which pops up a buffer and shows... the warning :D  Some packages use it when they do a breaking update to let users know.  `ido-ubiquitious` comes to mind.

Simply put it somewhere on top-level and save into a variable that you already shown it in this session, or, for bonus points, run a predicate that can tell if the problem is present or not (e.g. check the timestamps of the `.elc` files or some such).

This way you can roll out the change and gracefully inform users about the problem.

Wilfred:
@cpitclaudel I see you wrote a fabulous bit of elisp to recompile dependencies of flycheck: https://lists.gnu.org/archive/html/emacs-devel/2018-07/msg01159.html

Could you just do this if an old function is still defined? I'd think a feature test, e.g. `(fboundp 'flycheck-error-region-for-mode)` would meet your needs of knowing that flycheck was old. If you want the recompile to run at-most-once, you could `makunbound` afterwards.

Profpatsch:
Am I correct in that this patch is essentially feature-complete and tested, just not merged yet because of the dependency-problem?

fmdkdd:
@Profpatsch The dependency problem is not small.  If we can avoid breaking setups, we should try our best.

@cpitclaudel Following @Wilfred suggestion, if a known obsolete function exists, then we need to recompile.  If it doesn't, then we don't?  Wouldn't that be a way to force the recompilation of dependencies only once?

Profpatsch:
> The dependency problem is not small. If we can avoid breaking setups, we should try our best.

From what I understand it’s a general problem of Emacs’s statefulness. I’ve had other packages break symbols on updates, usually it fixed itself after a few days and another update. Not a giant problem in practice.

jbaum98:
Just wanted to bump this again. I know we were worried about breaking people's installs. Is that still an issue, and if so, is there anything I can do to make progress on it?

Maybe I never fully understand the issue, but is there a reason why we can't tell people to recompile their bytecode and restart emacs? It's obviously not ideal, but if it seems intractable to preserve bytecode compatibility or to automatically recompile, I don't think it's worth not making these changes.

Seeing as this PR has already waited a while though, I do think it makes sense to batch this with other breaking changes, like something related to #774, as we discussed about a year ago.

fmdkdd:
@jbaum98 Agreed, I would like to merge this as well.

AFAIK, there is no solution that would allow us to recompile all of Flycheck dependencies automatically.  @cpitclaudel's snippet works only for `package.el`, and is still missing the detection part.  Which leaves us with mitigation.

I would be in favor of:

1. Displaying a warning when loading Flycheck, as suggested by @Fuco1.
2. Providing a snippet to recompile Flycheck dependencies (could be in `flycheck.el`, or could be in the manual)
3. Adding instructions to the `Troubleshooting` section of the manual on how to deal with these bytecode incompatibilities issues

For the warning, we could do something like checking if any loaded feature starts with `flycheck-`.

#774 sounds nice, and I'm sure the LSP folks could also benefit from that.  I don't want to delay this PR much further though.

@cpitclaudel How does that sound for you?

cpitclaudel:
I agree with you. Sorry for the low responsivity.
I'm really excited to get this in, and my daily workflow with OCaml would be approximately infinitely better.
And it's terrible to have the great work of @jbaum98 just sit around unused.

Here is one more thought.  Since Flycheck is large, maybe we could very kindly ask @purcell to do a no-change rebuild of Flycheck dependencies after we push our update? (:angel: :bowing_man:)

purcell:
> Here is one more thought. Since Flycheck is large, maybe we could very kindly ask @purcell to do a no-change rebuild of Flycheck dependencies after we push our update?

I might be missing something, but I don't think that would achieve anything, since byte compilation happens in Emacs, not on the MELPA server. There's no way to force an artificially-new version of packages in MELPA, since the version comes solely from the last-modified time of the most-recently-updated packaged file.

cpitclaudel:
> I might be missing something, but I don't think that would achieve anything, since byte compilation happens in Emacs, not on the MELPA server. 

I was thinking that if MELPA marked packages depending on flycheck as updated, it would cause users to redownload them, and hence recompile them.

> There's  no way to force an artificially-new version of packages in MELPA, since  the version comes solely from the last-modified time of the  most-recently-updated packaged file.

Ah, that's too bad.

amosbird:
Does this patch help move cursor to the EOL errors?

image

Will checkers need to be rewritten?

jbaum98:
@amosbird I think the patch does address this issue. I also (if memory serves correctly) think that most checkers are already written to take advantage of columns, and it was the implementation of the primitives on which they relied that was failing to do so. So I think many checkers will get the better behavior "for free".

That being said this code was written a while ago, and I think it ended up being written mostly by @cpitclaudel so he might be able to answer better than me.

kaz-yos:
This will be very useful in flycheck-stan (https://github.com/stan-dev/stan-mode). I’m looking forward to it being merged.

cpitclaudel:
OK, I've refreshed a simplified version of this PR that shouldn't have backwards-compatibility issues; see #1676)

fmdkdd:
That's great!  After merging, that would be a good time to finally make a release.

fmdkdd:
Is there any reason to keep this one open now that #1676 has been merged?

cpitclaudel:
I don't think so :) I'm working on further tweaks (adding error patterns for end-line and end-column and updating checkers) in a separate branch

PR #1404, opened 2018-02-16
Add "-" to phpcs arguments.

hirayama-onagigawa-computing:
On PHP_CodeSniffer (phpcs) 3.x or later, "-" must be added as the last argument to wait for STDIN.

https://github.com/squizlabs/PHP_CodeSniffer/blob/master/package.xml#L2237

On Linux, "-" is not necessary for pipe input.
But on Windows, "-" needs to be added to work properly.

Adding "-" has no effect for Linux systems.
For phpcs 2.9 or older, it is just ignored. (https://github.com/squizlabs/PHP_CodeSniffer/blob/2.9/CodeSniffer/CLI.php#L454)
It seems to be safe.

Thanks in advance.


cpitclaudel:
Looks great, thanks.

PR #1406, opened 2018-02-16
Better error for non writable temporary files

### LINKED ISSUE #1401: source-inplace errors in unwritable directories
When I `find-function` in emacs it pops me into an unwritable directory (`/usr/local/stow/emacs-25.2/share/emacs/25.2/lisp/font-lock.el.gz` on my system).  When flycheck tries to check the elisp source it errors out because the directory is not writable.

I think in such case it should just do nothing (the code is "unfixable" anyway by virtue of being unwritable).

I think this is also a problem for other checkers which use the `source-inplace` but I have not tested it.

I can modify the predicate on the checker but I think a better solution would be to handle this at the source interpretation (i.e. when trying to create the input file).

fmdkdd:
A proposal to fix #1401.

This disables checker that use `source-inplace` when the temporary file that would be created by `temp-file-name` cannot be written.

It also adds a message in `verify-setup`.  Though I'm not quite happy with how it turns out.  There is not a lot of room for verification result messages (and even less so for labels).  Ideally, I would have liked to write a one-line reason ("need writable permissions on `dir` for writing temporary files) for why the checker has been disabled, and link to (or have some clickable thingy that reveals) a longer explanation of source-inplace.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1118,8 +1118,6 @@ Otherwise this function creates a temporary file with
 `flycheck-temp-prefix' and a random suffix.  The path of the file
 is added to `flycheck-temporaries'.
 
-Add the path of the file to `flycheck-temporaries'.
-
 Return the path of the file."
   (let ((tempfile (convert-standard-filename
                    (if filename
@@ -1148,6 +1146,29 @@ Return the path of the file."
         (push tempfile flycheck-temporaries)
         tempfile)
     (flycheck-temp-file-system filename)))
+
+(defun flycheck-temp-directory (checker)
+  "Return the directory where CHECKER writes temporary files.
+
+Return nil if the CHECKER does not write temporary files."
+  (let ((args (flycheck-checker-arguments checker)))
+    (cond
+     ((memq 'source args) temporary-file-directory)
+     ((memq 'source-inplace args)
+      (if buffer-file-name (file-name-directory buffer-file-name)
+        temporary-file-directory))
+     (t nil))))
+
+(defun flycheck-temp-files-writable-p (checker)
+  "Whether CHECKER can write temporary files.
+
+If CHECKER has `source' or `source-inplace' in its `:command',
+return whether flycheck has the permissions to create the
+respective temporary files.
+
+Return t if CHECKER does not use temporary files."
+  (let ((dir (flycheck-temp-directory checker)))
+    (or (not dir) (file-writable-p dir))))
 
 (defun flycheck-save-buffer-to-file (file-name)
   "Save the contents of the current buffer to FILE-NAME."
@@ -4666,6 +4687,7 @@ default `:verify' function of command checkers."
           (plist-put properties :enabled
                      (lambda ()
                        (and (flycheck-find-checker-executable symbol)
+                            (flycheck-temp-files-writable-p symbol)
                             (or (not enabled) (funcall enabled))))))
 
     (apply #'flycheck-define-generic-checker symbol docstring
@@ -5071,7 +5093,13 @@ CHECKER."
             (list (flycheck-verification-result-new
                    :label "configuration file"
                    :message (if path (format "Found at %S" path) "Not found")
-                   :face (if path 'success 'warning))))))))
+                   :face (if path 'success 'warning)))))
+      ,@(when (not (flycheck-temp-files-writable-p checker))
+          (list (flycheck-verification-result-new
+                 :label "temp directory"
+                 :message (format "%s is not writable"
+                                  (flycheck-temp-directory checker))
+                 :face 'error))))))
 
 
 ;;; Process management for command syntax checkers


cpitclaudel:
It actually looks fairly good to me :)

I have two suggestions:

* extract the check to a separate function, maybe `flycheck-source-directory-writable-p CHECKER`.
* Change the message from `no` to "Directory …. is not writable" (and add the error font)

WDYT?

fmdkdd:
Yes and yes.  I've done that, and also checked for temporary files for when `source` is used.  

Actually, "directory is not writable" is slightly misleading: the check will return nil if the temporary *files* cannot be written . It can be because there already exist non-writable files (though very unlikely), or some other reason (not enough space on disk?).

At least, it highlights the most probable cause.  But I'd still like the ability to show more information (like the actual locations of the files that cannot be written) for easier troubleshooting.

Any idea how to provide that extra info?  Make the error a button?  Add some kind of foldable?

cpitclaudel:
> Yes and yes. I've done that, and also checked for temporary files for when source is used.

OK, neat.

> Actually, "directory is not writable" is slightly misleading: the check will return nil if the temporary files cannot be written . It can be because there already exist non-writable files (though very unlikely), or some other reason (not enough space on disk?).

I don't think a lack of space on disk would be detected by your code here, since we don't actually try to write.  Ultimately, I'm fine with not catching such exceptional cases.

> At least, it highlights the most probable cause. But I'd still like the ability to show more information (like the actual locations of the files that cannot be written) for easier troubleshooting. Any idea how to provide that extra info? Make the error a button? Add some kind of foldable?

I'd include the info directly in the message rather than hiding it.  Since the message is only shown in fairly problematic conditions, I wouldn't worry about it being a tad too long.

flycheck.el
```diff
@@ -1149,6 +1147,20 @@ Return the path of the file."
...
+If CHECKER has `source' or `source-inplace' in its `:command',
+return whether flycheck has the permissions to create the
+respective temporary files.
+
+Return t if CHECKER does not use temporary files."
+  (let ((args (flycheck-checker-arguments checker)))
+    (and (or (not (memq 'source args))
+             (file-writable-p (flycheck-temp-file-system buffer-file-name)))
+         (or (not (memq 'source-inplace args))
+             (file-writable-p (flycheck-temp-file-inplace buffer-file-name))))))
```
cpitclaudel: Would this read better like this?

```elisp
(cond
  ((memq 'source args) (file-writable-p …))
  ((memq 'source-inplace args) (file-writable-p …))
  (t t))
```

flycheck.el
```diff
@@ -5071,7 +5084,12 @@ CHECKER."
             (list (flycheck-verification-result-new
                    :label "configuration file"
                    :message (if path (format "Found at %S" path) "Not found")
-                   :face (if path 'success 'warning))))))))
+                   :face (if path 'success 'warning)))))
+      ,@(when (not (flycheck-temp-files-writable-p checker))
+          (list (flycheck-verification-result-new
+                 :label "directory"
+                 :message "is not writable"
+                 :face 'error))))))
```
cpitclaudel: Maybe `temp directory` rather than just directory, and `not writable (<path>)`?

flycheck.el
```diff
@@ -1149,6 +1147,20 @@ Return the path of the file."
...
+return whether flycheck has the permissions to create the
+respective temporary files.
+
+Return t if CHECKER does not use temporary files."
+  (let ((args (flycheck-checker-arguments checker)))
+    (and (or (not (memq 'source args))
+             (file-writable-p (flycheck-temp-file-system buffer-file-name)))
+         (or (not (memq 'source-inplace args))
+             (file-writable-p (flycheck-temp-file-inplace buffer-file-name))))))
+
```
cpitclaudel: One issue with this implementation is that it adds the files to `flycheck-temporaries`.

flycheck.el
fmdkdd: I agree it's not optimal, but some checkers have both `source` and `source-inplace` in their arguments.

flycheck.el
fmdkdd: I'll try that yeah.

flycheck.el
fmdkdd: Ah right.  We could `pop flycheck-temporaries`... but that would be bad.

So I guess we should split the existing functions to have one return the temporary filename that would be written, and the other actually create the file.

However, `make-temp-file` actually *creates* the temporary file, so it's still not quite correct.  We might have to wait for an error when creating the files after all.

cpitclaudel: > So I guess we should split the existing functions to have one return the temporary filename that would be written, and the other actually create the file.

I'm always a bit scared of that, because race conditions.  It probably doesn't matter too much for Flycheck, though.

> However, make-temp-file actually creates the temporary file, so it's still not quite correct. We might have to wait for an error when creating the files after all.

Hm. What about just moving the push into flycheck-temporaries to flycheck-save-buffer-to-temp?

flycheck.el
cpitclaudel: Oh? Hmm. Do we actually have any like that currently? (I couldn't find any) When is that useful?

fmdkdd: Ah you're right, we don't.  I was thinking of #1354 which uses `source-original` and `source`.  Combining `source` and `source-inplace` does not make much sense after all.

flycheck.el
fmdkdd: > Hm. What about just moving the push into flycheck-temporaries to flycheck-save-buffer-to-temp?

That's good I think.  But isn't it better to keep track of these temporaries actually?  At least, if we can clean them at the end of `temporaries-writable-p` that way?

cpitclaudel: Ah, sorry, I didn't understand the make-temp-file comment at first.
… thinks for a while …
Could we just use file-writable-p on the appropriate directory?  A quick look at the C sources suggests that it should work on GNU/Linux systems.

fmdkdd: > Could we just use file-writable-p on the appropriate directory? 

I'll try that, that seems better in the end yes.

fmdkdd:
I've updated to use `file-writable-p` directly on the directory where the temporary file would be created.  Here is what the message looks like:

screen

Let me know what you think.

cpitclaudel:
Looks great

flycheck.el
```diff
@@ -1149,6 +1147,28 @@ Return the path of the file."
...
+(defun flycheck-temp-files-writable-p (checker)
+  "Whether CHECKER can write temporary files.
+
+If CHECKER has `source' or `source-inplace' in its `:command',
+return whether flycheck has the permissions to create the
+respective temporary files.
+
+Return t if CHECKER does not use temporary files."
+  (file-writable-p (flycheck-temp-directory checker)))
```
cpitclaudel: Won't this raise an error if `(flycheck-temp-directory checker)` returns nil?

fmdkdd: Indeed!

So I could add a guard to `flycheck-temp-files-writable-p`, or I could just dispense with this function altogether and just use `flycheck-temp-directory` directly in the couple of places the former is called.

WDYT?

cpitclaudel: I think a guard is fine :)

fmdkdd: Done.

fmdkdd:
Thanks for reviewing!

Fuco1:
Thank you for working on this @fmdkdd , @cpitclaudel :heart: Very neat how it turned out!

Issue #1395, opened 2018-02-03
flycheck for groovy gets stuck, high cpu usage


philiparvidsson:
# How to reproduce:

1. Open Emacs
2. Open .groovy-file
3. Activate groovy-flycheck
4. Look at cpu usage

# Details

Windows 10
Emacs 25.3.1
Latest flycheck
JRE 9 (with java.exe in path, otherwise it doesn't seem to happen!)

# More info

https://github.com/Groovy-Emacs-Modes/groovy-emacs-modes/issues/106

cpitclaudel:
Thanks for the report! And sorry for the trouble. How odd.
I don't run windows, so your debugging help will be much appreciated. Does the same thing happen when you run the checker explicitly with C-c ! C-c? What if you interrupt the check using C-u C-c ! C? Does that return things to normal? What about when you run the checker by hand from the command line? And is it just one file, or any file? If you set M-x toggle-debug-on-quit and then press C-g while emacs is stressing you CPU, what does the backtrace say? And, does Emacs freeze while the CPU usage is high?

philiparvidsson:
* *Does the same thing happen when you run the checker explicitly with C-c ! C-c?*
**This seems to do nothing.**

* *What if you interrupt the check using C-u C-c ! C?*
**This kills the cmd.exe process running at 100% CPU usage, so everything returns to normal (though flycheck obviously doesn't check anything).**

* *What about when you run the checker by hand from the command line?*
**Can you tell me how to do that?**

* *And is it just one file, or any file?*
**Any Groovy source file.**

* *If you set M-x toggle-debug-on-quit and then press C-g while emacs is stressing you CPU, what does the backtrace say?*
**No idea if I did this correct. I did the following:**
  - M-x toggle-debug-on-quit
  - Load Groovy-file (issue occurs)
  - Press C-g

  **Nothing happens except Emacs says "Quit" in the mode line.**

* *And, does Emacs freeze while the CPU usage is high?*
**No, I can still work as I normally would (though flycheck is waiting for the checker process to complete, which never happens)**

cpitclaudel:
Great, thanks a lot! It looks like Groovy is spinning endlessly then.
C-c ! C-c is flycheck-compile; can you try M-x flycheck-compile, and select groovy at the prompt?
This will run the checker in a separate window, and print the command explicitly; then you'll be able to copy that command into a terminal.

philiparvidsson:
[...]

PR #716, opened 2015-08-04
Add checker for groovy programming language.

mbezjak:
groovyc cannot be used here because it cannot be made to stop at syntax
check. Instead use groovy compiler API and compile up to the phase
`CONVERSION`.

References for an idea:

- http://www.groovy-lang.org/mailing-lists.html#nabble-td365810
- https://github.com/groovy/groovy-core/blob/master/src/main/org/codehaus/groovy/control/ErrorCollector.java

Fix #715.

PR #1407, opened 2018-02-17
Check Groovy using standard-input.

### LINKED ISSUE #1396: Groovy checker could use stdin
The small script that we have in the Groovy checker could use 

```java
addSource(java.lang.String name, java.io.InputStream stream)
```

instead of the current call to addSource with a file name.

scolej:
Fixes #1396.

And possibly also provides a workaround for #1395, but I can't verify this till Monday, when I have access to a Windows machine.

Another alternative might be to just use `groovyc`. I know in the original PR which added the Groovy checker ( #716), the author identified that they couldn't use `groovyc` because there was no way to make it stop it syntax checking... but maybe the simplicity of using it is worth the overhead?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7738,19 +7738,17 @@ See URL `http://www.groovy-lang.org'."
   :command ("groovy" "-e"
             "import org.codehaus.groovy.control.*
 
-file = new File(args[0])
 unit = new CompilationUnit()
-unit.addSource(file)
+unit.addSource(\"input\", System.in)
 
 try {
     unit.compile(Phases.CONVERSION)
 } catch (MultipleCompilationErrorsException e) {
     e.errorCollector.write(new PrintWriter(System.out, true), null)
-}
-"
-            source)
+}")
+  :standard-input t
   :error-patterns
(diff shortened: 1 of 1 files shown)


cpitclaudel:
>  the author identified that they couldn't use groovyc because there was no way to make it stop it syntax checking... but maybe the simplicity of using it is worth the overhead?

Maybe, but we've historically tried to avoid this — a full compilation cycle is usually pretty costly.

scolej:
Thanks for merging and thanks for Flycheck! 😃

philiparvidsson:
Unfortunately doesn't seem to fix #1395. :-(

scolej:
> Unfortunately doesn't seem to fix #1395. :-(

No, but I did manage to successfully run the checker on Windows using `C-c ! C-c` which is really strange; ie: I get different behaviour running manually compared to when the checker runs automatically.

(Additionally, the way the checker-process interprets `JAVA_HOME` seems to be different between automatic and manual runs. I had to use backslashes in `JAVA_HOME` to get the auto-checker to start at all (and then to hang indefinitely)).

I'll try to dig around a bit more...

PR #1414, opened 2018-02-25
As discussed in GH-1055. Should our CI have caught the documentation issue?

cpitclaudel:
As discussed in GH-1055. Should our CI have caught the documentation issue?

### DIFF
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -862,6 +862,14 @@ to view the docstring of the syntax checker.  Likewise, you may use
    Flycheck checks Python with `python-flake8` or `python-pylint`, and falls
    back to `python-pycompile` if neither of those is available.
 
+   All Python checkers are invoked indirectly using ``python -c ...`` (rather
+   than a direct call to ``flake8`` or ``pylint``) to make it easier to switch
+   between Python 2 and 3.  For example, you can use ``(setq-local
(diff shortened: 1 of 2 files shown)


doc/languages.rst
```diff
@@ -863,6 +862,14 @@ to view the docstring of the syntax checker.  Likewise, you may use
    Flycheck checks Python with `python-flake8` or `python-pylint`, and falls
    back to `python-pycompile` if neither of those is available.
 
+   All Python checkers are invoked indirectly using ``python -c ...`` (rather
```
fmdkdd: You could also put that in an `important ::` block for better visibility.

fmdkdd:
Thanks, LGTM.

> Should our CI have caught the documentation issue?

You mean the CI should fail on the documentation warnings, like the duplicate links and unknown `ref` for `declare-function`?

I would be okay with that.  I think they were disabled because the manual was a work in progress, and some references were still undefined.

I noticed that we can also ignore links without using inline code syntax (double backticks) by adding them in `doc/conf.py`.  I think the intent is: all referenced flycheck functions should eventually be documented, while other elisp functions may link to other Emacs manuals.  Non-elisp code should probably use the inline code syntax.

cpitclaudel:
Thanks for the review and for pointing out the conf.py trick. I updated accordingly and merged.

PR #1419, opened 2018-03-04
Avoid nil for the default argument of ‘completing-read’.

### LINKED ISSUE #1418: Take completing-read quirk into account
`flycheck-completing-read` doesn't take the quirk described at https://github.com/DarwinAwardWinner/ido-completing-read-plus#why-does-ret-sometimes-not-select-the-first-completion-on-the-list--why-is-there-an-empty-entry-at-the-beginning-of-the-completion-list--what-happened-to-old-style-default-selection into account: it regularly passes a nil default, but a non-nil `require-match`. This causes command such as `flycheck-disable-checker` to use an empty checker symbol by default if `ido-completing-read+` is enabled, directly triggering an error. You can reproduce this by enabling `ido-ubiquitous-mode` and then evaluating something like

```elisp
(read-flycheck-checker "prompt: ")
```

Probably the callers of `flycheck-completing-read` should pass the first element of the collection instead of a nil default.

phst:
This avoids the quirk described at
https://github.com/DarwinAwardWinner/ido-completing-read-plus#why-does-ret-sometimes-not-select-the-first-completion-on-the-list--why-is-there-an-empty-entry-at-the-beginning-of-the-completion-list--what-happened-to-old-style-default-selection.

Fixes #1418

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1364,7 +1364,7 @@ a default on its own."
                                   (flycheck-defined-checkers property))))
          (default (and default (symbol-name default)))
          (input (flycheck-completing-read
-                 prompt candidates default
+                 prompt candidates (or default (car candidates))
                  'read-flycheck-checker-history)))
     (when (string-empty-p input)
       (unless default
@@ -1384,7 +1384,8 @@ to the default levels."
                           (flycheck-error-list-current-errors)))
          (levels-with-defaults (append '(info warning error) levels))
          (uniq-levels (seq-uniq levels-with-defaults))
-         (level (flycheck-completing-read prompt uniq-levels nil)))
+         (level (flycheck-completing-read prompt uniq-levels
+                                          (car uniq-levels))))
     (and (stringp level) (intern level))))
 
 


flycheck.el
```diff
@@ -1384,7 +1384,8 @@ to the default levels."
                           (flycheck-error-list-current-errors)))
          (levels-with-defaults (append '(info warning error) levels))
          (uniq-levels (seq-uniq levels-with-defaults))
-         (level (flycheck-completing-read prompt uniq-levels nil)))
+         (level (flycheck-completing-read prompt uniq-levels
+                                          (car uniq-levels))))
```
cpitclaudel: This look wrong: we do want to offer an empty entry (it corresponds to setting a nil level)

flycheck.el
```diff
@@ -1364,7 +1364,7 @@ a default on its own."
                                   (flycheck-defined-checkers property))))
          (default (and default (symbol-name default)))
          (input (flycheck-completing-read
-                 prompt candidates default
+                 prompt candidates (or default (car candidates))
```
cpitclaudel: The code below explicitly deals with the case of no checker selected case, so do we really need this?

cpitclaudel:
Closing as this patch has gone stale.  If the original concerns can be addressed, I'll be happy to have a new look.

cpitclaudel:
I created a separate PR, #1706, with a different fix

Issue #1397, opened 2018-02-05
rust-cargo munges paths incorrectly for project with nested crates


msherry:
I have noticed that `rust-cargo` fails to produce output for projects with nested crates, and have produced a minimal repro test case.

I created a project structured like this:
```
├── Cargo.toml
├── project
│   └── subproj
│       ├── Cargo.toml
│       └── src
│           └── lib.rs
```
Editing `lib.rs` to contain just the string `xxx` gives no error output, when it should be giving an error message. I believe that flycheck is setting `subproj` to be the working directory, but `cargo` is giving error messages relative to the top-level `Cargo.toml` file.

I placed a debugging `(message)` statement at the end of `flycheck-fill-and-expand-error-file-names`, after filename expansion had been performed -- this was the output:
```
(#s(flycheck-error lib.rs|subproj rust-cargo /Users/msherry/projects/rust-debug/project/subproj/project/subproj/src/lib.rs 1 1 expected one of `!` or `::`, found `<eof>` (expected one of `!` or `::` here) error nil group))
```
Note that the path is given as `/Users/msherry/projects/rust-debug/project/subproj/project/subproj/src/lib.rs`, which has a duplicate `project/subproj/` component -- the full path should be `/Users/msherry/projects/rust-debug/project/subproj/src/lib.rs`.

I have temporarily worked around this by redefining `flycheck-fill-and-expand-error-file-names` as follows, but obviously this is just a local hack:
```
(defun flycheck-fill-and-expand-error-file-names (errors directory)
  "Fill and expand file names in ERRORS relative to DIRECTORY.

Expand all file names of ERRORS against DIRECTORY.  If the file
name of an error is nil fill in the result of function
`buffer-file-name' in the current buffer.

Return ERRORS, modified in-place."
  (seq-do (lambda (err)
            (setf (flycheck-error-filename err)
                  (buffer-file-name)))
          errors)
  errors)
```

`M-x flycheck-verify-setup`:
```
Syntax checkers for buffer lib.rs|subproj in rust-mode:

  rust-cargo
    - may enable:  yes
    - predicate:   t
    - executable:  Found at /Users/msherry/src/client/virtual_env/rust/bin/cargo
    - Cargo.toml:  Found
    - Crate type:  lib
    - Binary name: Not required

  rust
    - may enable: yes
    - predicate:  t
    - executable: Found at /Users/msherry/src/client/virtual_env/rust/bin/rustc

[...]

PR #1422, opened 2018-03-07
Changed flycheck-rust-manifest-directory to use 'cargo metadata'

kngwyu:
This commit fixes https://github.com/flycheck/flycheck/issues/1397.
By the recent change of cargo(https://github.com/rust-lang/cargo/pull/4788), flycheck can't detect error files in multi-crate project of Rust.
So I changed to use `workspace_root` in `cargo metadata` 's outputs, as working directory.
If the user's cargo is a bit old and there's no 'workspace_root', `flycheck-rust-manifest-directory` search a directory with `Cargo.toml`, as ever.

I've seen @fmdkdd's comment in https://github.com/rust-lang/cargo/issues/4998, but I wrote this patch just because I need this, so please feel free to abort this PR if you don't like.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9480,6 +9480,22 @@ or if the current buffer has no file name."
   (and buffer-file-name
        (locate-dominating-file buffer-file-name "Cargo.toml")))
 
+(defun flycheck-rust-cargo-metadata ()
+  "Run 'cargo metadata' and return the result as parsed JSON object."
+  (with-temp-buffer
+    (call-process "cargo" nil t nil
+                  "metadata" "--no-deps" "--format-version" "1")
+    (goto-char (point-min))
+    (json-read)))
+
+(defun flycheck-rust-cargo-workspace-root ()
+  "Return the path to the workspace root of a Rust Cargo project.
+
+Return nil if the workspace root does not exist (for Rust
+versions inferior to 1.25)."
+  (let-alist (flycheck-rust-cargo-metadata)
+    .workspace_root))
+
 (defun flycheck-rust-cargo-has-command-p (command)
   "Whether Cargo has COMMAND in its list of commands.
 
@@ -9515,7 +9531,15 @@ This syntax checker requires Rust 1.17 or newer.  See URL
             (eval flycheck-cargo-check-args)
             "--message-format=json")
   :error-parser flycheck-parse-cargo-rustc
-  :error-filter flycheck-rust-error-filter
+  :error-filter (lambda (errors)
+                  ;; In Rust 1.25+, filenames are relative to the workspace
+                  ;; root.
+                  (let ((root (flycheck-rust-cargo-workspace-root)))
+                    (seq-do (lambda (err)
+                              (setf (flycheck-error-filename err)
+                                    (expand-file-name
+                                     (flycheck-error-filename err) root)))
+                            (flycheck-rust-error-filter errors))))
   :error-explainer flycheck-rust-error-explainer
   :modes rust-mode
   :predicate flycheck-buffer-saved-p
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3999,7 +3999,6 @@ The manifest path is relative to
      '(3 1 warning "function is never used: `main`" :checker rust-cargo :id "dead_code")
      '(3 1 info "#[warn(dead_code)] on by default" :checker rust-cargo :id "dead_code")
      '(4 9 warning "unused variable: `x`" :checker rust-cargo :id "unused_variables")
-     '(4 9 info "#[warn(unused_variables)] on by default" :checker rust-cargo :id "unused_variables")
      '(4 9 info "to avoid this warning, consider using `_x` instead" :checker rust-cargo :id "unused_variables"))))
 
 (flycheck-ert-def-checker-test rust-cargo rust default-target
@@ -4012,7 +4011,6 @@ The manifest path is relative to
      '(3 1 warning "function is never used: `main`" :checker rust-cargo :id "dead_code")
      '(3 1 info "#[warn(dead_code)] on by default" :checker rust-cargo :id "dead_code")
      '(4 9 warning "unused variable: `x`" :checker rust-cargo :id "unused_variables")
-     '(4 9 info "#[warn(unused_variables)] on by default" :checker rust-cargo :id "unused_variables")
      '(4 9 info "to avoid this warning, consider using `_x` instead" :checker rust-cargo :id "unused_variables"))))
 
 (flycheck-ert-def-checker-test rust-cargo rust lib-main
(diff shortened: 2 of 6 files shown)


fmdkdd:
Thanks for reminding me about this.  It looks like we don't have a choice, we have to call `cargo metadata` after all, as the alternative rust-lang/rust#47939 has been dropped.

And thank you for writing the patch!  I'll try to review it soon.

kngwyu:
I'm sorry I checked this change cause 
```
Suspicious state from syntax checker rust-cargo: Flycheck checker rust-cargo returned non-zero exit code 101, but its output contained no errors: error: no library targets found
```
in the project like
```
├── Cargo.toml
├── crate1
│   ├── Cargo.lock
│   ├── Cargo.toml
│   ├── src
│   └── target
├── crate2
│   ├── Cargo.toml
│   └── src
├── rustfmt.toml
├── target
```
so it needs to be fixed!

kngwyu:
This error occurs in a Project which has both `bin` and `lib` (and etc..).
In addition, changing `working-directory` into `workspace_root` can make compile time much more longer!
Hmm, now I don't feel this is a good way....

Is there any good idea to separate the directory where flycheck run command, from working directory?

fmdkdd:
> Is there any good idea to separate the directory where flycheck run command, from working directory?

The working directory *is* the directory where flycheck runs the command.

But for fixing #1397, the only thing that we need to change is to expand the names relatively to the workspace root, whereas previously they were relative to the working directory.  We don't want to change the working directory.

So the change should probably rather go to `:error-filter` or `:error-parser`.

kngwyu:
@fmdkdd 
Thanks for your advice.
I'll try it.

kngwyu:
I modified to use completely new way.
It works well in my test directory, but I'm not sure it's a good way.

fmdkdd:
@kngwyu Thanks!  That's closer to what the fix should be.  I agree that we should expand into an absolute path, only if the workspace root is present.

But doing the expansion in `flycheck-parse-rustc-diagnostic` will break the plain `rust` checker, which uses the same parser.  I think we should rather do the expansion into absolute paths in the  `:error-filter` of `rust-cargo`, so we don't mess with parsing and the other checker.

I've pushed some changes in that direction on your branch.  I've also fixed the tests, and added a new test for the workspace situation.  It works with nightly and stable.

kngwyu:
@fmdkdd Oh, I didn't consider `rust` checker, sorry, and thanks for your all work!

fmdkdd:
@kngwyu No worries, that's what reviews are for.

I've changed `alist-get` to `let-alist` as well, since `alist-get` is not available in Emacs 24.  I think we should be good.

@cpitclaudel Can you take a look?

flycheck.el
```diff
@@ -9480,6 +9480,26 @@ or if the current buffer has no file name."
   (and buffer-file-name
        (locate-dominating-file buffer-file-name "Cargo.toml")))
 
+(defun flycheck-rust-cargo-metadata ()
+  "Exec 'cargo metadata' and return the result as parsed JSON object.
```
cpitclaudel: I'd say `Run` rather than `Exec`, but that's nitpicking :)

flycheck.el
```diff
@@ -9480,6 +9480,26 @@ or if the current buffer has no file name."
...
+
+Return nil if the command failed or if a parse error occured."
```
cpitclaudel: Do we really want to fail silently for failed calls to `cargo metadata` and for parse errors? (IIUC this is for backwards compatibility, but I'm afraid it'll create harder-to-debug bugs if things change unexpectedly)

flycheck.el
```diff
@@ -9480,6 +9480,26 @@ or if the current buffer has no file name."
...
+  (let ((metadata-lines
```
cpitclaudel: `-when-let` here would remove the need for the `and` below

flycheck.el
```diff
@@ -9480,6 +9480,26 @@ or if the current buffer has no file name."
...
+         (ignore-errors
+           (process-lines "cargo" "metadata" "--no-deps" "--format-version" "1"))))
+    (and metadata-lines
+         (ignore-errors
+           (json-read-from-string (string-join metadata-lines ""))))))
```
cpitclaudel: That call to string-join should use `\n`, not ``, I think (though it might not matter in this particular case?)
Besides, would it read nicer to use `call-process` and `json-read` instead of `process-lines` + `string-join` + `json-read-from-string`?

flycheck.el
```diff
@@ -9480,6 +9480,26 @@ or if the current buffer has no file name."
...
+
+(defun flycheck-rust-cargo-workspace-root ()
+  "Return the path to the workspace root of a Rust Cargo project.
+
+Return nil if the workspace root does not exist (for Rust
+versions inferior to 1.25)."
+  (let ((metadata (flycheck-rust-cargo-metadata)))
+    (let-alist metadata
```
cpitclaudel: No need to bind `metadata` here, I think (`(let-alist (flycheck-rust-cargo-metadata) …)`)

flycheck.el
kngwyu: I think it's only because 'cargo' 's bug if parse error occurs, so we can allow error here, but I'm not sure.

kngwyu:
Thanks for review.
I modified a bit, but for `flycheck-rust-cargo-metadata` I'm not sure how we should handle errors and parse json.

flycheck.el
fmdkdd: In this case, we may just use `flycheck-parse-json`?

kngwyu: @fmdkdd 
`flycheck-parse-json` returns a list of parsed JSON lines, so isn't it too complex for this case, where what we want is only `metadata` ?

kngwyu:
I simplified `flycheck-rust-cargo-metadata` and changed not to use `ignore-errors` in this function.
The newest option `--no-deps` in  `cargo metadata --no-deps --format-version 1` was added [2 years ago](https://github.com/rust-lang/cargo/commit/80e3c05beefad3a19b2cd694d0acfd4e09041fb0#diff-24a11135fb787144156765e41edea25b), so backwards compatibility is OK.
Thus function never fail unless `cargo` doesn't exist or breaking change is done.

flycheck.el
fmdkdd: It may be overkill, yes.  Using `call-process` and `json-read` directly is fine :)

flycheck.el
```diff
@@ -9480,6 +9480,24 @@ or if the current buffer has no file name."
   (and buffer-file-name
        (locate-dominating-file buffer-file-name "Cargo.toml")))
 
+(defun flycheck-rust-cargo-metadata ()
+  "Run 'cargo metadata' and return the result as parsed JSON object.
+
+Return nil if the command failed or if a parse error occured."
```
fmdkdd: I guess this line can go now that we don't ignore errors anymore.

fmdkdd: Oh and it looks like the file is misformatted.  Please run `make format`, which should take care of the issue.

kngwyu:
Modified.
Thanks for your review!

flycheck.el
```diff
@@ -9483,7 +9483,7 @@ or if the current buffer has no file name."
 (defun flycheck-rust-cargo-metadata ()
   "Run 'cargo metadata' and return the result as parsed JSON object.
 
-Return nil if the command failed or if a parse error occured."
+We don't ignore any errors in this function."
```
fmdkdd: I don't think we need to spell out that we don't do any error catching, as that is a detail of implementation.  Anyone caring about implementation can see in the code that no error checking is done.

Please remove this line :)

kngwyu:
Modified.

fmdkdd:
Great!  Thank you very much for this patch! :clap:

Issue #1420, opened 2018-03-05
tslint not locating package.json correctly


xdl:
I have tslint configured for Flycheck and it's reading in the correct tslint.json file:

```
Syntax checker in buffer Navigation.tsx in web-mode:

  typescript-tslint
    - major mode:         `web-mode' supported
    - may enable:         yes
    - executable:         Found at /Users/xiaodili/dev/myapp/node_modules/tslint/bin/tslint
    - configuration file: Found at "/Users/xiaodili/dev/myapp/tslint.json"

Flycheck can use this syntax checker for this buffer.

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled
checkers.

--------------------

Flycheck version: 32snapshot (package: 20180224.1046)
Emacs version:    25.3.1
System:           x86_64-apple-darwin13.4.0
Window system:    ns
```

However, I get the following errors:

```
     1  36 warning  no-im… Module 'material-ui' is not listed as dependency in package.json (typescript-tslint)
     2  24 warning  no-im… Module 'react' is not listed as dependency in package.json (typescript-tslint)
     4  18 warning  no-em… An empty interface is equivalent to `{}`. (typescript-tslint)
```

Despite having listed these modules as dependencies in the `package.json` file in the root of my project. When I run tslint on the terminal for that file, I get:

```
node_modules/tslint/bin/tslint src/components/Navigation.tsx

ERROR: src/components/Navigation.tsx[4, 18]: An empty interface is equivalent to `{}`.
```

Which suggests that Flycheck's tslint isn't able to pick up the package.json location. I had a quick Google and was wondering if it might be related to this issue: https://github.com/w0rp/ale/issues/1101 but my elisp isn't solid enough to dig into this.

fmdkdd:
If the issue is about picking up implicit dependencies in the same folder as the source file being checked, then yes, it's likely that is the cause.

In flycheck.el, can you change the `source` to `source-inplace` in the `:command` property of `typescript-tslint` (about [here](https://github.com/flycheck/flycheck/blob/master/flycheck.el#L10153))?  Evaluate that (`C-M-x` with the cursor inside the checker definition) and try to see if that fixes the issue.

xdl:
Thanks @fmdkdd, changing `source` to `source-inplace` has fixed the issue, thanks so much! Will you include this change in a future release?

fmdkdd:
> Will you include this change in a future release?

[...]

PR #1425, opened 2018-03-08
Change source to source-inplace for TSLint

Simplify:
Fixes and closes GH-1420

- Changes `source` to `source-inplace` in `tslint` command. This way `tslint` can locate `package.json`.
- Added rule `no-implicit-dependencies` into test `tslintrc.json` and `package.json` to test this.
- Removed `use-strict` rule because it does not exist in `tslint` any more.
- Removed rule `no-unused-variable` because this now in `tslint` also requires `tsconfig.json`.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -10150,7 +10150,7 @@ See URL `https://github.com/palantir/tslint'."
             (config-file "--config" flycheck-typescript-tslint-config)
             (option "--rules-dir" flycheck-typescript-tslint-rulesdir)
             (eval flycheck-tslint-args)
-            source)
+            source-inplace)
   :error-parser flycheck-parse-tslint
   :modes (typescript-mode))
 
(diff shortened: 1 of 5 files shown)


cpitclaudel:
Thanks. I'm happy to trust your experience on all things tslint, but I'm suspicious still: must we really move to source-inplace? Isn't a working directory change enough?
If not, can we open a bug report with the tslint people?

fmdkdd:
> Isn't a working directory change enough?

@Simplify tested that in #1420:

> it did not work, I'll make PR for source-inplace.

> If not, can we open a bug report with the tslint people?

There is already palantir/tslint#1590 for accepting files on standard input, but the issue is not progressing on their front.

That seems to leave us with few options.

cpitclaudel:
> @Simplify tested that in #1420:

Yup, I saw that; I was surprised about it, though, and I wondered whether that part specifically was worth a report.  But I'm not so convinced about that anymore, so I'm satisfied with this approach :)

Thanks for clarifying!

Simplify:
Lets wait until tomorrow with merging, I'll take another look to working directory problem, just to be sure.

fmdkdd:
@Simplify Understood.

Simplify:
@cpitclaudel @fmdkdd After some more testing I can confirm again that setting `working-directory` does not help at all. For now, only option is to use `source-inplace`.

fmdkdd:
Good.  Thank you for making sure :+1:

PR #1426, opened 2018-03-10
Add a way to control suffix of temporary files, and use it for dmd checker

CyberShadow:
Please see the individual commits' commit messages for details.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7503,8 +7503,9 @@ See URL `https://www.commonwl.org/v1.0/SchemaSalad.html'."
 (defun flycheck-d-base-directory ()
   "Get the relative base directory path for this module."
   (let* ((file-name (buffer-file-name))
-         (module-file (if (string= (file-name-nondirectory file-name)
-                                   "package.d")
+         (module-file (if (and file-name
+                               (string= (file-name-nondirectory file-name)
+                                        "package.d"))
                           (directory-file-name (file-name-directory file-name))
                         file-name)))
     (flycheck-module-root-directory
--- a/flycheck.el
+++ b/flycheck.el
@@ -7553,7 +7553,7 @@ Requires DMD 2.066 or newer.  See URL `https://dlang.org/'."
             (eval (concat "-I" (flycheck-d-base-directory)))
             (option-list "-I" flycheck-dmd-include-path concat)
             (eval flycheck-dmd-args)
-            source)
+            (source ".d"))
   :error-patterns
   ((error line-start
           (file-name) "(" line "," column "): Error: " (message)
--- a/flycheck.el
+++ b/flycheck.el
@@ -1230,33 +1230,35 @@ Return the path of the directory"
     (push tempdir flycheck-temporaries)
     tempdir))
 
-(defun flycheck-temp-file-system (filename)
+(defun flycheck-temp-file-system (filename &optional suffix)
   "Create a temporary file named after FILENAME.
 
 If FILENAME is non-nil, this function creates a temporary
 directory with `flycheck-temp-dir-system', and creates a file
 with the same name as FILENAME in this directory.
 
-Otherwise this function creates a temporary file with
-`flycheck-temp-prefix' and a random suffix.  The path of the file
-is added to `flycheck-temporaries'.
+Otherwise this function creates a temporary file starting with
+`flycheck-temp-prefix'.  If present, SUFFIX is appended;
+otherwise, a random suffix is used.  The path of the file is
+added to `flycheck-temporaries'.
 
 Return the path of the file."
   (let ((tempfile (convert-standard-filename
                    (if filename
                        (expand-file-name (file-name-nondirectory filename)
                                          (flycheck-temp-dir-system))
-                     (make-temp-file flycheck-temp-prefix)))))
+                     (make-temp-file flycheck-temp-prefix nil suffix)))))
     (push tempfile flycheck-temporaries)
     tempfile))
 
-(defun flycheck-temp-file-inplace (filename)
+(defun flycheck-temp-file-inplace (filename &optional suffix)
   "Create an in-place copy of FILENAME.
 
 Prefix the file with `flycheck-temp-prefix' and add the path of
 the file to `flycheck-temporaries'.
 
-If FILENAME is nil, fall back to `flycheck-temp-file-system'.
+If FILENAME is nil, fall back to `flycheck-temp-file-system' with
+the specified SUFFIX.
 
 Return the path of the file."
   (if filename
@@ -1268,7 +1270,7 @@ Return the path of the file."
                                           (file-name-directory filename)))))
         (push tempfile flycheck-temporaries)
         tempfile)
-    (flycheck-temp-file-system filename)))
+    (flycheck-temp-file-system filename suffix)))
 
 (defun flycheck-temp-directory (checker)
   "Return the directory where CHECKER writes temporary files.
@@ -5006,6 +5008,8 @@ this error to produce the explanation to display."
   (pcase arg
     ((pred stringp) t)
     ((or `source `source-inplace `source-original) t)
+    (`(,(or `source `source-inplace) ,suffix)
+     (stringp suffix))
     ((or `temporary-directory `temporary-file-name) t)
     (`null-device t)
     (`(config-file ,option-name ,config-file-var)
@@ -5258,6 +5262,12 @@ STRING
      syntax checker needs other files from the source directory,
      such as include files in C.
 
+`(source SUFFIX)', `(source-inplace SUFFIX)'
+     Like `source' and `source-inplace', but ensure generated
+     file names end with the given suffix.  Use this when the
+     checker requires that file names on its command line have a
+     certain suffix (file extension).
+
 `source-original'
      Return the path of the actual file to check, or an empty
      string if the buffer has no file name.
@@ -5354,6 +5364,12 @@ are substituted within the body of cells!"
      (list (flycheck-save-buffer-to-temp #'flycheck-temp-file-system)))
     (`source-inplace
      (list (flycheck-save-buffer-to-temp #'flycheck-temp-file-inplace)))
+    (`(source ,suffix)
+     (list (flycheck-save-buffer-to-temp
+            (lambda (filename) (flycheck-temp-file-system filename suffix)))))
+    (`(source-inplace ,suffix)
+     (list (flycheck-save-buffer-to-temp
+            (lambda (filename) (flycheck-temp-file-inplace filename suffix)))))
     (`source-original (list (or (buffer-file-name) "")))
     (`temporary-directory (list (flycheck-temp-dir-system)))
     (`temporary-file-name
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3224,6 +3224,19 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
    '(13 1 info "instantiated from here: `U!()`"
         :checker d-dmd)))
 
+(flycheck-ert-def-checker-test d-dmd d non-d-extension
+  (skip-unless (fboundp 'd-mode))
+  (unless (version<= "24.4" emacs-version)
+    (ert-skip "Skipped because CC Mode is broken on 24.3.
+See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
+  (flycheck-ert-with-temp-buffer
+    (insert "!invalid")
+    (d-mode)
+    (flycheck-ert-buffer-sync)
+    (flycheck-ert-should-errors
+     '(1 1 error "declaration expected, not `!`"
+         :checker d-dmd))))
+
 (flycheck-ert-def-checker-test dockerfile-hadolint dockerfile error
   (flycheck-ert-should-syntax-check
    "language/dockerfile/Dockerfile.error" 'dockerfile-mode
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3199,7 +3199,7 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
     (flycheck-ert-should-syntax-check
      "language/d/src/dmd/warning.d" 'd-mode
      '(9 5 warning "statement is not reachable" :checker d-dmd)
-     '(20 17 warning "function dmd.warning.bar is deprecated"
+     '(20 17 warning "function `dmd.warning.bar` is deprecated"
           :checker d-dmd))))
 
 (flycheck-ert-def-checker-test d-dmd d missing-import
@@ -3208,7 +3208,7 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
 See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
   (flycheck-ert-should-syntax-check
    "language/d/src/dmd/warning.d" 'd-mode
-   '(4 8 error "module external_library is in file 'external_library.d' which cannot be read"
+   '(4 8 error "module `external_library` is in file 'external_library.d' which cannot be read"
        :checker d-dmd)))
 
 (flycheck-ert-def-checker-test d-dmd d continuation-line
@@ -3217,11 +3217,11 @@ See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
 See https://github.com/flycheck/flycheck/issues/531 and Emacs bug #19206"))
   (flycheck-ert-should-syntax-check
    "language/d/src/dmd/continuation.d" 'd-mode
-   '(5 12 error "undefined identifier 'invalid'"
+   '(5 12 error "undefined identifier `invalid`"
        :checker d-dmd)
-   '(10 12 error "template instance continuation.T!() error instantiating"
+   '(10 12 error "template instance `continuation.T!()` error instantiating"
         :checker d-dmd)
-   '(13 1 info "instantiated from here: U!()"
+   '(13 1 info "instantiated from here: `U!()`"
         :checker d-dmd)))
 
 (flycheck-ert-def-checker-test dockerfile-hadolint dockerfile error


cpitclaudel:
Thanks a lot; this is a great PR.   I'll post a more detailed review soon, but can you clarify when the cdr of the new `--suffixes` variable is used?

flycheck.el
```diff
@@ -1110,6 +1110,22 @@ Return the path of the directory"
     (push tempdir flycheck-temporaries)
     tempdir))
 
+(defvar flycheck--forced-suffix nil
+  "List of forced filename suffixes.
+
+Dynamic variable.  If non-nil, ensure generated file names have
+one of the suffixes in this list.  Activated using
+a (force-suffix ...) clause in `flycheck-substitute-argument'.")
```
cpitclaudel: I'm not a huge fan of dynamic variables; though I'm not sure whether there's a uch better alternative yet.

flycheck.el
```diff
@@ -1110,6 +1110,22 @@ Return the path of the directory"
...
+
+(defun flycheck--enforce-suffix (filename)
+  "Ensure FILENAME ends with one of the suffixes in `flycheck--forced-suffix'."
+  (concat
+   filename
+   (when (and flycheck--forced-suffix
+              (-none? (lambda (suffix) (string-suffix-p suffix filename))
+                      flycheck--forced-suffix))
```
cpitclaudel: Ah, I see now why there's a list.

cpitclaudel: We tend to use the cl-lib functions when possible

flycheck.el
```diff
@@ -4792,6 +4814,16 @@ STRING
...
 
+`(force-suffix BODY SUFFIX...)'
+     Ensure generated file names end with one of the given
+     suffixes.  If not, append the first suffix in the list.  Use
+     this when the checker requires that file names on its
+     command line have a certain suffix (file extension).
+
+     BODY is evaluated as any other
+     `flycheck-substitute-argument' argument.  Affects `source',
+     `source-inplace', and `temporary-file-name'.
```
cpitclaudel: I wonder if we need to support temporary-file-name — I don't think we have even a single checker in Flycheck that actually uses it.

CyberShadow:
> There's something that confuses me a bit: for source and source-inplace, we already maintain the file's original extension, right? Your PR solves the case when we're checking a buffer whose name ends with a specific extension, but that extension is rejected by the linter. Right?

Yep, that's right.

> Can you say a bit more about the use case? 

I mentioned this in the first commit's commit message: "Even though editing real files whose filenames do not conform to their respective languages' toolchains' requirements is not expected to be a common use case, this situation can apply when editing code in transient buffers created by Emacs, such as those created by Org-Babel's org-edit-src-code."

One can think of other use cases, such as changing the major mode of the scratch bufffer to code up a code fragment that's to be used elsewhere.

Generally speaking, most other language facilities already work in this situation - syntax highlighting, and in D's case, even auto-completion, so from the end-user's perspective, there's no reason why we wouldn't want to make error highlighting "just work" as well.

> Given the added complexity, I'd like to get a better understanding of what makes D special (most linters/compilers reject unknown extensions, but I don't expect we'd want to make a similar change to all other non-stdin checkers)

D is not at all special, it's just the use case I'm working on and familiar with. I do think this should be applied to all checkers which require certain extensions.

flycheck.el
CyberShadow: It could be made a property of the checker, though that would give less fine-grained control over which filenames are affected (in case we want to generate more than one filename in an invocation).

flycheck.el
CyberShadow: What is the cl-lib equivalent? I found `cl-notany`, but that's apparently in `cl-extra` which isn't currently `require`d in flycheck.

flycheck.el
cpitclaudel: Indeed, a property of the checker might work.  I don't think we ever generate more than one filename — can we think of a case in which we would want to?

flycheck.el
cpitclaudel: I think just `(not (cl-some …))` :) Although here `(not (seq-some …))` would be even better, sorry :)

cpitclaudel:
> Generally speaking, most other language facilities already work in this situation - syntax highlighting, and in D's case, even auto-completion, so from the end-user's perspective, there's no reason why we wouldn't want to make error highlighting "just work" as well.

OK. I feel -0.5 on this feature, but +1 on a nice contribution.  Maybe we can simplify the implementation together a bit, and I'll be more enthusiastic about maintaining it :)  I'd also like to hear from @fmdkdd 

Some thoughts:

* Could/should do anything when the current buffer already has a file name? Or should this feature could be restricted to the case in which the the current buffer doesn't have a filename?

  For some reason, org-mode temporary buffers have a file name set; I'm not sure why. It comes from commit [4b6988bf36cb458c9d113ee4332e016990c1eb04](https://code.orgmode.org/bzg/org-mode/commit/4b6988bf36cb458c9d113ee4332e016990c1eb04) in the org-mode repo.  Maybe we should ask on the org mailing list?
    I think changing this would allow us to drop `--enforce-suffix`, and would make it so that we don't have to teach Flycheck about all valid extensions for a particular compiler).

* If we only apply this to source-inplace and source, could we (maybe) make it an argument to `source` and `source-inplace`?  You could then write `(source ".el")`, for example, and `(source)` would be just the same as `(source nil)`.  If I'm reading the code correctly, this would keep the flexibility of the current approach while getting rid of the dynamic variable.

WDYT?

CyberShadow:
> Could/should do anything when the current buffer already has a file name?

Don't know. Does a lot of Emacs userspace/ecosystem care about this distinction? I found out about it only while developing this patch.

> I think changing this would allow us to drop --enforce-suffix, and would make it so that we don't have to teach Flycheck about all valid extensions for a particular compiler).

Sorry, I don't understand how that conclusion follows.

> If we only apply this to source-inplace and source, could we (maybe) make it an argument to source and source-inplace? You could then write (source ".el"), for example, and (source) would be just the same as (source nil). If I'm reading the code correctly, this would keep the flexibility of the current approach while getting rid of the dynamic variable.

Sure @ the syntax change, but not sure how that allows dropping the dynamic variable. The challenge is getting the information from `flycheck-substitute-argument` to `flycheck-temp-file-system` and `flycheck-temp-file-inplace`. It can't be higher (in the call stack) because we need the suffix for `make-temp-file`. Doing this without a dynamic variable would require modifying the signatures of those two functions, plus their caller (`flycheck-save-buffer-to-temp`) and its invocations, so the dynamic variable seemed like the less intrusive change.

cpitclaudel:
> > Could/should do anything when the current buffer already has a file name?

> Don't know. Does a lot of Emacs userspace/ecosystem care about this distinction? I found out about it only while developing this patch.

Sorry, I think I didn't explain myself well :) I meant: when a buffer has buffer-file-name non-nil, maybe we don't need to do anything in terms of adding an extension: IOW, if the buffer is already backed by a file, we wouldn't touch its extension. 

> > I think changing this would allow us to drop --enforce-suffix, and would make it so that we don't have to teach Flycheck about all valid extensions for a particular compiler).

> Sorry, I don't understand how that conclusion follows.

Does the above help? Sorry for being unclear at first.

> > If we only apply this to source-inplace and source, could we (maybe) make it an argument to source and source-inplace? You could then write (source ".el"), for example, and (source) would be just the same as (source nil). If I'm reading the code correctly, this would keep the flexibility of the current approach while getting rid of the dynamic variable.

> Sure @ the syntax change, but not sure how that allows dropping the dynamic variable. The challenge is getting the information from `flycheck-substitute-argument` to `flycheck-temp-file-system` and `flycheck-temp-file-inplace`. It can't be higher (in the call stack) because we need the suffix for `make-temp-file`. Doing this without a dynamic variable would require modifying the signatures of those two functions, plus their caller (`flycheck-save-buffer-to-temp`) and its invocations, so the dynamic variable seemed like the less intrusive change.

I agree that the variable is less intrusive; but I'm more concerned about the complexity of the final code than the size of this particular change.  WDYT?

CyberShadow:
> Sorry, I think I didn't explain myself well :) I meant: when a buffer has buffer-file-name non-nil, maybe we don't need to do anything in terms of adding an extension: IOW, if the buffer is already backed by a file, we wouldn't touch its extension.

Right, on the same page here; but, as you said, temporary org-mode buffers apparently do have a file name. Is that a quirk specific to org-mode, or could it be something more prevalent? Is this distinction a sufficiently meaningful signal that flycheck can use, or is it too noisy to be useful, and flycheck shouldn't care whether a buffer has an associated filename or not?

> Does the above help? Sorry for being unclear at first.

Sorry, not quite. When you wrote "would allow us to drop --enforce-suffix", did you mean that it would make this entire pull request unnecessary, or only some part of it?

> I agree that the variable is less intrusive; but I'm more concerned about the complexity of the final code than the size of this particular change. WDYT?

Well, this is pretty subjective, but in such situations I'd be generally not in favor of adding function arguments, because they grow a complexity metric of the function's contract in a non-sustainable way - functions with many arguments become overly unwieldy to use. Having to pass this information across multiple layers implies modifying each layer's contract / documentation / implementation / invocations, so a side channel like a dynamic variable might be more preferable.

cpitclaudel:
> Is that a quirk specific to org-mode, or could it be something more prevalent? Is this distinction a sufficiently meaningful signal that flycheck can use, or is it too noisy to be useful, and flycheck shouldn't care whether a buffer has an associated filename or not?

I'd favor starting small, and expanding as needed.  I will email the org-mode list.

> Sorry, not quite. When you wrote "would allow us to drop --enforce-suffix", did you mean that it would make this entire pull request unnecessary, or only some part of it?

Not the whole pull request: just that one particular function.

> Having to pass this information across multiple layers implies modifying each layer's contract / documentation / implementation / invocations, so a side channel like a dynamic variable might be more preferable.

Indeed, there's always room for debate.  We don't use the dynamic variable style in Flycheck, so at least there's a case to be made that consistency is good.  Besides, the suffic argument is already available in the functions that we call in the end, so mirroring it in the callers doesn't bother me much.

flycheck.el
fmdkdd: We don't have any.  However a Github search turns out a few hits on external checkers.  It's probably not an issue if we leave `temporary-file-name` out of this PR.

fmdkdd:
This is a well thought-out PR, thank you. :+1:   I have a couple of concerns.

> from the end-user's perspective, there's no reason why we wouldn't want to make error highlighting "just work" as well.

That is a good point.  One reason we might not want to do that is if the costs (complexity) outweight the benefits (i.e., niche feature).  In this case, I'm not convinced that the use-case is so prevalent, so the complexity would have to be low to make the cut.  Currently, there are two things that bug me.

> You could then write (source ".el"), for example, and (source) would be just the same as (source nil)

I would prefer this solution as well.  `source` alone would not force any suffix (current behavior), and `(source ".el")` would be equivalent to (but more readable than) `(force-suffix source ".el")`.

Concerning the dynamic variable: the only place where the `-temp-file` functions are called is in `flycheck-save-buffer-to-temp`.  This is expanded in `flycheck-substiute-arguments`, so it would be easy (with the above change) to just pass the suffix to `flycheck-save-buffer-to-temp`.

> Well, this is pretty subjective, but in such situations I'd be generally not in favor of adding function arguments, because they grow a complexity metric of the function's contract in a non-sustainable way 

Dynamic variables can be justified in a situation of deep call stacks (and parameterization from distant calls), and/or already long parameter list.  Here we have neither.

Besides, the dynamic variable changes the function's behavior in a subtle way.  I'd argue that the contract should make this behavior explicit.  And, in this case, it wouldn't be an overwhelming change.

CyberShadow:
Thank you for the detailed review. I've pushed amendments implementing the changes proposed above.

> Not the whole pull request: just that one particular function.

Ah, I understand what you mean now. It would apply only to the case where the `filename` argument to `flycheck-temp-file-system` and `flycheck-temp-file-inplace` is nil, thus the suffix list is only used for `make-temp-file`, in which case there doesn't even need to be a list.

> Dynamic variables can be justified in a situation of deep call stacks (and parameterization from distant calls), and/or already long parameter list. Here we have neither.

Now that I've actually tried implementing it, I agree - we don't even need to change the signature of `flycheck-save-buffer-to-temp`.

cpitclaudel:
I like it :)  But if possible, I'd feel even better without `--enforce-suffix` nor the list of suffixes. @CyberShadow, based on the following comment, I think you agreed?

> Ah, I understand what you mean now. It would apply only to the case where the filename argument to flycheck-temp-file-system and flycheck-temp-file-inplace is nil, thus the suffix list is only used for make-temp-file, in which case there doesn't even need to be a list.

The relevant org-mode thread is at https://lists.gnu.org/archive/html/emacs-orgmode/2018-03/msg00242.html .

CyberShadow:
> But if possible, I'd feel even better without --enforce-suffix nor the list of suffixes.

Well, without that part, this patch loses the intended effect that led to me writing it: make flycheck work in `org-edit-src-code` buffers. You're of course welcome to use the changes here in full or in part, but, unless the org-mode discussion results in a change in org-mode, there is no immediate use case motivating any further work on this issue.

Let's see how the org-mode thread unfolds, perhaps?

cpitclaudel:
> Let's see how the org-mode thread unfolds, perhaps?

Yup, absolutely.

cpitclaudel:
Nicolas is open to a change on the org-mode side: https://lists.gnu.org/archive/html/emacs-orgmode/2018-03/msg00324.html
I don't know if I'll time to work on the (small) patch to make this happen, though :/ @CyberShadow , WDYT?  We could merge your patch as is, or we could DTRT and patch org-mode ^^

CyberShadow:
I can have a stab at it but not soon, sorry! Though, TBH it seems like you have a much better grasp of the situation than I do :)

In any case, feel free to close this for the time being.

cpitclaudel:
I have good news :) The Org-mode folks (thanks Bastien!) made the change, so org-mode doesn't set buffer-file-name anymore.  @CyberShadow, still motivated about this PR? :)

https://lists.gnu.org/archive/html/emacs-orgmode/2018-04/msg00486.html

matthew-piziak:
I'm hoping this will make `flycheck-emacs-lisp` usable for Org-mode-literate Emacs configuration files. Has this change landed? I compile `org-plus-contrib` from source and it doesn't work for me.

matthew-piziak:
`buffer-file-name` in an Org SRC buffer evals to `nil` for me.

matthew-piziak:
If I perform `(setq-local buffer-file-name "test.el")` in the buffer, then flycheck starts working for me. I worry about the implications of such a hack though...

matthew-piziak:
Here's how I hacked it in.

```
(add-hook 'emacs-lisp-mode-hook 'flycheck-mode)
(defun set-dummy-elisp-backing-file ()
  (if (and (eq major-mode 'emacs-lisp-mode) (bound-and-true-p org-src-mode) (null buffer-file-name))
      (setq-local buffer-file-name "org-src.el")))
(advice-add 'org-edit-src-save :after #'set-dummy-elisp-backing-file)
```

matthew-piziak:
As you might imagine that doesn't work too well. Error navigation tries to go to a nonexistent file, and flycheck doesn't know where to find variables defined elsewhere in the config.

rileyrg:
So this still is not working? (well, its not with Melpa org-mode and emacs 27.0.50)

cpitclaudel:
The patch needs to be updated based on the discussion above.  @rileyrg, would you like to do it?

rileyrg:
> The patch needs to be updated based on the discussion above. @rileyrg, would you like to do it?

I would but tbh I'm not skilled enough in elisp and the patching process.  It would generate more work than it saves. I might have a peek tomorrow.

rileyrg:
Interestingly the flyspell checker emacs-lisp-checkdoc works fine when editing an elisp src block in an elisp buffer.

CyberShadow:
> I have good news :) The Org-mode folks (thanks Bastien!) made the change, so org-mode doesn't set buffer-file-name anymore. @CyberShadow, still motivated about this PR? :)
> 
> https://lists.gnu.org/archive/html/emacs-orgmode/2018-04/msg00486.html

In Emacs 26.3 and Org-Mode 9.1.9, the value of `buffer-file-name` for an Org Src buffer editing a D program is now:

`/path/to/my-org-file.org[*Org Src my-org-file.org[ d ]*]`

This causes the checker to, of course, complain: `Error: unrecognized file extension org[ d ]*]`

So, either the above change has been reverted / superseded, or it was committed in a branch for a future release so distant that it has not occurred during the 21 months since that message.

cpitclaudel:
I see this on master:

```diff
commit 25f99522adf1488f934e23328c735e879391a1c9
Author: Bastien <bzg@gnu.org>
Date:   Fri Apr 27 00:25:15 2018 +0200

    org-src.el: Don't set `buffer-file-name'
    
    * lisp/org-src.el (org-src-mode-configure-edit-buffer): Don't
    set `buffer-file-name'.
    
    Thanks to Clément Pit-Claudel for reporting this.

diff --git a/lisp/org-src.el b/lisp/org-src.el
--- a/lisp/org-src.el
+++ b/lisp/org-src.el
@@ -668,12 +668,10 @@
 
 (defun org-src-mode-configure-edit-buffer ()
+  "Configure the src edit buffer."
   (when (bound-and-true-p org-src--from-org-mode)
     (add-hook 'kill-buffer-hook #'org-src--remove-overlay nil 'local)
     (if (bound-and-true-p org-src--allow-write-back)
 	(progn
 	  (setq buffer-offer-save t)
-	  (setq buffer-file-name
-		(concat (buffer-file-name (marker-buffer org-src--beg-marker))
-			"[" (buffer-name) "]"))
 	  (setq-local write-contents-functions '(org-edit-src-save)))
       (setq buffer-read-only t))))
```

cpitclaudel:
Is anyone interested in updating this patch, or should we close this?

CyberShadow:
@cpitclaudel Please see my previous comment. I can reproduce a non-null `buffer-file-name` from `emacs -Q`, just type a `#+BEGIN_SRC` block with any language (`lisp` works), press `C-c '`, then `M-: buffer-file-name`.

cpitclaudel:
> Please see my previous comment. 

I saw it, but I also linked to the on master; can you check whether your version of org-mode includes it?

> I can reproduce a non-null buffer-file-name from emacs -Q, just type a #+BEGIN_SRC block with any language (lisp works), press C-c ', then M-: buffer-file-name.

I cannot.  Which version of Emacs are you using?  Are you installing org-mode from ELPA or using the bundled version?

CyberShadow:
> can you check whether your version of org-mode includes it?

It does not.

> I cannot. Which version of Emacs are you using?

This information can be found in the comment I posted above.

I must assume that you misunderstood the intention of said comment. Continuing with the direction that you proposed makes no sense as long as the org-mode change is not widely available. We have to either wait until it ends up in a generally available release, however more years that takes, or choose another path.

cpitclaudel:
> I must assume that you misunderstood the intention of said comment. 

Look, I'm doing my best here.  Let's remember that this is volunteer work, and let's try to up the kindness a bit.

> This information can be found in the comment I posted above.

Sorry for missing that.  Surely it wouldn't have hurt too much to re-post it.    In previous comments, you said: 

> So, either the above change has been reverted / superseded, or it was committed in a branch for a future release so distant that it has not occurred during the 21 months since that message.

> Continuing with the direction that you proposed makes no sense as long as the org-mode change is not widely available. We have to either wait until it ends up in a generally available release

I pointed to the commit on master to show that, as far as I could tell, it hadn't been reverted.  Here's what I see on my machine:

```
$ git tag --contains 25f99522adf1488f934e23328c735e879391a1c9
release_9.2
release_9.2.1
release_9.2.2
release_9.2.3
release_9.2.4
release_9.2.5
release_9.2.6
release_9.3
release_9.3.1
release_9.3.2
````

```
$ git log --tags --simplify-by-decoration --pretty="format:%ai %d"
2020-01-26 11:05:20 +0100  (tag: release_9.3.2)
2019-12-26 19:16:10 +0100  (tag: release_9.3.1)
2019-12-03 16:25:32 +0100  (tag: release_9.3)
2019-09-04 06:53:49 +0200  (tag: release_9.2.6)
2019-08-02 09:55:53 +0200  (tag: release_9.2.5)
2019-05-26 00:58:28 -0700  (tag: release_9.2.4)
2019-04-02 08:30:19 +0200  (tag: release_9.2.3)
2019-02-22 18:32:31 -0800  (tag: release_9.2.2)
2019-02-03 14:54:58 +0100  (tag: release_9.2.1)
2018-12-26 16:08:13 +0100  (tag: release_9.2)
…
2018-01-03 15:25:33 +0100  (tag: release_9.1.6)
```

You're to be using a version of org that's 2 years old at this point.  The first release to include the fix was org 9.2, released on ELPA about 15 months ago.  That's 12 months prior to your comment mentioning "a future release so distant that it has not occurred during the 21 months since that message."  I understand that Emacs 26.3 bundles 9.1.9 and thus doesn't include the fix, but isn't it straightforward to update org from ELPA?  Am I missing something?

**Comment by CyberShadow** [comment]
> Look, I'm doing my best here. Let's remember that this is volunteer work, and let's try to up the kindness a bit.

We're a bit past the point of "I'm proposing a change in your project that I want" and are more in the "I'm doing what the project maintainers want me to do" territory, and of course after 2 years I don't have an immediate use case, so I'm feeling the same way. I hope you understand that it's frustrating to have your comments skimmed over or having the PR marked as abandoned when the ball is not in your court.

> You're using a version of org that's 2 years old at this point.

Unfortunately we have no control over the Org/Emacs integration schedule or Emacs release schedule. As far as I can see, the *effective* latest version of Org mode is the one that's shipped with the latest released version of Emacs. The one time I tried to update Org mode to a different version, everything broke horribly, so I had to assume that that wasn't something that's expected of users to be doing, and those who wish to have a stable Org experience should just use the version bundled with their Emacs version. It's also not impossible that by the time the Emacs/Org maintainers decide to update the Org version bundled with Emacs, another change lands in Org which invalidates the fix referenced above. As such, I could do the change, but I just don't see it actually solving any problem that anyone has right now.

**Comment by CyberShadow** [comment]
Closing this for now so it doesn't waste flycheck maintainers' time.

**Comment by cpitclaudel** [comment]
> I hope you understand that it's frustrating to have your comments skimmed over or having the PR marked as abandoned when the ball is not in your court.

Yes, I understand, and I apologize about this.  Please don't take that label badly; I've spent quite a bit of time triaging through PRs lately, and the word choice was poor.  I know you do a lot of community work, and that work is very appreciated, so sorry for the frustration.

> and of course after 2 years I don't have an immediate use case, so I'm feeling the same way.

Yes, I'm hoping that we can have a better turnaround in the future.  

> Unfortunately we have no control over the Org/Emacs integration schedule or Emacs release schedule.

Thanks, that clarifies a lot. I got confused by your first comment about not seeing this commit and it having been committed to a far-future branch.

It is my understanding that the next Emacs will include Org 9.3, which includes this fix:

```
commit 165f7383822086d465519ebe6e4283723923f097
Author: Bastien <bzg@gnu.org>
Date:   Tue Dec 3 23:27:04 2019 +0100

    Update Org to 9.3
```

Additionally, I think it's common to run Emacs from ELPA, but I don't have hard numbers to back this up.  I think the recommended installation procedure is from ELPA, and I believe many of the issues that used to exist with updating have been worked on in the last few years.

> As such, I could do the change, but I just don't see it actually solving any problem that anyone has right now.

In January @rileyrg mentioned running into this problem despite running org from ELPA, so I think a fix would in fact help other users.

**Comment by cpitclaudel** [comment]
Ah, I hadn't seen that you had closed the PR.

**Comment by CyberShadow** [comment]
> In January @rileyrg mentioned running into this problem despite running org from ELPA, so I think a fix would in fact help other users.

Thank you for pointing that out. I will try to have a look at this soon. Will reopen once this is ready for review again.

**Comment by cpitclaudel** [comment]
Thanks a lot.

In the longer term, I wonder if it would be useful for temporary org buffers to have a way to specify which buffer they came from.  For example in the ELisp checker we load user packages if the file being checked is .emacs or emacs.d/init.el, and so if you edit your .emacs as a literate org file you'll get warnings about missing packages.

**Comment by CyberShadow** [comment]
I think this is ready for review now. Flycheck now seems to work properly in `d-mode` buffers created by Org 9.3.6's `org-edit-src-code`.

File: flycheck.el [hunk 4]
```diff
@@ -5354,6 +5364,12 @@ are substituted within the body of cells!"
      (list (flycheck-save-buffer-to-temp #'flycheck-temp-file-system)))
     (`source-inplace
      (list (flycheck-save-buffer-to-temp #'flycheck-temp-file-inplace)))
+    (`(source ,suffix)
+     (list (flycheck-save-buffer-to-temp
+            (lambda (filename) (flycheck-temp-file-system filename suffix)))))
+    (`(source-inplace ,suffix)
+     (list (flycheck-save-buffer-to-temp
+            (lambda (filename) (flycheck-temp-file-inplace filename suffix)))))
```
cpitclaudel: You could use let to bind variables in patterns to avoid duplicated branches (`(or (and `source (let suffix nil)) `(source ,suffix))`), but I'm not sure it would be more readable here.

cpitclaudel:
Thank you very much for your patience and your perseverance.  I've very happy that we got this PR done :)

CyberShadow:
Thank you too @cpitclaudel :)

PR #1427, opened 2018-03-11
Support errors found in other files

purcell:
This change revives the abandoned PR #1191 by @TyOverby, which in turn built on @chrisdone's PR #972 to address #966. There was some excellent feedback on #1191 by @fmdkdd in [this comment](https://github.com/flycheck/flycheck/pull/1191#issuecomment-273471143), which this PR has incorporated.

- Include errors (but not warnings) from other files in the error list
- When clicking on the error in the error list, jump to the other file
- Having jumped, force a re-check if and only if the other file's
  buffer does not already have that error listed
- Display overlays for errors from other files on the first line
- When showing such errors in the echo area, show the filename too

The hiding of other-file errors is an ongoing source of pain for Flycheck users in some language environments (Haskell, Elm, Rust), so I'm hoping we can get something merged soon.

Having spent a few hours on this, these changes seem to work well in local testing and feel quite natural to me - I'm keen to hear any further feedback! :tada:

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -813,6 +813,12 @@ This variable is a normal hook.  See Info node `(elisp)Hooks'."
   :group 'flycheck-faces
   :package-version '(flycheck . "0.16"))
 
+(defface flycheck-error-list-filename
+  '((t :inherit font-lock-variable-name-face))
+  "Face for filenames in the error list."
+  :group 'flycheck-faces
+  :package-version '(flycheck . "32"))
+
 (defface flycheck-error-list-id
   '((t :inherit font-lock-type-face))
   "Face for the error ID in the error list."
@@ -3083,8 +3089,12 @@ non-whitespace character of the error line, if ERR has no error column."
 (defun flycheck-error-format-message-and-id (err)
   "Format the message and id of ERR as human-readable string."
   (let ((id (flycheck-error-id err))
-        (message (flycheck-error-message err)))
-    (if id (format "%s [%s]" message id) message)))
+        (filename (flycheck-error-filename err)))
+    (concat (when (and filename (not (equal filename (buffer-file-name))))
+              (format "In \"%s\":\n" (file-relative-name filename default-directory)))
+            (flycheck-error-message err)
+            (when id
+              (format " [%s]" id)))))
 
 (defun flycheck-error-format (err &optional with-file-name)
   "Format ERR as human-readable string, optionally WITH-FILE-NAME.
@@ -3187,6 +3197,14 @@ Return ERRORS, modified in-place."
           errors)
   errors)
 
+(defun flycheck-relevant-error-other-file-p (err)
+  "Determine whether ERR is a relevant error for another file."
+  (let ((file-name (flycheck-error-filename err)))
+    (and file-name
+         (or (null buffer-file-name)
+             (not (flycheck-same-files-p buffer-file-name file-name)))
+         (eq 'error (flycheck-error-level err)))))
+
 (defun flycheck-relevant-error-p (err)
   "Determine whether ERR is relevant for the current buffer.
 
@@ -3196,12 +3214,14 @@ otherwise."
     (let ((file-name (flycheck-error-filename err))
           (message (flycheck-error-message err)))
       (and
-       ;; The error is relevant for the current buffer if it's got no file-name
-       ;; and the current buffer has no file name, too, or if it refers to the
-       ;; same file as the current buffer.
-       (or (and (not file-name) (not buffer-file-name))
-           (and buffer-file-name file-name
-                (flycheck-same-files-p file-name buffer-file-name)))
+       (or
+        ;; Neither the error nor buffer have a file name
+        (and (not file-name) (not buffer-file-name))
+        ;; Both have files, and they match
+        (and buffer-file-name file-name
+             (flycheck-same-files-p file-name buffer-file-name))
+        ;; This is a significant error from another file
+        (flycheck-relevant-error-other-file-p err))
        message
        (not (string-empty-p message))
        (flycheck-error-line err)))))
@@ -3716,8 +3736,13 @@ Return the created overlay."
   ;; We must have a proper error region for the sake of fringe indication,
   ;; error display and error navigation, even if the highlighting is disabled.
   ;; We erase the highlighting later on in this case
-  (pcase-let* ((`(,beg . ,end) (flycheck-error-region-for-mode
-                                err (or flycheck-highlighting-mode 'lines)))
+  (pcase-let* ((`(,beg . ,end)
+                (if (flycheck-relevant-error-other-file-p err)
+                    ;; Display overlays for other-file errors on the first line
+                    (cons (point-min)
+                          (save-excursion (goto-char (point-min)) (point-at-eol)))
+                  (flycheck-error-region-for-mode
+                   err (or flycheck-highlighting-mode 'lines))))
                (overlay (make-overlay beg end))
                (level (flycheck-error-level err))
                (category (flycheck-error-level-overlay-category level)))
@@ -3877,10 +3902,10 @@ beginning of the buffer, otherwise advance from the current
 position.
 
 Intended for use with `next-error-function'."
-  (let ((pos (flycheck-next-error-pos n reset)))
-    (if pos
-        (goto-char pos)
-      (user-error "No more Flycheck errors"))))
+  (-if-let* ((pos (flycheck-next-error-pos n reset)))
+      (let ((err (get-char-property pos 'flycheck-error)))
+        (flycheck-jump-to-error err))
+    (user-error "No more Flycheck errors")))
 
 (defun flycheck-next-error (&optional n reset)
   "Visit the N-th error from the current point.
@@ -3940,7 +3965,8 @@ message to stretch arbitrarily far."
             message checker-name)))
 
 (defconst flycheck-error-list-format
-  `[("Line" 5 flycheck-error-list-entry-< :right-align t)
+  `[("File" 6)
+    ("Line" 5 flycheck-error-list-entry-< :right-align t)
     ("Col" 3 nil :right-align t)
     ("Level" 8 flycheck-error-list-entry-level-<)
     ("ID" 6 t)
@@ -4060,6 +4086,7 @@ string with attached text properties."
 Return a list with the contents of the table cell."
   (let* ((level (flycheck-error-level error))
          (level-face (flycheck-error-level-error-list-face level))
+         (filename (flycheck-error-filename error))
          (line (flycheck-error-line error))
          (column (flycheck-error-column error))
          (message (or (flycheck-error-message error)
@@ -4071,7 +4098,12 @@ Return a list with the contents of the table cell."
          (msg-and-checker (flycheck-error-list-make-last-column flushed-msg checker))
          (explainer (flycheck-checker-get checker 'error-explainer)))
     (list error
-          (vector (flycheck-error-list-make-number-cell
+          (vector (flycheck-error-list-make-cell
+                   (if filename
+                       (file-name-nondirectory filename)
+                     "")
+                   'flycheck-error-list-filename)
+                  (flycheck-error-list-make-number-cell
                    line 'flycheck-error-list-line-number)
                   (flycheck-error-list-make-number-cell
                    column 'flycheck-error-list-column-number)
@@ -4233,22 +4265,38 @@ LEVEL is either an error level symbol, or nil, to remove the filter."
 
 POS defaults to `point'."
   (interactive)
-  (-when-let* ((error (tabulated-list-get-id pos))
-               (buffer (flycheck-error-buffer error)))
-    (when (buffer-live-p buffer)
-      (if (eq (window-buffer) (get-buffer flycheck-error-list-buffer))
-          ;; When called from within the error list, keep the error list,
-          ;; otherwise replace the current buffer.
-          (pop-to-buffer buffer 'other-window)
-        (switch-to-buffer buffer))
-      (let ((pos (flycheck-error-pos error)))
-        (unless (eq (goto-char pos) (point))
-          ;; If widening gets in the way of moving to the right place, remove it
-          ;; and try again
-          (widen)
-          (goto-char pos)))
-      ;; Re-highlight the errors
-      (flycheck-error-list-highlight-errors 'preserve-pos))))
+  (-when-let* ((error (tabulated-list-get-id pos)))
+    (flycheck-jump-to-error error)))
+
+(defun flycheck-jump-to-error (error)
+  "Go to the location of ERROR."
+  (let ((error-copy (copy-flycheck-error error))
+        (buffer (find-file-noselect (flycheck-error-filename error))))
+    (setf (flycheck-error-buffer error-copy) buffer)
+    (flycheck-jump-in-buffer buffer error-copy)
+    ;; When jumping to an error in another file, it may not have
+    ;; this error available for highlighting yet, so we trigger a check
+    ;; if necessary.
+    (with-current-buffer buffer
+      (unless (seq-contains flycheck-current-errors error-copy 'equal)
+        (when flycheck-mode
+          (flycheck-buffer))))))
+
+(defun flycheck-jump-in-buffer (buffer error)
+  "In BUFFER, jump to ERROR."
+  (if (eq (window-buffer) (get-buffer flycheck-error-list-buffer))
+      ;; When called from within the error list, keep the error list,
+      ;; otherwise replace the current buffer.
+      (pop-to-buffer buffer 'other-window)
+    (switch-to-buffer buffer))
+  (let ((pos (flycheck-error-pos error)))
+    (unless (eq (goto-char pos) (point))
+      ;; If widening gets in the way of moving to the right place, remove it
+      ;; and try again
+      (widen)
+      (goto-char pos)))
+  ;; Re-highlight the errors
+  (flycheck-error-list-highlight-errors 'preserve-pos))
 
 (defun flycheck-error-list-explain-error (&optional pos)
   "Explain the error at POS in the error list.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,10 @@
   - ``rust-cargo`` now uses ``cargo check`` and ``cargo test`` [GH-1289]
   - ``python-pylint`` and ``python-flake8`` are now invoked with ``python -c``,
     to make it easier to change between Python 2 and Python 3 [GH-1113]
+  - When a checker returns errors for another file, they will be displayed
+    instead of ignored.  They can be navigated to from the error list.
+    This change helps with compiled languages, where an error in another file
+    may cause the current file to be considered invalid. [GH-1427]
 
 31 (Oct 07, 2017)
 =================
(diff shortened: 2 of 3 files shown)


cpitclaudel:
That moment when you get a PR from Steve Purcell :heart_eyes: :1st_place_medal: :heart: 
This looks superb.  I should have time to do a proper review tomorrow. Thanks a lot!

purcell:
> That moment when you get a PR from Steve Purcell

In theory I'm an admin here, but I haven't been active for ages. :-)

cpitclaudel:
> In theory I'm an admin here

I know :) Doesn't reduce the excitement to review your code ^^

purcell:
> I know :) Doesn't reduce the excitement to review your code ^^

Please lower your expectations. ;-)

purcell:
I made a few further tweaks such that only errors from other files are shown, since in Elm - at least - warnings from other files would be displayed in the current file when not strictly necessary. In fact, a more conservative version of this change could choose to show errors from other files only if there are no errors listed for the current file. (I'd suggest it's not necessary to be so conservative.)

fmdkdd:
That looks very exciting!  Thank you @purcell for working on this.

flycheck.el
```diff
@@ -3083,8 +3089,12 @@ non-whitespace character of the error line, if ERR has no error column."
 (defun flycheck-error-format-message-and-id (err)
   "Format the message and id of ERR as human-readable string."
   (let ((id (flycheck-error-id err))
-        (message (flycheck-error-message err)))
-    (if id (format "%s [%s]" message id) message)))
+        (filename (flycheck-error-filename err)))
+    (concat (when (and filename (not (equal filename (buffer-file-name))))
+              (format "In \"%s\": " (file-relative-name filename default-directory)))
```
cpitclaudel: Maybe add a newline after the colon?

cpitclaudel: Should we show the path relative to the default directory, or to the current file? (I'm not sure whether default-directory here is the checker's `:working-directory` or the current buffer's folder)

flycheck.el
```diff
@@ -3187,6 +3197,13 @@ Return ERRORS, modified in-place."
           errors)
   errors)
 
+(defun flycheck-relevant-error-other-file-p (err)
+  "Determine whether ERR is a relevant error for another file."
+  (let ((file-name (flycheck-error-filename err)))
+    (and buffer-file-name file-name
+         (not (flycheck-same-files-p buffer-file-name file-name))
+         (eq 'error (flycheck-error-level err)))))
```
cpitclaudel: Should we make the error level for other files customizable? It might be best to wait until the feature is requested, though.

flycheck.el
```diff
@@ -3187,6 +3197,13 @@ Return ERRORS, modified in-place."
           errors)
   errors)
 
+(defun flycheck-relevant-error-other-file-p (err)
+  "Determine whether ERR is a relevant error for another file."
+  (let ((file-name (flycheck-error-filename err)))
+    (and buffer-file-name file-name
```
cpitclaudel: Do we really need to have a a buffer-file-name here?

flycheck.el
```diff
@@ -3196,12 +3213,14 @@ otherwise."
...
-           (and buffer-file-name file-name
-                (flycheck-same-files-p file-name buffer-file-name)))
+       (or
+        ;; Neither the error nor buffer have a file name
+        (and (not file-name) (not buffer-file-name))
+        ;; Both have files, and they match
+        (and buffer-file-name file-name
+             (flycheck-same-files-p file-name buffer-file-name))
+        ;; This is a significant error from another file
+        (flycheck-relevant-error-other-file-p err))
```
cpitclaudel: This logic feels a bit convoluted: `flycheck-relevant-error-other-file-p` is used in two places: here it simplifies to `(eq 'error (flycheck-error-level err))`, and… (see below)

flycheck.el
```diff
@@ -3716,8 +3735,14 @@ Return the created overlay."
   ;; We must have a proper error region for the sake of fringe indication,
   ;; error display and error navigation, even if the highlighting is disabled.
   ;; We erase the highlighting later on in this case
-  (pcase-let* ((`(,beg . ,end) (flycheck-error-region-for-mode
-                                err (or flycheck-highlighting-mode 'lines)))
+  (pcase-let* ((`(,beg . ,end)
+                (if (flycheck-relevant-error-other-file-p err)
```
cpitclaudel: (continued) …here it simplifies to having different file names, since the error has already been filtered.  I wondered if it'd be better to inline it in both places and simplify accordingly.

flycheck.el
```diff
@@ -3716,8 +3735,14 @@ Return the created overlay."
...
+                    ;; Display overlays for other-file errors on the first line
+                    (cons (point-min) (save-excursion
+                                        (goto-char (point-min))
+                                        (end-of-line) (point)))
```
cpitclaudel: Maybe `point-at-eol`?

flycheck.el
```diff
@@ -3940,7 +3965,8 @@ message to stretch arbitrarily far."
             message checker-name)))
 
 (defconst flycheck-error-list-format
-  `[("Line" 5 flycheck-error-list-entry-< :right-align t)
+  `[("File" 6)
+    ("Line" 5 flycheck-error-list-entry-< :right-align t)
```
cpitclaudel: It it worth adding a column to the error list, vs. adding the file name to the error message? Arguably it's nice to be able to sort by file, but in most cases it'll take space somewhat needlessly.  Could we hide that column when all errors point to the same file?

flycheck.el
```diff
@@ -4233,22 +4265,38 @@ LEVEL is either an error level symbol, or nil, to remove the filter."
...
+      ;; When called from within the error list, keep the error list,
+      ;; otherwise replace the current buffer.
+      (pop-to-buffer buffer 'other-window)
+    (switch-to-buffer buffer))
+  (let ((pos (flycheck-error-pos error)))
+    (unless (eq (goto-char pos) (point))
+      ;; If widening gets in the way of moving to the right place, remove it
+      ;; and try again
+      (widen)
+      (goto-char pos)))
```
cpitclaudel: I wonder why that function didn't just check whether `(<= (point-min) pos (point-max))` (I know it's existing code).

flycheck.el
purcell: I agree we should wait for a request.

flycheck.el
purcell: I'll add the newline and check whether the default directory is the right one here.

purcell: I checked and the default-directory is the correct one, i.e. it's the directory of the current buffer's file.

flycheck.el
purcell: Yes, because `buffer-file-name` can be nil, in which case it's not safe to call the function which checks that the files are the same.

flycheck.el
purcell: See comment below.

flycheck.el
purcell: No, it doesn't simplify to that - either the error's file or the current buffer's file name could be empty, and duplicating that logic from `flycheck-relevant-error-file-p` therefore didn't feel right here. So I extracted `flycheck-relevant-error-other-file-p`, with the result that that function checks the presence of those two filenames - in the context in which it is called in `flycheck-relevant-error-file-p` that check is redundant.

flycheck.el
purcell: Nice!

flycheck.el
purcell: Yes, there are probably some things we could do here, but I'm also keen not to try to gold-plate this before it has seen use in the wild. My suggestion would be to go with this simple approach for now.

flycheck.el
purcell: Hmmm... good point!

purcell:
> I've made a few comments here and there. The main thing I'm unsure about is overlays on the first line of the file. The nice bit about that trick is that it reuses the existing navigation infrastructure quite smoothly, so I could live with it :)

Yeah, that's what I like about it too. It's simple, and it makes those other-file errors obvious, at least.

> It might be nice to add a few tests for navigation (previous/next-error) with other files, too.

Yep, that would be nice for completeness. The existing tests still pass, which is encouraging. ;-)

purcell:
I've pushed those couple of little fixes.

flycheck.el
cpitclaudel: OK, makes sense. Thanks!

flycheck.el
cpitclaudel: Right, but we could also not call that function in that case (if I'm editing a scratch file and that file doesn't compile because of an error in an import, I might still be interested in that error)

flycheck.el
cpitclaudel: OK

flycheck.el
```diff
@@ -4071,7 +4097,12 @@ Return a list with the contents of the table cell."
          (msg-and-checker (flycheck-error-list-make-last-column flushed-msg checker))
          (explainer (flycheck-checker-get checker 'error-explainer)))
     (list error
-          (vector (flycheck-error-list-make-number-cell
+          (vector (flycheck-error-list-make-cell
+                   (if filename
+                       (file-name-nondirectory filename)
+                     "")
```
cpitclaudel: I think this reduces to `(file-name-nondirectory (or filename ""))`… not sure it reads better.

purcell-nec: It has the same effect and is more concise, but it calls `file-name-nondirectory` unnecessarily on `""`, so I'd say keep the `if`. (This came from the original PR, and I didn't find a better way to express it.)

flycheck.el
purcell-nec: Ah, I see! I think you're right. Will fix that in a few hours - currently at work. :-)

cpitclaudel: Thanks!

purcell:
I've tried really hard to find somewhere to add relevant tests, but gave up. With the PR in its current form, there's no real effect on error navigation, either in-buffer, or when clicking on items in the error list. It's not clear how best to set up a situation where errors are generated for a depended-upon file -- from what I can see, this isn't the case for the language resources used by existing tests, and there's no existing coverage of the `flycheck-relevant-error-p`. So I'm feeling a bit defeated on that front, and am inclined to merge regardless. :-/

cpitclaudel:
> am inclined to merge regardless. :-/

So am I :) This is a great PR, and I've nitpicked enough at this point ^^ @fmdkdd, thoughts/comments?  I'll merge tonight from home otherwise.

fmdkdd:
@cpitclaudel I haven't had a chance to take a look yet.  I *should* be able to do so in the next 24h.

cpitclaudel:
No pressure :) I just meant to confirm that I was ready for merging on my end.

purcell:
Alrighty, I'm going for it. If there are any resulting issues which appear related, please CC me into them!

purcell:
Sorry @fmdkdd - if you end up taking a closer look and find anything that bugs you, just follow up here with any feedback and I'll be happy to tinker further.

flycheck.el
```diff
@@ -3877,10 +3902,10 @@ beginning of the buffer, otherwise advance from the current
 position.
 
 Intended for use with `next-error-function'."
-  (let ((pos (flycheck-next-error-pos n reset)))
-    (if pos
-        (goto-char pos)
-      (user-error "No more Flycheck errors"))))
+  (-if-let* ((pos (flycheck-next-error-pos n reset)))
+      (let ((err (get-char-property pos 'flycheck-error)))
```
fmdkdd: Isn't that `let` superfluous since we already have `-if-let*` above?

purcell: Yes, good catch. Fixed in 0ff3c723.

purcell:
> Adding error overlays from other files on the first line actually breaks the integration tests for Rust (and would probably affect other project-based checkers depending on the tests we have). If we want to keep that solution, we'll have to update the tests infrastructure to consider that case.
> What happens if we don't add these overlays? The errors from other files still appear in the error list, and in the modeline, and are caught by next-error, right? Would it be that confusing to not display the squiggles in that case?

I personally think that the overlay on the first line is the single most important part of this change. Flycheck is _most_ commonly used _without_ the error-list, from what I've observed of my own and others' usage, and in that mode of usage it's not sufficient to see only the error count in the modeline: you need an overlay to jump to, and to trigger the display of the error in the echo area.

It's certainly a little ad-hoc to bundle errors onto the first line, but in practical use it seems fairly optimal.

Sorry it broke the Rust tests. Do they run automated somewhere? I'll happily see if I can run them locally and fix them up accordingly.

> In any case, I'm excited to have that feature finally land. I would have moved the changelog item much further up ;)

I actually went ahead and did that, too, because it makes sense to me to mention it before readers zone out. :-)

fmdkdd:
> Do they run automated somewhere? I'll happily see if I can run them locally and fix them up accordingly.

No, unfortunately they don't run automated anywhere.  I've fixed them in #1433, since it was a good opportunity to refresh them.

> you need an overlay to jump to

Ah that's the crucial point.  So `next-error` does not work without that overlay.  In that case then I agree that putting the error on the first line is acceptable.

Thank you again for pushing this :)

fmdkdd:
@chrissound It is enabled by default, if you update Flycheck from MELPA (*not* stable).  Errors from other files will appear on the first line of your buffer (and in the error list), provided that 1) the checker actually emits errors for other files, and 2) you have no errors in the current buffer (if you have only warnings or info messages in the current buffer, errors from other files should appear).

ntc2:
@fmdkdd said:
>>  Adding error overlays from other files on the first line actually breaks the integration tests for Rust (and would probably affect other project-based checkers depending on the tests we have). If we want to keep that solution, we'll have to update the tests infrastructure to consider that case.
>>  What happens if we don't add these overlays? The errors from other files still appear in the error list, and in the modeline, and are caught by next-error, right? Would it be that confusing to not display the squiggles in that case?

@purcell said:
> I personally think that the overlay on the first line is the single most important part of this change. Flycheck is most commonly used without the error-list, from what I've observed of my own and others' usage, and in that mode of usage it's not sufficient to see only the error count in the modeline: you need an overlay to jump to, and to trigger the display of the error in the echo area.

Some feedback after using this feature for a few days now: I like having the overlay on the first line for errors in other files, but I'd like to ability to jump to the overlay (with `flycheck-{previous,next}-error`) without actually jumping to the other file containing the error. I find it jarring/surprising when I'm editing one file, I try to jump to the next error, and suddenly I'm in another file. I think it would be less surprising if the first jump brought to the overlay, and I then had to jump again to actually change files.

fmdkdd:
> I find it jarring/surprising when I'm editing one file, I try to jump to the next error, and suddenly I'm in another file.

Well, maybe it's because you are not used to it yet? ;)

>  I think it would be less surprising if the first jump brought to the overlay, and I then had to jump again to actually change files.

We could add an option along these lines: if it's an error from another file, jump to the in-buffer overlay, /unless/ the cursor is already on the overlay, and in that case we jump to the other file instead.

Do you want to make a PR?

ntc2:
On Tue, May 22, 2018 at 1:48 AM, fmdkdd <notifications@github.com> wrote:
>
> I find it jarring/surprising when I'm editing one file, I try to jump to the next error, and suddenly I'm in another file.
>
> Well, maybe it's because you are not used to it yet? ;)

Perhaps :)

> I think it would be less surprising if the first jump brought to the overlay, and I then had to jump again to actually change files.
>
> We could add an option along these lines: if it's an error from another file, jump to the in-buffer overlay, /unless/ the cursor is already on the overlay, and in that case we jump to the other file instead.
>
> Do you want to make a PR?

Sure. I've never worked on an Emacs package locally before (I use Cask
to install flycheck). Do you have a pointer to some docs on developing
and testing Emacs packages in general, or flycheck specifically?

fmdkdd:
> Do you have a pointer to some docs on developing and testing Emacs packages in general, or flycheck specifically?

We [do](http://www.flycheck.org/en/latest/contributor/contributing.html#the-build-system).  The Emacs Lisp manual inside Emacs is also invaluable if you want to brush up on Elisp programming. 

For this specific feature, you want to look at `flycheck-jump-to-error` in `flycheck.el`.  You can use `flycheck-overlay-errors-at` to get the errors under the cursor, and compare that to the error you need to jump to.  

If you want to add tests, you should add them in `specs/test`, maybe under new file `test-error-navigation.el`.

Do not hesitate to ask further questions.

ntc2:
@fmdkdd I just submitted PR #1473 in relation to our discussion.

I did not add any tests, because I don't understand how the tests work, and above @purcell [mentioned he couldn't figure out where to add tests](https://github.com/flycheck/flycheck/pull/1427#issuecomment-372220605) for his new first-line overlays feature. I'm testing locally with two Haskell files, where one imports the other, and the imported file doesn't type check.

As part of my testing, I found a bug in `flycheck-first-error` / `flycheck-next-error-pos`, and I'm looking into fixing that separately. The problem is that `flycheck-next-error-pos` assumes that the first position in the file, `(point-min)`, is not part of an error overlay, but that's untrue for the new first-line error overlays. So, tests for the new first-line error overlays might be a good idea.

I didn't ask very clearly, but when I asked about docs for developing Emacs packages, what I wanted to know was how to run my locally modified version of the package, preferably without affecting my global Emacs installation. I found good docs for that in [this Stack Overflow answer](https://emacs.stackexchange.com/questions/30044/how-do-i-patch-an-emacs-package/30060#30060), and ended up using the `emacs -L <path to my local flycheck repo>` approach. I would like to add a pointer to this in the contributor docs, if you're OK with that.

ntc2:
@fmdkdd I fixed the `flycheck-first-error` problem I mentioned above (fix: https://github.com/flycheck/flycheck/pull/1473/commits/3f0788f5760406df59803f5687136ab6ce4fb90a) and a similar `flycheck-next-error` problem (fix: https://github.com/flycheck/flycheck/pull/1473/commits/d174f558ce7cb2b7c108720be5aef17c3af53bcc) that I discovered after that. Both are related to the new other-file overlays introduced by #1427, so more evidence that some tests would be good.

The current specification for how Flycheck error navigation (`flycheck-{next,previous,first}-error`) works with relation to other-file error overlays is:

- if the cursor is already on an other-file overlay, then that overlay is considered to be the next/previous overlay. This makes jumping work correctly in this case, i.e. by jumping to the other file that already contains the error.

- if the cursor is not already on an other-file error overlay, then the next/previous/first overlay is the next/previous in relation to the cursor, or first, respectively. If the overlay determined in this way is an other-file overlay, then what happens when you jump to it is determined by the setting of `flycheck-jump-to-other-files-immediately`: if that setting is non-nil, the default, then jump to the error in the other file; if that setting is nil, then jump to the other-file overlay in the current file.

fmdkdd:
> I would like to add a pointer to this in the contributor docs, if you're OK with that.

That's a good idea, please do.  I've used `emacs -l/-L` in for reproducing bugs as well.  The file `test/init.el` can also be used to launch an Emacs session with Flycheck, without impacting the local Emacs configuration.

For the rest, I'll move the discussion to #1427 directly.

raxod502:
> what I wanted to know was how to run my locally modified version of the package, preferably without affecting my global Emacs installation

Another option for this is [using `straight.el`](https://github.com/raxod502/straight.el/tree/55f966b11bed4f82b3f91ab15edbe0f890c75532#using-straightel-to-reproduce-bugs). The "without affecting my global Emacs installation" part is covered because you can either:

1. Just check out the feature branch, and run your usual Emacs configuration, or
2. Check out the feature branch, and then load the package from `emacs -Q`,

and in neither case is any change to your `~/.emacs.d` necessary.

PR #1432, opened 2018-03-12
added scheme-chicken args variable

aeggenberger:


fmdkdd:
Thank you for the contribution. We try to shy away from catch-all arguments variables, if we can avoid them.

Do you have a specific use-case in mind?  Specific variables that you want to pass to the checker?

aeggenberger:
The compiler has a few options that enable [type checking](https://wiki.call-cc.org/man/4/Types): `-specialize`,  `-lfa2` and `-strict-types`. Would it be better to have a Boolean variable that adds them when truthy?

flycheck.el
```diff
@@ -9726,11 +9726,16 @@ See URL `http://www.scalastyle.org'."
                        ((not config-file) '(bold error))
                        (t 'success)))))))
 
+(flycheck-def-args-var flycheck-scheme-chicken-args scheme-chicken
+  :package-version '(scheme-chicken . "0.31"))
```
fmdkdd: `:package-version` records the *Flycheck* version this variable has been added.  So it should rather be `'(flycheck . "32")`.

fmdkdd:
> The compiler has a few options that enable type checking: -specialize, -lfa2 and -strict-types. Would it be better to have a Boolean variable that adds them when truthy?

Ah well, in that case it might not be worth it.  WDYT @cpitclaudel ?

flycheck.el
aeggenberger: Changed to `'(flycheck . "32")`.

flycheck.el
```diff
@@ -9726,11 +9726,16 @@ See URL `http://www.scalastyle.org'."
                        ((not config-file) '(bold error))
                        (t 'success)))))))
 
+(flycheck-def-args-var flycheck-scheme-chicken-args scheme-chicken
+  :package-version '(scheme-chicken . "0.32"))
```
fmdkdd: Replace `scheme-chicken` by `flycheck`, and `0.32` by `32`.

cpitclaudel:
> Ah well, in that case it might not be worth it. WDYT @cpitclaudel ?

I think I like individual options better :) But I'm fine either way

fmdkdd:
Last thing @aeggenberger: please squash your commits into one before we can merge.

fmdkdd:
> I think I like individual options better :) But I'm fine either way

One thing we can do after we add these catch-all variables is to use the Github code search to find out what kind of flags come up often, and add these as proper options.

aeggenberger:
Squashed and ready, I hope.

fmdkdd:
Yep, thank you! :+1:

PR #1439, opened 2018-03-24
Add flycheck-sh-bash-args

scop:


fmdkdd:
Do you have an example of useful arguments that you pass to bash?

scop:
Sure. `-O extglob` is what I would need in https://github.com/scop/bash-completion. Currently I'm using this fragile hack: https://github.com/scop/bash-completion/commit/b49a182cc68a9d69d702754c3adc0b17d50e52f6#diff-939eb0c23e9efb46517d67d1760df731

fmdkdd:
Ok thanks! :+1:

Thaodan:
How did  you add -O extglob when bash is executed using this?

PR #1440, opened 2018-03-28
Don't autoquote lambdas in `flycheck-define-checker`

### LINKED ISSUE #1398: error-filter is not a function
I'm using Spacemacs and I have a custom checker script based off of clang-checker. When Spacemacs loads, I get this error:

```
(Spacemacs) Error in dotspacemacs/user-config: :error-filter (function (lambda (errors) (let ((errors (flycheck-sanitize-errors errors))) (let ((--dolist-tail-- errors) err) (while --dolist-tail-- (setq err (car --dolist-tail--)) (let* ((v err)) (\(setf\ flycheck-error-message\) (or (flycheck-error-message err) "no message") v)) (setq --dolist-tail-- (cdr --dolist-tail--)))) (flycheck-fold-include-levels errors "In file included from")))) of syntax checker myscript is not a function
```

This is my Spacemacs user-config code:
```
(defun dotspacemacs/user-config ()
  (with-eval-after-load "flycheck"
    (flycheck-define-checker myscript
      :command ("myscript" source-original)
      :standard-input t
      :error-patterns
      ((error line-start
              (message "In file included from") " " (or "<stdin>" (file-name))
              ":" line ":" line-end)
       (info line-start (or "<stdin>" (file-name)) ":" line ":" column
             ": note: " (optional (message)) line-end)
       (warning line-start (or "<stdin>" (file-name)) ":" line ":" column
                ": warning: " (optional (message)) line-end)
       (error line-start (or "<stdin>" (file-name)) ":" line ":" column
              ": " (or "fatal error" "error") ": " (optional (message)) line-end))
      :error-filter
      (lambda (errors)
        (let ((errors (flycheck-sanitize-errors errors)))
          (dolist (err errors)
            ;; Clang will output empty messages for #error/#warning pragmas without
            ;; messages.  We fill these empty errors with a dummy message to get
            ;; them past our error filtering
            (setf (flycheck-error-message err)
                  (or (flycheck-error-message err) "no message")))
          (flycheck-fold-include-levels errors "In file included from")))
      :modes c++-mode))
  )
```
I have also tried replacing `with-eval-after-load` with
```
(use-package flycheck
    :config
   (flycheck-define-checker myscript <...>)   
)
```
but I still see the error.

When I execute the expression manually after Spacemacs has loaded, I do not see the error.

fmdkdd:
In `flycheck-define-checker`, we function quoted (`#'`) all arguments
except `:command`.  This made for more readable checker definitions,
since we could pass lambdas or unquoted symbols to functions.

While this works in Flycheck itself, quoting lambdas fails when using
`with-eval-after-load`.  In the body of the latter, macros will be
expanded, and a `(lambda ...)` form is macro-expanded to
`(function (lambda ...)` resulting in a doubly-quoted lambda, which is
incorrect.

This commit solves the issue by not auto-quoting the arguments to
`flycheck-define-checker`.  For backward compatibility, it
function-quotes unquoted symbols, and emits a warning indicating that
this use is deprecated.

Fix GH-1398

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6100,9 +6100,17 @@ objects)."
   ;; future, `flycheck-define-checker' should not accept unquoted symbols, and
   ;; thus `flycheck-maybe-quote' will be unnecessary.
   ;; See https://github.com/flycheck/flycheck/issues/1398
-  (defun flycheck-maybe-quote (arg)
+  (defun flycheck-maybe-quote-function (arg)
     "Function quote ARG if it is a symbol, return ARG otherwise."
-    (if (and (symbolp arg) arg) `(function ,arg) arg)))
+    (if (and (symbolp arg) arg) `(function ,arg) arg))
+
+  ;; We also quote non-function properties for backward-compatibility.  In the
+  ;; future, unquoted arguments should be invalid.
+  (defun flycheck-maybe-quote (arg)
+    "Quote ARG unless it is already quoted, or is a boolean."
+    (if (or (booleanp arg) (and (listp arg) (eq 'quote (car arg))))
+        arg
+      `(quote ,arg))))
 
 (defmacro flycheck-define-checker (symbol docstring &rest properties)
   "Define SYMBOL as command syntax checker with DOCSTRING and PROPERTIES.
@@ -6112,37 +6120,42 @@ be quoted.  Also, implicitly define the executable variable for
 SYMBOL with `flycheck-def-executable-var'."
   (declare (indent 1)
            (doc-string 2))
-  (let ((command                           (plist-get properties :command))
-        (parser      (flycheck-maybe-quote (plist-get properties :error-parser)))
-        (filter      (flycheck-maybe-quote (plist-get properties :error-filter)))
-        (explainer   (flycheck-maybe-quote (plist-get properties :error-explainer)))
-        (predicate   (flycheck-maybe-quote (plist-get properties :predicate)))
-        (enabled-fn  (flycheck-maybe-quote (plist-get properties :enabled)))
-        (verify-fn   (flycheck-maybe-quote (plist-get properties :verify)))
-        (work-dir-fn (flycheck-maybe-quote (plist-get properties :working-directory))))
+  (let ((command        (flycheck-maybe-quote (plist-get properties :command)))
+        (patterns       (flycheck-maybe-quote (plist-get properties :error-patterns)))
+        (modes          (flycheck-maybe-quote (plist-get properties :modes)))
+        (next-checkers  (flycheck-maybe-quote (plist-get properties :next-checkers)))
+        (standard-input (flycheck-maybe-quote (plist-get properties :standard-input)))
+        (parser      (flycheck-maybe-quote-function (plist-get properties :error-parser)))
+        (filter      (flycheck-maybe-quote-function (plist-get properties :error-filter)))
+        (explainer   (flycheck-maybe-quote-function (plist-get properties :error-explainer)))
+        (predicate   (flycheck-maybe-quote-function (plist-get properties :predicate)))
+        (enabled-fn  (flycheck-maybe-quote-function (plist-get properties :enabled)))
+        (verify-fn   (flycheck-maybe-quote-function (plist-get properties :verify)))
+        (work-dir-fn (flycheck-maybe-quote-function (plist-get properties :working-directory))))
 
     `(progn
        (flycheck-def-executable-var ,symbol ,(car command))
 
        (flycheck-define-command-checker ',symbol
          ,docstring
-         :command ',command
+         :command ,command
          ,@(when parser
              `(:error-parser ,parser))
-         :error-patterns ',(plist-get properties :error-patterns)
+         :error-patterns ,patterns
          ,@(when filter
              `(:error-filter ,filter))
          ,@(when explainer
              `(:error-explainer ,explainer))
-         :modes ',(plist-get properties :modes)
+         :modes ,modes
          ,@(when predicate
              `(:predicate ,predicate))
-         :next-checkers ',(plist-get properties :next-checkers)
+         ,@(when next-checkers
+             `(:next-checkers ,next-checkers))
          ,@(when enabled-fn
              `(:enabled ,enabled-fn))
          ,@(when verify-fn
              `(:verify ,verify-fn))
-         :standard-input ',(plist-get properties :standard-input)
+         :standard-input ,standard-input
          ,@(when work-dir-fn
              `(:working-directory ,work-dir-fn))))))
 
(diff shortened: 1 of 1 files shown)


fmdkdd:
This is one solution to #1398.

I've opted to emit a deprecation warning right away.   This would mean that we need to update all Flycheck checkers to avoid the warning.  I can do that in the same PR if we agree on this fix.

fmdkdd:
Ah yes, `eval-and-compile` did the trick.  That byte compiler is killing me.

So we can go either way, using `eval-and-compile` or `cl-flet`.  We already have two instances of the former, none of the latter.

cpitclaudel:
I'd vote for eval-and-compile here

fmdkdd:
Sure thing, updated.

flycheck.el
```diff
@@ -6082,6 +6082,18 @@ objects)."
 
 
 ;;; Convenience definition of command-syntax checkers
+(eval-and-compile
+  ;; We need the byte-compiler to be aware of this function when compiling
+  ;; `flycheck-define-checker'.
```
cpitclaudel: Technically, when macroexpanding flycheck-define-checker.
I think it'd also be good to document why this function is a problem (namely, the byte-compiler needs to be aware of it, but when-eval-after-load can run before that.

flycheck.el
```diff
@@ -6082,6 +6082,18 @@ objects)."
...
 

 ;;; Convenience definition of command-syntax checkers
+(eval-and-compile
+  ;; We need the byte-compiler to be aware of this function when compiling
+  ;; `flycheck-define-checker'.
+  (defun flycheck-maybe-quote (arg)
+    "Quote ARG if it is a symbol, return ARG otherwise."
+    (let ((a arg))
+      (if (and (symbolp a) (not (null a)))
```
cpitclaudel: Isn't `not null a` the same as `a`?
Also, why do you let-bind `a`?

flycheck.el
```diff
@@ -6082,6 +6082,18 @@ objects)."
...
+          (progn
+            (warn "Passing unquoted function symbols to `flycheck-define-checker' is deprecated")
```
cpitclaudel: I'd vote to delay this warning and ping other checkers about the needed fix.

cpitclaudel (commented):
Thanks. I think this one will be a bit of an issue backwards-compatibility wise.   I don't think external checkers will want to update until they can guarantee that their clients are on a new Flycheck release, right? Their fixed checkers won't be compatible with the old Flycheck (it will double-quote), right?

Given this, we'll need to provide them with a new stable release with the new quoting behavior, and have them update their dependency declaration.

IIUC, this means that we shouldn't start warning yet, because we don't have a safe fix yet.  Once we have a stable release with the autoquoting behavior optional we can deprecate the current form and warn, and after a while drop support for the old form.

flycheck.el
fmdkdd: Ah, I started wrting it as a macro, then found it was unnecessary.

fmdkdd:
I hadn't thought of the dependency on stable version.  I agree that in that case we should wait.

I've maybe-quoted `:working-directory` as well, since it's usually a function.  However, there is an argument to be made to allow lambdas for the remaining properties (`:next-checkers`, `:standard-input` and `:modes`), see [this example](https://gitter.im/flycheck/flycheck?at=5a43b0b529ec6ac3118c7a49).

cpitclaudel:
> I've maybe-quoted :working-directory as well, since it's usually a function. However, there is an argument to be made to allow lambdas for the remaining properties (:next-checkers, :standard-input and :modes), see this example.

I'm not sure I follow. Don't we want to allow lambdas for all properties that currently accept functions?
I'm also not sure I understand the example :/ It looks like the OP wants the definition to update as the flycheck-chaining-error-level setting is updated, but IIUC that won't work (the value will be read at definition time, and won't update automatically).

fmdkdd:
> I'm not sure I follow. Don't we want to allow lambdas for all properties that currently accept functions?

We do, and that's why I've added `working-directory` to the mix.

>  looks like the OP wants the definition to update as the flycheck-chaining-error-level setting is updated, 

Ah well, my understanding of that example was that using the value of their custom variable at definition time was enough.  But that was not possible (and still isn't with this PR currently) as we invariable quote the non-function properties like `:next-checker`.

So, since we are here, we could consider removing that quote to allow that scenario.  Although, we could also leave it as-is, as all checker properties can be modified after definition using `(setf (flycheck-checker-get ...`.

One last thing for this PR is whether we want to function-quote all the unquoted function symbols in our own checker definitions.  Now that I'm considering it, it would be slightly less readable than our current checker definitions.  But, if we want to deprecate that usage in the future, then might as well do the right thing now.

cpitclaudel:
> We do, and that's why I've added working-directory to the mix.

Super, thanks.

> Ah well, my understanding of that example was that using the value of their custom variable at definition time was enough. But that was not possible (and still isn't with this PR currently) as we invariable quote the non-function properties like :next-checker.

That sounds suspicious to me — having to reload a package for a setting to apply seems rather painful.

> So, since we are here, we could consider removing that quote to allow that scenario. Although, we could also leave it as-is, as all checker properties can be modified after definition using (setf (flycheck-checker-get ....

I like that option better.

> One last thing for this PR is whether we want to function-quote all the unquoted function symbols in our own checker definitions. Now that I'm considering it, it would be slightly less readable than our current checker definitions. But, if we want to deprecate that usage in the future, then might as well do the right thing now.

Agreed. We probably should.


But part of me still isn't convinced that this is a bug we should fix at all. AFAIC, this is an Emacs bug :/

fmdkdd:
> But part of me still isn't convinced that this is a bug we should fix at all. AFAICT, this is an Emacs bug :/

Ah well, I agree that the eager macroexpansion under unknown symbols done by the byte-compiler is a semantic trap.  But I recall the conclusion on emacs-devel was "it's not supposed to work".  Lisps in general have a bad macro-hygiene story, so I'm not even sure that fixing the root cause in Emacs would be feasible without breaking stuff left and right.

Since double-quoting lambdas result in Flycheck users being unable to add `flycheck-define-checker` calls to their own config, and since we can actually fix that on our side, landing that patch seems to me the best course of action.

cpitclaudel:
> landing that patch seems to me the best course of action.

Agreed.

flycheck.el
```diff
@@ -6085,6 +6085,22 @@ objects)."
...
+;; `flycheck-define-checker'.
+(eval-and-compile
+  ;; We allow unquoted function symbols and lambdas as arguments to
+  ;; `flycheck-define-checker'.  When that macro is defined, like in this file,
+  ;; it works fine.  However, when wrapping `flycheck-define-checker' in a
+  ;; `with-eval-after-load flycheck', the body of the latter is macroexpanded
+  ;; *before* flycheck.el has been loaded, and thus before the macro is defined.
+  ;; Unquoted function symbols are still unquoted, but (lambda ...) forms are
+  ;; macroexpanded into (function (lambda ...)), and thus should *not* be
+  ;; quoted.  We use `flycheck-maybe-quote' to avoid double-quoting.
```
cpitclaudel: I'd add a link to one of the Github issues and a note that we intend to get rid of this function after a while.

flycheck.el
```diff
@@ -6085,6 +6085,22 @@ objects)."
...
+  (defun flycheck-maybe-quote (arg)
+    "Function quote ARG if it is a symbol, return ARG otherwise."
+    (if (and (symbolp arg) arg) `#',arg arg)))
```
cpitclaudel: I think this would read better as `` `(function ,arg)``

cpitclaudel (commented):
Thanks. This looks good.
I have one more question: since we're quoting function, should we also go ahead and quote other properties (modes, etc), too?

fmdkdd:
> should we also go ahead and quote other properties (modes, etc), too?

If we quote them in checker definitions, then we also need to make sure we don't double-quote them, since these properties are always quoted in `flycheck-define-checker`.

I would be in favor of allowing them to be quoted, if only because it could confuse users to see that some properties need to be quoted, and some shouldn't.

fmdkdd:
@cpitclaudel Updated.  Let me know what you think.

The last thing that is automatically quoted is the checker symbol.  If we want to turn `flycheck-define-checker` into a regular function in the future, we should accept a quoted version as well.  Shall I make the change?

cpitclaudel:
I've tried an alternative fix in #1710.  With that fix in place, I think it would be enough to deprecate flycheck-define-checker in favor of flycheck-define-command-checker (though it wouldn't be needed).  So I'm in favor of closing this.

purcell:
Commenting here without reading the details, but this sort of change is good, IMO. From the outside, args passed to a macro should work pretty much like args passed to a function, otherwise people will naturally get confused. The existing evaluation rules always seemed odd to me. It's not even just about lambdas: Autoquoting also stops people from passing forms which - when evaluated - will return a function.

cpitclaudel:
From the outside, args passed to a macro should work pretty much like args passed to a function, otherwise people will naturally get confused.

I agree, but we already have flycheck-define-command-checker, which is the function version of that macro.  Adding complexity to this macro to allow it being used in both forms doesn't help much, I think.

> The existing evaluation rules always seemed odd to me. It's not even just about lambdas: Autoquoting also stops people from passing forms which - when evaluated - will return a function.

Agreed to, so we could push people in the direction of flycheck-define-command-checker (the function) instead of flycheck-define-checker (the macro)

fmdkdd:
I'm fine with dropping this to avoid breaking setups.  We should document the quoting issue and the use of `flycheck-define-command-checker` in the manual though.

cpitclaudel:
> I'm fine with dropping this to avoid breaking setups. We should document the quoting issue and the use of flycheck-define-command-checker in the manual though.

Yup, sounds good.

fmdkdd:
I've added a section to the developer guide.

fmdkdd:
Closing now that we have mitigated the issue.

PR #1443, opened 2018-04-01
Prevent `stack` from installing GHC

### LINKED ISSUE #1417: Emacs freezes when opening up a haskell file with flycheck on
I installed `flycheck` a couple days ago and I really love it, but whenever I open up a Haskell file, with [[https://github.com/haskell/haskell-mode/][haskell-mode]] on, Emacs would just freeze. I tried opening up `htop` to see if there was any `stack` process running, but that is not the case. I can unfreeze Emacs by typing `C-g` repeatedly, and it says `flycheck` is on, but no syntax checking is provided; although, whever I start typing something it freezes again until I stop it with `C-g`. I can't get an output for `M-x flycheck-verify-setup` because, again, it freezes until I repeatedly type `C-g`. I had `flycheck-haskell` installed and removed it, but nothing changed. The same exact issue repeats itself with `emacs -Q`. I could grab this error message from the `*Messages*` buffer (this was suggested by a member of the `flycheck` Gitter chat), it said:

```
Error in post-command-hook (flycheck-perform-deferred-syntax-check): (quit)
Waiting for process to die...done
Quit
```

Operating system: Void Linux
Flycheck version: 32snapshot (package: 2018224.1046)
GNU Emacs 25.3.1 (x86_64-unknown-linux-gnu, X toolkit) of 2017-12-16

fmdkdd:
This should fix #1417.

Although since we have an `ignore-errors` in there, I'm not sure what happens if `stack` cannot find GHC.  Probably `output` is nil, then we return the location of the cabal file, if any.

Maybe we need to pass `--no-install-ghc` to the `stack` invocation in `haskell-stack-ghc` as well?

WDYT @flycheck/haskell ?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7924,7 +7924,9 @@ contains a cabal file."
          "stack.*\\.yaml\\'"))
       (-when-let* ((stack (funcall flycheck-executable-find "stack"))
                    (output (ignore-errors
-                             (process-lines stack "path" "--project-root")))
+                             (process-lines stack
+                                            "--no-install-ghc"
+                                            "path" "--project-root")))
                    (stack-dir (car output)))
         (and (file-directory-p stack-dir) stack-dir))))
     (_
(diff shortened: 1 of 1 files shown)


purcell:
> Maybe we need to pass --no-install-ghc to the stack invocation in haskell-stack-ghc as well?

Yes, I'd suggest adding that to any checker which uses `stack`.

fmdkdd:
@flycheck/haskell Updated.  Do you think we can land this?

PR #1444, opened 2018-04-03
Call `ansi-color-filter-apply` in nix checker error parser

marsam:
Nix 2.0 outputs ANSI terminal colors, we strip those colors before passing it to the default behavior.

tests:

```
make LANGUAGE=nix integ
cask exec emacs -Q --batch -L .  --load test/flycheck-test --load test/run.el -f flycheck-run-tests-main '(and (tag external-tool) (language nix))'
Running tests on Emacs 25.3.1, built at 2018-02-19
Running 1 tests (2018-04-03 09:45:41-0500)
   passed  1/1  flycheck-define-checker/nix/default

Ran 1 tests, 1 results as expected (2018-04-03 09:45:42-0500)

```

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -89,6 +89,7 @@
 ;; Tell the byte compiler about autoloaded functions from packages
 (declare-function pkg-info-version-info "pkg-info" (package))
 
+(autoload 'ansi-color-filter-apply "ansi-color")
 
 ;;; Compatibility
 (eval-and-compile
@@ -9217,6 +9218,9 @@ See URL `https://github.com/mivok/markdownlint'."
 See URL `https://nixos.org/nix/manual/#sec-nix-instantiate'."
   :command ("nix-instantiate" "--parse" "-")
   :standard-input t
+  :error-parser
+  (lambda (output checker buffer)
+    (flycheck-parse-with-patterns (ansi-color-filter-apply output) checker buffer))
   :error-patterns
   ((error line-start
           "error: " (message) " at " (file-name) ":" line ":" column


fmdkdd:
Thanks.  Shouldn't `nix` have an option to *not* output these colors instead?  I appreciate that it's not hard to do on our side, but other tools usually have that option.

benley:
This is getting fixed upstream in Nix, but it isn't in a stable release version yet and it might not be for a little while.  In the mean time it would be lovely to have Flycheck handle it :-)

Fix is in https://github.com/NixOS/nix/commit/486872150638d56483c2bc429ba9e137d9974ee8, discussion in https://github.com/NixOS/nix/issues/2045

flycheck.el
```diff
@@ -9217,6 +9218,9 @@ See URL `https://github.com/mivok/markdownlint'."
 See URL `https://nixos.org/nix/manual/#sec-nix-instantiate'."
   :command ("nix-instantiate" "--parse" "-")
   :standard-input t
+  :error-parser
+  (lambda (output checker buffer)
+    (flycheck-parse-with-patterns (ansi-color-filter-apply output) checker buffer))
```
cpitclaudel: I wonder whether it's best to add an autoload or to just require `'ansi-color` in that lambda.

benley: fyi the `erlang-rebar3` checker (which does exactly the same as this parser) has `(require 'ansi-color)`, so you may want to change them together for consistency's sake

marsam:
Hi, sorry for the late response, I was really busy these days. Looks like nix 2.0.1 (already in stable channels) included https://github.com/NixOS/nix/commit/486872150638d56483c2bc429ba9e137d9974ee8 which fixes this issue

Issue #1447, opened 2018-04-10
command detection cannot handle rbenv shims


kaikuchn:
Hey folks,

I recently updated all my packages and am now seeing warnings about suspicious state for the ruby-reek syntax checker. Flycheck does find a ruby-reek executable, albeit in a different Ruby version, and thus enables the syntax checker. When it then tries to call the executable a `command not found` with exit status 127 is returned hinting that in another installed Ruby version the command is available.

It seems that the availability check of the command cannot deal with rbenv shims. This is only a problem if in one of the installed ruby versions the executable exists. By deleting it from that version I can circumvent this problem.

Here's the flycheck-verify-setup output that already includes the requested version information:

```
Syntax checkers for buffer foo.rb in ruby-mode:

  chef-foodcritic (disabled)
    - may enable: Automatically disabled!
    - predicate:  nil
    - executable: Not found

  ruby-rubocop
    - may enable:         yes
    - executable:         Found at /Users/kaikuchn/.rbenv/shims/rubocop
    - configuration file: Not found

  ruby-reek
    - may enable:         yes
    - executable:         Found at /Users/kaikuchn/.rbenv/shims/reek
    - configuration file: Not found

  ruby-rubylint (disabled)
    - may enable:         Automatically disabled!
    - executable:         Not found
    - configuration file: Not found

  ruby
    - may enable: yes
    - executable: Found at /Users/kaikuchn/.rbenv/shims/ruby

  ruby-jruby
    - may enable: Automatically disabled!
    - executable: Not found

Flycheck Mode is enabled. Use SPC u C-c ! x to enable disabled checkers.

--------------------

Flycheck version: 32snapshot (package: 20180326.836)
Emacs version:    25.3.1
System:           x86_64-apple-darwin17.0.0
Window system:    ns
```

Edit: Rbenv provides [a way to check](https://github.com/rbenv/rbenv#rbenv-which) if the currently active ruby version does have the executable.

fmdkdd:
I'm confused.   How do you call `reek` when rbenv shims are in use.  Isn't just `reek`?  In other words, how should flycheck invoke `reek` in this case?

fmdkdd:
Thanks for the explanation.  Flycheck simply looks up for `reek` on your `PATH` and calls that if it is found.  We can see that Flycheck found the `reek` shim created by rbenv and called that.  The call failed because `reek` is not installed for the currently active ruby version. 

[...]

PR #1449, opened 2018-04-11
[WIP] Handle rbenv shims for Ruby checkers

fmdkdd:
This is a proof of concept for fixing #1447.

@kaikuchn Can you try this patch and see if it works for you?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9344,6 +9344,27 @@ See URL `http://batsov.com/rubocop/'."
   :safe #'string-or-null-p
   :package-version '(flycheck . "30"))
 
+(defun flycheck-ruby-rbenv-shim-p (program)
+  "Whether PROGRAM is an rbenv shim.
+
+See URL `https://github.com/rbenv/rbenv'."
+  (-when-let (rbenv-root (getenv "RBENV_ROOT"))
+    (string-prefix-p (expand-file-name "shims" rbenv-root)
+                     (funcall flycheck-executable-find program))))
+
+(defun flycheck-ruby-rbenv-shim-path (program)
+  "Return the full path to a Ruby SHIM.
+
+If PROGRAM is an rbenv shim, return the value of `rbenv which
+PROGRAM'.  If the shim is not available for the current version,
+or if the program is not a shim, return nil."
+  (when (flycheck-ruby-rbenv-shim-p program)
+    (let* ((ret)
+           (path (with-output-to-string
+                   (setq ret (call-process "rbenv" nil standard-output nil
+                                           "which" program)))))
+      (and (= ret 0) path))))
+
 (flycheck-define-checker ruby-reek
   "A Ruby smell checker using reek.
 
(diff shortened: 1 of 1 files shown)


kaikuchn:
This patch mostly works for me. One caveat is that switching versions seems to break things. The checker will remain active, but the executable isn't found? Let me explain in detail:

Installing `reek` for Ruby `2.3.1` and using Ruby `2.5.0`, `flycheck-verify-setup` returns
```
  ruby-reek (disabled)
    - may enable:         yes
    - executable:         Found at /Users/kaikuchn/.rbenv/shims/reek
    - configuration file: Not found
```

The check is disabled as expected.

If I change the local version for the `~/foo` directory to `2.3.1` and then start emacs in a shell with PWD `~/foo`, then the ruby-reek checker is enabled as expected and it works.

So far so good.

Now there are two cases where things break:
1. When I switch Ruby to `2.3.1` for a __shell session__ (`rbenv shell 2.3.1`) and then start emacs in that shell, the checker is enabled, as is expected. __However__, the active Ruby version that emacs uses seems to be the local version (or global if no local override exists, in my case it is the local `2.5.0`) and __not__ the shell session specific Ruby version. Thus the executable again is not available and the ruby-reek checker does not work.
2. When I start emacs while the local version is `2.3.1` the ruby-reek checker is enabled, switching to `2.5.0` then switches the version emacs uses but the checker will remain active. Thus the exectuable is again not available.

fmdkdd:
Wait, I thought you only wanted to enhance `flycheck-verify-setup` for detecting the case where `reek` was installed for another Ruby version.  It seems that instead you expected the patch to be able to always run the one version of `reek` that is installed, regardless of the curent Ruby version.

Am I misunderstanding?

Simplify:
@fmdkdd I'm not sure we should bother with this PR. `ruby-reek` is not only one affected but anything ruby related that `rbenv` makes shims of: `rubocop`, `rubylint`, `ruby`, `jruby`, `sass` and `scss` that I know of. This is something that user can solve by installing checker in current ruby version or by disabling checker in Flycheck. This is also hard to solve right because user can change ruby version in other terminal and that will affect Emacs that is already running as described in previous comment by @kaikuchn. 

If you want to run command that is not installed in your current rbenv ruby version, but in some other, you can try to add this to `.dir-locals.el`:

``` elisp
((ruby-mode . ((eval . (setq flycheck-command-wrapper-function
                             (lambda (command)
                               (append '("RBENV_VERSION=x.y.z") command)))))))
```

That should set rbenv ruby version to x.y.z during command execution.

kaikuchn:
> Wait, I thought you only wanted to enhance flycheck-verify-setup for detecting the case where reek was installed for another Ruby version.

Yes. I merely pointed out that in this proof of concept switching between versions causes havok.

However, I'm now inclined to agree with @Simplify that you guys should not bother with this problem. I'm sorry for wasting everyone's time, I did not know what this innocent looking issue would lead to.

It seems to me now that the best course of action when a rbenv shim returns command not found is to either
* install the respective executable in the corresponding - or even all - ruby versions as [suggested](https://github.com/flycheck/flycheck/issues/1447#issuecomment-380405809) by the creator of rbenv.
* [disable the checker](http://www.flycheck.org/en/latest/user/syntax-checkers.html#disable-syntax-checkers) e.g., via directory local variables

Trying to solve this automatically is probably just not worth the effort, seeing how this would depend on rbenv internals and is difficult to get right. Again I apologise for wasting your time and thank you for being so responsive and helpful!

fmdkdd:
@kaikuchn  It's alright.  Your reporting of the issue provided solutions and workarounds to this problem for future users.   All is not lost ;)

I agree that the complexity cost may balloon up if we have to take all Ruby executables into account.  Let's drop this.

PR #1451, opened 2018-04-13
Fix `cargo metadata` parse error

fmdkdd:
`cargo metadata` can emit plain-text warnings alongside the JSON otuput,
as `cargo rustc` does.  This causes `json-read` to fail.

This commit uses `flycheck-parse-json` instead to handle this mixed
output.

Fix flycheck/flycheck-rust#65.


fmdkdd:
@cpitclaudel Can we merge this?

PR #1455, opened 2018-04-23
Whitelist web-mode to CSS checkers

wyuenho:
web-mode can be used for CSS and CSS preprocessor files as well. This PR enables these CSS and CSS preprocessor linters on web-mode.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -284,7 +284,7 @@ Syntax checkers in this list must be defined with
   :type '(repeat (symbol :tag "Checker"))
   :risky t)
 
-(defcustom flycheck-disabled-checkers nil
+(defcustom flycheck-disabled-checkers '(html-tidy)
   "Syntax checkers excluded from automatic selection.
 
 A list of Flycheck syntax checkers to exclude from automatic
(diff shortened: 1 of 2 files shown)


fmdkdd:
Thanks, but I'm wondering how this affects the list of checkers allowed in `web-mode`.  Can you show us what `flycheck-verify-setup` looks like in a `web-mode` buffer with this change?

wyuenho:
Thanks, I've fixed the parens. Here's a screenshot:

screen shot 2018-04-24 at 15 50 29

fmdkdd:
So when you don't have `tide` installed, the default checker for `web-mode` would be `csslint`.  Shouldn't we have an HTML checker in there as well?

cpitclaudel:
I'm confused by this PR.  If I run csslint on a webpage (.html), I get a bunch of `unexpected token` errors.  Is this PR different?

wyuenho:
@cpitclaudel I've added a bunch of checks, see if this makes more sense now.

@fmdkdd  Orginally I didn't want to turn on HTML tidy because it's basically useless and lots of people name their templates in `.html`. My preference is to still leave it disabled by default but mark it as may be enabled for `web-mode`, and I've just done that in the latest push.

wyuenho:
I just realized there's a an undocumented function `(flycheck-add-mode)` that'll get me what I want more sensibly. Closing.

PR #1459, opened 2018-05-09
Support mhtml-mode

xuchunyang:
Since Emacs 26.1, mhtml-mode is now the default mode for html files, also see https://github.com/emacs-mirror/emacs/blob/766b057e41/etc/NEWS.26#L1052


cpitclaudel:
Note that mhtml-mode breaks flycheck in embedded regions, though (it sets major-mode, which confuses flycheck).

PR #1460, opened 2018-05-11
Separate options for pylint with equals signs.

phst:
The documentation only documents equals signs for long
options (https://pylint.readthedocs.io/en/latest/user_guide/run.html#command-line-options),
so using separate arguments only works by accident.


fmdkdd:
+1 on this change, if only because I like long arguments better.

And wow, these are some funky errors in the CI.  Probably #1462 will help there as well, since the Cask version we use is frozen.

PR #1466, opened 2018-05-18
Use header-line-format for the Message / Checker column

### LINKED ISSUE #1293: Error-list header does not respect header-line face
When using the error-list, the header-line mostly uses the correct face (as specified with `header-line`), except for the `Message (Checker)` column (same behaviour with -Q and the default color theme):

screen shot 2017-07-23 at 18 52 22

I would expect to see the same header-line face be used over the whole line. I am using emacs 25.2.1 on macOS 10.11 and flycheck snapshot 31 (20170717.2300).

I think the relevant [section](https://github.com/flycheck/flycheck/blob/master/flycheck.el#L3883) in the code is:
```elisp
(defun flycheck-error-list-make-last-column (message checker)
  "Compute contents of the last error list cell.
MESSAGE and CHECKER are displayed in a single column to allow the
message to stretch arbitrarily far."
  (let ((checker-name (propertize (symbol-name checker)
                                  'face 'flycheck-error-list-checker-name)))
    (format (propertize "%s (%s)" 'face 'default)
            message checker-name)))

(defconst flycheck-error-list-format
  `[("Line" 5 flycheck-error-list-entry-< :right-align t)
    ("Col" 3 nil :right-align t)
    ("Level" 8 flycheck-error-list-entry-level-<)
    ("ID" 6 t)
    (,(flycheck-error-list-make-last-column "Message" 'Checker) 0 t)]
"Table format for the error list.")
```
Is there a specific reason `flycheck-error-list-make-last-column` has to be called to show the column header? Wouldn't it be enough to just use the un`propertize`d strings?

d4ncer:
Fixes #1293

~As outlined in #1293 it would be good to use the `header-line-format` face across the entire tabulated list header. The other columns respect this face and the last one should as well.~

EDIT: I am now only propertizing the checker name, and not the entire column.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3984,8 +3984,7 @@ MESSAGE and CHECKER are displayed in a single column to allow the
 message to stretch arbitrarily far."
   (let ((checker-name (propertize (symbol-name checker)
                                   'face 'flycheck-error-list-checker-name)))
-    (format (propertize "%s (%s)" 'face 'default)
-            message checker-name)))
+    (format "%s (%s)" message checker-name)))
 
 (defconst flycheck-error-list-format
   `[("File" 6)
@@ -4061,7 +4060,8 @@ message to stretch arbitrarily far."
 
 (define-button-type 'flycheck-error-list
   'action #'flycheck-error-list-button-goto-error
-  'help-echo "mouse-2, RET: goto error")
+  'help-echo "mouse-2, RET: goto error"
+  'face nil)
 
 (defun flycheck-error-list-button-goto-error (button)
   "Go to the error at BUTTON."


d4ncer:
Oops, that was my bad. I meant to use `header-line`, not `header-line-format`.

d4ncer:
I did not; let me try that.

d4ncer:
Works great, and much easier to comprehend. Thanks for the suggestion @cpitclaudel!

joewreschnig:
I think the propertization was because the default button face is often overwhelming (too many rows of underlines) and ugly when only partially applied to a line (underline up through the `(` but not the checker name).
screenshot from 2018-05-21 13-18-19

Since `flycheck-error-list-button` is its own face already themes can fix this, but as-is I think this change will affect a lot of people's display.

cpitclaudel:
@joewreschnig Excellent point, thanks!
This should be fixed in the definition of the flycheck button category, though:

```
(define-button-type 'flycheck-error-list
  'action #'flycheck-error-list-button-goto-error
  'help-echo "mouse-2, RET: goto error")
```

That is, we should get rid of the propertize call and make sure that the `flycheck-error-list-button` category doesn't set a face (or uses default).

cpitclaudel:
The following seems to be enough:

```
(define-button-type 'flycheck-error-list
  'action #'flycheck-error-list-button-goto-error
  'help-echo "mouse-2, RET: goto error"
  'face nil)
```

@fmdkdd, objections? The current code lets `button` add a button-face that then gets overriden by the existing propertize. The suggested fix it to configure `button` to not add that face and then to remove the `propertize` call.

fmdkdd:
Sorry for the delay.  This seems to have slipped by.

> The following seems to be enough:

Looks good!

@d4ncer can you make the change in the PR?

d4ncer:
Done!

d4ncer:
Done :) Thanks for all the guidance!

cpitclaudel:
:tada: Thanks for your patience and for fixing a longstanding issue!
Congrats on your first Flycheck commit — looking forward to many more :)

fmdkdd:
Yeah!  Thank you @d4ncer :clap:

PR #1467, opened 2018-05-23
Hadolint has a new home

lorenzo:
Hadolint development was moved to an organization in github, and the code
has been improved since. flycheck should take advantage of this


fmdkdd:
Thanks for noticing.  LGTM.

PR #1470, opened 2018-05-25
Add a check for long lines in `make check`

fmdkdd:
We did not check that previously on CI, now we do.

Of course, we now have over a hundred lines longer than 80 characters (a number we can still change).

There may be a better way to check for long lines than rolling our own function.  I thought about using `whitespace-mode` which highlights long lines in a custom face, but seeing as I wanted to make some exceptions (e.g., for the first line of a package), I wrote the code directly.

Before merging, we should make sure that the long lines *can* be made shorter.  For example, the first line of `flycheck-format.el`  is too long, but I don't think we can split it without breaking the lexical-binding setting.   I've added an exception to the first line of files in this commit.

In `flycheck.el`, we have the `Package-Requires` line that is too long.  Can we split that?  Or do we need an exception?

We can also raise the max length to 100 to catch overly long lines.  In this case, we mostly have long lines in `flycheck-test.el` which I think should be splitteable without exceptions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,8 +56,7 @@ script:
   case $JOB in
     emacs)
       # Check formatting (our formatting rules don’t work on Emacs 24).
-      # FIXME: this should run on Emacs 26 as well
-      if [[ $EMACS_VERSION == 25.* ]]; then
+      if [[ $EMACS_VERSION != 24.* ]]; then
         docker run --volume "$TRAVIS_BUILD_DIR":/flycheck \
                    --workdir /flycheck \
                    flycheck/emacs-cask:$EMACS_VERSION \
--- a/flycheck-ert.el
+++ b/flycheck-ert.el
@@ -57,11 +57,12 @@ Defaults to `error'."
       (let ((conditions
              (if (consp parent)
                  (apply #'append
-                        (mapcar (lambda (parent)
-                                  (cons parent
-                                        (or (get parent 'error-conditions)
-                                            (error "Unknown signal `%s'" parent))))
-                                parent))
+                        (mapcar
+                         (lambda (parent)
+                           (cons parent
+                                 (or (get parent 'error-conditions)
+                                     (error "Unknown signal `%s'" parent))))
+                         parent))
                (cons parent (get parent 'error-conditions)))))
         (put name 'error-conditions
              (delete-dups (copy-sequence (cons name conditions))))
@@ -196,7 +197,8 @@ should use to lookup resource files."
   (let ((tests (ert-select-tests t t)))
     ;; Select all tests
     (unless tests
-      (error "No tests defined.  Call `flycheck-ert-initialize' after defining all tests!"))
+      (error "No tests defined.  \
+Call `flycheck-ert-initialize' after defining all tests!"))
 
     (setq flycheck-ert--resource-directory resource-dir)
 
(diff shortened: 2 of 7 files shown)


cpitclaudel:
Looking at the implementation of `lm-header`, it seems that these lines can't be split.
We could start checking for long lines after the `;;; Commentary` line; WDYT?

fmdkdd:
> We could start checking for long lines after the ;;; Commentary line; WDYT?

Not all files may have this line.  For files that do, we can start checking after it, for those that don't, we can check the whole file.  I think that would take care of most unsplittable lines.

Shall we settle on 80 columns?  I'm tempted to up the limit for test files (they all contain `test` in their name), as too much splitting there may hinder readability, but we'll see how it goes.

cpitclaudel:
> Not all files may have this line. For files that do, we can start checking after it, for those that don't, we can check the whole file. 

Or we could just skip that test entirely on files that don't have this line, for simplicity — all well-formatted files should have it, so if a file doesn't it might be OK for it to violate other guidelines like line length.

> Shall we settle on 80 columns?

Sounds super.

> I'm tempted to up the limit for test files (they all contain test in their name), as too much splitting there may hinder readability, but we'll see how it goes.

I'm tempted to only apply this rule to flycheck.el, in fact.

fmdkdd:
>  all well-formatted files should have it, so if a file doesn't it might be OK for it to violate other guidelines like line length.

Turns out, we have a bunch of files with a proper header!  Even test files have them.  But I applied this logic still.

> I'm tempted to only apply this rule to flycheck.el, in fact.

I wasn't courageous enough to split the 100 or so long lines in `flycheck-test.el`.  A quick look at them tells them we wouldn't gain much in readibility from splitting them.  I added a `flycheck/can-have-long-lines` predicate to `flycheck-format.el` to skip these.

But, some other files still contain code where the 80 chars limit makes sense.  I've fixed a few of them.

There is only *one line* I don't know how to split.  This URL:

```elisp
object that represents a message from Cargo.  The format of
messages emitted by Cargo is described there:

https://github.com/rust-lang/cargo/blob/master/src/cargo/util/machine_message.rs"
```

It's *one char* over the limit.  Putting the end quote `"` on the following line triggers this warning:

```
flycheck.el:6021: Documentation strings should not start or end with whitespace
```

I can of course split the URL with a newline or backslash, but then it's a lot less useful (`browse-url-at-point` is confused for instance).  Any ideas?

fmdkdd:
Brilliant!  I didn't know Github had an "official" URL shortener. :+1: 

I allowed 26.1 to run `make check` as well.  There was one place in `flycheck.el` were 25.3 and 26.1 did not agree on the format, but moving things around did the trick.

Should be good to review.

.travis.yml
```diff
@@ -56,8 +56,7 @@ script:
   case $JOB in
     emacs)
       # Check formatting (our formatting rules don’t work on Emacs 24).
-      # FIXME: this should run on Emacs 26 as well
-      if [[ $EMACS_VERSION == 25.* ]]; then
+      if [[ $EMACS_VERSION != 24.* ]]; then
```
cpitclaudel: Fancy, this is beyond my own bash-fu.  Are Travis script guaranteed to run in bash, or is this POSIX-sh compliant too?

fmdkdd: Good question.  `!=` itself [may be POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html#tag_20_128_05), but the wildcard may not be, and `[[ ]]` isn't, I think.  I can't find a definite source saying whether Travis uses bash (maybe dash?).  But it works as intended.

cpitclaudel: > But it works as intended.

Let's merge :)

PR #1473, opened 2018-06-02
Add option `flycheck-jump-to-other-file-immediately`

ntc2:
This options controls whether Flycheck error navigation jumps to other
files immediately when the selected error is in another file. This is
related to PR #1427 that added overlays on the first line of the
current file when errors in other files prevent checking the current
file.

When the new option `flycheck-jump-to-other-files-immediately` is
nil (it defaults to non-nil), error navigation in the current file
moves the cursor to first-line overlay, and then you have to jump
again to actually go to the other file containing the error. The idea
is that it's surprising for Flycheck error navigation to change files,
and stopping on the first-line overlay first makes it less surprising.

There is more context for this commit here:
https://github.com/flycheck/flycheck/pull/1427#issuecomment-390912510.

The current specification for how Flycheck error navigation (`flycheck-{next,previous,first}-error`) works with relation to other-file error overlays is:

- if the cursor is already on an other-file overlay, then that overlay is considered to be the next/previous overlay, and jumping to it jumps to corresponding error in the other file.

- if the cursor is not already on an other-file error overlay, then the next/previous/first overlay is the next/previous in relation to the cursor, or first, respectively. If the overlay determined in this way is an other-file overlay, then what happens when you jump to it is determined by the setting of `flycheck-jump-to-other-files-immediately`: if that setting is non-nil, the default, then jump to the error in the other file; if that setting is nil, then jump to the other-file overlay in the current file.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -595,6 +595,30 @@ where variable `flycheck-mode' is already non-nil."
   :package-version '(flycheck . "0.15")
   :safe #'booleanp)
 
+(defcustom flycheck-jump-to-other-files-immediately t
+  "Whether to jump to other files immediately in `flycheck-next-error'.
+
+When an error in another file prevents checking the current file,
+Flycheck puts on error overlay on the first line of the current
+file that references the errors in the other file. This option
+determines how error navigation treats that first-line overlay
+when the cursor is not already on the first line.
+
+If non-nil, Flycheck error navigation jumps to the other file
+immediately, regardless of the position of the cursor in the
+current file.
+
+If nil, Flycheck error navigation jumps to that first-line
+overlay when the cursor is not already on the first line.
+
+If the cursor is already on the first-line error overlay, then
+this option has no effect: Flycheck error navigation will always
+jump to the other file containing the error in that case."
+  :group 'flycheck
+  :type 'boolean
+  :package-version '(flycheck . "0.32")
+  :safe #'booleanp)
+
 (define-widget 'flycheck-minimum-level 'lazy
   "A radio-type choice of minimum error levels.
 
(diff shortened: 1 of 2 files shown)


flycheck.el
```diff
@@ -3850,17 +3874,26 @@ overlays."
             (flycheck-error-level-severity (flycheck-error-level err)))
       t)))
 
-(defun flycheck-next-error-pos (n &optional reset)
-  "Get the position of the N-th next error.
+(defun flycheck-next-error-overlay-pos (n &optional reset)
+  "Get the position of the N-th next error overlay in buffer.
```
ntc2: I renamed this function because I found the name misleading, since after #1427 overlays may now point to errors in other files, and so the position of an overlay and the corresponding error need no longer be the same. However, there may be other places where this new distinction between overlays and errors also applies, and I didn't fix any other occurrences. So, maybe it's better to not change the name here?

cpitclaudel:
Thanks for the patch. I don't have time right now for a full review, so I'll stick to high-level questions

* Should we change the default behavior?  This feature isn't old enough for changing the default to be an issue, I think.
* Should we show a message, like "press <key> again to visit <file>", after landing on an overlay that points to another file?
* What should we do when we're already on an overlay, but we didn't get there with `flycheck-{previous,next}-error`?  Should we always require two presses, or should we require just one in that case?  (We could jump only when the previous command is already a flycheck motion).

ntc2:
On Sat, Jun 2, 2018 at 4:12 PM, Clément Pit-Claudel
<notifications@github.com> wrote:
> Thanks for the patch. I don't have time right now for a full review, so I'll
> stick to high-level questions
>
> - Should we change the default behavior? This feature isn't old enough for
> changing the default to be an issue, I think.

I think changing the default is a good idea, but I might be biased,
since the whole point of my PR was to make it possible to get
(currently) non-default behavior :)

> - Should we show a message, like "press again to visit ", after landing on an
> overlay that points to another file?

+1

> - What should we do when we're already on an overlay, but we didn't get there
> with flycheck-{previous,next}-error? Should we always require two presses,
> or should we require just one in that case? (We could jump only when the
> previous command is already a flycheck motion).

I'd say always jump when we're already on an overlay, since that's
simpler than the alternative, and currently other-file overlays are
obvious anyway (they're on the first line, and cover the entire first
line).

cpitclaudel:
@ntc2 Are you interested in updating this PR based on the changes we discussed previously?

ntc2:
I'm traveling in SE Asia and things are very uncertain. I might have a
chance to look at this in the next few days, but I haven't used flycheck
for 6+ months (not working currently) so it's not a big priority for me
right now.

cpitclaudel:
Thanks for the update. I'll close this for now, and we can pick it up when the world is a bit less crazy. Stay safe out there.

PR #1476, opened 2018-06-10
[WIP] Fix unstable order of displayed errors 

### LINKED ISSUE #1469: Error order is not stable
If there are multiple errors at point, the error messages may not always be reported in the same order.  See:

bug

I created an empty buffer, then executed the following:

```emacs-lisp
(progn
 (flycheck-clear-errors)
 (flycheck-mark-all-overlays-for-deletion)
 (flycheck-delete-marked-overlays)
 (flycheck-report-current-errors
  (list
   (flycheck-error-new-at 3 1 'error "error 1")
   (flycheck-error-new-at 3 1 'error "error 2")
   (flycheck-error-new-at 3 1 'error "error 3"))))
```

When moving point *down* to the overlays, I get the order `1 2 3`, but when moving point *up* from below the overlays, I get `3 2 1`.

Depending on the checker, the order of errors can be meaningful.  For `rust-cargo`, errors come first, then warnings, but both kinds may come with interspersed info:

```
error 1
info for error 1
another info for error 1
warning 2
info for warning 2
```

We should preserve this order.  And, as this case shows, we cannot simply sort by error level.
Arguably, for other checkers where the order may not matter, it would still be desirable to keep the order stable.

Digging into this, it seems the order we use for displaying errors comes from the order of the overlays at point given by `overlays-at`.  There is an extra `SORTED` argument which sorts overlay by priority.  Setting it does keep the order stable, but as far as I  understand the manual, there is no guarantee that this should always be the case, since these overlays have the same priority (`110` seemingly by default) currently.

So, one solution would be to assign priorities to overlays which correspond to the position of the error in `flycheck-current-errors`.  Then, use `overlays-at pos 'sorted` to guarantee that the overlays, and the corresponding flycheck errors, are returned in the intended order.

fmdkdd:
This is a tentative fix for #1469.

I've added an index to the `flycheck-error` struct, and uses it to sort the errors instead of relying on the unstable overlay order.

This seems to solve the situation highlighted in #1469.  I need to add tests, and check it works with multiple checkers.

Adding a field to `flycheck-error` is unfortunately a breaking change which requires downstream packages to be byte-recompiled.

This is likely to break all integration tests as well.

Closes #1469.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4136,6 +4136,16 @@ If LEVEL is omitted if the current buffer has any errors at all."
 
 
 ;;; Error overlays in the current buffer
+(defvar-local flycheck--last-overlay-index 0
+  "Last index given to a Flycheck overlay.
+
+These indices are used to preserve error order (Emacs doesn't
+preserve overlay order when calling `overlays-at').")
+
+(defun flycheck--next-overlay-index ()
+  "Compute the index to assign to a new Flycheck overlay."
+  (cl-incf flycheck--last-overlay-index))
+
 (defun flycheck-add-overlay (err)
   "Add overlay for ERR.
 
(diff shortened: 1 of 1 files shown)


cpitclaudel:
I just pushed a new implementation based on overlay properties instead of an error field, which should be backwards-compatible. Can you give the patch a look?

fmdkdd:
Github won't let me approve my own PR, so go ahead ;)

PR #1478, opened 2018-06-23
Enhance `verify-setup` buffer

fmdkdd:
1.  Separate first checker from other checkers in verify-setup.
      
    This makes it clearer that only one checker can be run, even though
    multiple checkers may be available.

2.  Display next checkers in verify-setup when applicable.

    This makes it easier to debug checker chaining.

Here are some screenshots:

elisp

The checker that will effectively run is the first, and the available checkers for the buffer are below.  Note that the first checker is not repeated in the available checkers.  You can also see the new `next checkers` verification result.

Here is what it looks like with larger chains:

more-checkers

I've added two checkers `foobar` and `foobar2` for testing.  `foobar` was added as `next-checker` to `c/c++-clang`, and `foobar2` as `next-checker` to `foobar`.  This is all explicitly visible in the screenshot.  The first checker will show the full chain, but the description of `foobar` still shows that `foobar2` comes after it.

And the last example is for a buffer without any checker:

no-checkers

### DIFF
--- a/doc/user/flycheck-versus-flymake.rst
+++ b/doc/user/flycheck-versus-flymake.rst
@@ -226,8 +226,8 @@ files with PHP CLI first to find syntax errors, then with PHP MessDetector to
 additionally find idiomatic and semantic errors, and eventually with PHP
 CheckStyle to find stylistic errors.  The user will see all errors reported by
 all of these tools in the buffer.  However, if the first checker reported at
-least one error, then the subsequent checkers would not be run (by default; this
-behavior is configurable).
+least one error, then the subsequent checkers would not be run.  This behavior
+is fully configurable; see :ref:`flycheck-checker-chains`.
 
 Errors
 ------
--- a/doc/user/syntax-checkers.rst
+++ b/doc/user/syntax-checkers.rst
@@ -274,3 +274,57 @@ installed into a local ``node_modules`` directory:
 
       :infonode:`(emacs)Directory Variables`
          Information about directory variables.
+
+.. _flycheck-checker-chains:
+
+Configuring checker chains
+==========================
+
+In any given buffer where Flycheck is enabled, only one checker may be run at a
+time.  However, any number of checkers can be run in sequence.  In such a
+sequence, after the first checker has finished running and its errors have been
+reported, the next checker of the sequence runs and its errors are reported,
+etc. until there are no more checkers in the sequence.  This sequence is called
+a *checker chain*.
+
+Some checkers chains are already setup by default in Flycheck: e.g.,
+`emacs-lisp` will be followed by `emacs-lisp-checkdoc`, and `python-mypy` will
+be followed by `python-flake8`.
+
+When defining a checker, you can specify which checkers may run after it by
+setting the ``:next-checkers`` property (see the docstring of
+`flycheck-define-generic-checker`).
+
+For a given checker, several next checkers may be specified.  Flycheck will run
+the first (in order of declaration) whose error level matches (see below) and
+which can be used in the current buffer.
+
+You can also customize the next checker property by calling
+`flycheck-add-next-checker` in your Emacs configuration file.
+
+.. defun:: flycheck-add-next-checker checker next &optional append
+
+   Set *next* to run after *checker*.  Both arguments are syntax checker
+   symbols.
+
+   For example, the following will make `python-pylint` run after
+   `python-flake8`:
+
+   .. code-block:: elisp
+
+      (flycheck-add-next-checker 'python-flake8 'python-pylint)
+
+   *Next* may also be a cons cell ``(level . next-checker)``, where
+   *next-checker* is a symbol denoting the syntax checker to run after
+   *checker*, and *level* is an error level.  The *next-checker* will then only
+   be run if there is no current error whose level is more severe than *level*.
+   If *level* is ``t``, then *next-checker* is run regardless of the current
+   errors.
+
+   For instance, if you wanted to run `python-pylint` only if `python-flake8`
+   produced no errors (only warnings and info diagnostics), then you would
+   rather use:
+
+   .. code-block:: elisp
+
+      (flycheck-add-next-checker 'python-flake8 '(warning . python-pylint))
--- a/flycheck.el
+++ b/flycheck.el
@@ -1,6 +1,6 @@
 ;;; flycheck.el --- On-the-fly syntax checking -*- lexical-binding: t; -*-
 
-;; Copyright (C) 2017-2018 Flycheck contributors
+;; Copyright (C) 2017-2019 Flycheck contributors
 ;; Copyright (C) 2012-2016 Sebastian Wiesner and Flycheck contributors
 ;; Copyright (C) 2013, 2014 Free Software Foundation, Inc.
 ;;
@@ -915,6 +915,12 @@ This variable is a normal hook.  See Info node `(elisp)Hooks'."
   '((t :inherit highlight))
   "Flycheck face to highlight errors in the error list."
   :package-version '(flycheck . "0.15")
+  :group 'flycheck-faces)
+
+(defface flycheck-verify-select-checker
+  '((t :box (:style released-button)))
+  "Flycheck face for the 'select' button in the verify setup buffer."
+  :package-version '(flycheck . "32")
   :group 'flycheck-faces)
 
 (defvar flycheck-command-map
@@ -1975,7 +1981,7 @@ nil otherwise."
 (define-button-type 'help-flycheck-checker-def
   :supertype 'help-xref
   'help-function #'flycheck-goto-checker-definition
-  'help-echo "mouse-2, RET: find Flycheck checker definition")
+  'help-echo "mouse-1, RET: find Flycheck checker definition")
 
 (defconst flycheck-find-checker-regexp
   (rx line-start (zero-or-more (syntax whitespace))
@@ -2133,22 +2139,43 @@ Return a list of `flycheck-verification-result' objects."
 (define-button-type 'help-flycheck-checker-doc
   :supertype 'help-xref
   'help-function #'flycheck-describe-checker
-  'help-echo "mouse-2, RET: describe Flycheck checker")
+  'help-echo "mouse-1, RET: describe Flycheck checker")
 
-(defun flycheck--verify-princ-checker (checker buffer &optional with-mm)
+(define-button-type 'flycheck-checker-select
+  :supertype 'help-xref
+  'help-function (lambda (buffer checker)
+                   (with-current-buffer buffer
+                     (flycheck-select-checker checker))
+                   ;; Revert the verify-setup buffer since it is now stale
+                   (revert-buffer))
+  'help-echo "mouse-1, RET: select Flycheck checker"
+  'face 'flycheck-verify-select-checker)
+
+(defun flycheck--verify-princ-checker (checker buffer
+                                               &optional with-mm with-select)
   "Print verification result of CHECKER for BUFFER.
 
 When WITH-MM is given and non-nil, also include the major mode
-into the verification results."
+into the verification results.
+
+When WITH-SELECT is non-nil, add a button to select this checker."
   (princ "  ")
   (insert-button (symbol-name checker)
                  'type 'help-flycheck-checker-doc
                  'help-args (list checker))
   (when (with-current-buffer buffer (flycheck-disabled-checker-p checker))
     (insert (propertize " (disabled)" 'face '(bold error))))
+  (when (eq checker (buffer-local-value 'flycheck-checker buffer))
+    (insert (propertize " (explicitly selected)" 'face 'bold)))
+  (when with-select
+    (princ "  ")
+    (insert-text-button "select"
+                        'type 'flycheck-checker-select
+                        'help-args (list buffer checker)))
   (princ "\n")
   (let ((results (with-current-buffer buffer
-                   (flycheck-verify-generic-checker checker))))
+                   (append (flycheck-verify-generic-checker checker)
+                           (flycheck--verify-next-checkers checker)))))
     (when with-mm
       (with-current-buffer buffer
         (let ((message-and-face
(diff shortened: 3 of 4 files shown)


cpitclaudel:
Thanks, the screenshot look very nice.  Should we also show the other checkers in the current chain in the first section? We often run more than one checker on a single buffer, so it seems misleading to highlight just the first one.

In the past we've also discussed clarifying 'automatically disabled' vs 'may enable': maybe we could add an 'enable now' link next to 'may enable'?  An in fact, if we do that, we could remove that line entirely and only show the 'enable now' link next to the thing that says 'automatically disabled'?

fmdkdd:
>  Should we also show the other checkers in the current chain in the first section? We often run more than one checker on a single buffer, so it seems misleading to highlight just the first one.

I tend to agree.  I can put them in sequence, although it might be misleading since we cannot know if all checkers will be run (depending on the value of `current-errors`).  At least that gets you a distinction between "these checkers may run in this buffer, and these other checkers won't definitely run until you change your config".

Plus `:next-checkers` can actually be multiple checkers, and only one will eventually be picked, depending on current errors .  For now I'm using `flycheck-get-next-checker-for-buffer` which picks just *one* checker from `:next-checkers`.   A quick grep tells me that we don't have any checker with multiple `:next-checkers` set, so this point may be moot.

> In the past we've also discussed clarifying 'automatically disabled' vs 'may enable': 

:+1:  This is confusing, I'll look into it.  An `enable now` is a much better solution indeed.

fmdkdd:
@m-cat If you want to help, can you check out this PR and report on its usability, what could be improved, etc?  I've rebased on latest master.

m-cat:
Sure @fmdkdd, will be glad to help once I find some time 👍

doc/user/syntax-checkers.rst
```diff
@@ -274,3 +274,57 @@ installed into a local ``node_modules`` directory:
...
+
+In any given buffer where Flycheck is enabled, only one checker may be run at a
+time.  However, any number of checkers can be run in sequence.  In such a
+sequence, after the first checker has finished running and its errors have been
+reported, the next checker of the sequence runs and its errors are reported,
+etc. until there are no more checkers in the sequence.  This sequence is called
+a *checker chain*.
+
+Some checkers chains are already setup by default in Flycheck: e.g.,
+`emacs-lisp` will follow with `emacs-lisp-checkdoc`, and `python-mypy` will
```
cpitclaudel: Maybe `will be followed by`?

doc/user/syntax-checkers.rst
```diff
@@ -274,3 +274,57 @@ installed into a local ``node_modules`` directory:
...
+follow with `python-flake8`.
+
+When defining a checker, you can specify which checkers may run after it by
+setting the ``:next-checkers`` property (see the docstring of
+`flycheck-define-generic-checker`).
+
+For a given checker, several next checkers may be specified.  Flycheck will run
+the first (in order of declaration) whose error level matches (see below) and
+whose checker can be used in the current buffer.
```
cpitclaudel: `whose checker` → `which`

flycheck.el
```diff
@@ -2130,11 +2130,14 @@ into the verification results."
   (insert-button (symbol-name checker)
                  'type 'help-flycheck-checker-doc
                  'help-args (list checker))
-  (when (with-current-buffer buffer (flycheck-disabled-checker-p checker))
+  (when (buffer-local-value (flycheck-disabled-checker-p checker) buffer)
```
cpitclaudel: :clap:

flycheck.el
```diff
@@ -2130,11 +2130,14 @@ into the verification results."
...
     (insert (propertize " (disabled)" 'face '(bold error))))
+  (when (eq checker (buffer-local-value 'flycheck-checker buffer))
+    (insert (propertize "(explicitly selected)" 'face 'bold)))
```
cpitclaudel: Do you need a space here?

flycheck.el
```diff
@@ -2159,10 +2162,43 @@ into the verification results."
         (princ (make-string (- message-column (current-column)) ?\ ))
         (let ((message (flycheck-verification-result-message result))
               (face (flycheck-verification-result-face result)))
-          (insert (propertize message 'face face)))
+          ;; If face is nil, using propertize erases the face already contained
+          ;; by the message.  We don't want that, since this would remove the
+          ;; button face from the checker chain result.
+          (if face
+              (insert (propertize message 'face face))
+            (insert message)))
```
cpitclaudel: But when face is non-nil, won't it override the existing face? Maybe you want `font-lock-append-text-property`?

flycheck.el
```diff
@@ -2159,10 +2162,43 @@ into the verification results."
...
   (princ "\n"))
 
+(defun flycheck-checker-chain (checker)
+  "Return the chain of checkers following CHECKER.
+
+The return value is a list of checkers, not including CHECKER."
+  (let ((chain))
+    (while (setq checker (flycheck-get-next-checker-for-buffer checker))
+      (push checker chain))
+    (nreverse chain)))
```
cpitclaudel: Do we only want the usable ones here? Wouldn't it better instead to include also the ones that may not be used for this buffer? (`-get-next-checker-for-buffer` calls `flycheck-may-use-next-checker`, which checks the current max error level, etc)

flycheck.el
cpitclaudel: Oh I think I understand now.  Just a small change then:
`(insert (if face (propertize message …) message))` (no need to repeat the insert)

flycheck.el
```diff
@@ -2159,10 +2162,43 @@ into the verification results."
...
+      :message (string-join
+                (split-string
+                 (with-temp-buffer
+                   (dolist (checker chain)
+                     (insert-text-button (symbol-name checker)
+                                         'type 'help-flycheck-checker-doc
+                                         'help-args (list checker))
+                     (insert " "))
+                   (buffer-string)))
+                ", ")))))
```
cpitclaudel: Cool trick: `make-text-button` can take a string and propertize it; hence this should work:

```elisp
 (string-join
  (seq-map (lambda (checker)
             (make-text-button (symbol-name checker) nil
                               'type 'help-flycheck-checker-doc
                               'help-args (list checker)))
           chain)
  ", ")
```

flycheck.el
```diff
@@ -2179,7 +2215,7 @@ buffer."
     (insert-button (symbol-name mode)
                    'type 'help-function
                    'help-args (list mode)))
-  (princ ":\n\n"))
+  (princ ".\n\n"))
```
cpitclaudel: ?

flycheck.el
```diff
@@ -2245,34 +2281,46 @@ Display a new buffer listing all syntax checkers that could be
 applicable in the current buffer.  For each syntax checkers,
 possible problems are shown."
   (interactive)
+  ;; Save to make sure checkers that only work on saved buffers will pass the
+  ;; verification
   (when (and (buffer-file-name) (buffer-modified-p))
-    ;; Save the buffer
     (save-buffer))
```
cpitclaudel: Ugh, I didn't realize we did that.  Maybe we should do it through `save-some-buffers` instead?
I'd be happy with just a FIXME; no need to fix it now.

flycheck.el
```diff
@@ -2245,34 +2281,46 @@ Display a new buffer listing all syntax checkers that could be
...
+
+    ;; Print all applicable checkers for this buffer
     (with-help-window help-buffer
       (with-current-buffer standard-output
         (flycheck--verify-print-header "Syntax checkers for buffer " buffer)
-        (unless checkers
+
+        ;; First display the checkers that will run for this buffer
+        (when checker-chain
+          (progn
```
cpitclaudel: No need for progn I think

flycheck.el
```diff
@@ -2245,34 +2281,46 @@ Display a new buffer listing all syntax checkers that could be
...
+            (princ "Enabled checkers:\n\n")
+            (dolist (checker checker-chain)
+              (flycheck--verify-princ-checker checker buffer))))
+
+        ;; Then display other checkers
+        (when other-checkers
+          (progn
```
cpitclaudel: Same here

flycheck.el
```diff
@@ -2245,34 +2281,46 @@ Display a new buffer listing all syntax checkers that could be
...
+            (princ "Available checkers:\n\n")
+            (dolist (checker other-checkers)
+              (flycheck--verify-princ-checker checker buffer))))
+
+        ;; If we have no checkers at all, that's worth mentioning
+        (when (and (not checker-chain) (not other-checkers))
```
cpitclaudel: `(unless (or checker-chain other-checkers)`?

flycheck.el
fmdkdd: I do!

flycheck.el
fmdkdd: Added a FIXME.  `save-some-buffers` would let the user choose, but then we would have a few `:predicate` failing due to `flycheck-buffer-saved-p`.  Ideally we would annotate these in the verify buffer, but how do we know that `flycheck-buffer-saved-p` is used in the predicate?  Macro magic in `define-checker`?

flycheck.el
fmdkdd: I always have trouble thinking about `unless` :)

flycheck.el
fmdkdd: Hmm the problem with the static view of available next checkers is that it is a tree, not a list.

We could display the immediate next checkers of the current one; that is easy and not ambiguous.

cpitclaudel: Indeed :thinking:
I wonder if the simplest wouldn't be to just show the list of next checkers for each checker.

flycheck.el
cpitclaudel: I think we can't know; the best we could do would be to show a warning if the current buffer is unsaved.

flycheck.el
fmdkdd: Actually, that would have worked with the `flycheck-disabled-checkers` variable, but it doesn't with the predicate :sweat:.

fmdkdd:
@cpitclaudel I made some changes.  Now we display:

1. The first checker to run.   I added back the message when there isn't any first checker.
2. All the checkers that are part of the first checker's chain.  That is a *static* view, not using `-get-next-checker`, so we disregard the current error level.
3. All checkers that may be used in this buffer, but are not part of the current chain.  I added a `select` button on these to easily toggle from the verify setup buffer.
4. Any remaining (misconfigured) checker for the major mode.

doc/user/flycheck-versus-flymake.rst
```diff
@@ -225,9 +225,8 @@ run in sequence rather than concurrently.  For instance, Flycheck will check PHP
 files with PHP CLI first to find syntax errors, then with PHP MessDetector to
 additionally find idiomatic and semantic errors, and eventually with PHP
 CheckStyle to find stylistic errors.  The user will see all errors reported by
-all of these tools in the buffer.  However, if the first checker reported at
-least one error, then the subsequent checkers would not be run (by default; this
-behavior is configurable).
+all of these tools in the buffer.  This behavior is fully configurable; see
+:ref:`flycheck-checker-chains`.
```
cpitclaudel: I think the 'this behavior' part is unclear with the previous sentence removed; don't you think?

flycheck.el
```diff
@@ -2173,10 +2200,45 @@ into the verification results."
...
+  (mapcar (lambda (c) (if (consp c) (cdr c) c))
+          (flycheck-checker-get checker 'next-checkers)))
+
+(defun flycheck-all-next-checkers (checker)
+  "Return all checkers that may follow CHECKER.
+
+Return the transitive closure of the next-checker relation.  The
+return value is a list of checkers, not including CHECKER."
+  (let ((next (flycheck-get-next-checkers checker)))
+    (seq-uniq (append next (seq-mapcat #'flycheck-all-next-checkers next)))))
```
cpitclaudel: This is going to infloop if someone adds a checker as its own next-checker, right? I think our verification code will likely infloop as well, though, so this is not worse

flycheck.el
```diff
@@ -2173,10 +2200,45 @@ into the verification results."
...
+
+(defun flycheck--verify-next-checkers (checker)
+  "Return a verification result for the next checkers of CHECKER."
+  (-when-let (next (flycheck-get-next-checkers checker))
+    (list
+     (flycheck-verification-result-new
+      :label "next checkers"
+      ;; We use `insert-text-button' to preserve the button properties in
```
cpitclaudel: That comment needs to be updated :)

flycheck.el
```diff
@@ -2173,10 +2200,45 @@ into the verification results."
...
+      ;; the string
+      :message (string-join
+                (seq-map (lambda (checker)
+                           (make-text-button (symbol-name checker) nil
+                                             'type 'help-flycheck-checker-doc
+                                             'help-args (list checker)))
+                         next)
+                ", ")))))
```
cpitclaudel: Not that I look at this again, I think this is an ideal candidate for mapconcat :)

doc/user/flycheck-versus-flymake.rst
fmdkdd: I agree.  I'm not sure why I removed it in the first place.

flycheck.el
fmdkdd: Yes, this assumes we don't have cycles.  Ideally we want to ensure there are no cycles by construction, but I'm not sure how to do that in `define-checker` and when the `:next-checkers` property is altered after.

Do you want me to harden this function to ensure we don't revisit checkers twice?

fmdkdd: @cpitclaudel I've changed the code to avoid visiting checkers twice.  Can you look?

cpitclaudel:
Thanks for the update. I just gave it a try. It looks great.
One thing that I noticed: in my Python buffer, I have `flycheck-disabled-checkers` set to `(python-flake8 python-mypy)`; in the old code I get `(disabled)` next to these checkers, but in the new version they don't have this mark any more.  Is that intended?

fmdkdd:
>  I get (disabled) next to these checkers, but in the new version they don't have this mark any more. Is that intended?

Maybe you are using an old version of the PR?  I did have that bug, but I fixed it.

cpitclaudel:
> Maybe you are using an old version of the PR? I did have that bug, but I fixed it.

Yes, that seems to be it!  It looks gorgeous :)
Please merge :rocket: Don't forget the changelog entry!

Issue #1482, opened 2018-07-16
flycheck-previous-error does not work in *scratch*


raxod502:
From `emacs -Q`, here is what I do:

* Load revision 0995b0bb3f518c7350f1428860c0fc7eeaee3c46 of Flycheck.
* Delete the text in the `*scratch*` buffer.
* Run `M-x python-mode`.
* Insert `foo.` in the `*scratch*` buffer, which is then highlighted by Flycheck as an error.
* Press `C-c ! p`. I get the following backtrace with `debug-on-error`:

```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  expand-file-name(nil)
  find-file-noselect(nil)
  (let ((error-copy (copy-flycheck-error error)) (buffer (find-file-noselect (progn (or (and (memq (type-of error) cl-struct-flycheck-error-tags) t) (signal 'wrong-type-argument (list 'flycheck-error error))) (aref error 3))))) (progn (or (and (memq (type-of error-copy) cl-struct-flycheck-error-tags) t) (signal 'wrong-type-argument (list 'flycheck-error error-copy))) (let* ((v error-copy)) (aset v 1 buffer))) (flycheck-jump-in-buffer buffer error-copy) (save-current-buffer (set-buffer buffer) (if (seq-contains flycheck-current-errors error-copy 'equal) nil (if flycheck-mode (progn (flycheck-buffer))))))
  flycheck-jump-to-error(#s(flycheck-error :buffer #<buffer *scratch*> :checker python-pycompile :filename nil :line 1 :column nil :message "invalid syntax" :level error :id nil :group nil))
  (if err (flycheck-jump-to-error err) (user-error "No more Flycheck errors"))
  (let ((err (get-char-property pos 'flycheck-error))) (if err (flycheck-jump-to-error err) (user-error "No more Flycheck errors")))
  (if pos (let ((err (get-char-property pos 'flycheck-error))) (if err (flycheck-jump-to-error err) (user-error "No more Flycheck errors"))) (user-error "No more Flycheck errors"))
  (let ((pos (flycheck-next-error-pos n reset))) (if pos (let ((err (get-char-property pos 'flycheck-error))) (if err (flycheck-jump-to-error err) (user-error "No more Flycheck errors"))) (user-error "No more Flycheck errors")))
  flycheck-next-error-function(-1 nil)
  flycheck-next-error(-1)
  flycheck-previous-error(nil)
  funcall-interactively(flycheck-previous-error nil)
[...]

PR #1483, opened 2018-07-16
Fix `flycheck-previous-error` from failing in buffers without files

fmdkdd:
Fix GH-1482.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4297,20 +4297,28 @@ POS defaults to `point'."
 
 (defun flycheck-jump-to-error (error)
   "Go to the location of ERROR."
-  (let ((error-copy (copy-flycheck-error error))
-        (buffer (find-file-noselect (flycheck-error-filename error))))
-    (setf (flycheck-error-buffer error-copy) buffer)
-    (flycheck-jump-in-buffer buffer error-copy)
-    ;; When jumping to an error in another file, it may not have
-    ;; this error available for highlighting yet, so we trigger a check
-    ;; if necessary.
-    (with-current-buffer buffer
-      (unless (seq-contains flycheck-current-errors error-copy 'equal)
-        (when flycheck-mode
-          (flycheck-buffer))))))
+  (let* ((error-copy (copy-flycheck-error error))
+         (filename (flycheck-error-filename error))
+         (other-file-error (flycheck-relevant-error-other-file-p error))
+         (buffer (if filename
+                     (find-file-noselect filename)
+                   (flycheck-error-buffer error))))
+    (when (buffer-live-p buffer)
+      (setf (flycheck-error-buffer error-copy) buffer)
+      (flycheck-jump-in-buffer buffer error-copy)
+      ;; When jumping to an error in another file, it may not have
+      ;; this error available for highlighting yet, so we trigger a check
+      ;; if necessary.
+      (when other-file-error
+        (with-current-buffer buffer
+          (unless (seq-contains flycheck-current-errors error-copy 'equal)
+            (when flycheck-mode
+              (flycheck-buffer))))))))
 
 (defun flycheck-jump-in-buffer (buffer error)
   "In BUFFER, jump to ERROR."
+  ;; FIXME: we assume BUFFER and the buffer of ERROR are the same.  We don't
+  ;; need the first argument then.
   (if (eq (window-buffer) (get-buffer flycheck-error-list-buffer))
       ;; When called from within the error list, keep the error list,
       ;; otherwise replace the current buffer.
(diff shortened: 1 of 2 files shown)


fmdkdd:
Integration tests are failing due to updated tools.  I will add a commit that fixes them.

flycheck.el
```diff
@@ -4297,8 +4297,10 @@ POS defaults to `point'."
 
 (defun flycheck-jump-to-error (error)
   "Go to the location of ERROR."
-  (let ((error-copy (copy-flycheck-error error))
-        (buffer (find-file-noselect (flycheck-error-filename error))))
+  (let* ((error-copy (copy-flycheck-error error))
+         (filename (flycheck-error-filename error))
+         (buffer (or (and filename (find-file-noselect filename))
+                     (flycheck-error-buffer error))))
```
cpitclaudel: Shouldn't we use -when-let? The buffer could have been deleted

fmdkdd: Good idea.

fmdkdd:
@cpitclaudel Can you take another look?  I've used `(when (buffer-live-p` instead, as `when-let` would have failed when `filename` was nil (and in that case we might still find a buffer).

flycheck.el
```diff
@@ -4297,17 +4297,20 @@ POS defaults to `point'."
...
-    ;; this error available for highlighting yet, so we trigger a check
-    ;; if necessary.
-    (with-current-buffer buffer
-      (unless (seq-contains flycheck-current-errors error-copy 'equal)
-        (when flycheck-mode
-          (flycheck-buffer))))))
+  (let* ((error-copy (copy-flycheck-error error))
+         (filename (flycheck-error-filename error))
+         (buffer (or (and filename (find-file-noselect filename))
+                     (flycheck-error-buffer error))))
```
cpitclaudel: LGTM, though shouldn't we only use `flycheck-error-buffer` if there's no filename set? (i.e. shouldn't we just exit if find-file-noselect returns nil?)

flycheck.el
```diff
@@ -4297,17 +4297,20 @@ POS defaults to `point'."
...
+    (when (buffer-live-p buffer)
+      (setf (flycheck-error-buffer error-copy) buffer)
+      (flycheck-jump-in-buffer buffer error-copy)
+      ;; When jumping to an error in another file, it may not have
+      ;; this error available for highlighting yet, so we trigger a check
+      ;; if necessary.
+      (with-current-buffer buffer
+        (unless (seq-contains flycheck-current-errors error-copy 'equal)
+          (when flycheck-mode
+            (flycheck-buffer)))))))
```
cpitclaudel: It would be nice to restrict this bit of code to run only if we've changed buffers.  Otherwise we scan the whole error list at every move, in most cases for nothing.

fmdkdd:
@cpitclaudel Fixed.

flycheck.el
```diff
@@ -4297,17 +4297,23 @@ POS defaults to `point'."
...
+      (setf (flycheck-error-buffer error-copy) buffer)
+      (flycheck-jump-in-buffer buffer error-copy)
+      ;; When jumping to an error in another file, it may not have
+      ;; this error available for highlighting yet, so we trigger a check
+      ;; if necessary.
+      (when other-file-error
+        (with-current-buffer buffer
+          (unless (seq-contains flycheck-current-errors error-copy 'equal)
+            (when flycheck-mode
+              (flycheck-buffer))))))))
```
cpitclaudel: Looks great now.  We could move the error copy down into the "unless", but it's a minor point.

fmdkdd: The point of the error copy is for the `flycheck-jump-in-buffer` call above.  We would have to duplicate that path as well?

cpitclaudel: No, I think we could just pass the original error to `flycheck-jump-in-buffer`.

cpitclaudel: Sorry, I was misreading `flycheck-jump-in-buffer`. I got confused by the fact that we pass both a buffer and an error to that function, and it uses both the buffer we pass it *and* the buffer in the error (and presumably expects them to be the same?)

fmdkdd: > it uses both the buffer we pass it and the buffer in the error (and presumably expects them to be the same?)

Yes.  Now that you mention it, we could probably refactor that.  The point is that `flycheck-error-pos` needs the error struct to have the right buffer, or the position will be wrong.

cpitclaudel: Yep, it's very clear now.  Does it make sense to call that function with another buffer than the one in the error? If not, we could drop that extra argument.
 
But let's not delay this independent fix too much, either.  Your patch is quite fine :) You could just push it with a FIXME.

PR #1485, opened 2018-07-28
Allow checker executables to be set to a relative path

CyberShadow:
Hi,

Essentially, my goal is to put `(flycheck-CHECKER-executable . "relative/path/to/checker-script")` in `.dir-locals.el`, include said `checker-script` with the project, and have things "just work".

Below is one way to address this, but I'm open to suggestions.

```
Previously, flycheck-CHECKER-executable could not be set to a relative
path. This is useful when a checker (compatible substitute or wrapper)
is included with the project.

The setting didn't work because the relative path was passed as-is to
executable-find, which returns nil for inputs containing a directory
component.

Address this by introducing flycheck--find-executable, a new function
which checks whether its input has a directory component before
forwarding to executable-find, and update flycheck-executable-find's
default value to point to flycheck--find-executable instead of
executable-find.
```

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -453,23 +453,38 @@ commands through `bundle exec', `nix-shell' or similar wrappers."
   :package-version '(flycheck . "0.25")
   :risky t)
 
-(defcustom flycheck-executable-find #'executable-find
+(defcustom flycheck-executable-find #'flycheck-default-executable-find
   "Function to search for executables.
 
 The value of this option is a function which is given the name or
 path of an executable and shall return the full path to the
 executable, or nil if the executable does not exit.
 
-The default is the standard `executable-find' function which
-searches `exec-path'.  You can customize this option to search
-for checkers in other environments such as bundle or NixOS
+The default is `flycheck-default-executable-find', which searches
+`exec-path' when given a command name, and resolves paths to
+absolute ones.  You can customize this option to search for
+checkers in other environments such as bundle or NixOS
 sandboxes."
   :group 'flycheck
   :type '(choice
-          (const :tag "Search executables in `exec-path'" executable-find)
+          (const :tag "Search executables in `exec-path'"
+                 flycheck-default-executable-find)
           (function :tag "Search executables with a custom function"))
-  :package-version '(flycheck . "0.25")
+  :package-version '(flycheck . "32")
   :risky t)
+
+(defun flycheck-default-executable-find (executable)
+  "Resolve EXECUTABLE to a full path.
+
+If given just a command name (no directory component), search
+`exec-path' using the standard `executable-find' function;
+otherwise, resolve any relative paths to absolute ones."
+  ;; file-name-directory returns non-nil iff the given path has a
+  ;; directory component.
+  (if (file-name-directory executable)
+      (when (file-executable-p executable)
+        (expand-file-name executable))
+    (executable-find executable)))
 
 (defcustom flycheck-indication-mode 'left-fringe
   "The indication mode for Flycheck errors and warnings.


CyberShadow:
> How did you solve the problem prior to this patch?

By placing a wrapper in my PATH which checks PWD and changes behavior depending on the detected current project. This was a solution only I benefited from, and not easily extendable to other project contributors.

> The patch allows for any relative path, but the documentation only talks about absolute paths. Which is it?

Sorry, how did you come to the conclusion that the documentation only talks about absolute paths?

> but I'm not fond of the overloading of the checker command

Sorry, what do you mean by this? Do you mean that the user needs to choose between using the provided checker wrapper (by accepting the `dir-locals` setting), or using the system-wide one (by ignoring said setting), but not both?

fmdkdd:
> This was a solution only I benefited from, and not easily extendable to other project contributors.

Sure, I understand the convenience of the PR.  I just wanted a concrete look at how much effort you would save.

> Sorry, how did you come to the conclusion that the documentation only talks about absolute paths?

You're right, I misread.  The docstrings are clear on that point.

> Sorry, what do you mean by this? 

I meant using the checker command property in two ways.  Previously it had only one meaning:  the name of the checker executable, and we relied on a properly set PATH.  With this patch, the command can also be the *path* to the checker executable, and in that case we override the PATH mechanism.

It's a minor point, but assigning multiple meanings to definitions can create confusion down the road.  Although, in that case you could argue that the way Python checkers now have `python` as command creates a precedent.  But again, minor point ;)

CyberShadow:
> With this patch, the command can also be the path to the checker executable, and in that case we override the PATH mechanism.

Ah, I see what you mean. I don't see the distinction myself, as the behavior is consistent with what you get when you type a command at a shell prompt, or when you execute a command is a shell script, or when you call the C `execvp` function. From this point, it seems that the previous behavior is unexpected - generally, places which allow specifying a command that is then searched in `PATH` also allow specifying a relative or absolute path to an executable.

cpitclaudel:
> I meant using the checker command property in two ways. Previously it had only one meaning: the name of the checker executable, and we relied on a properly set PATH. With this patch, the command can also be the path to the checker executable, and in that case we override the PATH mechanism.

:thinking: don't we already allow absolute paths there?

CyberShadow:
Ah, OK, so `executable-find` works with absolute paths but not relative paths. That's curious. Its documentation doesn't mention it, and I didn't think to check.

fmdkdd:
> I don't see the distinction myself, as the behavior is consistent with what you get when you type a command at a shell prompt

That's a good point. 

> Ah, OK, so executable-find works with absolute paths but not relative paths. 

In any case, you also want the relative paths to work, right?

CyberShadow:
> In any case, you also want the relative paths to work, right?

Yes, that's my goal.

fmdkdd:
> don't we already allow absolute paths there?

It works, but it's not documented as far as I can tell.

flycheck.el
```diff
@@ -460,17 +460,32 @@ The value of this option is a function which is given the name or
...
   :group 'flycheck
   :type '(choice
-          (const :tag "Search executables in `exec-path'" executable-find)
+          (const :tag "Search executables in `exec-path'"
+                 flycheck--find-executable)
           (function :tag "Search executables with a custom function"))
   :package-version '(flycheck . "0.25")
   :risky t)
 
+(defun flycheck--find-executable (executable)
```
fmdkdd: Since we are exposing this function in a configuration variable, it should be public (no double hyphens `--`).

flycheck.el
```diff
@@ -460,17 +460,32 @@ The value of this option is a function which is given the name or
 path of an executable and shall return the full path to the
 executable, or nil if the executable does not exit.
 
-The default is the standard `executable-find' function which
-searches `exec-path'.  You can customize this option to search
-for checkers in other environments such as bundle or NixOS
+The default is `flycheck--find-executable', which searches
```
fmdkdd: You haven't changed the default value though, only one of the options for customize.  I'm fine with changing the default since it should be backwards compatible.

If we change the default, we should bump the `package-version` to `32`.

CyberShadow:
Fixed and fixed :)

cpitclaudel:
> It works, but it's not documented as far as I can tell.

It seems to be fairly common: https://github.com/search?q=flycheck+executable&type=Code , so it'll be good to document it :)

fmdkdd:
> Should I do the package-version bump, or should I leave that to a maintainer?

Go ahead and bump it.

Please also add a line to CHANGES.rst to advertize this change.

flycheck.el
```diff
@@ -453,24 +453,39 @@ commands through `bundle exec', `nix-shell' or similar wrappers."
   :package-version '(flycheck . "0.25")
   :risky t)
 
-(defcustom flycheck-executable-find #'executable-find
+(defcustom flycheck-executable-find #'flycheck-find-executable
```
cpitclaudel: I'm not super happy with the names (we have `flycheck-executable-find`, `flycheck-find-executable`, and `executable-find`).  What do you think of `flycheck-default-executable-find`?

cpitclaudel:
This PR looks good to me, thanks @CyberShadow! I've made one suggestion regarding naming.

CyberShadow:
Sorry for the delayed response! I've renamed the function and bumped `flycheck-executable-find`'s `package-version` as suggested.

fmdkdd:
Looks good.  No idea what's going on with the cla-assistant though.  Please squash your commits; maybe it will wake up the CLA.

PR #1486, opened 2018-07-28
Configure chaining for Python checkers

apiology:
I noticed that I wasn't getting mypy and pylint messages in my Python files - it appears that if you have flake8 installed, it will run exclusively and not attempt to run any other tools.

Tools like mypy (for typechecking) and pylint (for style issues) are complementary to tools like flake8, and I think should run alongside unless there's a big syntax-level error that something like flake8 or pycompile finds.

Changes inside:

- Run Python syntax checkers before running deeper checkers
- Don't bypass pylint or mypy even if other tools installed
- Skip further checkers when syntax-level errors reported
- Fix integration tests accordingly


cpitclaudel:
Hi there,

Thanks a lot for the contribution.

If i understand correctly, you pull request enables mypy for everyone by default, but I'm not sure that's the right default behavior: even if a user has mypy installed, it's not obvious that we should run it for every file.

IIRC, that's why things are currently set up like this.  Can we find a solution that makes it easier to enable mypy without enabling it for everyone by default?

Thanks again!

apiology:
It seems to me that if a solution is to be found, we're going to need to map out what the goals are for that solution.

It sounds like the policy that you're looking for is to make the user take some positive action in their config to enable mypy, rather than have it turned on automatically if installed, like other tools.  Are there other Python tools you'd like to apply the same policy to?  e.g., just because someone has pylint (or even flake8) installed, does it mean they want to see messages from it?  It'd be helpful to understand the intent here - do you remember any more of the context there on why we'd want to disable mypy by default and how that context would apply to other tools?

For the 'how' part of the equation, is there some standard way to make a tool require positive action before it runs, even if it's the first checker in `flycheck-checkers` for the given language?  Sounds like whatever policy we want to apply to mypy should apply regardless of what else is or is not installed, right?  I think maybe python is a bit of an exception case since python-pycompile can always run, but perhaps this situation has come up for other languages?

fmdkdd:
> do you remember any more of the context there on why we'd want to disable mypy by default and how that context would apply to other tools?

From #1354:

>  I think it shouldn't be enabled for all files by default, though, since it's pretty slow. Or maybe it should just run last?

> In any case, if we put python-mypy after python-pycompile, then we force users to opt-in by selecting mypy in their own config

The checker is available, but yes, users have to select it, either by disabling other checkers for `python-mode`, or by tweaking the `next-checker` properties of other Python checkers, as you did in your PR.

As for the policy we follow: We just try to make sensible defaults, and then give users the ability to customize Flycheck as they see fit.  So in that case, we felt that `mypy` was too slow to be included in the default chain of Python checkers.

> is there some standard way to make a tool require positive action before it runs?

I think that @cpitclaudel was suggesting to include an `:enabled` predicate to determine whether we should enable `mypy` for the current file.  I don't think the suggestion was to prompt the user.  

It's true that configuring `next-checker` requires writing some ELisp, and that heavy customization of checker chains is maybe harder than it should be (that is, if you stray too far from the defaults).  But it's the scope of #1072, not this PR.

cpitclaudel:
> It seems to me that if a solution is to be found, we're going to need to map out what the goals are for that solution.

I like this approach :)

> It sounds like the policy that you're looking for is to make the user take some positive action in their config to enable mypy, rather than have it turned on automatically if installed, like other tools.

I think that's right: usually, it's fairly easy to know whether one tool should apply to a given file; but for mypy, it's tricky: most Python files don't have type annotations, and for these files mypy is useless.

> For the 'how' part of the equation, is there some standard way to make a tool require positive action before it runs, even if it's the first checker in flycheck-checkers for the given language? Sounds like whatever policy we want to apply to mypy should apply regardless of what else is or is not installed, right?

I don't think we have a good story for this.  One of the issues is that flycheck-disabled-checkers is hard to set buffer- or dir-locally.

fmdkdd:
> One of the issues is that flycheck-disabled-checkers is hard to set buffer- or dir-locally.

Do you mean, hard to set interactively?  It's just a list of symbols, and it's already buffer-local, so I don't see why it's hard to set using ELisp.

We could have a nicer UI for toggling that easily without having to mess with buffer-, file- or dir- locals though.

dakra:
For my projects I have this in my `.dir-locals.el`:

``` emacs-lisp
((python-mode . ((flycheck-checker . python-mypy))))
```

To activate mypy and pylint checking.

But imo I like this PR better where just all checkers
are enabled if they're available.

Most python devs create a separate virtualenv for
their projects and if they don't want mypy checking
then they could just *not* install mypy in the virtualenv (and globally).

fmdkdd:
> I guess the problem I'm pointing out is that list variables are too easy to override entirely, and not easy enough to add or remove things from.

Ah yes, I see the problem now.  It's good to keep in mind for #1072, but do we want to tackle that for this PR though?

fmdkdd:
> In conclusion, I continue favor to this PR as I implemented it, as I think the performance impact times the likelyhood of including mypy for folks who got it accidentally included in their python environments is a pretty small total area.

A reasonable argument.  I'm not against merging, but I'll let @cpitclaudel have the final say.

cpitclaudel:
> I dug into this - unfortunately, I can't think of a reasonable way to indicates type annotations exist in a given file.

Would it be particularly bad to request people who need mypy to add a file-local setting for flycheck-checker?  Just something like `# -*- flycheck-checker: python-mypy -*-` at the beginning of the file.

If it's too much, I'm mostly fine with merging as is.  But I have another concern if we do: what will then be the recommended way to turn off flake8 and mypy globally, and enable mypy only in certain specific files?  AFAICT, flycheck-disabled-checkers takes precedence over flycheck-checker.

apiology:
> Would it be particularly bad to request people who need mypy to add a file-local setting for flycheck-checker? Just something like `# -*- flycheck-checker: python-mypy -*-` at the beginning of the file.

I'm sure that'll be on OK solutions for some developers - especially folks who are working on solo projects, especially where those projects are made of one or two files.  

For open source or private projects that involve multiple developers, I think that many users of other editors would object to Emacs-specific syntax in their otherwise editor-agnostic source files.

Additionally, for projects that involve a ton of files, needing to remember that annotation for every file gets excessive.  

As I noted above, the dir-local solution above would be more appropriate compromise for multi-file or multi-developer python projects, as at least it's only a single file being added per project.  My argument is that we can do better as a default config trade-off.

> If it's too much, I'm mostly fine with merging as is. But I have another concern if we do: what will then be the recommended way to turn off flake8 and mypy globally, and enable mypy only in certain specific files? AFAICT, flycheck-disabled-checkers takes precedence over flycheck-checker.

I think that scenario is pretty unlikely, so in that scenario, I think it'd be fine to ask someone to tailor their next-checker setup the way they'd like it.

Here's my reasoning on why it's unlikely--keep me honest:

The scenario you're laying out, if I understand it, is someone who:

* Has flake8 and mypy tools installed into their Python environment (otherwise, flycheck would not try to use them).
* Has a global Python package environment and not project-specific virtual environments (otherwise, they could uninstall flake8 from the project-specific virtual environment, or, more likely, would have never installed it)
* Has a project with Python files in it, and wants to run mypy against only some of them (and so dir-local variables wouldn't be appropriate).
* Works on a team, so can't use buffer-local variables on common source files.

Am I understanding the scenario correctly?  Or are we (e.g.) thinking of a scenario where file-local variables are allowed by the team?  I'm happy to think it through once I have an understanding of the scenario you're considering.

If that's the scenario, I do continue to think that the small likelihood of situations like this times the reasonable burden of running mypy (see above for timing) makes the overall harm small compared to the overall benefit of mypy users having flycheck work out of the box when they want to use mypy with Emacs.

fbergroth:
I'd like to see this PR merged.

Currently if you use the mypy checker, flake8 only runs on file-save. If mypy is set as a next-checker after flake8, you still get flake8 while typing, and the mypy linting when saving.

cpitclaudel:
> I'd like to see this PR merged.

So would I :) I am currently on holidays.  I'll respond next Tuesday.  Sorry for the delay!

fmdkdd:
@cpitclaudel Do you still want this merged?

cpitclaudel:
Yes, let's merge this.

fmdkdd:
Closing in favor of #1679.

PR #1491, opened 2018-08-10
Fix to make error with no line number relevant

DamienCassou:
When a command checker does not provide any line number for an error, the line is by default set to nil. This makes the error disappear because of `flycheck-relevant-error-p`. This PR makes the default line number to be 1 instead of nil.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6076,7 +6076,7 @@ otherwise."
             (message (match-string 4 err))
             (id (match-string 5 err)))
         (flycheck-error-new-at
-         (flycheck-string-to-number-safe line)
+         (or (flycheck-string-to-number-safe line) 1)
          (flycheck-string-to-number-safe column)
          level
          (unless (string-empty-p message) message)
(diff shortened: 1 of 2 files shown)


fmdkdd:
Usually we use `flycheck-fill-empty-line-numbers` as `:error-filter` when we want this behavior.  Look for instance at the `emacs-lisp` checker.  This replaces nil with 0 instead of 1 (hence why the test on 24.5 currently fails).

This way, the checker definition has to be explicit in giving meaning to errors without line numbers.  Instead, if we default to 1 as the PR does, then it's harder to make these errors irrelevant in `:error-filter`, right?

Does using `:error-filter` satisfies your use case?

DamienCassou:
I agree. I went for a different solution in https://github.com/purcell/flycheck-ledger/pull/8.

You may close the PR. Sorry for the noise.

fmdkdd:
> You may close the PR. Sorry for the noise.

It's fine, don't sweat it.  :+1:

PR #1495, opened 2018-08-16
Add column numbers to python-mypy

fbergroth:
Just a minor enhancement. I also had to update some tests for pylint2.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9014,12 +9014,15 @@ See URL `https://docs.python.org/3.4/library/py_compile.html'."
 
 See URL `http://mypy-lang.org/'."
   :command ("mypy"
+            "--show-column-numbers"
             (config-file "--config-file" flycheck-python-mypy-ini)
             (option "--cache-dir" flycheck-python-mypy-cache-dir)
             source-original)
   :error-patterns
-  ((error line-start (file-name) ":" line ": error:" (message) line-end)
-   (warning line-start (file-name) ":" line ": warning:" (message) line-end))
+  ((error line-start (file-name) ":" line ":" column ": error:" (message)
+          line-end)
+   (warning line-start (file-name) ":" line ":" column  ": warning:" (message)
+            line-end))
   :modes python-mode
   ;; Ensure the file is saved, to work around
   ;; https://github.com/python/mypy/issues/4746.
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -3676,7 +3676,7 @@ Why not:
     (flycheck-ert-with-env '(("LC_ALL" . nil))
       (flycheck-ert-should-syntax-check
        "language/text.txt" '(text-mode markdown-mode)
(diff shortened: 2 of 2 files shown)


fbergroth:
The following change seems to work fine as well, which let's you get mypy errors as you type, and not only on-save. Should I include it as well? The predicate is the same as `flycheck-buffer-saved-p`, except we don't check `(not (buffer-modified-p buffer))`.
```diff
--- a/flycheck.el
+++ b/flycheck.el
@@ -9012,24 +9012,27 @@ See URL `https://docs.python.org/3.4/library/py_compile.html'."
 (flycheck-define-checker python-mypy
   "Mypy syntax and type checker.  Requires mypy>=0.580.
 
 See URL `http://mypy-lang.org/'."
   :command ("mypy"
             "--show-column-numbers"
+            "--shadow-file" source-original source
             (config-file "--config-file" flycheck-python-mypy-ini)
             (option "--cache-dir" flycheck-python-mypy-cache-dir)
             source-original)
   :error-patterns
   ((error line-start (file-name) ":" line ":" column ": error:" (message)
           line-end)
    (warning line-start (file-name) ":" line ":" column  ": warning:" (message)
             line-end))
   :modes python-mode
   ;; Ensure the file is saved, to work around
   ;; https://github.com/python/mypy/issues/4746.
-  :predicate flycheck-buffer-saved-p
+  :predicate (lambda (&optional buffer)
+               (let ((file-name (buffer-file-name buffer)))
+                 (and file-name (file-exists-p file-name))))
   :next-checkers '(t . python-flake8))
```

fmdkdd:
> The following change seems to work fine as well, which let's you get mypy errors as you type, and not only on-save. Should I include it as well? 

This `--shadow-file` has been confusing me from the start.  If you provide the option, `mypy` does look at the current buffer contents, and not just the file on disk, right?

fbergroth:
> One question though: does the --show-column-numbers require a version fresher than 0.580?

Nope, it was added in a previous version.

> This --shadow-file has been confusing me from the start. If you provide the option, mypy does look at the current buffer contents, and not just the file on disk, right?

Yeah, the buffer is saved to a temporary file, and that file will be read instead. From the mypy docs:

> --shadow-file SOURCE_FILE SHADOW_FILE
> When mypy is asked to type check SOURCE_FILE, this flag makes mypy read from and type check the contents of SHADOW_FILE instead. However, diagnostics will continue to refer to SOURCE_FILE.

But reading #1354, @cpitclaudel chose to not include this, so I'd like to hear his input.

fmdkdd:
> But reading #1354, @cpitclaudel chose to not include this, so I'd like to hear his input.

Sure.  I'll just merge this one as-is, and we can continue the discussion on `--shadow-file` here, and eventually open a new PR if we want to include the additional patch.

cpitclaudel:
:+1: thanks.  I'm in favor of using shadow-file, as long as we work around the bug related to unsaved files.

PR #1496, opened 2018-08-18
Improve handling of Chicken Scheme errors

jwhitbeck:
This commit adds two-error patterns to the Chicken Scheme flycheck
checker.

On Chicken-scheme version 4.12.0 (rev 6ea24b6), the compiler generates
error and warning messages that aren't properly parsed by flycheck.

Here are two examples:

```
Warning: in local unknown procedure,
  in toplevel procedure `foo':
  (foo.scm:36) in procedure call to `read-line', expected argument #1 of type `input-port' but was given an argument of type `output-port'
```

```
Error: foo.scm:43: procedure `hash-passwd' called with wrong number of arguments
```


flycheck.el
```diff
@@ -9881,6 +9881,10 @@ See URL `http://call-cc.org/'."
             "Warning: " (zero-or-more not-newline) ":\n"
             (one-or-more (any space)) "(" (file-name) ":" line ") " (message)
             line-end)
+   (warning line-start
+            "Warning: " (zero-or-more not-newline) ":\n"
+            (one-or-more (any space)) "(" (file-name) ":" line ") " (message)
+            line-end)
```
fmdkdd: I might be missing something, but it looks identical to the pattern just above it?

fmdkdd:
Thanks for the contribution :+1: 

It would be great if you could add example tests under `test/flycheck-test.el`, backed by Scheme files under `test/resources/language/chicken`.

I'll fix the tests reported for other checkers on the CI in the meantime.

flycheck.el
jwhitbeck: Good catch :) I copied to the wrong form from my personal config. I just pushed a new version.

flycheck.el
```diff
@@ -9890,9 +9895,21 @@ See URL `http://call-cc.org/'."
...
           line-end)
+   ;; A of, version 4.12.0, the chicken compiler doesn't currently provide a
+   ;; line number for this error.
+   (error line-start "Syntax error: illegal atomic form\n\n"
+          (message (one-or-more not-newline)
+                   (zero-or-more "\n"
+                                 (zero-or-more space)
+                                 (zero-or-more not-newline))
+                   (one-or-more space) "<--")
+          line-end)
```
jwhitbeck: The chicken compiler immediately crashes if a single empty `()` form is present in the buffer, thereby disabling flycheck entirely for the buffer. I added this form to at least ignore the noisy errors (this pops up every few seconds when coding with paredit), but I don't think this is the best way to proceed. Is there a way of preventing flycheck from updating if a `()` form is present in the buffer?

jwhitbeck: Example compiler output:

```
Syntax error: illegal atomic form

	()

	Expansion history:

	library.scm:3158: ##sys#print	  
	library.scm:3188: case-sensitive	  
	library.scm:3189: keyword-style	  
	library.scm:3190: ##sys#print-length-limit	  
	library.scm:3281: outstr	  
	library.scm:3188: g5138	  
	extras.scm:630: loop	  
	extras.scm:621: newline	  
	library.scm:3154: ##sys#write-char/port	  
	library.scm:3146: ##sys#write-char-0	  
	library.scm:3139: g4941	  
	extras.scm:630: loop	  
	support.scm:122: print-call-chain	  
	library.scm:3882: ##sys#get-call-chain	  
	library.scm:3834: ##sys#make-vector	  
	library.scm:1371: ##sys#allocate-vector	  	<--

Error: shell command terminated with non-zero exit status 17920: '/usr/bin/chicken' 'test/resources/language/chicken/io-type-warning.scm' -output-file 'test/resources/language/chicken/io-type-warning.c' -analyze-only -local
```

jwhitbeck: There seems to be a class of compiler errors that just exit immediately without producing usable output.

Here is another one:

```
Syntax error: malformed expression

	("foo" . "bar")

	Expansion history:

	library.scm:3348: outchr	  
	library.scm:3188: g5148	  
	library.scm:3327: outchr	  
	library.scm:3188: g5148	  
	library.scm:3359: outchr	  
	library.scm:3188: g5148	  
	extras.scm:630: loop	  
	extras.scm:621: newline	  
	library.scm:3154: ##sys#write-char/port	  
	library.scm:3146: ##sys#write-char-0	  
	library.scm:3139: g4941	  
	extras.scm:630: loop	  
	support.scm:122: print-call-chain	  
	library.scm:3882: ##sys#get-call-chain	  
	library.scm:3834: ##sys#make-vector	  
	library.scm:1371: ##sys#allocate-vector	  	<--

Error: shell command terminated with non-zero exit status 17920: '/usr/bin/chicken' 'test/resources/language/chicken/io-type-warning.scm' -output-file 'test/resources/language/chicken/io-type-warning.c' -analyze-only -local
```

jwhitbeck:
Thanks for the feedback. 

After adding some tests, I ended removing one of the error-patterns I added, as I couldn't reproduce the issue. If it becomes a problem, I'll open a new PR.

I just added a test for the IO type mismatch error.

Finally, what is the best way to handle the case where the compiler crashes before being able to produce a useful message (see my comment on the relevant error-pattern)?

flycheck.el
fmdkdd: I think the good approach here is to match on all `Syntax error: ...`, not just the `illegal atomic form`.

Too bad `chicken` doesn't give a line/column information for these, as that would be more useful.

test/flycheck-test.el
```diff
@@ -4273,12 +4273,25 @@ The manifest path is relative to
...
   (flycheck-ert-should-syntax-check
    "language/chicken/syntax-error.scm" 'flycheck/chicken-mode
    '(1 nil error "not enough arguments\n\n\t(define)\n\n\tExpansion history:\n\n\textras.scm:630: loop\t  \n\textras.scm:633: get-output-string\t  \n\tlibrary.scm:3655: ##sys#substring\t  \n\tlibrary.scm:603: ##sys#make-string\t  \n\tlibrary.scm:511: ##sys#allocate-vector\t  \n\textras.scm:633: ##sys#print\t  \n\tlibrary.scm:3188: case-sensitive\t  \n\tlibrary.scm:3189: keyword-style\t  \n\tlibrary.scm:3190: ##sys#print-length-limit\t  \n\tlibrary.scm:3319: ##sys#number?\t  \n\tlibrary.scm:3349: outstr\t  \n\tlibrary.scm:3188: g5138\t  \n\tsupport.scm:122: print-call-chain\t  \n\tlibrary.scm:3882: ##sys#get-call-chain\t  \n\tlibrary.scm:3834: ##sys#make-vector\t  \n\tlibrary.scm:1371: ##sys#allocate-vector\t  \t<--"
        :checker scheme-chicken)))
 
+(flycheck-ert-def-checker-test scheme-chicken scheme syntax-error-no-line-number
+  (flycheck-ert-should-syntax-check
+   "language/chicken/syntax-error-no-line-number.scm" 'flycheck/chicken-mode
+   ;; As the chicken-compiler doesn't print a line number for these errors,
```
fmdkdd: But we probably do want the syntax error to be reported in the Flycheck buffer, lest we think everything is okay.

Since chicken does not report any line number, we can use a default one.  Please add an `:error-filter` to the checker definition like this:

```emacs-lisp
  :error-filter flycheck-fill-empty-line-numbers
```

This will add a default line number.

jwhitbeck: That's a much better experience! Thanks for pointing this out.
I've added that line, reworked the message capture for the matching error pattern and updated the relevant test accordingly.
Please take another look.

jwhitbeck:
I just again encountered the third break-case I had noticed. Added an error-pattern and a test to handle it.

flycheck.el
```diff
@@ -9890,10 +9903,20 @@ See URL `http://call-cc.org/'."
                                  (zero-or-more not-newline))
                    (one-or-more space) "<--")
           line-end)
+   ;; A of version 4.12.0, the chicken compiler doesn't provide a
+   ;; line number for this error.
+   (error line-start "Syntax error: "
```
fmdkdd: So, to be clear: the pattern above this one shows there are some syntax errors which *do* have line numbers, while some other do not?

flycheck.el
```diff
@@ -9877,11 +9877,24 @@ See URL `http://call-cc.org/'."
...
+            (one-or-more (any space)) "(" (file-name) ":" line ") " (message)
+            line-end)
    (warning line-start
             "Warning: " (zero-or-more not-newline) ":\n"
             (one-or-more (any space)) "(" (file-name) ":" line ") " (message)
             line-end)
    (error line-start "Error: (line " line ") " (message) line-end)
+   ;; A of version 4.12.0, the chicken compiler doesn't provide a
+   ;; line number for this error.
+   (error line-start "Error: "
```
fmdkdd: I'd prefer if we moved that pattern just after the other "Error: " one (the last one), since it would be easier to compare the two.

flycheck.el
jwhitbeck: That's correct...

flycheck.el
jwhitbeck: Done. Please take another look.

flycheck.el
fmdkdd: I just wanted to make sure, as it seems a strange behavior.  But at least now we can report these error to the user.

fmdkdd:
LGTM.  Thank you for the quick turn-around, and for adding tests.

:+1:

PR #1512, opened 2018-10-16
Fix flycheck-default-executable-find to work with exec-suffixes

### LINKED ISSUE #1509: Cannot enable emacs-lisp flychecker
I just installed flycheck. When I go to a .el file and run flycheck-verify-setup I get:

```
Syntax checkers for buffer anagram.el in emacs-lisp-mode:

  emacs-lisp (disabled)
    - may enable: Automatically disabled!
    - predicate:  t
    - executable: Not found

  emacs-lisp-checkdoc (disabled)
    - may enable: Automatically disabled!
    - predicate:  t
    - executable: Not found

checkers.

--------------------

Flycheck version: 32snapshot (package: 20180907.1319)
Emacs version:    26.1
System:           x86_64-w64-mingw32
Window system:    w32
```

I didn't know Emacs lisp would require a binary...? Or is it that Emacs needs to be in my $PATH?

Same problem with emacs -Q

(Yes, I know I'm using Windows...)

CyberShadow:
Change `flycheck-default-executable-find`'s strategy, so that it works on platforms with implicit executable suffixes, such as Win32.

Fixes #1509.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,9 @@
   - Add ``flycheck-sh-bash-args`` to pass arguments to ``sh-bash`` [GH-1439].
   - Add ``flychjeck-eslint-args`` to pass arguments to ``javascript-eslint``
     [GH-1360]
+  - Add ``flycheck-default-executable-find``, the new default value for
+    ``flycheck-executable-find``, to allow using relative paths to checkers
+    (set e.g. in file or dir-local variables). [GH-1485]
 
 - Improvements
 
--- a/flycheck.el
+++ b/flycheck.el
@@ -476,15 +476,17 @@ sandboxes."
 (defun flycheck-default-executable-find (executable)
   "Resolve EXECUTABLE to a full path.
 
-If given just a command name (no directory component), search
-`exec-path' using the standard `executable-find' function;
-otherwise, resolve any relative paths to absolute ones."
+Like `executable-find', but supports relative paths.
+
+Attempts invoking `executable-find' first; if that returns nil,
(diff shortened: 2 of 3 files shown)


fmdkdd:
Oh, regarding the CLA situation, I finally found the issue: the gist containing the actual CLA text is now a 404.  I'll resurrect it.

fmdkdd:
The annoying side-effect of re-adding the CLA was that we lost all signatures.  But at least it should work now.

CyberShadow:
Added a changelog entry.

It would be nice to confirm that this does in fact fix the original symptoms in #1509, i.e. emacs-lisp not working on Windows with the Emax64 flavor when Emacs is invoked without an extension in its `argv[0]`.

CyberShadow:
> Added a changelog entry.

BTW, the bug was introduced in the current release. Should the changelog be documenting fixes to regressions introduced in the same release they were fixed?

fmdkdd:
> Should the changelog be documenting fixes to regressions introduced in the same release they were fixed?

Probably not.  The changelog entry should state that we added the `flycheck-default-executable-find` instead, and how it is useful.

cpitclaudel:
Great work, thanks a lot. And I like the new code even better than the original implementation, so thumbs up all around :)

CyberShadow:
Changelog amended, how does that look?

fmdkdd:
> Changelog amended, how does that look?

Great.  Thank you.

fmdkdd:
@beetleb Thanks!

@CyberShadow Thanks for the quick fix.

PR #1518, opened 2018-11-21
add org-mode to proselint modes

gdkrmr:


fmdkdd:
Hmm. `org-mode` is derived from `outline-mode` which is derived from `text-mode`.  Couldn't we just tell Flycheck to enable proselint in all text-mode derived modes?  @cpitclaudel WDYT?

gdkrmr:
is there a way to force-enable a linter for a mode? This would be useful for inline documentation of code.

fmdkdd:
@gdkrmr You can add a checker to a mode with `flycheck-add-mode`, but if there is already a checker selected for that mode, the one you add probably won't get picked up unless you also use `flycheck-add-next-checker`.   But that becomes tedious to do if you want to run proselint in all your programming modes.

cpitclaudel:
Sorry for the long silence. I think we should merge this patch, and discuss the option of enabling proselint more globally in a separate issue.

Issue #1517, opened 2018-11-20
`r-lintr` checker should support ess-r-mode


ashiklom:
The `r-lintr` checker no-longer works with the current version of Emacs ESS. I _think_ this is because the `r-lintr` checker is configured around `ess-mode`, but as far as I can tell, recent changes to ESS (perhaps [this commit?](https://github.com/emacs-ess/ESS/commit/59686509231c03cf4af0bdcda855f42dc64cc98d)) changed the mode inheritance so `ess-r-mode` (the default mode for R scripts) is now separate from `ess-mode`.

When I changed the `r-lintr` `:mode` from `ess-mode` to `ess-r-mode` and reloaded the script, the checker started working again.

Incidentally, the new inheritance logic may allow you to simplify some of the `r-lintr` logic -- for instance, you may not need the `:predicate` that checks the current `ess-language` anymore.

cpitclaudel:
Thanks a lot for this report. I'd love to merge a change along these lines. Would you like to open a pull request?

ashiklom:
Closed via #1521.

PR #1521, opened 2018-11-27
Add ess-r-mode to r-lintr modes

ashiklom:
Temporary patch for #1517. Should preserve back compatibility with earlier versions of ESS. In the longer term, a more comprehensive revision of `r-lintr` may be possible (see discussion in #1517).


fmdkdd:
Let's continue the discussion of the proper fix in #1517.

PR #1525, opened 2018-12-05
Update invocation of `go tool vet`

### LINKED ISSUE #1523: go-vet is broken because "go vet tool" was deprecated
`go-vet` checker does not work anymore because `go vet tool` stopped working after recent refactoring s in `vet`. For more details see https://github.com/golang/go/issues/28869.

Also `-printfuncs`, `-shadow`, and `-tags` parameters are deprecated in `go vet`. `-all` is ignored.

One of proposed solutions is to use `go vet` instead of `go vet tool`. I have tried that locally. The problem is when the source file has syntax errors and `go vet` returns only those syntax errors and does not display vet warnings. Otherwise it works. 

Any ideas are welcome.

fmdkdd:
Fix #1523, which see.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7644,7 +7644,7 @@ enabled. "
 
 See URL `https://golang.org/cmd/go/' and URL
 `https://golang.org/cmd/vet/'."
-  :command ("go" "tool" "vet" "-all"
+  :command ("go" "vet" "-all"
             (option "-printfuncs=" flycheck-go-vet-print-functions concat
                     flycheck-option-comma-separated-list)
             (option-flag "-shadow" flycheck-go-vet-shadow)


fmdkdd:
Hmm, probably we need to update the Go version in [docker-tools](https://github.com/flycheck/docker-tools), but Go 1.12 hasn't been released yet.

ysmolsky:
go tool vet supported options that no longer supported by `go vet`. For example `-all` does not do anything. 2nd: output format from go vet is different. It has columns added to it.

fmdkdd:
Oh, so they are actually *different tools* that happen to have a very close name? 

@ysmolsky Feel free to create another PR if you have a patch ready, otherwise we can work on this one.

Besides updating the arguments and parser, we also need to update the `verify` and `predicate` which call `go tool` to check for the presence of `vet`.  I'm assuming `go vet` is bundled and always present?

dominikh:
`go vet` is always present, unless your Go setup is broken.

r-medina:
there are other references to `go tool vet` https://github.com/flycheck/flycheck/pull/1525/files#diff-374cdd9563b826c712b1dc6a373a0bc1R7618

juanwolf:
@fmdkdd I made some changes on your PR (https://github.com/flycheck/flycheck/compare/fix-1523-go-tool-vet-deprecated...juanwolf:fix-1523-go-tool-vet-deprecated). Some feedback would be appreciated as those are my first lines ever of lisp.

Regarding the integration tests, I ran: `make LANGUAGE=go integ` and but got few errors. 

Anyway, there's some big changes from the `go tool vet`. Even to get _shadow_ checks some changes are needed regarding the [Go Changelog](https://golang.org/doc/go1.12#vet). Maybe we could have a check to verify if `golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow` is installed  but I am not quite sure how to implement it.

vincekd:
Changed all references to `go tool vet` and updated verification here https://github.com/flycheck/flycheck/pull/1548

fmdkdd:
Closing in favor of #1548.

fmdkdd:
@juanwolf Thanks, we'll probably drop `shadow` at first, since it looks unwieldly.  Feel free to contribute on #1548.

PR #1536, opened 2019-01-20
python-mypy: Add patterns for errors without cols

jbaum98:
Even with the --show-columns flag, mypy sometimes doesn't emit columns. We should have patterns for errors/warnings with and without columns so they can still be parsed.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9197,10 +9197,10 @@ See URL `http://mypy-lang.org/'."
             (option "--cache-dir" flycheck-python-mypy-cache-dir)
             source-original)
   :error-patterns
-  ((error line-start (file-name) ":" line ":" column ": error:" (message)
-          line-end)
-   (warning line-start (file-name) ":" line ":" column  ": warning:" (message)
-            line-end))
+  ((error line-start (file-name) ":" line (optional ":" column)
+          ": error:" (message) line-end)
+   (warning line-start (file-name) ":" line (optional ":" column)
+            ": warning:" (message) line-end))
   :modes python-mode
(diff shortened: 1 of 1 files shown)


fmdkdd:
> Even with the --show-columns flag, mypy sometimes doesn't emit columns.

Sounds like a bug with `mypy`.  Can you check upstream if this behavior is intended?

jbaum98:
I haven't asked them, but https://github.com/python/mypy/blob/f03ad7094e7a9bd174011e2ae3bc0c7204c95e22/mypy/errors.py#L43 seems to suggest that they don't always have a column.

flycheck.el
```diff
@@ -9199,8 +9199,10 @@ See URL `http://mypy-lang.org/'."
   :error-patterns
   ((error line-start (file-name) ":" line ":" column ": error:" (message)
           line-end)
+   (error line-start (file-name) ":" line ": error:" (message) line-end)
```
fmdkdd: Instead of duplicating the patterns, why not make the `":" column` part optional with the `(opt ` or `(? ` rx patterns?

jbaum98: Good point! I followed an example from somewhere else in the source code and used `optional`; not sure what the difference is between `opt`, `optional` and `?` is.

fmdkdd: By the `rx` docstring, `opt`  and `optional` are aliases, but `?` causes that part to be greedy regexp, but I'm not sure it makes a difference here.

fmdkdd:
LGTM.  Thanks!

PR #1541, opened 2019-01-27
Implement a checker for the golang `staticcheck` tool

Gastove:
### Commit Message:
This commit implements `go-staticcheck`, a checker wrapping the `staticcheck`
tool. `staticcheck` is the successor to `megacheck`, which is now
deprecated.

This commit includes:

1. A checker definition for `go-staticcheck`.
2. The introduction of a new variable, `flycheck-go-version`, which can be
passed to `staticcheck` to specify the version of the Go compiler to perform
checks compatible with.
3. A test for `staticcheck`.
4. Updates to appropriate documentation.
5. Updates to other Golang syntax checkers to prefer falling back to
`staticcheck` before `megacheck`.

The documentation pass includes:

1. `rst` docs for `go-staticcheck` itself.
2. An update to the documentation for `megacheck`, specifying that it's
deprecated.
3. Documentation for the new `flycheck-go-version` variable.
4. Updating the definition of `flycheck-go-build-tags` to correctly declare the
full list of checkers that can use it.

### A Note about this PR

So! Went to get this tested, and in doing so, discovered a few things. While the specs all pass on my local machine, the integration tests for go... do not. This is because the tests for Go tooling are hardcoded with the `go` binary installed in `/usr/local/bin`; on Fedora, at least, `go` is installed in `/usr/bin`, so those tests fail as well. 

I'm currently testing changes to the [all-tools Dockerfile](https://github.com/flycheck/docker-tools/blob/master/all-tools/Dockerfile); I'll make sure that PR is linked to this one once I've got everything together. My worldview here is that this means the tests on this PR wont pass until my PR is merged in to `docker-tools`, which is fine. 

I'll look in to updating the tests to more flexibly find the `go` binary if I have time!

Thanks for a great tool, y'all, and my complements on _superb_ contributor docs!

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5972,6 +5972,35 @@ about Reek."
            errors))))
     (nreverse errors)))
 
+(defun flycheck-parse-go-staticcheck (output checker buffer)
+  "Parse staticheck warnings from JSON OUTPUT.
+
+CHECKER and BUFFER denote the CHECKER that returned OUTPUT and
+the BUFFER that was checked respectively.
+
+See URL `https://staticcheck.io/docs/formatters' for more
+information about staticheck."
+  (let ((errors nil))
+    (dolist (msg (flycheck-parse-json output))
+      (let-alist msg
+        (push
+         (flycheck-error-new-at
+          .location.line
+          .location.column
+          (pcase .severity
+            (`"error"   'error)
+            (`"warning" 'warning)
+            (`"ignored" 'info)
+            ;; Default to warning for unknown .severity
+            (_          'warning))
+          .message
+          :id .code
+          :checker checker
+          :buffer buffer
+          :filename .location.file)
+         errors)))
+    (nreverse errors)))
+
 (defun flycheck-parse-tslint (output checker buffer)
   "Parse TSLint errors from JSON OUTPUT.
 
@@ -7955,14 +7984,11 @@ recommended to migrate to `staticcheck'.
 versions of go\". `staticheck' can target earlier versions (with
 limited features) if `flycheck-go-version' is set. See URL
 `https://staticcheck.io/'."
-  :command ("staticcheck"
+  :command ("staticcheck" "-f" "json"
             (option-list "-tags" flycheck-go-build-tags concat)
-            (option "-go" flycheck-go-version)
-            ;; Run in current directory to make staticcheck aware of symbols
-            ;; declared in other files.
-            ".")
-  :error-patterns
-  ((warning line-start (file-name) ":" line ":" column ": " (message) line-end))
+            (option "-go" flycheck-go-version))
+
+  :error-parser flycheck-parse-go-staticcheck
   :modes go-mode)
 
 (flycheck-define-checker groovy
--- a/flycheck.el
+++ b/flycheck.el
@@ -190,6 +190,7 @@ attention to case differences."
     go-errcheck
     go-unconvert
     go-megacheck
+    go-staticcheck
     groovy
     haml
     handlebars
@@ -7717,6 +7718,7 @@ See URL `https://golang.org/cmd/gofmt/'."
                   (warning . go-build) (warning . go-test)
                   (warning . go-errcheck)
                   (warning . go-unconvert)
+                  (warning . go-staticcheck)
                   (warning . go-megacheck)))
 
 (flycheck-define-checker go-golint
@@ -7778,7 +7780,17 @@ See URL `https://golang.org/cmd/go/' and URL
                   ;; Fall back if `go build' or `go test' can be used
                   go-errcheck
                   go-unconvert
-                  go-megacheck))
+                  go-staticcheck
+                  go-megacheck)
+  :verify (lambda (_)
+            (let* ((go (flycheck-checker-executable 'go-vet))
+                   (have-vet (member "vet" (ignore-errors
+                                             (process-lines go "tool")))))
+              (list
+               (flycheck-verification-result-new
+                :label "go tool vet"
+                :message (if have-vet "present" "missing")
+                :face (if have-vet 'success '(bold error)))))))
 
 (flycheck-def-option-var flycheck-go-build-install-deps nil (go-build go-test)
   "Whether to install dependencies in `go build' and `go test'.
@@ -7789,13 +7801,26 @@ while syntax checking."
   :safe #'booleanp
   :package-version '(flycheck . "0.25"))
 
-(flycheck-def-option-var flycheck-go-build-tags nil go-build
+(flycheck-def-option-var flycheck-go-build-tags nil
+                         (go-build go-test go-errcheck
+                                   go-megacheck go-staticcheck)
   "A list of tags for `go build'.
 
 Each item is a string with a tag to be given to `go build'."
   :type '(repeat (string :tag "Tag"))
   :safe #'flycheck-string-list-p
   :package-version '(flycheck . "0.25"))
+
+
+(flycheck-def-option-var flycheck-go-version nil go-staticcheck
+  "The version of go that should be targeted by `staticcheck'.
+
+Should be a string representing a version, like 1.6 or 1.11.4.
+See `https://staticcheck.io/docs/#targeting-go-versions' for
+details."
+  :type 'string
+  :safe #'stringp
+  :package-version '(flycheck . "0.32"))
 
 (flycheck-define-checker go-build
   "A Go syntax and type checker using the `go build' command.
@@ -7835,6 +7860,7 @@ Requires Go 1.6 or newer.  See URL `https://golang.org/cmd/go'."
                     (not (string-suffix-p "_test.go" (buffer-file-name)))))
   :next-checkers ((warning . go-errcheck)
                   (warning . go-unconvert)
+                  (warning . go-staticcheck)
                   (warning . go-megacheck)))
 
 (flycheck-define-checker go-test
@@ -7857,6 +7883,7 @@ Requires Go 1.6 or newer.  See URL `https://golang.org/cmd/go'."
                   (string-suffix-p "_test.go" (buffer-file-name))))
   :next-checkers ((warning . go-errcheck)
                   (warning . go-unconvert)
+                  (warning . go-staticcheck)
                   (warning . go-megacheck)))
 
 (flycheck-define-checker go-errcheck
@@ -7886,6 +7913,7 @@ See URL `https://github.com/kisielk/errcheck'."
   :modes go-mode
   :predicate (lambda () (flycheck-buffer-saved-p))
   :next-checkers ((warning . go-unconvert)
+                  (warning . go-staticcheck)
                   (warning . go-megacheck)))
 
 (flycheck-define-checker go-unconvert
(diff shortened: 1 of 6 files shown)


fmdkdd:
Hey, sorry for the delay.  This is a very nice PR, thank you.  I'll try to review it soon enough.

flycheck.el
```diff
@@ -7941,6 +7960,26 @@ Requires Go 1.6 or newer. See URL
...
+`staticcheck' is explicitly fully compatible with \"the last two
+versions of go\". `staticheck' can target earlier versions (with
+limited features) if `flycheck-go-version' is set. See URL
+`https://staticcheck.io/'."
+  :command ("staticcheck"
+            (option-list "-tags" flycheck-go-build-tags concat)
+            (option "-go" flycheck-go-version)
+            ;; Run in current directory to make staticcheck aware of symbols
+            ;; declared in other files.
+            ".")
```
fmdkdd: So if you pass the current file to to staticcheck, it will not look at other imported files?

Gastove: Mmmm, that's a great question. `staticcheck` evolved out of `megacheck`, which (to my understanding) was persnickity about these things. Likely, `staticcheck` isn't, and I'm being overly cautious/cargo-cult-y. Will confirm how this works.

Gastove: So: checking in on this has turned up two things. 

First, to answer your question: if you just give it a file, `staticcheck` will report errors, but not warnings, for other files imported by the file in question. This seems fine to me, and updating the checker to not specify `"."` is easy enough. However, I also remark that this checker, as I've written it, does not exhibit a behaviour I see in other checkers where errors in other files are  reported in the file currently under check. I think it would be correct for this checker to have that behaviour; I'll go try and learn how to do that, but if you have the time to point me in a direction, it would be appreciated!

Second thing I found: `staticcheck` _does_ have a notion of "severity", but it only reports severity if you format its output as JSON. I think it would make this checker dramatically more useful if severity were correctly reported, so I'm going to go see if I can figure out how to parse the JSON output. I don't know that that needs to block this PR; setting the default level of the checker to `warning` or `error` seems like an OK version one to me, if it works for you. But, maybe it's quite easy to ingest JSON in flycheck? I have no idea.

Gastove: Ah so! Three birds, one stone: it turns out, flycheck makes parsing JSON errors extremely easy! (I should have guessed that this would be true -- flycheck has been a lovely project to contribute to.) Modifying this checker to 1) not use "." and 2) to correctly parse JSON output has _also_ resulted in flycheck reporting errors in other files. Fantastic!

I'm adding a buttercup test for the new JSON parser. I've also realized I managed to never commit or push the test files required for the go-staticheck tests I wrote -- so that's embarrassing, but also an easy fix. Will get all the changes pushed very soon here :+1:

flycheck.el
```diff
@@ -7941,6 +7989,23 @@ Requires Go 1.6 or newer. See URL
...
+`megacheck'; while the latter isn't fully deprecated yet, it's
+recommended to migrate to `staticcheck'.
+
+`staticcheck' is explicitly fully compatible with \"the last two
+versions of go\". `staticheck' can target earlier versions (with
+limited features) if `flycheck-go-version' is set. See URL
+`https://staticcheck.io/'."
+  :command ("staticcheck" "-f" "json"
+            (option-list "-tags" flycheck-go-build-tags concat)
+            (option "-go" flycheck-go-version))
```
fmdkdd: Since we are not providing a file to check using `source` or its variants, `staticcheck` will be called with the directory of the current file as working directory.  Is this the correct behavior?  I don't know how Go projects are usually structured, but if you have:

```
├── main.go
└── module
    └── foo.go
```
And you open `foo.go`, does the checker still work as expected?  Or should `staticcheck` always run in the top-level directory of your project?  If the latter, then we should pass the `:working-directory` property to the checker.

flycheck.el
fmdkdd: Thanks for looking into this.  I left another question in the comments below.

flycheck.el
Gastove: I think the most correct behavior is the current one -- which I will call "current-directory". I can see arguments for `source-inplace`, but I think it leads to usability breakdowns. @fmdkdd, you might not care about all the gory details, but I want to spell them out anyways if anyone ever reads this PR and wants to know what's up. So, here's a tl;dr and also a Long Version.

### tl;dr
Go allows many files to be part of the same package, and it doesn't consider those files "different" -- it only sees the package itself. Using `source-inplace` causes `staticcheck` to be unable to "see" declarations in other files in the same package, causing spurrious "undeclared name" errors.

While we would get correct check results by running in the top-level directory of the project, Go very commonly uses vendoring as its dependency management approach. It's easy to clear a million, two million lines of dependencies. This bogs the checker down hellatiously.

### The full story

For the sake of documenting all of this, here's my reasoning:

A useful thing to know about Go is this: most typically, a directory defines a package, and every file within that directory is part of that package. For example: 

```
├── pkg
│  └── lib
│     ├── boop.go
│     └── lib.go
```

Both `lib.go` and `boop.go` declare themselves as being part of package `lib`. Even in totally "flat" Go projects (i.e. all files at the root level, see [Logrus](https://github.com/sirupsen/logrus) as a good example), all Go files are part of the same package (or else, `package_test`, a special case of the same thing). Importantly: files within the same package can use symbols declared by that package, without declaring an import, _no matter which file they're declared in_. Put another way: Go cares about packages, not specific files. Files are there to allow humans to separate a package in to logical hunks, but Go only sees the package. 

Current-directory and `source-inplace` will both report errors on any Go code directly imported by the file-or-package currently under check; this is good. However: `source-inplace` only checks symbols in the current file, cutting off access to every other file in the package. While this _does_ mean that the checker might report a great number of errors in files outside the one currently under check, I think for Go that behavior is the most correct one.

fmdkdd: Thansk for the thorough answer.  The "current-directory" behavior looks good to me.

fmdkdd:
LGTM, but you will need to clear the CI before we can merge.  Currently it's failing on formatting errors in `test-go.`el.  I've merged https://github.com/flycheck/docker-tools/pull/2 so your checker tests should also run now.

mpolden:
Great work, @Gastove! Any chance of fixing the formatting errors and getting this merged soon?

Gastove:
Hello! Apologies -- Life *happened*. Back now! Will get this cleaned up quick as I can :+1:

Gastove:
Hokay. Now passing format and checkdoc; the buttercup tests are passing as well. If I'm reading the (pretty extensive) CI output correctly, I'm currently failing the integration tests. Let me see what I can do about that.

Gastove:
Okay. I'm down to one last failing test. I'm.... pretty baffled by the error I'm getting, however. Can anyone advise? @fmdkdd? 

The error originates on line 3436 of `flycheck-test.el`. Here's the test:

```emacs-lisp
(flycheck-ert-def-checker-test go-staticcheck go nil
  :tags '(language-go external-tool)
  (let ((flycheck-disabled-checkers '(go-golint go-megacheck)))
    (flycheck-ert-with-env
        `(("GOPATH" . ,(flycheck-ert-resource-filename "language/go")))
       "language/go/src/staticcheck/staticcheck1.go" 'go-mode
       '(8 6 error "should omit values from range; this loop is equivalent to `for range ...`"
           :checker go-staticcheck :id "S1005")
       '(12 21 error "calling strings.Replace with n == 0 will return no results, did you mean -1?"
            :checker go-staticcheck :id "SA1018")
       '(16 6 error "func unused is unused"
            :checker go-staticcheck :id "U1000")))))
```
The error from `ert` is `different-types`, like so:
```shell
Test flycheck-define-checker/go-staticcheck/default condition:
    (ert-test-failed
     ((should
       (equal
	(mapcar #'flycheck-error-without-group expected)
	(mapcar #'flycheck-error-without-group flycheck-current-errors)))
      :form
      (equal
       (#s(flycheck-error #<killed buffer> go-staticcheck "/flycheck/test/resources/language/go/src/staticcheck/staticcheck1.go" 8 6 "should omit values from range; this loop is equivalent to `for range ...`" error "S1005" nil)
	  #s(flycheck-error #<killed buffer> go-staticcheck "/flycheck/test/resources/language/go/src/staticcheck/staticcheck1.go" 12 21 "calling strings.Replace with n == 0 will return no results, did you mean -1?" error "SA1018" nil)
	  #s(flycheck-error #<killed buffer> go-staticcheck "/flycheck/test/resources/language/go/src/staticcheck/staticcheck1.go" 16 6 "func unused is unused" error "U1000" nil))
       nil)
      :value nil :explanation
      (different-types
       (#s(flycheck-error #<killed buffer> go-staticcheck "/flycheck/test/resources/language/go/src/staticcheck/staticcheck1.go" 8 6 "should omit values from range; this loop is equivalent to `for range ...`" error "S1005" nil)
	  #s(flycheck-error #<killed buffer> go-staticcheck "/flycheck/test/resources/language/go/src/staticcheck/staticcheck1.go" 12 21 "calling strings.Replace with n == 0 will return no results, did you mean -1?" error "SA1018" nil)
	  #s(flycheck-error #<killed buffer> go-staticcheck "/flycheck/test/resources/language/go/src/staticcheck/staticcheck1.go" 16 6 "func unused is unused" error "U1000" nil))
       nil)))
```
Can anyone advise? It _reads_ like ert is trying to compare the three forms provided (all correctly of type `flycheck-error`) to `nil` -- which, I agree, is a different type. However, if that's what's happening... how on _earth_ is `nil` getting in there?

cpitclaudel:
If I read this correctly, `nil` is just an empty list of errors — IOW, it seems that the checker is returning no errors, and the test suite complains about that.  Do you see errors if you just open that file and turn on Flycheck?

Gastove:
@cpitclaudel aaaahaha. Yes, okay --  that does, at least, help with the interpretation of the error message. However, if I load the checker, then the test files, I am seeing the expected triplet of errors. So. Just have to figure out why no errors are being parsed by the integration test.

Gastove:
Okay, I'm... a bit lost. If I don't disable `go-megacheck` in the test for `go-staticcheck`, then the test fails because there are unexpected `go-megacheck` errors in the test results. However, if I _do_ disable `go-megacheck`, then the test fails because no errors are produced. It seems like I must have wired something up incorrectly. So far, I'm not at all sure _what_.

Gastove:
THERE. Okay: for reasons I *do not* understand, disabling `go-unconvert` for my staticcheck test causes the test to pass. I _suspect_ that this has something to do with the way `:next-checker` has been defined for some of the Go family of checkers. However, I don't understand what's going on clearly enough to feel confident in proposing changes, _and_ I don't want to expand the scope of this PR any further. My inclination is to leave things where they are; please let me know if further changes are required!

cpitclaudel:
:clap: :tada: Thanks a lot! Looking forward to many more contributions :)

cpitclaudel:
(I should have requested a squash — or performed it myself — sorry for the commit history noise @fmdkdd !)

fmdkdd:
@cpitclaudel Arguably, the only benefit of squashing is when using `git blame`.  The downside is added friction when merging PRs.  (I've re-enabled the squash & merge button, so at least we can squash ourselves when commits are not signed).

@Gastove Adding a new checker to a chain will usually have effects on other checker tests, as you figured out.  I wish the test output would make that explicit, but the thing to remember is that tests are meant to mimic the behavior of Flycheck from a user standpoint: when you open a file, the first valid checker is selected, then its `:next-checker` will be run.  So if you add a checker, you may have to disable another one in order for your checker to be run.

Anyway, thanks for pulling through on this PR!  :+1:

Gastove:
@fmdkdd oh yes! I quite like `:next-checker` as a feature for users, and have made use of it myself! That said: the thing that was specifically bewildering me was that disabling `go-megacheck` in my tests _also_ appeared to disable `go-staticcheck`. That is: with `go-megacheck` enabled, my `go-staticcheck` test results would contain results for both linters (expected); with `go-megacheck` disabled, the linting check would return `nil`. The solution was to disable `go-unconvert`, which... I still don't grok. 

I can make a to-do for myself to be sure that behaviour isn't present for users, submit any follow-up work in a new PR. 

Given that `megacheck` is both deprecated _and_ its checks have been fully subsumed by `staticcheck`, I'd also be in favour of a "should activate" condition on `megacheck` that disables it if `staticcheck` is available, or similar. Thoughts?

fmdkdd:
If megacheck has been deprecated and superseded, we should just drop it altogether.  Users who wish to use the checker may still copy its definition to their own config.

If you have time for this, please do open a PR.

Gastove:
Works for me. I'll open an issue, just so I don't lose track of it; I should have time this weekend :+1:

PR #1546, opened 2019-02-26
WIP pylint: Add support for standard-input t

thisch:
The unreleased pylint 2.4 will contain support for linting
files from stdin (see https://github.com/PyCQA/pylint/pull/2746).

This PR updates the minimum pylint requirement to version 2.4 (requires python3!!).

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9514,9 +9514,10 @@ See URL `https://www.pylint.org/'."
                       "--msg-template={path}:{line}:{column}:{C}:{symbol}:{msg}"
                     "--msg-template={path}:{line}:{column}:{C}:{msg_id}:{msg}"))
             (config-file "--rcfile=" flycheck-pylintrc concat)
-            ;; Need `source-inplace' for relative imports (e.g. `from .foo
-            ;; import bar'), see https://github.com/flycheck/flycheck/issues/280
-            source-inplace)
+            ;; pylint takes the original file name as argument when reading
+            ;; from standard input
+            "--from-stdin" source-original)
+  :standard-input t
   :error-filter
   (lambda (errors)
     (flycheck-sanitize-errors (flycheck-increment-error-columns errors)))


cpitclaudel:
This looks great to me.  Thanks a lot!  Our usual policy is to not worry too much about being compatble with too many versions of a tool, since updating a linter is usually fairly easy, especially for something like pylint (through pip).  So I think we should just wait for a little while after a new pylint is released with that option, and make an unconditional switch.

cpitclaudel:
This looks great to me.  Thanks a lot!  Our usual policy is to not worry too much about being compatble with too many versions of a tool, since updating a linter is usually fairly easy, especially for something like pylint (through pip).  So I think we should just wait for a little while after a new pylint is released with that option, and make an unconditional switch.

thisch:
> This looks great to me. Thanks a lot! Our usual policy is to not worry too much about being compatble with too many versions of a tool, since updating a linter is usually fairly easy, especially for something like pylint (through pip). So I think we should just wait for a little while after a new pylint is released with that option, and make an unconditional switch.

Great. I've just updated the description.

thisch:
One problem with this change is that it requires the user of the pylint checker to run pylint in a python3 environment, because pylint-2.4 requires python 3.x. There is unfortunately an (unreleased) flycheck feature about supporting python2, whose changelog entry is 

```
- ``python-pylint`` and ``python-flake8`` are now invoked with ``python -c``,
    to make it easier to change between Python 2 and Python 3 [GH-1113]
```
What do we want to do here? Make a last release of flycheck that has py2/py3  support and drop it in the next release by merging this PR?

cpitclaudel:
> Make a last release of flycheck that has py2/py3 support and drop it in the next release by merging this PR?

I think that's reasonable.  We should make a new release soon.

thisch:
@cpitclaudel can you tell me where I can specify the version of pylint that is used for running the tests?

cpitclaudel:
Here, I think: https://github.com/flycheck/docker-tools/blob/e5a500277ed22a405f816c1e13e3e7a3dd38d3fe/all-tools/Dockerfile#L162

cpitclaudel:
@thisch I still very much want to merge this PR, and the sooner the better; especially so now that Python 2 is deprecated.  We're planning to make a release with the error ranges change soon, and then we can merge this.  Thanks for your patience.

thisch:
There is a fix in pylint that is unfurtunately not yet released: https://github.com/PyCQA/pylint/pull/3248

WIthout this fix, pylint raises an exception, if there is a syntaxerror in the file (can happen very easily while a file is edited).

@PCManticore would it be possible to backport https://github.com/PyCQA/pylint/pull/3248 to 2.4.x and create a new release with it?

cpitclaudel:
Thanks for the update

thisch:
pylint 2.5 has finally been released.  :tada: 

How do we want to inform users of `flycheck` that they have to update `pylint` now?

cpitclaudel:
Wonderful news!

> How do we want to inform users of flycheck that they have to update pylint now?

Typically, we don't ^^ We wait a bit to give people time to update naturally, and then we push the update :) (Typically we wait a bit more for packages that are hard to update: for example GCC has better output now but we don't use it yet because it's hard to update; but pylint is trivial to update, so we don't have to wait too long).

I was hoping to release Flycheck 32 before this, and then to get into a more regular release schedule, but there are a few PRs that I'd like to get into 32 (most importantly this: https://github.com/flycheck/flycheck/pull/1743) if @fmdkdd has time to look at it.

thisch:
> Typically, we don't ^^

Okay ;) Then we can merge this change. I've just added an entry to the breaking changes section in  the `changes.rst` file. That should be enough I guess.

PR #1547, opened 2019-03-04
Added ghdl option for setting IEEE library standard.

wastevensv:
Adds a common option to the syntax checker for VHDL, to specify which IEEE standard library to use. A typcial value is "--ieee=synopsys". I'm not sure if I should set that as the default, or leave it nil.

### DIFF
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -1430,6 +1430,10 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
          The directory to use for the file library.
 
+      .. defcustom:: flycheck-ghdl-ieee-library
+
+         The standard to use for the IEEE library.
+
 .. supported-language:: XML
 
    Flycheck checks XML with `xml-xmlstarlet` or `xml-xmllint`.
--- a/flycheck.el
+++ b/flycheck.el
@@ -10714,7 +10714,7 @@ pass the language standard via the `--std' option."
 (make-variable-buffer-local 'flycheck-ghdl-language-standard)
 
 (flycheck-def-option-var flycheck-ghdl-workdir nil vhdl-ghdl
-  "The directory to use for the file library
+  "The directory to use for the file library.
 
 The value of this variable is either a string with the directory
 to use for the file library, or nil, to use the default value.
@@ -10725,6 +10725,21 @@ When non-nil, pass the directory via the `--workdir' option."
   :package-version '(flycheck . "32"))
 (make-variable-buffer-local 'flycheck-ghdl-workdir)
 
+(flycheck-def-option-var flycheck-ghdl-ieee-library nil vhdl-ghdl
+  "The standard to use for the IEEE library.
+
+The value of this variable is either a string denoting an ieee library
+standard, or nil, to use the default standard.  When non-nil,
+pass the ieee library standard via the `--ieee' option."
+  :type '(choice (const :tag "Default standard" nil)
+                 (const :tag "No IEEE Library" "none")
+                 (const :tag "IEEE standard" "standard")
+                 (const :tag "Synopsys standard" "synopsys")
+                 (const :tag "Mentor standard" "mentor"))
+  :safe #'stringp
+  :package-version '(flycheck . "32"))
+(make-variable-buffer-local 'flycheck-ghdl-ieee-library)
+
 (flycheck-define-checker vhdl-ghdl
   "A VHDL syntax checker using GHDL.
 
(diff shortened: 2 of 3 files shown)


wastevensv:
It looks like the CI is failing because I'm missing some documentation. How do I correct this?

wastevensv:
I corrected the documentation issue, however during the integration tests, it looks like the ghdl binary is missing in the builder image.

doc/languages.rst
```diff
@@ -1413,6 +1413,10 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
          The directory to use for the file library.
 
+      .. defcustom:: flycheck-ghdl-ieee-library
+
+         The standard to use for the ieee library
```
fmdkdd: Missing  a colon at sentence end.

Also, shouldn't `ieee` be `IEEE` when in plain text?  (Also in the docstring below).

flycheck.el
```diff
@@ -10668,6 +10668,18 @@ When non-nil, pass the directory via the `--workdir' option."
   :package-version '(flycheck . "32"))
 (make-variable-buffer-local 'flycheck-ghdl-workdir)
 
+(flycheck-def-option-var flycheck-ghdl-ieee-library nil vhdl-ghdl
+  "The standard to use for the ieee library
```
fmdkdd: First sentence of a docstring always ends with a colon.

flycheck.el
```diff
@@ -10668,6 +10668,18 @@ When non-nil, pass the directory via the `--workdir' option."
...
+
+The value of this variable is either a string denoting an ieee library
+standard, or nil, to use the default standard.  When non-nil,
+pass the ieee library standard via the `--ieee' option."
+  :type '(choice (const :tag "Default standard" nil)
+                 (string :tag "IEEE library standard"))
```
fmdkdd: Is there an exhaustive list of standards?  If so, we could hardcode them to make the customization more useful.

fmdkdd:
> I'm not sure if I should set that as the default, or leave it nil.

Depends on how common it is to use it.  The default should be helpful, but not annoying.  I don't use GHDL so I'll let you decide.  Leave it nil is fine.

>  it looks like the ghdl binary is missing in the builder image.

Yep, I ran out of steam when I first built the image.  Help is always welcome though ;)

flycheck.el
wastevensv: Do you mean a period? I don't see a colon at the end of the other docstrings. I also noticed the docstring for the option above my addition (flycheck-ghdl-workdir) is missing a period. Should I add that or make a separate PR,  since it doesn't have to do with the option I'm adding?

fmdkdd: Yes of course, period.  My bad.

Go ahead and fix the other docstring.  No need for a separate PR :)

flycheck.el
wastevensv: Found a list [in ghdl docs](https://ghdl.readthedocs.io/en/latest/using/InvokingGHDL.html#cmdoption-ghdl-ieee). Changed to hardcoded options in latest commit.

flycheck.el
wastevensv: Fixed both docstrings in latest commits.

doc/languages.rst
wastevensv: Fixed in latest commits.

wastevensv:
I might add ghdl to the builder image in another PR whenever I have the time. What would I need to edit to make this change?

fmdkdd:
> What would I need to edit to make this change?

See https://github.com/flycheck/docker-tools#how-to-add-a-tool

flycheck.el
```diff
@@ -10668,6 +10668,22 @@ When non-nil, pass the directory via the `--workdir' option."
...
+
+The value of this variable is either a string denoting an ieee library
+standard, or nil, to use the default standard.  When non-nil,
+pass the ieee library standard via the `--ieee' option."
+  :type '(choice (const :tag "Default standard" nil)
+                 (const :tag "No IEEE Library" "none")
+                 (const :tag "IEEE standard" "standard")
+                 (const :tag "Synopsys standard" "synopsys")
+                 (const :tag "Mentor standard" "mentor")
+                 )
```
fmdkdd: Please wrap this lonely paren with the line above :)

fmdkdd:
The CI is failing due to unrelated Go changes.  I'll fix that on master; you'll need to rebase once that is done.

wastevensv:
Rebased onto master and added those changes. Should I squash all these commits into one?

fmdkdd:
@wastevensv Yes please, do squash the commits.

cpitclaudel:
:+1: :tada:

fmdkdd:
Thanks @wastevensv !

PR #1548, opened 2019-03-06
change 'go tool vet' to 'go vet'

### LINKED ISSUE #1523: go-vet is broken because "go vet tool" was deprecated

vincekd:
[fixes #1523]

Update command/docs/checks to replace `go tool vet` with `go vet`.

Removed deprecated command options/flycheck variables

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7769,31 +7769,16 @@ See URL `https://golang.org/cmd/go/' and URL
   :command ("go" "vet"
             (option "-printf.funcs=" flycheck-go-vet-print-functions concat
                     flycheck-option-comma-separated-list)
-            ;; (option-flag "-shadow" flycheck-go-vet-shadow) ;; DEPRECATED
-            ;; (option-list "-tags=" flycheck-go-build-tags concat) ;; DEPRECATED
-            ;; (eval (when (eq flycheck-go-vet-shadow 'strict) "-shadowstrict")) ;; DEPRECATED
             source)
   :error-patterns
   ((warning line-start (file-name) ":" line ": " (message) line-end))
   :modes go-mode
-  ;; We must explicitly check whether the "vet" command is available
-  :predicate (lambda ()
-               (let ((go (flycheck-checker-executable 'go-vet)))
-                 (ignore-errors (process-lines go "help" "vet"))))
   :next-checkers (go-build
                   go-test
                   ;; Fall back if `go build' or `go test' can be used
                   go-errcheck
                   go-unconvert
-                  go-megacheck)
-  :verify (lambda (_)
-            (let* ((go (flycheck-checker-executable 'go-vet))
-                    (have-vet (ignore-errors (process-lines go "help" "vet"))))
-              (list
-               (flycheck-verification-result-new
-                :label "go vet"
-                :message (if have-vet "present" "missing")
-                :face (if have-vet 'success '(bold error)))))))
+                  go-megacheck))
 
 (flycheck-def-option-var flycheck-go-build-install-deps nil (go-build go-test)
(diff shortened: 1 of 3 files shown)


flycheck.el
```diff
@@ -7774,24 +7762,24 @@ enabled. "
...
-            (eval (when (eq flycheck-go-vet-shadow 'strict) "-shadowstrict"))
+            ;; (option-flag "-shadow" flycheck-go-vet-shadow) ;; DEPRECATED
+            ;; (option-list "-tags=" flycheck-go-build-tags concat) ;; DEPRECATED
+            ;; (eval (when (eq flycheck-go-vet-shadow 'strict) "-shadowstrict")) ;; DEPRECATED
             source)
   :error-patterns
   ((warning line-start (file-name) ":" line ": " (message) line-end))
   :modes go-mode
-  ;; We must explicitly check whether the "vet" tool is available
+  ;; We must explicitly check whether the "vet" command is available
```
fmdkdd: I thought `go vet` was always present as long as `go` was installed?

flycheck.el
```diff
@@ -7774,24 +7762,24 @@ enabled. "
...
 `https://golang.org/cmd/vet/'."
-  :command ("go" "tool" "vet" "-all"
-            (option "-printfuncs=" flycheck-go-vet-print-functions concat
+  :command ("go" "vet"
+            (option "-printf.funcs=" flycheck-go-vet-print-functions concat
                     flycheck-option-comma-separated-list)
-            (option-flag "-shadow" flycheck-go-vet-shadow)
-            (option-list "-tags=" flycheck-go-build-tags concat)
-            (eval (when (eq flycheck-go-vet-shadow 'strict) "-shadowstrict"))
+            ;; (option-flag "-shadow" flycheck-go-vet-shadow) ;; DEPRECATED
```
fmdkdd: You can remove these outright, no need to to leave dead code in comments.

flycheck.el
```diff
@@ -7748,18 +7748,6 @@ take an io.Writer as their first argument, like Fprintf,
                  (string :tag "function"))
   :safe #'flycheck-string-list-p)
 
-(flycheck-def-option-var flycheck-go-vet-shadow nil go-vet
```
fmdkdd: You'll need to remove the option from the documentation in `doc/languages.rst`.

flycheck.el
vincekd: Unless your go package is broken, I guess. 

Should I remove it or just leave it in as a check that go is install properly?

fmdkdd: We already checks the `:command` part exists, you can remove the `:predicate` and `:verify` which are redundant then.

fmdkdd:
Good, just one minor change in the tests in order to pass the CI now.  See [here](https://travis-ci.org/flycheck/flycheck/jobs/502686558#L1435), there's a stray `IDENT` in the output that should be removed from [there](https://github.com/flycheck/flycheck/blob/master/test/flycheck-test.el#L3332).

andyafter:
Just asking, when will this be merged? Needing this desparately

fmdkdd:
LGTM.  Thank you @vincekd!

PR #1553, opened 2019-03-18
Add HiDPI support

### LINKED ISSUE #1552: flycheck-fringe-bitmap-double-arrow is tiny on HiDPI screen
Hi all—thanks for the great package!

I've recently run into a minor (but annoying) issue: the error/warning/info indicators in the right fringe are tiny on my display (see screenshot, depicting a normal 80-char line of text).  I am sure that this is due to the fact that my screen is a HighDPI screen; indeed, it appears that this is a [long-standing issue](https://www.reddit.com/r/emacs/comments/51wm7d/emacs_on_hidpi_displays/) with HDPI screens and flycheck.

That same link also describes how I can fix the issue, but I wanted to mention it here since it sounds like it's impacted others and it will only become more of an issue as HDPI screens become more common.

codesections:
This PR checks whether the current screen is a HiDPI screen.  If it is, the code doubles the size (in pixels) of the `flycheck-fringe-bitmap-double-arrow` bitmap (which results in the
normal visual size for the icon).

If the screen isn't a HiDPI screen, this code prints the normal `>>` icon.

If merged, this resolves #1552

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3556,7 +3556,10 @@ The following PROPERTIES constitute an error level:
         (plist-get properties :compilation-level))
   (setf (get level 'flycheck-overlay-category)
         (plist-get properties :overlay-category))
-  (setf (get level 'flycheck-fringe-bitmap-double-arrow)
+  (setf (get level
+        (if (flycheck-high-dpi-screen)
+             'flycheck-fringe-bitmap-double-arrow-hdpi
+             'flycheck-fringe-bitmap-double-arrow))
         (plist-get properties :fringe-bitmap))
   (setf (get level 'flycheck-fringe-face)
         (plist-get properties :fringe-face))
@@ -3630,8 +3633,7 @@ show the icon."
       (if (> (* (/ (float resx) sizex) 25.4) 150) t nil))))
 
 (when (fboundp 'define-fringe-bitmap)
-  (if (flycheck-high-dpi-screen)
-    (define-fringe-bitmap 'flycheck-fringe-bitmap-double-arrow
+    (define-fringe-bitmap 'flycheck-fringe-bitmap-double-arrow-hdpi
         (vector #b0000000000000000
                 #b0000000000000000
                 #b0000000000000000
@@ -3681,7 +3683,7 @@ show the icon."
                 #b00000000
                 #b00000000
                 #b00000000
-                #b00000000))))
+                #b00000000)))
 
 (setf (get 'flycheck-error-overlay 'face) 'flycheck-error)
 (setf (get 'flycheck-error-overlay 'priority) 110)
--- a/flycheck.el
+++ b/flycheck.el
@@ -3677,7 +3677,7 @@ show the icon."
                 #b00000000
                 #b00000000
                 #b00000000
-                #b00000000)))
+                #b00000000))))
 
 (setf (get 'flycheck-error-overlay 'face) 'flycheck-error)
 (setf (get 'flycheck-error-overlay 'priority) 110)
--- a/flycheck.el
+++ b/flycheck.el
@@ -3621,13 +3621,7 @@ show the icon."
          (res (cdr (assoc 'geometry attrs)))
          (resx (- (caddr res) (car res)))
          dpi)
-    (catch 'exit
-      ;; in terminal
-      (unless sizex
-        (throw 'exit 10))
-      ;; on big screen
-      (when (> sizex 1000)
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -3613,25 +3613,71 @@ show the icon."
...
+  "Return t if the user has a HiDPI screen"
+  (let* ((attrs (car (display-monitor-attributes-list)))
+         (size (assoc 'mm-size attrs))
+         (sizex (cadr size))
+         (res (cdr (assoc 'geometry attrs)))
+         (resx (- (caddr res) (car res)))
+         dpi)
+    (catch nil
+      ;; if dpi > 150 (standard dpi: 96; hdpi: 192)
+      (if (> (* (/ (float resx) sizex) 25.4) 150) t nil))))
```
cpitclaudel: This looks good! But I think it could be simplified by just using `x-display-mm-width` and `x-display-pixel-width`.

codesections: I agree that using `(x-display-name)` and `(x-display-mm-width)` would be simpler.  However, the documentation for those function notes


```
On "multi-monitor" setups this refers to the width in millimeters for
all physical monitors associated with TERMINAL.  To get information
for each physical monitor, use ‘display-monitor-attributes-list’.
```

So, to support multi-monitor setups, I think it would be better to stick with `(display-monitor-attributes-list)`.  I added a comment to this effect.

codesections:
I updated the PR in response to your comments above. 

> it would be best (I think) to do the check when we set the fringe bitmap, rather than when we create it.

I *think* my revised version does this but, as I mentioned in the initial issue, I am *very* new to elisp so I'd appreciate a second pair of eyes on this.

flycheck.el
cpitclaudel: Thanks for pointing this out.  I emailed emacs-devel to ask about this.

flycheck.el
```diff
@@ -3556,7 +3556,10 @@ The following PROPERTIES constitute an error level:
         (plist-get properties :compilation-level))
   (setf (get level 'flycheck-overlay-category)
         (plist-get properties :overlay-category))
-  (setf (get level 'flycheck-fringe-bitmap-double-arrow)
+  (setf (get level
+        (if (flycheck-high-dpi-screen)
+             'flycheck-fringe-bitmap-double-arrow-hdpi
+             'flycheck-fringe-bitmap-double-arrow))
```
cpitclaudel: This looks good, but for performance reasons we should not call the high-dpi-screen function for each overlay.  Also, the function should likely be called `flycheck--screen-high-dpi-p`, since it's a private (`--`) predicate (`-p`).

flycheck.el
```diff
@@ -3614,7 +3617,11 @@ show the icon."
 
 ;;; Built-in error levels
 (defun flycheck-high-dpi-screen ()
-  "Return t if the user has a HiDPI screen"
+  "Return t if the user has a HiDPI screen."
```
cpitclaudel: Instead of `Return t if …`, ELispers typically write things like `Check whether …`)

flycheck.el
cpitclaudel: To clarify, what this code does is attach to the symbol stored in `level` a property named `flycheck-fringe-bitmap-double-arrow`, with value `'flycheck-fringe-bitmap-double-arrow` (easy Lisp symbol has a property list, or plist, essentially a dictionary of static atttributes).  The idea is that users can redefine the bitmap for a given level by changing the value mapped to the key `flycheck-fringe-bitmap-double-arrow` in the level's plist (and the default value happens to be the bitmap called `'flycheck-fringe-bitmap-double-arrow`, and yes it's kind of confusing that they are named the same).

The best would probably be to attach *two* properties, once for the lowdpi marker and one for hidpi one.  Then we can pick when constructing each overlay.

cpitclaudel: @fmdkdd opinions?

fmdkdd: Yes, adding a new property sounds the right thing to do.  Fringes don't allow anything else other than bitmaps, do they?

neuhalje:
I would really love to see this implemented - any progress here?

cpitclaudel:
@neuhalje not much; would you like to refresh the PR?

neuhalje:
@cpitclaudel sorry, not right now :-(

cpitclaudel:
Does anyone want to finish the implementation of this PR?

cpitclaudel:
Done in #1744.

PR #1560, opened 2019-04-04
Add profile to rebar3 for erlang

dgud:
To be able to compile test files profile needs to be set to "test",
with the "default" profile test files are not compiled.

Test files are found in the application "test" sub-directory.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7749,9 +7749,34 @@ Return the absolute path to the directory"
   "Return directory where rebar.config is located."
   (locate-rebar3-project-root buffer-file-name))
 
+(flycheck-def-option-var flycheck-erlang-rebar3-profile nil erlang-rebar3
+  "The rebar3 profile to use.
+
+The profile used when compiling, if VALUE is nil \"test\" will be used
+when the file is located in test directory, otherwise \"default\" will be
+used as profile."
+  :type 'string
+  :safe #'stringp
+  :package-version '(flycheck . "32"))
+
+(defun flycheck-erlang-rebar3-get-profile ()
+  "Return rebar3 profile.
+
+Use flycheck-erlang-rebar3-profile if set, otherwise use test profile if
+dirname is test or else default."
+  (cond
+   (flycheck-erlang-rebar3-profile flycheck-erlang-rebar3-profile)
+   ((and buffer-file-name
+         (string= "test"
+                  (file-name-base
+                   (directory-file-name
+                    (file-name-directory buffer-file-name)))))
+    "test")
+   (t "default")))
+
 (flycheck-define-checker erlang-rebar3
   "An Erlang syntax checker using the rebar3 build tool."
-  :command ("rebar3" "compile")
+  :command ("rebar3" "as" (eval (flycheck-erlang-rebar3-get-profile)) "compile")
   :error-parser
   (lambda (output checker buffer)
     ;; rebar3 outputs ANSI terminal colors, which don't match up with
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,6 +54,8 @@
   - Add ``flycheck-ghdl-ieee-library`` to select which standard IEEE
     library to use for ghdl. [GH-1547]
   - ``javascript-eslint`` checker now supports ``typescript-mode`` by default.
+  - Add ``flycheck-erlang-rebar3-profile`` to select which profile to
+    use when compiling erlang with rebar3.
 
 - Improvements
 
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -396,6 +396,12 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       Check Erlang with the `rebar3 <https://www.rebar3.org/>`_ build tool.
 
+      .. defcustom:: flycheck-erlang-rebar3-profile
+
+         The profile to use when compiling, e.g. "default" or "test".
+         The default value is nil which will use test profile
+         in test directories and default profile otherwise.
+
 .. supported-language:: ERuby
 
    .. syntax-checker:: eruby-erubis


dgud:
Or is there a better solution? @joedevivo @tsloughter

dgud:
And I need some help with the travis errors, which I don't understand.

tsloughter:
So this uses the test profile only if the file being modified is in a test directory?

fmdkdd:
> So this uses the test profile only if the file being modified is in a test directory?

Only if the file being modified is in a directory named "test" rather.  My question is then: does Erlang/rebar forces to name your test directory "test", or are you free to name it something else?

tsloughter:
rebar3 defaults to looking in "test" and Erlang convention is "test".

I don't know that I've seen a different directory used before. The only thing is this won't technically catch all modification of test code because some put unit tests directly in the module being tested with macros:

```erlang
-ifdef(TEST).
... unit tests...
-endif.
```

`TEST` is set by the `test` profile. So if the user is modifying a unit test like this it would still compile in the default profile. But I don't think this should stop the change.

fmdkdd:
> TEST is set by the test profile. So if the user is modifying a unit test like this it would still compile in the default profile.

And there's no way to run the `test` profile alongside the `default` one?  Or maybe the `test` profile is a superset of the `default` one?

dgud:
I do think the only difference for rebar (when profile is test) is to define TEST macro and compile files (including files in test dir) but Tristan knows this better, I'm just guessing here.

fmdkdd:
Why I'm asking is that the usual way we would provide the functionality of this PR would be through a new variable `flycheck-rebar3-profile`, which would be `test` or `default`.  Then we would usually let a Flycheck extension (another package, which defines a function that would be added to `flycheck-mode-hook`) set the value of this variable depending on the current buffer (here, using the buffer filename).

The advantage of using a variable instead of the `(eval (rebar3-profile))` is that a variable can be set manually, through buffer- or file-locals, and can be overridden easily. 

The downside is that you would need to write and publish a separate extension, just for a very simple function, which seems a bit silly.

If you feel that there is no valid use case for customizing the behavior of the `rebar3-profile`, then we don't need a variable and can go through with your implementation.

dgud:
Hmm, thinking about what you wrote, why can't we always run with the 'test' profile?
Could there be a problem with that @tsloughter ?

```
:command ("rebar3" "as" "test" "compile")
```

tsloughter:
That is what I thought this PR was at first. I'm not sure it is a good default. Maybe most people will want it that way, either way if it is default or not it should be easily configurable.

fmdkdd:
> either way if it is default or not it should be easily configurable.

Then in that case we can more simply have a configurable variable for the profile to use, with `default` as the default value.  No need to check the directory the file resides in.

tsloughter:
Eh, I like the use of test profile if you are in the test directory. It should be configurable for those who want to always use the test profile, for instance if they include tests in modules directly.

dgud:
Yeah I can try, give me a couple of days, I've been sick so it will take a while until I caught up with things.

fmdkdd:
@dgud No pressure; please take time to recover.  This PR isn't going anywhere ;)

dgud:
Some questions, are there more profiles we should support or are they user defined 'prod'/'native' @tsloughter ?

What should package-version be?

And please review my lisp code and comments I'm not fluent in either language :-)

fmdkdd:
> What should package-version be?

32.  You can look at the top of the file: "32-cvs" means we are currently working towards version 32.

flycheck.el
```diff
@@ -7647,9 +7647,35 @@ Return the absolute path to the directory"
...
+          (const :tag "Use test rebar3 profile" test))
+  :safe (lambda (value) (or (not value) (memq value '(default test))))
+  :package-version '(flycheck . "0.24"))
+
+(defun rebar3-profile ()
+  "Return rebar3 profile.
+
+Use flycheck-erlang-rebar3-profile if set, otherwise use test profile if
+dirname is test or else default."
+  (cond
```
fmdkdd: I don't think this does what you want.  `cond` is a like a `switch`.  Here you are saying: "Test the value of `(flycheck-erlang-rebar3-profile  (symbol-name flycheck-erlang-rebar3-profile))` (which will error, since `flycheck-erlang-rebar3-profile` is not a function), then if `(string= "test" ...` is true (which is unrelated to the value of the option), use `"test"`, if not, use `"default"`".

Did you test that code?

You could use `cond`, but there is an even simpler idiom.  Here is what you want:
```elisp
(or flycheck-erlang-rebar3-profile
    (and (string= "test" (file-name-base (directory-file-name (file-name-directory buffer-file-name))))
         "test")
    "default")
```
`or` and `and` are short-cicuiting and return the first (resp. last) non-nil value.  If `flycheck-erlang-rebar3-profile` is non-nil, then `or` returns it.  If it is nil, then we try to auto-detect.  If the `string=` succeeds (non-nil), then we return `"test"`.  If not, then `and` fails and we fallback to `"default"` (which is non-nil, and returned by `or`).

Also, `buffer-file-name` can be nil (some buffers have no file), so I would add:
```elisp
(and buffer-file-name
     (string= "test" (file-name-base (directory-file-name (file-name-directory buffer-file-name))))
     "test")
```

Please double-check that code by testing it on your setup.

dgud: I did test the code but everything work with "test" profile so that might fail to show errors in my code :-)

I expected the cond clause to be (CONDITION BODY) i.e. flycheck-erlang-rebar3-profile to be interpreted as the CONDITION there, i.e. not nil will evaluate the BODY.

Also I want strings to be returned from the function and flycheck-erlang-rebar3-profile is set to a  symbol? (atom in my world) will that work with your 'or' alternative is there some auto-conversion in lisp?

dgud: Hmm I tested some more and from what I can see it works as I expect it to do.

fmdkdd: > I expected the cond clause to be (CONDITION BODY) i.e. flycheck-erlang-rebar3-profile to be interpreted as the CONDITION there, i.e. not nil will evaluate the BODY.

You are right.   I don't know what I was thinking there.  The `cond` does what you want, and my suggestion doesn't (it's missing `symbol-name`).  I apologize for the brain fart.

> will that work with your 'or' alternative is there some auto-conversion in lisp?

There's no implicit conversion, so no it wouldn't work as is.  On the other hand we could also use strings instead of symbols for the option, but symbols are fine.

fmdkdd: Ah wait, do we want the user to be able to provide an arbitrary profile beyond `default` and `test` or not?

flycheck.el
```diff
@@ -7647,9 +7647,36 @@ Return the absolute path to the directory"
...
+   (flycheck-erlang-rebar3-profile
+    (symbol-name flycheck-erlang-rebar3-profile))
+   ((and buffer-file-name
+         (string= "test" (file-name-base (directory-file-name
+                                          (file-name-directory buffer-file-name)))))
```
fmdkdd: You want to line feed after `"test"` and reindent in order to pass the `make check` test.

fmdkdd:
Thanks for the changes.  You'll need to document the new option in the manual as well to pass the CI.  Add a `defcustom` line in `doc/languages.rst` (see other checkers around for examples).

flycheck.el
dgud: phwee I was mightily embarrassed there for not testing enough, I hate when I get untested PR's :-)
Anyway thanks for your patience, I write lisp every 5 years or so, so stumble around.

dgud:
> Ah wait, do we want the user to be able to provide an arbitrary profile beyond default and test or not?

That was the question to @tsloughter  ?

flycheck.el
fmdkdd: Ahh sorry for the embarrassment :(

fmdkdd:
> That was the question to @tsloughter ?

Yes indeed.  If we do want more profiles then we may want to use strings instead for the option, and allow for an arbitrary `(string` case.  If we only care about `test` and `default`, then we can stick to symbols.

tsloughter:
@fmdkdd that could be useful, yea, good idea.

dgud:
Changed to arbitrary strings.

fmdkdd:
LGTM.  Please add a line to `CHANGES.rst` introducing the new option.

dgud:
Added line to Changes, squashed commits and rebased to master.

fmdkdd:
Nice work!  Thank you. :clap:

Issue #748, opened 2015-09-17
No obvious way to re-enable a checker after the error count transiently exceeds error limit


ngoldbaum:
Often when editing a python file, a buffer will transiently exceed `flycheck-checker-error-threshold`. According to the resolution of #476, `flycheck-checker-error-threshold` was added to prevent syntax checking on buffers that produce thousands of errors.

Fair enough, but sometimes I will *transiently* introduce thousands of errors by e.g. typing a function definiton. Before I've finished typing, I trigger hundreds of syntax errors in the file, and flycheck disables the checker I'm using using until I restart emacs.  I see that I could remove the checker from `flycheck-disabled-checkers` as well, but that isn't clear from the error message, which looks like:

```
Syntax checker in buffer plot_window.py in python-mode:

  python-flake8
    - major mode: `python-mode' supported
    - predicate:  t
    - executable: Found at /usr/local/bin/flake8

Flycheck cannot use this syntax checker for this buffer.
```

Maybe this could be resolved by changing the message to explain that the reason why flycheck can't use the checker I'm looking for is that the checker is in `flycheck-disabled-checkers`. Maybe if I explicitly try switching the checker back to the one I want by using `M-x flycheck-select-checker`, flycheck could try running the checker to see if the number of errors still exceeds `flycheck-checker-error-threshold`.

cpitclaudel:
Indeed, we can make it easier to re-enable flycheck in these cases. Re-enabling flycheck automatically is a lot harder, as simply parsing the list of errors is a costly process if it is long. Thus we can't just keep running the checker until we get a smaller list.

@lunaryorn: maybe one thing that we could do would be to stop processing the process output if it gets too big? In effect this would put a threshold not only on the number of errors, but also on the size of the error output; since the output is generated and received asynchronously, that threshold would not be very costly to enforce, and we could cancel the check without disabling the checker if the output size went over the size threshold.

lunaryorn:
I tend to blame `flake8`.  As far as I know it's the only syntax checker that regularly hits this limit _transiently_, and I would not like to add workarounds for the issues of just one single syntax checker.

At most, I'd try to disable syntax checkers less aggressively, i.e. by averaging the number of errors over the last three or so checks, and only disable a checker if this average is higher than the limit.  That'd prevent this specific problem.

[...]

PR #1562, opened 2019-04-11
Distinguish automatically disabled checkers

### LINKED ISSUE #1542: State inconsistency in flycheck when switching environments with pyvenv
**OS**: Linux Mint 19
**Emacs version**: GNU Emacs 25.2.2 (x86_64-pc-linux-gnu, GTK+ Version 3.22.21) of 2017-09-22, modified by Debian
**Flycheck version**: 32snapshot (package: 20190108.351)

pyvenv and flycheck aren't playing nicely together. Here's what's happening:

1. Open python buffer
2. `M-x flycheck-verify-setup`
3. Observe:
   ```
     python-flake8 (disabled)
       - may enable:         Automatically disabled!
       - executable:         Found at /home/clemons/anaconda3/bin/python
       - configuration file: Found at "/home/clemons/Coding/test/.flake8rc"
       - `flake8' module:    Missing
   ```
4. `M-x pyvenv-workon`
5. Select conda environment that has flake8 module
6. `M-x flycheck-verify-setup`
7. Observe:
   ```
     python-flake8 (disabled)
       - may enable:         yes
       - executable:         Found at /home/clemons/anaconda3/envs/myenv/bin/python
       - configuration file: Found at "/home/clemons/Coding/test/.flake8rc"
       - `flake8' module:    Found at "/home/clemons/anaconda3/envs/myenv/lib/python3.7/site-packages/flake8/__init__.py"
   ```

Notice in step 7 that python-flake8 is disabled even though the flake8 module has been discovered in the `myenv` conda environment. And as expected, the value of flycheck-disabled-checkers is (python-flake8). So, it appears that flycheck is not reenabling the python-flake8 checker after changing python environments.

Workarounds:
- Run `M-x revert-buffer` in the python buffer after step 5
- Run steps 4 and 5 before step 1
- Write a hook (`pyvenv-post-activate-hooks`?) to reverify the checkers after pyvenv activates a new environment
  - Is there a more generic way to accomplish this without requiring flycheck to know about pyvenv?

Other thoughts:
- this may be related to #1416

fmdkdd:
This commit does the following in order to resolve #1542:

1. Add a variable `-automatically-disabled-checkers` which distinguish
between manually disabled checkers and automatically disabled ones.
This allows us to reset the automatically disabled status without
overwriting user preference from `-disabled-checkers`.

2. Add a function `-reset-enabled-checker` to reset the enabled
or automatically disabled status of a checker.  This allows us to retest
the `:enabled` predicate for a checker.  This function is also useful
for external plugins that may have knowledge that circumstances that led
to a failed `:enabled` test have changed, and a new check must be
done (see #1542).

3. Call `-reset-enabled-checker` in `-verify-checker` so that manually
triggering a verification checks the enabled state again.

Closes #1542.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -313,6 +313,17 @@ respectively."
   :safe #'flycheck-symbol-list-p)
 (make-variable-buffer-local 'flycheck-disabled-checkers)
 
+(defvar-local flycheck--automatically-disabled-checkers nil
+  "List of syntax checkers automatically disabled for this buffer.
+
+A checker can be automatically disabled in two cases:
+
+1. Its `:enabled' predicate returned false.
+2. It returned too many errors (see `flycheck-checker-error-threshold').
+
+To trigger a reverification from Emacs Lisp code, do not modify
+this variable: use `flycheck-reset-enabled-checker'.")
+
 (defvar-local flycheck-checker nil
   "Syntax checker to use for the current buffer.
 
@@ -1097,6 +1108,7 @@ Only has effect when variable `global-flycheck-mode' is non-nil."
       flycheck-mode]
      "---"
      ["Describe syntax checker" flycheck-describe-checker t]
+     ["Verify setup" flycheck-verify-setup t]
      ["Show Flycheck version" flycheck-version t]
      ["Read the Flycheck manual" flycheck-info t]))
   "Menu of command `flycheck-mode'.")
@@ -1563,7 +1575,16 @@ A checker is registered if it is contained in
 
 A checker is disabled if it is contained in
 `flycheck-disabled-checkers'."
-  (memq checker flycheck-disabled-checkers))
+  (or (memq checker flycheck-disabled-checkers)
+      (flycheck-automatically-disabled-checker-p checker)))
+
+(defun flycheck-automatically-disabled-checker-p (checker)
+  "Determine whether CHECKER has been automatically disabled.
+
+A checker has been automatically disabled if it is contained in
+`flycheck--automatically-disabled-checkers'."
+  (memq checker flycheck--automatically-disabled-checkers))
+
 
 
 ;;; Generic syntax checkers
@@ -1746,10 +1767,10 @@ are mandatory.
      current buffer.  Otherwise it shall return nil.
 
      If FUNCTION returns a non-nil value the checker is put in a
-     whitelist in `flycheck-enabled-checkers' to prevent further
-     invocations of `:enabled'.  Otherwise it is disabled via
-     `flycheck-disabled-checkers' to prevent any further use of
-     it.
+     whitelist in `flycheck--automatically-enabled-checkers' to
+     prevent further invocations of `:enabled'.  Otherwise it is
+     disabled via `flycheck--automatically-disabled-checkers' to
+     prevent any further use of it.
 
      If this checker has a `:working-directory' FUNCTION is
      called with `default-directory' bound to the checker's
@@ -1934,11 +1955,14 @@ Return non-nil if CHECKER supports MODE and nil otherwise."
   (let ((mode (or mode major-mode)))
     (memq mode (flycheck-checker-get checker 'modes))))
 
-(defvar-local flycheck-enabled-checkers nil
+(defvar flycheck--automatically-enabled-checkers nil
   "Syntax checkers included in automatic selection.
 
 A list of Flycheck syntax checkers included in automatic
-selection for current buffer.")
+selection for the current buffer.")
+(define-obsolete-variable-alias 'flycheck-enabled-checkers
+  'flycheck--automatically-enabled-checkers "32")
+(make-variable-buffer-local 'flycheck--automatically-enabled-checkers)
 
 (defun flycheck-may-enable-checker (checker)
   "Whether a generic CHECKER may be enabled for current buffer.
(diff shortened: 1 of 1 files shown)


fmdkdd:
Sending for reviewing @cpitclaudel.  If we agree on the specifics, I will add tests before merging.

flycheck.el
```diff
@@ -312,6 +312,22 @@ respectively."
...
+
+A checker can be automatically disabled in two cases:
+
+1. Its `:enabled' predicate returned false.
+2. It returned too many errors (see `flycheck-checker-error-threshold').
+
+The command `flycheck-verify-setup' will clear this variable and
+reverify all checkers for the current buffer.  Checkers that fail
+the `:enabled` test will remain automatically disabled.  Checkers
+that pass will be enabled.
```
cpitclaudel: I think this is a bit dangerous: don't we fear that we'll have heisenbugs were something doesn't work, we ask for the contents of `verify`, and and the bugs go away?

flycheck.el
```diff
@@ -312,6 +312,22 @@ respectively."
   :safe #'flycheck-symbol-list-p)
 (make-variable-buffer-local 'flycheck-disabled-checkers)
 
+(defvar-local flycheck-automatically-disabled-checkers nil
```
cpitclaudel: `flycheck--` instead of `flycheck-`, maybe?

flycheck.el
```diff
@@ -1556,7 +1572,16 @@ A checker is registered if it is contained in
...
-  (memq checker flycheck-disabled-checkers))
+  (or (memq checker flycheck-disabled-checkers)
+      (memq checker flycheck-automatically-disabled-checkers)))
+
+(defun flycheck-automatically-disabled-checker-p (checker)
+  "Determine whether CHECKER has been automatically disabled.
+
+A checker has been automatically disabled if it is contained in
+`flycheck-automatically-disabled-checkers'."
+  (memq checker flycheck-automatically-disabled-checkers))
```
cpitclaudel: Do you want to use that function in flycheck-disabled-checker-p?

flycheck.el
```diff
@@ -1945,9 +1970,26 @@ nil otherwise."
...
+(defun flycheck-reset-enabled-checker (checker)
+  "Reset the `:enabled' test of CHECKER.
+
+Forget that CHECKER has been enabled or automatically disabled
+from a previous `:enabled' test.  Once a checker has been enabled
+or automatically disabled, `flycheck-may-enable-checker' will
+always be constant (t or nil respectively).
+
+If you wish to test the `:enabled' predicate again, you must
+first reset its state using this function."
```
cpitclaudel: I think this description is a bit too technical.  Maybe just something along the lines of `clear Flycheck's cache of :enabled properties`?

EDIT: Oh, except I see that this removes the value from `flycheck-disabled-checkers`, too.  But see below

flycheck.el
```diff
@@ -1945,9 +1970,26 @@ nil otherwise."
...
+  (when (memq checker flycheck-automatically-disabled-checkers)
+    (setq flycheck-automatically-disabled-checkers
+          (remq checker flycheck-automatically-disabled-checkers)))
+  (when (memq checker flycheck-enabled-checkers)
+    (setq flycheck-enabled-checkers
+          (remq checker flycheck-enabled-checkers))))
```
cpitclaudel: I think we don't want that, since we call this function from `flycheck-verify-checker`

flycheck.el
```diff
@@ -2114,7 +2156,7 @@ Return a list of `flycheck-verification-result' objects."
         (enabled (flycheck-checker-get checker 'enabled))
         (verify (flycheck-checker-get checker 'verify)))
     (when enabled
-      (let ((result (funcall enabled)))
+      (let ((result (flycheck-may-enable-checker checker)))
```
cpitclaudel: Can you explain this part? (IIUC, this will return the same result, since we'll have reset the cache in the meantime?)

flycheck.el
```diff
@@ -1945,9 +1970,26 @@ nil otherwise."
                                 (funcall enabled)))))
     (if shall-enable
         (cl-pushnew checker flycheck-enabled-checkers)
```
cpitclaudel: Don't we want to rename this to `flycheck-automatically-enabled-checkers`, for symmetry? (We'll need an alias, since https://github.com/search?q=flycheck-enabled-checkers&type=Code some people seem to use that variable?!)

flycheck.el
fmdkdd: That is risk indeed.  This convenience may in fact mystify users.  Better to have the user trigger the reverification (using buttons like in #1478).

Besides, the reverification can delay opening the verify-setup buffer, as in js-mode where the enabled test for `eslint` takes its sweet time.

I'll remove this auto-reset and we'll deal with the interaction bit in #1478.

flycheck.el
fmdkdd: I forget we are using this convention now.  :+1:

flycheck.el
fmdkdd: Good catch.

flycheck.el
fmdkdd: That's the point though: if a checker has previously passed the `:enabled` test, it has been added to `flycheck-enabled-checkers` and won't be tested again unless we remove it from that list.

Resetting the `:enabled` state thus means removing the checker from these two lists.

flycheck.el
fmdkdd: `-verify-checker` ended up calling the `:enabled` predicate twice: once through `verify-generic-checker`, and once through `may-use-checker`.

Because `-may-enable` tests `:enabled` *and* caches its result, using `-may-enable` here ensures we run the predicate only once.

The `(funcall enabled)` above was obviously bypassing the cached result.

(Note that `-verify-checker` is seldom used though)

flycheck.el
fmdkdd: Yes, I agree that it makes sense for the symmetry.  I didn't do it precisely because I didn't want to break the API.

Maybe we add an alias and deprecate the former one?

flycheck.el
cpitclaudel: Oh, you're completely right. I read `disabled` instead of `enabled` here.

flycheck.el
cpitclaudel: An alias sounds good.

fmdkdd:
So apparently `defvaralias` can't handle local variables:
```elisp
(define-obsolete-variable-alias 'flycheck-enabled-checkers
  'flycheck-automatically-enabled-checkers "32")

-> (error "Don’t know how to make a localized variable an alias")
```
But, strangely, we can define the alias before making it buffer-local, and it seems to work:
```elisp
(defvar test-local-var 1)
(define-obsolete-variable-alias 'obsolete-var 'test-local-var)
(make-variable-buffer-local 'test-local-var)
(setq obsolete-var 2)
test-local-var -> 2
In other buffer:
test-local-var -> 1
```
Shall we go ahead with the workaround?

cpitclaudel:
Yes, a quick look at the C sources suggests that this should be fine.

fmdkdd:
I've added the alias, renamed occurences of the obsolete variable, and fixed a bug with `flycheck-disable-checker` that was unable to enable checkers that were automatically disabled.

fmdkdd:
Oh and I've also extended the error message for excessive checkers, as suggested in #748.  And added `verify-setup` to the modemap (and menubar) menu, to help discovery.

fmdkdd:
Didn't see #1580, let's go with you error message instead, it's clearer.

PR #1575, opened 2019-05-16
Update extensions.rst

michaelschiff:
adds bazel syntax checker

### DIFF
--- a/doc/community/extensions.rst
+++ b/doc/community/extensions.rst
@@ -49,6 +49,10 @@ directly to the languages that interest you:
 .. contents:: Languages
    :local:
 
+Bazel
+-----
+* :gh:`michaelschiff/flycheck-bazel` adds a syntax checker for files build with bazel (https://bazel.build/)
+
 Cadence
 -------
 
--- a/doc/community/extensions.rst
+++ b/doc/community/extensions.rst
@@ -51,7 +51,7 @@ directly to the languages that interest you:
 
 Bazel
 -----
-* :gh:`michaelschiff/flycheck-bazel` adds a syntax checker for files build with bazel (https://bazel.build/)
+* :gh:`michaelschiff/flycheck-bazel` adds a syntax checker for files built with bazel (https://bazel.build/)
 
 Cadence
 -------


doc/community/extensions.rst
```diff
@@ -49,6 +49,10 @@ directly to the languages that interest you:
 .. contents:: Languages
    :local:
 
+Bazel
+-----
+* :gh:`michaelschiff/flycheck-bazel` adds a syntax checker for files build with bazel (https://bazel.build/)
```
cpitclaudel: `s/build/built/`?

michaelschiff: good catch - fixed

cpitclaudel:
Great. Can you submit this into MELPA, too? This way users can install it very simply.  You may want to change the capitalization (`flycheck` → `Flycheck`)

michaelschiff:
Sure, ill adjust the capitalization.  I am planning on submitting this to MELPA, but currently it relies on a shell script that must be somewhere on the users path.  I didn't see a way to support that with distribution via MELPA.  Is there a better way to deal with that script?

cpitclaudel:
Yes: you can use an absolute path; here's how to retrieve your script's directory:

```
(defconst script-folder
  (file-name-directory (or load-file-name
                           (bound-and-true-p byte-compile-current-file)
                           (buffer-file-name))))
```

Then you can `expand-file-name` relative to that directory.
Also, your shell script is vulnerable to maliciously-crafted file names; whether you want to defend against this is up to you.

michaelschiff:
sorry for the delay.  Let me try the absolute path stuff.  I will submit my checker to MELPA once I get that working.

michaelschiff:
@cpitclaudel Sorry in advance if this is a dumb question - It seems that the arguments in the :command list are not being expanded.  Currently my command is defined as:

```
    :command ((eval (expand-file-name "bazel-flycheck-command" script-folder)) source-original)
```
where script-folder is the const defined according to your suggestion.  This errors in flycheck:
```
error: Command executable for syntax checker bazel must be a string: (eval (expand-file-name "bazel-flycheck-command" script-folder))
```
Looking at the code in flycheck, we take the first element of the command list and assert `stringp`.
```
    (unless (stringp (car command))
      (error "Command executable for syntax checker %s must be a string: %S"
             symbol (car command)))
```
My expectation was that this would be on the result of evaluating `(eval (expand-file-name "bazel-flycheck-command" script-folder))`, not the string of this sexp.  Thanks for the help

cpitclaudel:
No worries. There are multiple ways to go: 

- Use flycheck-define-command-checker instead of flycheck-define-checker
- Change the value of the flycheck-bazel-executable

michaelschiff:
@cpitclaudel Hey, I tried using `flycheck-define-command-checker` and get the following error:
```
Symbol's value as variable is void: bazel
```
Which actually kind of makes sense given the code, the symbol `bazel` has not been assigned any value.  However, thats no different from when I was calling `flycheck-define-checker` so this is a little confusing for me.  Also, I was reading the code for `flycheck-define-command-checker` and see that `flycheck-define-checker` is actually just a convenience method.  Given that, I'm not totally clear on how it fixes my original problem.

I may have misunderstood your suggestion regarding absolute path - my understanding was:
originally the first element of my command list was `"bazel-flycheck-command"`, a script that was expected to be put on the path by the user.  I understood your suggestion to mean that I should define a constant variable with value of the directory in which my script lives, and then expand-file-name relative to this, which would return a string `"/some/absolute/path/to/bazel-flycheck-command"`, which would replace the first element in my command list.  When I tried this, the error was that the first element was not a string, but instead was the sexp that I expected to return the full path.

So, I guess my questions would be:
* If I setup the same variables in ielm, the expression I pass as the first element of the command list does evaluate to a string, why does it seem like this expression is not evaluated when passed to flycheck
* How would calling `flycheck-define-command-checker` change this?

Thanks again

cpitclaudel:
> If I setup the same variables in ielm, the expression I pass as the first element of the command list does evaluate to a string, why does it seem like this expression is not evaluated when passed to flycheck

That's because `flycheck-define-checker` is a macro, so its arguments are not evaluated

> How would calling flycheck-define-command-checker change this?

flycheck-define-checker is a function, so its arguments are evaluated.

Can you show the attempt that gave the error about `bazel`?

michaelschiff:
> flycheck-define-checker is a macro, so its arguments are not evaluated
Ah! this makes much more sense.

This is what I have now - it is almost the same as before, but I have removed the additional `eval` since this was a misguided attempt at getting the macro arguments to be evaluated.
```
(defconst script-folder
  (file-name-directory (or load-file-name
                           (bound-and-true-p byte-compile-current-file)
                           (buffer-file-name))))
(flycheck-define-command-checker bazel
    "Bazel flychecker"
    :command ((expand-file-name "bazel-flycheck-command" script-folder) source-original)

    :error-patterns
    ((error line-start
            (one-or-more not-newline) "/" (file-name) ":" line
            ": error: "
            (message (one-or-more not-newline) "\n"
                     (one-or-more not-newline) "\n"
                     (one-or-more not-newline) "\n"
                     (one-or-more not-newline) "\n"
                     (one-or-more not-newline)) line-end))
    :modes java-mode scala-mode python-mode)
```

cpitclaudel:
Does the following work?

```elisp
(flycheck-define-command-checker 'bazel
  "Bazel flychecker"
  :command
  `(,(expand-file-name "bazel-flycheck-command" script-folder) source-original)
  :error-patterns
  '((error line-start
           (one-or-more not-newline) "/" (file-name) ":" line
           ": error: "
           (message (one-or-more not-newline) "\n"
                    (one-or-more not-newline) "\n"
                    (one-or-more not-newline) "\n"
                    (one-or-more not-newline) "\n"
                    (one-or-more not-newline)) line-end))
  :modes '(java-mode scala-mode python-mode))
```

fmdkdd:
@cpitclaudel It shouldn't, since #1440 hasn't been merged.

cpitclaudel:
Hmm? Why not?

fmdkdd:
Oh you were using `-define-command-checker`, my bad.  Ignore the comment then.

cpitclaudel:
@michaelschiff any updates?

michaelschiff:
@cpitclaudel apologies for falling off - I have not had time to experiment with the checker in the last few weeks.  I will post here as soon as I am able to try the suggestions above

cpitclaudel:
Gentle ping

michaelschiff:
eek I'm sorry - I've made 0 progress on this.  I can close the PR, or if another author wants to take over that would make sense.

cpitclaudel:
Lets' close it for now, and it can be refreshed and reopened in the future :)

PR #1576, opened 2019-05-17
Add support for eqc to flycheck-erlang-rebar3-profile

raghavkarol:
If the file is in the eqc directory use the eqc profile

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7838,6 +7838,12 @@ dirname is test or else default."
                    (directory-file-name
                     (file-name-directory buffer-file-name)))))
     "test")
+   ((and buffer-file-name
+         (string= "eqc"
+                  (file-name-base
+                   (directory-file-name
+                    (file-name-directory buffer-file-name)))))
+    "eqc")
    (t "default")))
 
 (flycheck-define-checker erlang-rebar3
--- a/flycheck.el
+++ b/flycheck.el
@@ -7831,19 +7831,12 @@ used as profile."
 Use flycheck-erlang-rebar3-profile if set, otherwise use test profile if
 dirname is test or else default."
   (cond
-   (flycheck-erlang-rebar3-profile flycheck-erlang-rebar3-profile)
-   ((and buffer-file-name
-         (string= "test"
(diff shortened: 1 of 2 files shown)


fmdkdd:
@dgud What do you think?

dgud:
eqc is for quickcheck ?

I don't know rebar3 enough, but I think that is handeld by a plugin to rebar3.
We don't use quickcheck in the erlang/OTP team (comercial product) so I can't say anything about directory name conventions for that.

I can't see that it does any harm, besides making flycheck code larger.

@raghavkarol what does the eqc profile do?

raghavkarol:
Sorry about the delayed response. 

Eqc is short for [QuickCheck](http://www.quviq.com/products/) which exists in both commercial and free versions. IIRC, there is also a provision to use full versions for eqc for open source projects. 

This PR is a clone of #1560, such that we can compile files in the`eqc` using the same rules as files in the `test` directory.

flycheck.el
```diff
@@ -7772,6 +7772,12 @@ dirname is test or else default."
                    (directory-file-name
                     (file-name-directory buffer-file-name)))))
     "test")
+   ((and buffer-file-name
```
fmdkdd: Instead of duplicating the second condition, we merge conditions 2 and 3 as:
```elisp
((seq-contains '("test" "eqc")
                 (and buffer-file-name
                      (file-name-base
                       (directory-file-name
                        (file-name-directory buffer-file-name)))))
```
Taking advantage of `cond` returning the value of `CONDITION` when `BODY` is empty.  We can also simplify the first condition as:
```elisp
(flycheck-erlang-rebar3-profile)
```
along the same lines.

raghavkarol: @fmdkdd - thanks for the suggestions, here you go

fmdkdd:
Thank you.  The docstring still needs to change though.

I've restarted the failing flaky test.

fmdkdd:
CI is acting up.  Trying again...

cpitclaudel:
Closing in favor of #1685

PR #1579, opened 2019-06-06
Add option to avoid showing other files' errors

ibizaman:
Since showing errors from other files was added recently, I'm adding an option to revert back to not showing them.

Note I tried to look at my changes in the documentation but the `make html-auto` step failed with the same error that appeared recently(?) https://travis-ci.org/flycheck/flycheck/jobs/531994058

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -56,8 +56,8 @@
   - ``javascript-eslint`` checker now supports ``typescript-mode`` by default.
   - Add ``flycheck-erlang-rebar3-profile`` to select which profile to
     use when compiling erlang with rebar3.
-  - Add ``flycheck-relevant-error-other-file-show`` to avoid showing other
-    file's error.
+  - Add ``flycheck-relevant-error-other-file-show`` to avoid showing errors
+    from other files.
 
 - Improvements
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -56,6 +56,8 @@
   - ``javascript-eslint`` checker now supports ``typescript-mode`` by default.
   - Add ``flycheck-erlang-rebar3-profile`` to select which profile to
     use when compiling erlang with rebar3.
+  - Add ``flycheck-relevant-error-other-file-show`` to avoid showing other
+    file's error.
 
 - Improvements
 
--- a/flycheck.el
+++ b/flycheck.el
@@ -707,6 +707,13 @@ display all errors from other files."
   :safe #'flycheck-error-level-p
   :package-version '(flycheck . "32"))
 
+(defcustom flycheck-relevant-error-other-file-show t
+  "Whether to show errors from other files."
+  :group 'flycheck
+  :type 'boolean
+  :package-version '(flycheck . "32")
+  :safe #'booleanp)
+
 (defcustom flycheck-completing-read-function #'completing-read
   "Function to read from minibuffer with completion.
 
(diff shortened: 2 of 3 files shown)


fmdkdd:
Thanks for the patch!  LGTM.

The Travis failure is surprising indeed.  I've found the culprit: the file 
http://mirror.keystealth.org/gnu/texinfo/htmlxref.cnf has a type on line 175:
```
 eww		node	$(EMACS}/html_node/eww/
```
The `$(EMACS}` should be `${EMACS}`.  This makes the python string template panic.

@cpitclaudel Do you know who to contact to fix this?

CHANGES.rst
```diff
@@ -56,6 +56,8 @@
   - ``javascript-eslint`` checker now supports ``typescript-mode`` by default.
   - Add ``flycheck-erlang-rebar3-profile`` to select which profile to
     use when compiling erlang with rebar3.
+  - Add ``flycheck-relevant-error-other-file-show`` to avoid showing other
+    file's error.
```
fmdkdd: other file's error -> errors from other files

flycheck.el
```diff
@@ -707,6 +707,13 @@ display all errors from other files."
   :safe #'flycheck-error-level-p
   :package-version '(flycheck . "32"))
 
+(defcustom flycheck-relevant-error-other-file-show t
+  "Whether to show errors from other files."
+  :group 'flycheck
+  :type 'boolean
+  :package-version '(flycheck . "32")
+  :safe #'booleanp)
```
cpitclaudel: I wonder if a `-p` suffix would be appropriate, since this is a boolean flag.  Additionally, the name is a bit of a mouthful, but maybe that's justified by the consistency with `-minimum-level`?

ibizaman: Another option could be to have add a new level to `flycheck-relevant-error-other-file-minimum-level` called `'never`.

I'm not totally sure it should have that `-p` suffix. I can definitely be wrong but I thought a predicate was only a function or here it's variable, right?

cpitclaudel: > Another option could be to have add a new level to flycheck-relevant-error-other-file-minimum-level called 'never. Having only one option could avoid confusion.

That could work, but it will conflict if someone defines an error level called never.  That would be weird, of course.

> Btw, I'm not totally sure it should have that -p suffix. I can definitely be wrong but I thought a predicate was only a function while here it's variable, right?

No, I think you're right, actually :)

ibizaman: And what about setting `flycheck-relevant-error-other-file-minimum-level` to nil means never show errors from other files? I think that would make sense.

Otherwise I'm totally up to renaming `flycheck-relevant-error-other-file-show` to something else. I think we should keep the `flycheck-relevant-error-other-file` prefix, to make it obvious it's "paired" with `flycheck-relevant-error-other-file-minimum-level`. I'm not the best at naming things though.

cpitclaudel: > And what about setting flycheck-relevant-error-other-file-minimum-level to nil means never show errors from other files? I think that would make sense.

It could, but I think that value is already used to display all errors from other files.    Let's go with the original name.  I'll merge as soon as CI is back on track.

cpitclaudel:
@fmdkdd I think the issue should be fixed soon. I reported the bug to the texinfo folks, and it was fixed today. An updated version should be up soon.

ibizaman:
Thank you both for your review!

fmdkdd:
@ibizaman Thank you for the patch.  Always happy to have good contributions :clap:

PR #1581, opened 2019-06-10
Disambiguate reference to exec-path in -executable-find

cpitclaudel:
The latest master complains about `flycheck.el:470: Disambiguate exec-path by preceding w/ function,command,variable,option or symbol.`


cpitclaudel:
I had to add an extra trivial commit to fix the build.

PR #1584, opened 2019-06-17
More helpful "Not found" messages

dandavison:
- Change the "Not found" message for executable and configuration file so that it displays the path or file name that was used when attempting to find the resource.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5501,14 +5501,26 @@ CHECKER."
     `(
       ,(flycheck-verification-result-new
         :label "executable"
-        :message (if executable (format "Found at %s" executable) "Not found")
+        :message
+        (if executable (format "Found at %s" executable)
+          (format
+           "Not found. %s"
+           (let ((candidate (flycheck-checker-executable checker)))
+             (if candidate (format "(Tried %s)" candidate)
+               "(No executable file was specified for this checker.)"))))
         :face (if executable 'success '(bold error)))
       ,@(when config-file-var
           (let* ((value (symbol-value config-file-var))
                  (path (and value (flycheck-locate-config-file value checker))))
             (list (flycheck-verification-result-new
                    :label "configuration file"
-                   :message (if path (format "Found at %S" path) "Not found")
+                   :message
+                   (if path (format "Found at %S" path)
+                     (format
+                      "Not found. %s"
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -5501,14 +5501,26 @@ CHECKER."
...
             (list (flycheck-verification-result-new
                    :label "configuration file"
-                   :message (if path (format "Found at %S" path) "Not found")
+                   :message
+                   (if path (format "Found at %S" path)
+                     (format
+                      "Not found. %s"
+                      (if value (format "(Tried %s)" value)
+                        (concat "(No configuration file was specified "
+                                "for this checker.)"))))
```
cpitclaudel: I think this logic could be extracted, instead of being repeated in both places.

flycheck.el
```diff
@@ -5501,14 +5501,26 @@ CHECKER."
...
       ,(flycheck-verification-result-new
         :label "executable"
-        :message (if executable (format "Found at %s" executable) "Not found")
+        :message
+        (if executable (format "Found at %s" executable)
+          (format
+           "Not found. %s"
+           (let ((candidate (flycheck-checker-executable checker)))
+             (if candidate (format "(Tried %s)" candidate)
+               "(No executable file was specified for this checker.)"))))
```
cpitclaudel: I worry that this message will mislead users into thinking that they must specify an executable for Flycheck to work.  I'd be more comfortable with displaying that extra message only if an executable was already specified, but couldn't be found.

dandavison: If a user has not specified an executable, then flycheck will use the value of `(flycheck-checker-default-executable checker)`. So my understanding is that we do not expect a user to see this message; but if they do, they do indeed need to specify an executable because no default was provided.

cpitclaudel: That doesn't sound right; what about the case where the user just hasn't installed the tool yet?

dandavison: Thanks! The ball is in my court here; I will think think this through more carefully in light of your review.

cpitclaudel:
Gentle ping

cpitclaudel:
Closing this for inactivity, but I'd love to merge a change like this one.

dandavison:
Thanks @cpitclaudel, sorry about the inactivity on this, makes sense to keep closed until further activity.

PR #1588, opened 2019-06-23
Fix `flycheck-eslint-config-exists-p` for ESLint 6

dywedir:
See https://github.com/eslint/eslint/pull/11885

> Since v4.1.0, ESLint has supported glob-based configuration. So --print-config with a directory path doesn't make sense anymore. Configurations are not per directory.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8621,8 +8621,10 @@ for more information about the custom directories."
 (defun flycheck-eslint-config-exists-p ()
   "Whether there is a valid eslint config for the current buffer."
   (let* ((executable (flycheck-find-checker-executable 'javascript-eslint))
-         (exitcode (and executable (call-process executable nil nil nil
-                                                 "--print-config" "."))))
+         (exitcode (and executable
+                        (call-process executable nil nil nil
+                                      "--print-config" (or buffer-file-name
+                                                           "index.js")))))
     (eq exitcode 0)))
 
 (defun flycheck-parse-eslint (output checker buffer)


flycheck.el
```diff
@@ -8622,7 +8622,7 @@ for more information about the custom directories."
   "Whether there is a valid eslint config for the current buffer."
   (let* ((executable (flycheck-find-checker-executable 'javascript-eslint))
          (exitcode (and executable (call-process executable nil nil nil
-                                                 "--print-config" "."))))
+                                                 "--print-config" "./__placeholder__.js"))))
```
fmdkdd: Wait, what?  Passing a directory is an error, but passing a phony filename isn't?  What is the reasoning behind this change in ESLint?

thomasf: It probably matches against some pattern, it check which config is valid for that pattern to actual code. A directory is not code.

knu: I think you can and should use `buffer-file-name` if available.  What about `(or buffer-file-name "index.js")` instead?

fmdkdd: I see.  But then shouldn't we pass the filename of the current buffer instead?  Since it could return a different configuration?

dywedir: Fixed!

fmdkdd:
Hmm Travis is acting up.  Can you refill the code so it stays within 80 chars?  This may also trigger a build.

flycheck.el
```diff
@@ -8621,8 +8621,10 @@ for more information about the custom directories."
 (defun flycheck-eslint-config-exists-p ()
   "Whether there is a valid eslint config for the current buffer."
   (let* ((executable (flycheck-find-checker-executable 'javascript-eslint))
-         (exitcode (and executable (call-process executable nil nil nil
-                                                 "--print-config" "."))))
+         (exitcode (and executable
+                        (call-process executable nil nil nil
+                                      "--print-config" (or buffer-file-name
+                                                           "index.js")))))
```
cpitclaudel: Will passing index.js actually work? If we don't have a buffer-file-name, how will eslint know which file we're checking when we actually call it? (source-original will return an empty string)

DamienCassou: I agree with @cpitclaudel. I think checking for `buffer-file-name` is enough.

fmdkdd: So if you don't have a backing file, `print-config` will fail and the checker can't be used?

DamienCassou: In which case do you expect flycheck-eslint to be enabled but have no backing file to send to eslint itself?

fmdkdd: We are piping the buffer via stdin to eslint, so I think the checker currently works without a backing file (using the default configuration for the working directory).

Not sure if it's used by many people though.

cjk: I use it :) eslint has become pretty essential for JS-development for many.

cpitclaudel: @cjk really? We're not talking about a buffer not yet saved to disk; rather,  we're discussing a buffer that doesn't even have a buffer-file-name.  
Also, what happens if we casll eslint with a non-existing file?

DamienCassou: > Also, what happens if we call eslint with a non-existing file?

```session
$ eslint foo.js

ESLint: 6.0.1.

No files matching the pattern "foo.js" were found.
Please check for typing mistakes in the pattern.

$ echo $?
2

$ eslint --print-config foo.js
{
  "env": { ... }
}

$ echo $?
0
```

cpitclaudel: OK.  Let's just use index.js then? I think that's fine by me.

DamienCassou: > OK. Let's just use index.js then? I think that's fine by me.

I suggest to keep `(or buffer-file-name "index.js")` because there might be some filename-specific configuration options. For example, files matching `tests-*.js` can have a different set of eslint rules.

cpitclaudel: yes yes of course, that's what I meant; sorry for being unclear ^^

fmdkdd: All in agreement then.  Thank you for being thorough, everyone.

fmdkdd:
In the end Travis did not manage to see this PR for some reason.  Had to force the merge after I tested it out locally.

dywedir:
Thanks!

flycheck.el
cjk: @cpitclaudel You're right, not using it on unbound buffers. Sorry for the confusion.

Issue #1600, opened 2019-07-08
Flycheck won't enable eslint 6


DamienCassou:
To detect if flycheck-eslint should be enabled, the command `eslint --print-config .`  is executed. Eslint 6 has disabled passing a directory to `--print-config`: https://github.com/eslint/eslint/issues/11874. The solution is to pass a filename instead.

cpitclaudel:
I think we already have https://github.com/flycheck/flycheck/issues/1593 :)

PR #1601, opened 2019-07-08
Fix eslint checker for eslint >= 6.

DamienCassou:
Fix [#1600].


cpitclaudel:
Duplicate of https://github.com/flycheck/flycheck/pull/1588, I think?

DamienCassou:
argh :-(. You are right.

PR #1606, opened 2019-07-28
flycheck-same-files-p avoid file-truename which is slow

liulk:
This change improves responsiveness when the buffer has a lot of
warnings or errors.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1191,17 +1191,23 @@ to a number and return it.  Otherwise return nil."
   (and (listp obj) (seq-every-p #'symbolp obj)))
 
 (defun flycheck-same-files-p (file-a file-b)
-  "Determine whether FILE-A and FILE-B refer to the same file."
-  (let ((file-a (expand-file-name file-a))
-        (file-b (expand-file-name file-b)))
-    ;; We must resolve symbolic links here, since some syntax checker always
-    ;; output canonical file names with all symbolic links resolved.  However,
-    ;; we still do a simple path compassion first, to avoid the comparatively
-    ;; expensive file system call if possible.  See
-    ;; https://github.com/flycheck/flycheck/issues/561
-    (or (string= (directory-file-name file-a) (directory-file-name file-b))
-        (string= (directory-file-name (file-truename file-a))
-                 (directory-file-name (file-truename file-b))))))
+  "Determine whether FILE-A and FILE-B refer to the same file.
+
+Files are the same if (in the order checked):
+
+- They have the same path.  Or,
+- They have the same inode and filesystem numbers, not following symlink.  Or,
+- After chasing symlinks, they result in the same path.
+
+This should work even on w32, where Emacs uses file index to
+emulate inode (see fstat() in emacs/src/w32.c)."
+  (or (string= file-a file-b)
+      (let ((attrs-a (file-attributes file-a))
+            (attrs-b (file-attributes file-b)))
+        (and attrs-a attrs-b  ;; Make sure both file-a and file-b exist.
+             (equal (nth 10 attrs-a) (nth 10 attrs-b))    ;; inode
+             (equal (nth 11 attrs-a) (nth 11 attrs-b))))  ;; filesystem
+      (string= (file-chase-links file-a) (file-chase-links file-b))))
 
 (defvar-local flycheck-temporaries nil
   "Temporary files and directories created by Flycheck.")


liulk:
I see that the test `flycheck-same-files-p/across-symlinks` failed. And I just realized that file-attriutes does not follow symlinks.

How do you feel about this heuristic?

0. If two file paths are identical, return true. This is just a string comparison that avoids any system calls and should be the fastest.
1. If file-attributes compare equal, return true. This works on regular files and should be the vast majority of all cases.
2. Else try file-chase-links and lookup the resulting file-attributes. If equal, we're done.
3. Fallback to file-truename?

I think it's better not to call file-chase-links right away since I fear that it would be expensive. And in the spirit of avoiding expensive operations, we can do a naive string comparison first. What do you think?

liulk

fmdkdd:
I like the idea of doing the fast tests first.  Could there be false positives though?

Also, do we really need to fallback to `file-truename` after all this?

liulk:
If two file paths are identical, then their `file-truename` and `file-attributes` must be equal as well.

The only case of false positive I'm aware of with `file-attributes` is when files are deleted, and the inode (or file index on Windows) is subsequently reused. But if Emacs keeps the file open, then reuse can't happen. On FAT, [file index] could change if the defragmentation relocates the first cluster of the file.

[file index]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/ns-fileapi-by_handle_file_information#remarks

Another case where `file-attributes` and `file-truename` could disagree is if the operating system resolves symlinks differently than `file-truename` which implements its own algorithm. But I think in that case, the operating system should be right. 😃

liulk

liulk:
Some very encouraging profiling results comparing `file-truename`, `file-attributes`, and the heuristic.

https://gist.github.com/liulk/91228e4e61f7e2ba3fce1eeee4915b8c

The total CPU usage for `flycheck-handle-signal` went down from 86%, to 59%, to 18%. Out of those, the amount used by `flycheck-same-files-p` went down from 84%, to 47%, to 14%.

liulk

liulk:
Note that I changed the heuristic so step (2) just compares the paths after `file-chase-links`. There is no step (3) since I realized that `file-chase-links` and `file-truename` serve approximately the same purpose even though the results differ slightly. There is no need to do `file-attributes` after `file-chase-links`.

liulk

fmdkdd:
Very encouraging numbers indeed.  Thank you for this nice patch.

I've fixed the integration tests errors on `master`, if you rebase it should pass the CI.

liulk:
There were additional test failures in CI from make integ, but I don't know how to read the output.

Here are the failed tests:

```
   FAILED  flycheck-define-checker/chef-foodcritic/default
   FAILED  flycheck-define-checker/go-build/complete-chain
   FAILED  flycheck-define-checker/go-build/missing-package
   FAILED  flycheck-define-checker/go-staticcheck/default
   FAILED  flycheck-define-checker/go-test/default
```

I was trying to reproduce these test failures on my machine but `make init` said `Some dependencies were not available: mmm-mode`. When I opportunistically ran `make integ`, the failed tests above were skipped, and new failures emerged. However, both `make specs` and `make unit` succeeded.

Got any pointers on how to setup the local machine for running flycheck integ tests?

liulk

fmdkdd:
You want to use the docker image.  I've updated the manual to describe how:

http://www.flycheck.org/en/latest/contributor/contributing.html#running-all-the-integration-tests

Let me know if it needs to be clarified.

liulk:
All checks have passed after fixing the handling of file not found and the rubocop error message. PTAL. Thanks!

flycheck.el
```diff
@@ -1191,17 +1191,23 @@ to a number and return it.  Otherwise return nil."
...
+- They have the same inode and filesystem numbers, not following symlink.  Or,
+- After chasing symlinks, they result in the same path.
+
+This should work even on w32, where Emacs uses file index to
+emulate inode (see fstat() in emacs/src/w32.c)."
+  (or (string= file-a file-b)
+      (let ((attrs-a (file-attributes file-a))
+            (attrs-b (file-attributes file-b)))
+        (and attrs-a attrs-b  ;; Make sure both file-a and file-b exist.
+             (equal (nth 10 attrs-a) (nth 10 attrs-b))    ;; inode
```
fmdkdd: Wouldn't `eq` suffice here?

liulk: The inode and file system numbers are cons cells (see [file-attributes]), since they may overflow a single integer. If using `eq`, that compares referential equality of the cons which will never return `t`. That's why we're using `equal` to compare value equality.

```lisp
(eq '(1 . 2) '(1 . 2))
nil
(equal '(1 . 2) '(1 . 2))
t
```

[file-attributes]: https://www.gnu.org/software/emacs/manual/html_node/elisp/File-Attributes.html

fmdkdd: I wasn't aware that it could overflow to a cons cell.  Nothing left to add then!

PR #1614, opened 2019-08-21
Capture ‘note’ messages in python-mypy checker

wbolster:
This extends the python-mypy checker to also collect ‘note’ messages,
which look like this:

```
somefile.py:123: note:     foo: builtins.dict[builtins.str, Any]
```

The format is identical to the error and warning messages which are
already handled by Flycheck, so just add another error pattern using
Flycheck's ‘info’ severity level.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9539,7 +9539,9 @@ See URL `http://mypy-lang.org/'."
   ((error line-start (file-name) ":" line (optional ":" column)
           ": error:" (message) line-end)
    (warning line-start (file-name) ":" line (optional ":" column)
-            ": warning:" (message) line-end))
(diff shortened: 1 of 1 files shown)


wbolster:
@fmdkdd No they are not. As an example of stand-alone `note` messages, mypy has a [`reveal_type()`](https://mypy.readthedocs.io/en/latest/common_issues.html?highlight=reveal_type#displaying-the-type-of-an-expression) to dump diagnostics. Without my changes these are not visible inside Emacs, and with this PR applied, they will be.

fmdkdd:
Ok, too bad.

wbolster:
thanks, that was quick!

PR #1619, opened 2019-09-11
Detect erlang-rebar3 projects with dynamic config

ehamberg:
This fixes an issue where flycheck's `erlang-rebar3` checker isn't correctly detecting a project that uses a dynamic rebar config.

Erlang projects can use a `rebar.config.script` file as an alternative to
`rebar.config` in order to have dynamic configuration. The script file will then
be evaluated by rebar3 and the result used as configuration.

This adds `rebar.config.script` as an alternative to `rebar.config` when
detecting an Erlang rebar3 project.


fmdkdd:
Thanks for the patch!  And thanks @marsam for reviewing.

PR #1624, opened 2019-09-21
Fix systemd-analyze parsing and unit test

kisaragi-hiu:
Fix several issues with the systemd-analyze checker:

- The unit test is broken. https://github.com/systemd/systemd/commit/71b21730d49ca53f0ccf1afc039c5962bbb85662 (released in systemd 241) and https://github.com/systemd/systemd/commit/28f30f4051c3424b5f0757c76856a1102763e04f (released in systemd 243) changed the error message for an unknown key name.

  This PR makes the unit test use the right message depending on which version is on the system.

- It cannot parse the error messages. This is because 1. `systemd-analyze verify` sometimes (at least on my machine) outputs colors, and 2. the message format has changed.

  This PR removes color from its output explicitly, like the `erlang-rebar3` checker (I've put it into a separate function, `flycheck-parse-out-color`.), and adds a new error pattern. This overlaps with #1275, but that PR has been blocked for over 2 years now, and didn't keep the old pattern (which, according to https://github.com/flycheck/flycheck/pull/1275#commitcomment-22970010, should be kept).

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -10854,7 +10854,8 @@ See URL
   :command ("systemd-analyze" "verify" source)
   :error-parser flycheck-parse-with-patterns-without-color
   :error-patterns
-  ((error line-start "[" (file-name) ":" line "] " (message) line-end))
+  ((error line-start (file-name) ":" (optional line ":") (message) line-end)
+   (error line-start "[" (file-name) ":" line "]" (message) line-end))
   :modes (systemd-mode))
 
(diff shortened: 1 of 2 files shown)


kisaragi-hiu:
I rebased the new commits after addressing the 2 reviews --- would it have been better if I rebased after the review process?

> but what about #1625 ?

#1625 requires some way to pass an environment variable to the command; I'm not sure how that should be done.

fmdkdd:
Thanks!

> #1625 requires some way to pass an environment variable to the command; I'm not sure how that should be done.

We can leave it for now.

PR #1628, opened 2019-09-27
Add `flycheck-` prefix to rebar3 defuns

fmdkdd:
Somehow missed that when reviewing.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -7809,12 +7809,13 @@ See URL `http://www.erlang.org/'."
   :modes erlang-mode
   :enabled (lambda () (string-suffix-p ".erl" (buffer-file-name))))
 
-(defun contains-rebar-config (dir-name)
+(defun flycheck-contains-rebar-config (dir-name)
   "Return DIR-NAME if DIR-NAME/rebar.config exists, nil otherwise."
   (when (file-exists-p (expand-file-name "rebar.config" dir-name))
     dir-name))
 
-(defun locate-rebar3-project-root (file-name &optional prev-file-name acc)
+(defun flycheck-locate-rebar3-project-root (file-name &optional
+                                                      prev-file-name acc)
   "Find the top-most rebar project root for source FILE-NAME.
 
 A project root directory is any directory containing a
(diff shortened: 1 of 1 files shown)


cpitclaudel:
LGTM; just needs he conflict fixed

purcell:
Doh, sorry, I just spotted and fixed this independently, plus a couple of other unprefixed symbols.

fmdkdd:
@purcell Was this spotted using the updated `package-lint`?  Good catch.

purcell:
> Was this spotted using the updated `package-lint`? Good catch.

Nope, just with the previous version. The prefix checks have been around for a while now.

I'd actually really like to run `package-lint` on everything as part of building MELPA packages, but a key limitation is that it currently only understands single-file packages, so it can't scan secondary `.el` files with an understanding of the overall package name and dependencies.

It's a good thing to use in CI builds, IMO, and hopefully fairly easy to do.

fmdkdd:
If it had no false positives, we would gladly had it to the CI.  That would prevent errors from getting into MELPA, and save you the work of reporting them back to us ;)

purcell:
> If it had no false positives, we would gladly had it to the CI. That would prevent errors from getting into MELPA, and save you the work of reporting them back to us ;)

That should *mostly* be the case once the PR is merged which bumps the minimum Emacs version. The only two remaining notes would be those about `less-css-mode` and `scss-mode`, which can be eliminated by converting (for example) `:modes (scss-mode)` to the equally-valid `:modes scss-mode`, which `package-lint` doesn't take to be a function call.

purcell:
Ah, not quite:
```
6803:1: error: `flycheck-c/c++-quoted-include-directory' contains a non-standard separator `/', use hyphens instead (see Elisp Coding Conventions).
```
This might be fixable in `package-lint`, though I have to be careful because people love to spam `/` into their symbol names... :-)

purcell:
Hmm, looks like that's not easy to fix without missing real cases I want `package-lint` to whine about.

cpitclaudel:
How hard would it be to extend package-lint to have an explicit exclude list?

purcell:
It wouldn't be hard, I guess. I'm not averse to the idea.

cpitclaudel:
That might be the easiest and mot flexible solution, then.  I'd be happy to have package-lint running on every build :)

Issue #1610, opened 2019-08-05
`__path__` should not be a NameError in `__init__.py`


glyph:
I think the problem here is that pyflakes / flake8 are being invoked with a temporary path name that doesn't include "`__init__.py`", since flake8 definitely doesn't report this on its own https://github.com/PyCQA/pyflakes/issues/458

```
Syntax checkers for buffer __init__.py in python-mode:

First checker to run:

  python-pyflakes
    - may enable:    yes
    - executable:    Found at /Users/glyph/.local/bin/pyflakes
    - next checkers: python-mypy

Checkers that may run as part of the first checker's chain:

  python-mypy
    - may enable:         yes
    - predicate:          t
    - executable:         Found at /Users/glyph/.local/bin/mypy
    - configuration file: Not found

Checkers that are compatible with this mode, but will not run until properly configured:

  python-flake8 (disabled)
    - may enable:         Automatically disabled!
    - executable:         Found at /usr/local/bin/python3.6
    - configuration file: Not found
    - `flake8' module:    Found at "/Users/glyph/Library/Python/3.6/lib/python/site-packages/flake8/__init__.py"
    - next checkers:      python-mypy

  python-pylint (disabled)
    - may enable:         Automatically disabled!
    - executable:         Found at /Homebrew/bin/python
    - configuration file: Not found
    - `pylint' module:    Missing

  python-pycompile (disabled)
    - may enable: Automatically disabled!
    - executable: Found at /Homebrew/bin/python

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled checkers.

--------------------

Flycheck version: 32snapshot (package: 20190709.1443)
Emacs version:    26.2
System:           x86_64-apple-darwin18.5.0
Window system:    mac
```

glyph:
The mypy bug is here: https://github.com/python/mypy/issues/1422

cpitclaudel:
I think your analysis is correct: it's not a temporary path name, but instead standard input:

```
$ flake8 - < __init__.py 
stdin:1:7: F821 undefined name '__path__'
```

Is there a way to tell flake8 which file standard input corresponds to? mypy has an option like that, for example.

cpitclaudel:
Yep, it's `flake8 --stdin-display-name __init__.py - < __init__.py`, and that doesn't print errors

cpitclaudel:
Could you give the branch that I just pushed a try?

cpitclaudel:
This was fixed in #1645

PR #1630, opened 2019-10-03
Set stdin-display-name in flake8 (fixes GH-1610)

cpitclaudel:
The tests might need to be updated. Let's see what CI says.


fmdkdd:
Hold on I'll fix the unrelated test cases.

fmdkdd:
I've pushed a branch.   It seems handlebars is still failing though.  We may check with upstream.

cpitclaudel:
Duplicate of #1645, which was merged.

PR #1636, opened 2019-10-21
Inhibit substitution of command keys and quotes in help-echo strings

cpitclaudel:
This long-standing issue is described in https://lists.gnu.org/archive/html/emacs-devel/2019-10/msg00090.html ; the tl;dr is that Emacs changes `` ` `` and ``'`` into ``‘`` and ``’`` in error messages, which isn't quite right.


fmdkdd:
I've fixed the go-staticheck test on master.

cpitclaudel:
Thanks for the test fix and the quick review :)

Issue #1294, opened 2017-07-28
*Flycheck error messages* buffer should have its own major mode


rgrinberg:
It would be nice if this buffer had its own major mode to easily create bindings & hooks.

cpitclaudel:
Are you thinking of flycheck-error-list-mode?

rgrinberg:
Hmm, that's right. I wonder why flycheck error buffer is in Fundamental mode when I hover over an error and get that popup. Enabling the mode manually works of course.

cpitclaudel:
Ah, I misunderstood you then

cpitclaudel:
You didn't mean the error list, right? You meant the buffer that pops up for long error messages?

fmdkdd:
If you are talking about the buffer that opens for long error messages, you can define your own error function with `flycheck-display-errors-function`.  There you could activate whatever you want, then delegate the call to the original function?

rgrinberg:
Yes, I misunderstood @cpitclaudel original comment. I see how `flycheck-display-errors-function` can be used to implement what I want but IMO it's a bit inconvenient and roundabout. For example, I want to add a quit hotkey to the long error messages popup. Seems like I'd have to go through a bit of boilerplate to accomplish that.

cpitclaudel:
I wouldn't object to a patch that sets a custom mode in flycheck-display-error-messages

fmdkdd:
@rgrinberg Do you want to send a PR for that?

rgrinberg:
I'm not that competent at elisp but can give it a try.

cpitclaudel:
This was done in #1638.

cpitclaudel:
(Thanks again @rgrinberg)

PR #1638, opened 2019-10-25
Create custom mode for error messages

rgrinberg:
Error messages that are too long, are displayed in their own buffer. We
make sure that this buffer has its own mode. This makes it easy to
customize it with hooks and keybindings.

This addressed my issue in #1294

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4385,7 +4385,7 @@ message to stretch arbitrarily far."
 
 (defun flycheck-error-list-update-source ()
   "Update the source buffer of the error list."
-  (when (not (eq (current-buffer) (get-buffer flycheck-error-list-buffer)))
+  (unless (eq (current-buffer) (get-buffer flycheck-error-list-buffer))
     ;; We must not update the source buffer, if the current buffer is the error
     ;; list itself.
     (flycheck-error-list-set-source (current-buffer))))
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -4862,9 +4866,13 @@ In the latter case, show messages in the buffer denoted by
 variable `flycheck-error-message-buffer'."
   (when (and errors (flycheck-may-use-echo-area-p))
     (let ((messages (seq-map #'flycheck-error-format-message-and-id errors)))
-      (display-message-or-buffer (string-join messages "\n\n")
-                                 flycheck-error-message-buffer
-                                 'not-this-window))))
+      (let ((result (display-message-or-buffer (string-join messages "\n\n")
+                                               flycheck-error-message-buffer
+                                               'not-this-window)))
+        (if (window-live-p result)
```
cpitclaudel: This could be a `when`

flycheck.el
```diff
@@ -4862,9 +4866,13 @@ In the latter case, show messages in the buffer denoted by
...
+            (with-current-buffer (window-buffer result)
+              (when (not (eq major-mode 'flycheck-error-message-mode))
```
cpitclaudel: This could be an `unless`, and I think the convention is to use `derived-mode-p`

flycheck.el
```diff
@@ -4849,6 +4849,10 @@ The echo area may be used if the cursor is not in the echo area,
 and if the echo area is not occupied by minibuffer input."
   (not (or cursor-in-echo-area (active-minibuffer-window))))
 
+(define-derived-mode flycheck-error-message-mode fundamental-mode
+  "Flycheck error messages"
+  "Major mode for extended error messages")
```
cpitclaudel: Missing `.` at the end of the docstring.
Also, should the mode derive from text-mode?

rgrinberg:
> LGTM, with a few minor tweaks. Feel free to merge after that ^^

Thanks for the review. I addressed your comments, but I don't have merging rights. A maintainer will need to do that.

PR #1641, opened 2019-10-28
Add a regression test for #1305

### LINKED ISSUE #1324: Add a regression test for #1305
Since the bug fixed by #1305 is subtle, but can throw Emacs into a loop, we should add a test for it in `flycheck-test.el`.  There are already a couple of test for `idle-change` in there that can serve as a base.

We could also add a test to ensure that `flycheck-handle-idle-change` takes no arguments, as there are many people who rely on this behavior, as discussed in #1305.

fmdkdd:
Closes #1324.

### DIFF
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -1199,6 +1199,23 @@
       (sleep-for 1.1)
       (should (flycheck-deferred-check-p)))))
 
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Thanks!

PR #1642, opened 2019-10-30
Ignore navigation-minimum-level if no error of that level is present

### LINKED ISSUE #1333: Navigate to next error should ignore warnings only if errors are present
Currently, if a buffer has errors and warnings, the default behavior of `flycheck-next-error` is to jump to the next error or warning (or info).

Following #398, we added `flycheck-navigation-minimum-level` which ignores errors that are below the given severity.  However, setting this variable to `error` means we will *always* skip warnings, unless we change its value at a later point.

It seems to me the more sensible default would be to skip warnings *only* if there are errors present in the file.  Once there are only warnings left, then surely `flycheck-next-error` should jump to them.

fmdkdd:
Closes #1333.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4203,11 +4203,15 @@ overlays."
   (flycheck-error-level-interesting-p (get-char-property pos 'flycheck-error)))
 
 (defun flycheck-error-level-interesting-p (err)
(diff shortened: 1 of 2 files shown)


fmdkdd:
Good point, fixed.

PR #1646, opened 2019-11-08
Fix error-pattern for yaml-yamllint checker

Namburgesas:
Checker was configured to use stdin, but its error-pattern was configured to expected
file-name instead. Updated its error-pattern to expect `stdin:...`

Tested using emacs 26.3 on macos.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -11262,9 +11262,9 @@ See URL `https://github.com/adrienverge/yamllint'."
             (config-file "-c" flycheck-yamllintrc))
   :error-patterns
   ((error line-start
-          (file-name) ":" line ":" column ": [error] " (message) line-end)
+          "stdin:" line ":" column ": [error] " (message) line-end)
    (warning line-start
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Wait, how does yamllint even know what the filename is?

Namburgesas:
@cpitclaudel That's right, it doesn't know anything about the file name because the file contents are passed in via stdin. This was exactly what the issue was for me. Parsing wasn't happening correctly because yamllint was printing `stdin:<line>:<col>:...`. I changed the error parsing to start with `stdin:`, similar to how the `yaml-ruby` checker is configured.

Sorry, I'll update the PR description to be a bit clearer.

cpitclaudel:
Oh, ouch, I read the diff in reverse in my Emacs :'(

cpitclaudel:
Thanks a lot! Sorry for the initial misunderstanding  :+1:

Namburgesas:
No problem at all, thanks @cpitclaudel !

Issue #1647, opened 2019-11-13
Markdownlint-cli checker error parsing incorrect


bleggett:
## Checklist
- [X] I have checked existing issues for potential duplicates before creating this one.
- [X] I have read the [Troubleshooting guide][].

## Bug description
Get this error with the latest version of `markdown-cli`
![Screen Shot 2019-11-12 at 8 15 41 PM](https://user-images.githubusercontent.com/854255/68724418-6caa9b00-0589-11ea-9ff3-77bc211945fe.png)

## Steps to reproduce
1. `npm install -g markdown-cli`
2. Open a markdown file with a lint error
3. See error

## System configuration
```
Syntax checkers for buffer Contributing.md in markdown-mode:

First checker to run:

  markdown-markdownlint-cli
    - may enable:         yes
    - executable:         Found at /usr/local/bin/markdownlint
    - configuration file: Not found

Checkers that could run if selected:

  markdown-mdl  select
    - may enable:         yes
    - executable:         Found at /Users/blegget1/.rbenv/shims/mdl
    - configuration file: Not found

  proselint  select
    - may enable: yes
    - executable: Found at /usr/local/bin/proselint

Checkers that are compatible with this mode, but will not run until properly configured:

  textlint (disabled)
    - may enable:         Automatically disabled!
    - executable:         Not found
    - configuration file: Not found
    - textlint plugin:    @textlint/markdown

Flycheck Mode is enabled. Use SPC u C-c ! x to enable disabled checkers.

--------------------

Flycheck version: 32snapshot
Emacs version:    26.3
System:           x86_64-apple-darwin18.7.0
Window system:    mac

```

Emacs configuration:
- [ ] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [X] Doom Emacs
- [ ] Other shared configuration

## Additional notes
This is almost certainly the offending change: https://github.com/igorshubovych/markdownlint-cli/commit/17b717ad9db1e2213db51811652a63496f7c7880

[Troubleshooting guide]:  https://www.flycheck.org/en/latest/user/troubleshooting.html

cpitclaudel:
Fixed by https://github.com/flycheck/flycheck/pull/1648 (Github didn't pick up the magic words from the issue description). Thanks a lot!

seagle0128:
The issue still exists in `flycheck-20191126.2142`. Does it work for you now?

cpitclaudel:
[...]

PR #1648, opened 2019-11-13
Update `markdownlint-cli` error syntax

bleggett:
This fixes issue reported here: https://github.com/flycheck/flycheck/issues/1647

The breaking change in the upstream checker is here: https://github.com/igorshubovych/markdownlint-cli/commit/17b717ad9db1e2213db51811652a63496f7c7880

### DIFF
--- a/Cask.24
+++ b/Cask.24
@@ -26,7 +26,7 @@
  (depends-on "groovy-mode")
  (depends-on "haml-mode")
  (depends-on "handlebars-mode")
- (depends-on "haskell-mode")
+ ;; (depends-on "haskell-mode")    ; requires Emacs 25.1+
  (depends-on "js2-mode")
  (depends-on "js3-mode")
  (depends-on "rjsx-mode")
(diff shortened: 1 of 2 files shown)


cpitclaudel:
Thanks a lot for assembling this PR.  Do we have tests covering the output of markdown-cli? If so, these should be updated too.
I'll merge your change soon, though I'd like to take the time to get the rest of the CI fixed (it broke due to updates in various checkers, including, I think, markdown-cli ^^

bleggett:
@cpitclaudel There are a few, but it didn't look as if they were testing the the complete error string.

I might have misunderstood the test rig, I'll take a look and see if I can grok the elisp well enough to fix/add tests

fmdkdd:
Rebasing should pass the two remaining tests.

Thanks for catching this one!

fmdkdd:
@bleggett I've removed `haskell-mode` from the Cask.24 file.  Rebasing again should do it!

Issue #1644, opened 2019-11-06
flake8 --exclude and --per-file-ignores options are incompatible


felker:
## Bug description
I have been using a `setup.cfg` file at the root of my project directory successfully for a while now to specify the `ignore` option of `flake8`, which flycheck respects. 

In January of 2019, `flake8` 3.7.0 added the `--per-file-ignores` option. See discussion https://stackoverflow.com/questions/48153886/flake8-ignore-specific-warning-for-entire-file

I wanted to use this option with flycheck, but it does not appear to respect this setting, nor the older `--exclude` setting. 
 
## Steps to reproduce
Here is an example `setup.cfg`, with an aggressive `exclude` rule that should actually ignore all Python files:

```
[flake8]
max-line-length = 79
exclude = *.py
ignore =
       E121,
       E123,
       E126,
       E226,
       E241,
       E704,
       E731,
       W503
per-file-ignores =
       test.py:E402
```

Example `test.py`:
```python
import sys
```

Opening `test.py`,  the line is highlighted by Flycheck with:
> 'sys' imported but unused [F401]

`flycheck-compile` on the checker:
```
python -c import\ sys\;sys.path.pop\(0\)\;import\ runpy\;runpy.run_module\(\"flake8\"\) --format\=default --config /home/kfelker/plasma-python/setup.cfg - < \
/home/kfelker/plasma-python/test.py
stdin:1:1: F401 'sys' imported but unused
```

## Expected behavior
From the command line, `flake8 test.py` correctly returns no violations.

felker:
Related #721, but I tried the workaround suggested in https://github.com/flycheck/flycheck/issues/721#issuecomment-129556743 and it did not work.

fmdkdd:
Thanks for the report, can confirm.

The issue is that we pass the content of the buffer via stdin, so flake8 doesn't know the filename:

```
$ flake8 - < test.py
stdin:1:1: F401 'sys' imported but unused
```

Luckily, the flag `--stdin-display-name` solves the problem:

```
$ flake8 --stdin-display-name=test.py - < test.py
```

Fixed in #1645.

Sent

felker:
Great! It may be worth documenting that this fix probably won't be compatible with the `flake8` options when they are set to relative paths, as opposed to simple file names.

fmdkdd:
@felker Can you elaborate with an example?

felker:
[...]

Issue #721, opened 2015-08-10
flake8 checker ignores project-specific configuration


ngoldbaum:
I've been using the flake8 checker along with flycheck to lint my code on-the-fly for a while now. Recently I started a project that has some per-project flake8 configuration settings defined in a `setup.cfg` file at the root of the project repository.

When I run the `flake8` executable outside of emacs, the configuration settings in this file are respected. However, it seems flycheck's flake8 checker does not respect these settings. I guess I have two questions about how flycheck sets things up that might clear this up.

1. Does flycheck execute flake8 in the same directory as the source code? According to the flake8 documentation, flake8 will take settings from a per-project configuration file if it finds a `[flake8]` section in a `setup.cfg` file that lives in the current directory or any parent directory.

2. Does flycheck have a special way for me to specify per-project configuration outside of flake8's built-in capabilities?

Thanks for your help!

lunaryorn:
Flycheck runs `flake8 from the default directory of the buffer.  Normally that's the directory the file is contained in.

I'm not sure what you mean with "flake8's built in capabilities ".  I've never used `flake8` and used to prefer pylint when I was writing large amounts of Python code, so my knowledge is very superficial here.  Off the top of my head, I don't know how Flycheck invokes flake8 exactly, but you can take a look at the definition of the syntax checker to find out.

fbergroth:
I usually put a `.dir-locals.el` file at the project root containing:
```el
((python-mode . ((flycheck-flake8rc . "setup.cfg"))))
```

lunaryorn:
@ngoldbaum Does that solve your problem?

ngoldbaum:
Yes, @fbergroth's suggestion works for me.

init-js:
There's also a way to set a global option it in your ~/.emacs, from the editor. You can do `M-x customize-group RET flycheck-config-files`, and edit the "Flycheck Flake8rc" setting to `.flake8`.

The previous default was `.flake8rc` in my case.

The description of the setting is:

>   If set to a string, locate the configuration file using the
   functions from `flycheck-locate-config-file-functions'.  If the
   file is found pass it to the syntax checker as configuration
   file.
> If no configuration file is found, or if this variable is set to
   nil, invoke the syntax checker without a configuration file.
> Use this variable as file-local variable if you need a specific
   configuration file a buffer.

I suspect this would apply to all projects then.

PR #1651, opened 2019-11-20
Use location of flake8 configuration file as working directory

fmdkdd:
Follow-up to https://github.com/flycheck/flycheck/issues/1644#issuecomment-553134516

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9928,7 +9928,7 @@ Requires Flake8 3.0 or newer. See URL
   :command ("python3"
             (eval (flycheck-python-module-args 'python-flake8 "flake8"))
             "--format=default"
-            (config-file "--config" flycheck-flake8rc)
+            (config-file "--append-config" flycheck-flake8rc)
             (option "--max-complexity" flycheck-flake8-maximum-complexity nil
                     flycheck-option-int)
             (option "--max-line-length" flycheck-flake8-maximum-line-length nil
--- a/flycheck.el
+++ b/flycheck.el
@@ -9912,6 +9912,12 @@ Update the error level of ERR according to
       (setf (flycheck-error-level err) level)))
   err)
 
+(defun flycheck-flake8--find-project-root (_checker)
+  "Find setup.cfg in a parent directory of the current buffer."
+  ;; This is a workaround for `https://gitlab.com/pycqa/flake8/issues/517'; see
+  ;; also `https://github.com/flycheck/flycheck/issues/1722'
+  (locate-dominating-file (or buffer-file-name default-directory) "setup.cfg"))
+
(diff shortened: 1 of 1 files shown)


felker:
Thanks! This worked for me, but to be clear, the user does need to have a `.dir-locals.el` in the project root directory as suggested in https://github.com/flycheck/flycheck/issues/721#issuecomment-129556743 for this extension to work.

E.g. `M-x flycheck-compile RET python-flake8 RET` executed in the buffer for `__init__.py` in my MWE from https://github.com/flycheck/flycheck/issues/1644#issuecomment-553134516


With the `flycheck-flake8rc` variable set by `.dir-locals.el`:
```
python -c import\ sys\;sys.path.pop\(0\)\;import\ runpy\;runpy.run_module\(\"flake8\"\)\
 --format\=default --config /Users/felker/Downloads/flycheck-flake-mwe/setup.cfg --stdi\
n-display-name\=/Users/felker/Downloads/flycheck-flake-mwe/test/__init__.py - < /Users/felker/Downloads/flycheck-flake-mwe/test/__init__.py
```

Without:
```
python -c import\ sys\;sys.path.pop\(0\)\;import\ runpy\;runpy.run_module\(\"flake8\"\) --format\=default --st\
din-display-name\=/Users/felker/Downloads/flycheck-flake-mwe/test/__init__.py - < /Users/felker/Downloads/flycheck-flake-mwe/test/__init__.py
/Users/felker/Downloads/flycheck-flake-mwe/test/__init__.py:1:1: F401 'sys' imported but unused
```

fmdkdd:
Yes, you need to tell `flycheck` the name of your flake8 configuration file.

I guess we could also try the usual ones, but that may be out of scope.

cpitclaudel:
> I guess we could also try the usual ones, but that may be out of scope.

We could do that, and there are just three such names, but I think this is a flake8 bug.  THese paths are meaningless if they are not relative to the config file.  So I'd rather see a flake8 bug report before we start implementing workarounds

felker:
@cpitclaudel good point; just found the `flake8` bug report:

https://gitlab.com/pycqa/flake8/issues/517

cpitclaudel:
@felker Thanks, then I think we'll just wait for this to be fixed on their side.  Feel free to reopen this if that issue gets stale there and we need to work around it.

cpitclaudel:
Reopened following #1722

cpitclaudel:
See my comments in #1722 though.  Just looking for any flake8 config file isn't quite right.

cpitclaudel:
Confirmed to work in 1722, and I think this rebase should address @felker's use case too.

fmdkdd:
Sure, go ahead with this (I can't review my own PR).

PR #1652, opened 2019-11-21
Fix displaying flycheck error in own buffer

rgrinberg:
display-message-or-buffer does not always return the window with thep
popup. We need to handle this case.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4874,10 +4874,13 @@ variable `flycheck-error-message-buffer'."
       (let ((result (display-message-or-buffer (string-join messages "\n\n")
                                                flycheck-error-message-buffer
                                                'not-this-window)))
+        ;; We cannot rely on `display-message-or-buffer' returning the right
+        ;; window. See URL `https://github.com/flycheck/flycheck/issues/1643'.
         (when (window-live-p result)
-          (with-current-buffer (window-buffer result)
-            (unless (derived-mode-p 'flycheck-error-message-mode)
-              (flycheck-error-message-mode))))))))
+          (-when-let ((buffer (get-buffer flycheck-error-message-buffer)))
+            (with-current-buffer buffer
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -4875,9 +4875,15 @@ variable `flycheck-error-message-buffer'."
...
-              (flycheck-error-message-mode))))))))
+          (let* ((maybe-popup-buffer (window-buffer result))
+                 (popup-buffer
+                  (if (string= (buffer-name maybe-popup-buffer)
+                               flycheck-error-message-buffer)
+                      maybe-popup-buffer
+                    (get-buffer flycheck-error-message-buffer))))
+            (with-current-buffer popup-buffer
+              (unless (derived-mode-p 'flycheck-error-message-mode)
+                (flycheck-error-message-mode)))))))))
```
cpitclaudel: I think this is a bit convoluted.  Can we just (1) call `display-message-or-buffer` and (2) use `(get-buffer flycheck-error-message-buffer)` to get a handle to the message buffer, and change the mode there?

rgrinberg:
Okay, I simplified it to just use (get-buffer ..). we simply need some sort of a check that we returned a window to make sure this is all necessary however.

flycheck.el
cpitclaudel: I think you want `-when-let` here rather than `cl-letf*`

rgrinberg: oh yeah. done

cpitclaudel:
Merged. Thanks for promptly coming up with a fix.
Btw, question: do we even need to check if display-buffer-or-window gives us a window back? We could just call it, then look for the error buffer and put it in error mode, unconditionally, right?

rgrinberg:
I suppose the check is optional. Feel free to remove it.

Issue #1661, opened 2019-12-27
Added buplicate checkers as next-checker


conao3:
## Checklist
- [x] I have checked existing issues for potential duplicates before creating this one.
- [x] I have read the [Troubleshooting guide][].

## Bug description
Duplicate next-checker added.

## Steps to reproduce
1. Add `flycheck-package-setup` to `flycheck-mode-hook`.
2. Turn on/off `flycheck-mode` multi times
3. Duplicate next-checker added.

## Expected behavior
Duplicate next-checker not added.

## Screenshots
None.

## System configuration
```
Syntax checkers for buffer flycheck-indent.el in emacs-lisp-mode:

First checker to run:

  emacs-lisp
    - may enable:    yes
    - predicate:     t
    - executable:    Found at /home/conao/.local/bin/emacs
    - next checkers: emacs-lisp-checkdoc, emacs-lisp-package, emacs-lisp-package, emacs-lisp-package, indent-elisp, emacs-lisp-package, indent-elisp

Checkers that may run as part of the first checker's chain:

  emacs-lisp-checkdoc
    - may enable:    yes
    - executable:    Found at /home/conao/.local/bin/emacs
    - next checkers: emacs-lisp-package, emacs-lisp-package, emacs-lisp-package, emacs-lisp-package, indent-elisp

  emacs-lisp-package
    - may enable:    yes
    - next checkers: indent-elisp

  indent-elisp
    - may enable: yes
    - executable: Found at /home/conao/.local/bin/emacs

Checkers that are compatible with this mode, but will not run until properly configured:

  emacs-lisp-elsa
    - may enable: yes
    - predicate:  nil
    - executable: Found at /home/conao/.local/tools/cask/bin/cask

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled
checkers.

--------------------

Flycheck version: 32snapshot (package: 20191125.1821)
Emacs version:    26.3
System:           x86_64-pc-linux-gnu
Window system:    x
```

Emacs configuration:
- [x] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs
- [ ] Other shared configuration

## Additional notes
My customization for `flycheck-package` is below using [leaf](https://github.com/conao3/leaf.el)

``` emacs-lisp
(leaf flycheck-package
  :ensure t
  :hook ((flycheck-mode-hook . flycheck-package-setup)))
```

[...]

PR #1662, opened 2019-12-27
Add next-checker only when not added

conao3:
Add next-checker only when not added.
This PR solves #1661.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2280,12 +2280,19 @@ When WITH-SELECT is non-nil, add a button to select this checker."
         (princ "\n"))))
   (princ "\n"))
 
+(defun flycheck--get-next-checker-symbol (next)
+  "Get the checker symmbol of NEXT checker.
+
+NEXT should be either a cons (NEXT-CHECKER . LEVEL) or a
+symbol."
+  (if (consp next) (cdr next) next))
+
 (defun flycheck-get-next-checkers (checker)
   "Return the immediate next checkers of CHECKER.
 
 This is a list of checker symbols.  The error levels of the
 `:next-checker' property are ignored."
-  (mapcar (lambda (c) (if (consp c) (cdr c) c))
+  (mapcar #'flycheck--get-next-checker-symbol
           (flycheck-checker-get checker 'next-checkers)))
 
 (defun flycheck-all-next-checkers (checker)
@@ -2497,6 +2504,23 @@ modified, or nil otherwise."
 
 
 ;;; Extending generic checkers
+(defun flycheck-remove-next-checker (checker next)
+  "After CHECKER remove a NEXT checker.
+
+CHECKER is a syntax checker symbol, from which to remove NEXT
+checker.
+
+NEXT is a cons or a symbol, as documented in
+`flycheck-add-next-checker'."
+  (unless (flycheck-valid-checker-p checker)
+    (error "%s is not a valid syntax checker" checker))
+  (let* ((next-symbol (flycheck--get-next-checker-symbol next)))
+    (setf
+     (flycheck-checker-get checker 'next-checkers)
+     (seq-remove
+      (lambda (next) (eq (flycheck--get-next-checker-symbol next) next-symbol))
+      (flycheck-checker-get checker 'next-checkers)))))
+
 (defun flycheck-add-next-checker (checker next &optional append)
   "After CHECKER add a NEXT checker.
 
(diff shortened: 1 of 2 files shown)


fmdkdd:
Thanks for the patch.

Ensuring there are no duplicates in `next-checker` sounds good, but what if the checker is added with another error level.  I.e, we should prevent this situation:
```
(error . emacs-lisp-checkdoc)
(warning . emacs-lisp-checkdoc)
```
as it makes little sense.

conao3:
> as it makes little sense.

What is your intention...?
This situation is a very rare case and it makes little sense?

I cannot understand that an example makes sense. Does it make sure that warnings and errors reported by checkers are unified as warnings/errors?

fmdkdd:
Sorry for not being clearer: your patch checks equality of checker using `memq`, but `flycheck-add-next-checker` can take the `next` argument as the form `(LEVEL . CHECKER)`, which will always fail `memq`.  So adding `(t . emacs-lisp-checkdoc)` as a next-checker to `emacs-lisp` adds a duplicate, which is not wanted.  Conversely, checkers in the `next-checkers` property can also be of the form `(LEVEL . CHECKER)`, so you would need to account for this case as well.

conao3:
Fix to handle your mentioned situation.
Please tell me if you came up with the better local variable name.

cpitclaudel:
Sorry for letting this patch slip through.  I think the right way to handle this is to create a separate function `flycheck-remove-next-checker`, and to call that from `flycheck-add-next-checker`.  Something like this:

```diff
---
 flycheck.el | 35 ++++++++++++++++++++++++++++++-----
 1 file changed, 30 insertions(+), 5 deletions(-)

diff --git a/flycheck.el b/flycheck.el
index c038c3f5..0ded6dcb 100644
--- a/flycheck.el
+++ b/flycheck.el
@@ -2280,12 +2280,19 @@ flycheck--verify-princ-checker
         (princ "\n"))))
   (princ "\n"))
 
+(defun flycheck--get-next-checker-symbol (next)
+  "Get the checker symmbol of NEXT checker.
+
+NEXT should be either a cons (NEXT-CHECKER . LEVEL) or a
+symbol."
+  (if (consp next) (cdr next) next))
+
 (defun flycheck-get-next-checkers (checker)
   "Return the immediate next checkers of CHECKER.
 
 This is a list of checker symbols.  The error levels of the
 `:next-checker' property are ignored."
-  (mapcar (lambda (c) (if (consp c) (cdr c) c))
+  (mapcar #'flycheck--get-next-checker-symbol
           (flycheck-checker-get checker 'next-checkers)))
 
 (defun flycheck-all-next-checkers (checker)
@@ -2497,6 +2504,21 @@ flycheck-buffer-saved-p
 
 
 ;;; Extending generic checkers
+(defun flycheck-remove-next-checker (checker next-checker)
+  "After CHECKER remove a NEXT-CHECKER.
+
+CHECKER is a syntax checker symbol, from which to remove NEXT
+checker.
+
+NEXT-CHECKER is a symbol denoting the syntax checker to not run after
+CHECKER any more."
+  (unless (flycheck-valid-checker-p checker)
+    (error "%s is not a valid syntax checker" checker))
+  (setf (flycheck-checker-get checker 'next-checkers)
+        (seq-filter (lambda (next) (not (eq (flycheck--get-next-checker-symbol next)
+                                       next-checker)))
+                    (flycheck-checker-get checker 'next-checkers))))
+
 (defun flycheck-add-next-checker (checker next &optional append)
   "After CHECKER add a NEXT checker.
 
@@ -2517,10 +2539,13 @@ flycheck-add-next-checker
   (unless (flycheck-valid-checker-p checker)
     (error "%s is not a valid syntax checker" checker))
   (flycheck-validate-next-checker next 'strict)
-  (if append
-      (setf (flycheck-checker-get checker 'next-checkers)
-            (append (flycheck-checker-get checker 'next-checkers) (list next)))
-    (push next (flycheck-checker-get checker 'next-checkers))))
+  (let ((next-checkers
+         (flycheck-remove-next-checker
+          (flycheck--get-next-checker-symbol next)
+          (flycheck-checker-get checker 'next-checkers))))
+    (setf (flycheck-checker-get checker 'next-checkers)
+          (if append (append next-checkers (list next))
+            (cons next next-checkers)))))
 
 (defun flycheck-add-mode (checker mode)
   "To CHECKER add a new major MODE.
```

flycheck.el
```diff
@@ -2507,10 +2507,16 @@ APPEND is non-nil."
...
+         (exists-next-checkers (flycheck-checker-get checker 'next-checkers))
+         (exists-next-checkers-sym (mapcar get-symbol exists-next-checkers)))
+    (if (memq nextsym exists-next-checkers-sym)
+        exists-next-checkers
       (setf (flycheck-checker-get checker 'next-checkers)
-            (append (flycheck-checker-get checker 'next-checkers) (list next)))
-    (push next (flycheck-checker-get checker 'next-checkers))))
+            (if append
+                (append exists-next-checkers (list next))
+              (append (list next) exists-next-checkers))))))
```
cpitclaudel: Doesn't this code lose levels?

conao3:
It has been a long time since I submitted this patch and it is difficult to understand the context.
If your patch solves the problem, please commit the patch to `master`.

cpitclaudel:
OK, I will do that, with proper credit.

cpitclaudel:
I think this is ready.  I've added tests and preserved authorship on the commits.  @fmdkdd, would you like to have a look?

cpitclaudel:
Thanks a lot!

conao3:
Thanks also from me!

PR #1664, opened 2020-01-09
Add file extension support for PHP Codesniffer

Sasanidas:
By default flycheck-phpcs doesn't check correctly with .module and .inc files, adding this custom option fix the issue.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9138,6 +9138,12 @@ or as path to a standard specification."
                  (string :tag "Standard name or file"))
   :safe #'stringp)
 
+(flycheck-def-option-var flycheck-phpcs-file-extension nil php-phpcs
+  "Define file extensions to use with PHP CodeSniffer"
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Thanks! Do these extensions also cover files included from the current one, or is it just the current one? If the latter, would it be better to just pass the extension of the current file as `--extensions`?

Sasanidas:
> Thanks! Do these extensions also cover files included from the current one, or is it just the current one? If the latter, would it be better to just pass the extension of the current file as `--extensions`?

It's just cover the current file, some php frameworks already defined a list of these file extensions for PHP Codesniffer, and it's quite easy to get that list, Drupal for example:
`
phpcs --standard=Drupal --extensions=php,module,inc,install,test,profile,theme,css,info,txt,md,yml /path/to/drupal/example_module
`
But it can be interesting to pass only the current extension , I will try it. Thank you!

PR #1670, opened 2020-01-24
[feat] Support auto-formatting/fixing eslint buffer code

rpatterson:
If a user adds `--fix-dry-run` to `flycheck-eslint-args` or otherwise gets `output` to
be returned in the eslint JSON output, then the buffer being checked will be updated
with the fixed output from the JSON.

The one remaining issue I'm aware of is that if there are fixes in the output then
updating the buffer triggers another run of the checker immediately after.  Is there a
way temporarily disable change tracking while calling `replace-buffer-contents`?

It may also be valuable to abstract this so it can be re-used with other
auto-formatters/fixers.

I couldn't run `$ make check specs unit` as it seems to be dependent on Mac OS X.  I did
as much testing as I could.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8730,6 +8730,14 @@ CHECKER and BUFFER denote the CHECKER that returned OUTPUT and
 the BUFFER that was checked respectively.
 
 See URL `https://eslint.org' for more information about ESLint."
+  (with-temp-buffer
+    (let ((output (let-alist (caar (flycheck-parse-json output)) .output)))
+      (if output
+          (progn
+            (insert output)
+            (let ((output-buffer (current-buffer)))
+              (with-current-buffer buffer
+                (replace-buffer-contents output-buffer)))))))
   (mapcar (lambda (err)
             (let-alist err
               (flycheck-error-new-at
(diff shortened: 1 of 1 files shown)


rpatterson:
Oh, FYI, I tested on Ubuntu 19.10 eoan

cpitclaudel:
We've discussed error fixers in the past: https://github.com/flycheck/flycheck/issues/530 ; but we don't currently have a good integration.  I don't think we'd want to swap out the whole buffer, though.
I think your code might make more sense in a third-party package.  WDYT?

rpatterson:
I think that the way I've done it here is optional and does nothing if `--fix-dry-run` isn't added to `flycheck-eslint-args`.  I think it's a good "for now" option for someone wanting to have their code fixed as they work and won't affect anyone else. @cpitclaudel

cpitclaudel:
That makes sense, but I'd prefer not to include it as-is; I'd rather see it packaged with another eslint package, or as a separate package.

cpitclaudel:
As mentioned before, let me reiterate my support for releasing this, just not bundled with flycheck at the moment.  In the long run, functionality like this should be part of #530, though replacing the full buffer seems agressive.  @rpatterson are you OK with closing this?

PR #1672, opened 2020-01-26
cuda-nvcc: add flag to support forcing CUDA file type as CUDA

idcrook:
Add option for checker `cuda-nvcc`

- option `flycheck-cuda-explicitly-specify-cuda-language`

In cases where editing header files (file name extensions of `.h` or `.huc`), the `cuda-nvcc` checker would complain 

```
Suspicious state from syntax checker cuda-nvcc: Flycheck checker cuda-nvcc returned non-zero exit code 1, but its output contained no errors: nvcc fatal   : Don't know what to do with '/tmp/flycheckMzNdps/vec3.cuh'

Try installing a more recent version of cuda-nvcc, and please open a bug report if the issue persists in the latest release.  Thanks!
```

the error is 

    `nvcc` "fatal   : Don't know what to do with file.cuh"

This patch adds an optional flag to pass to `nvcc` to force it to assume the file type is CUDA and then the `nvcc` command performs as required.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -179,8 +179,10 @@ attention to case differences."
     cwl
     d-dmd
     dockerfile-hadolint
+    elixir-credo
     emacs-lisp
     emacs-lisp-checkdoc
+    ember-template
     erlang-rebar3
     erlang
     eruby-erubis
@@ -240,6 +242,7 @@ attention to case differences."
     rst-sphinx
     rst
     ruby-rubocop
+    ruby-standard
     ruby-reek
     ruby-rubylint
     ruby
@@ -532,20 +535,20 @@ highlight them according to `flycheck-highlighting-mode'."
   "The highlighting mode for Flycheck errors and warnings.
 
 The highlighting mode controls how Flycheck highlights errors in
-buffers.  The following modes are known:
+buffers when a checker only reports the starting position of an
+error.  The following modes are known:
 
 `columns'
-     Highlight the error column.  If the error does not have a column,
+     Highlight a single character.  If the error does not have a column,
      highlight the whole line.
 
 `symbols'
-     Highlight the symbol at the error column, if there is any,
-     otherwise behave like `columns'.  This is the default.
+     Highlight a full symbol if there is any, otherwise behave like `columns'.
+     This is the default.
 
 `sexps'
-     Highlight the expression at the error column, if there is
-     any, otherwise behave like `columns'.  Note that this mode
-     can be *very* slow in some major modes.
+     Highlight a full expression, if there is any, otherwise behave like
+     `columns'.  Note that this mode can be *very* slow in some major modes.
 
 `lines'
      Highlight the whole line.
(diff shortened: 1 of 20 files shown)


idcrook:
This is a PR intended to address issue #1673

idcrook:
the `integ` test failures seem to fail with an unrelated error?

       FAILED  flycheck-define-checker/cwl/syntax-error

cpitclaudel:
Thanks. Would there be a downside if we always passed that flag, instead of conditioning it on a user option? Also, do we know why nvcc rejects header files?

idcrook:
My guess is `nvcc` tries to be clever about file names based on extensions. If it doesn't recognize as a CUDA-flavored object file, it doesn't expect to do anything with anything with it. (That's literally the error message : `Don't know what to do with *file*`)  

However many CUDA header files have CUDA inline code, macros, and APIs in them from my limited experience, so it is very convenient to be able to use flycheck on them.  It took me a couple days to find this flag that seems to do the trick in convincing `nvcc` should try to compile without its assuming it is unable to based on the filename extension.

I don't know if there is a downside to always setting it. I would expect that if you are in a CUDA-mode file, the flag can be set.  I didn't want to change the default behavior, mostly.

As it were, in my config, it is on for any "cuda" file, and seems to work as expected.  I did notice a bunch of filename.o files get generated for each filename.cuh/.h ran on, but those are already in my .gitignore files (`*.o`)...

cpitclaudel:
Then I'd be in favor of passing that flag in all cases.  Let's do that in a week if no one comments; can you remind me?

cpitclaudel:
Ping. Can we pass the flag in all cases?

idcrook:
It seems ok. Is there a post method cleanup way to delete the .o files left remaining on the file system? That is a downside I've noticed.

cpitclaudel:
Doesn't the problem with .o files already exist for .c files?  Maybe nvcc has a way to not produce output?

flycheck.el
```diff
@@ -7365,6 +7365,17 @@ See URL `http://stylelint.io/'."
...
+(flycheck-def-option-var flycheck-cuda-explicitly-specify-cuda-language
+    nil cuda-nvcc
+  "Explicitly specify the language for the input files to CUDA.
+
+Explicitly specify the language for the input files, rather than
+letting the compiler choose a default based on the file name
+suffix.  When non-nil, specify CUDA language mode, via `--x=cu'."
+  :type 'boolean
+  :safe #'booleanp
+  :package-version '(flycheck . "33"))
```
cpitclaudel: This should be 32, not 33

idcrook:
> Doesn't the problem with .o files already exist for .c files? Maybe nvcc has a way to not produce output?

I was wondering if there was a way to mark temporary files or output files for deleting? Trying to find such an approach in the flycheck source has not revealed to me such a practice.  Is that accurate?

cpitclaudel:
> Is that accurate?

Correct, we only track files produced by Flycheck itself (not the compilers).  Can't you add something like `-o /dev/null`?

idcrook:
OK.  latest change changes flycheck default to on (`t` instead of `nil`) for setting explicit language and sets output file to `/dev/null` as the rust checker was also doing.

Tested to behave as expected (flycheck-compile included new added arguments, ran compilation and detected injected errors, and no output .o file was observed, etc.)

cpitclaudel:
Thanks. Do you really think we need a defcustom? We always pass -x in the C/C++ checker, so it seems reasonable to do the same here. And it seems that turning off this flag will cause cuda-nvcc to exit with a complaint, so why not pass it always, without needing a defcustom?

idcrook:
The question above was "why not always enable it?". I've only tested on .cuh files in my setup, so with other scenarios, I do not know. If it is left in, it could be used as a local workaround hook per, right? e.g.,  if there's some other strange naming convention out there... 

>  And it seems that turning off this flag will cause cuda-nvcc to exit with a complaint

Why are you saying that?  Are you basing that on something you observed? It should not directly exit if this flag is not present-- `nvcc` will just assume the wrong file type for some file types, and there will be errors presented based on that incorrect assumption.  That is why this was introduced. 

For CUDA mode, I think you might be right however. It should be safe to always be on, since it is implicit assumption that a CUDA mode file should be recognized as such.

For now I'd rather leave it as is.  But we can change it to be hardcoded if required.

cpitclaudel:
> If it is left in, it could be used as a local workaround hook per, right? e.g., if there's some other strange naming convention out there...

I'd rather introduce the flag i such a convention surfaces. Since the checker is only used for cuda-mode, it seems reasonable to always set the flag.

>  Why are you saying that? Are you basing that on something you observed? 

That's what I understood from the original bug report ("Don't know what to do with …")

idcrook:
> > Why are you saying that? Are you basing that on something you observed?
> 
> That's what I understood from the original bug report ("Don't know what to do with …")

Oh good. Was making sure it was broken some way I haven't seen.

Made the change to remove the option wrapper based on feedback.

idcrook:
Oops.  I really biffed that last resulting push since I had merged with master along the way.  might be best for me to submit as another PR...

idcrook:
the two line change has been put forth in PR https://github.com/flycheck/flycheck/pull/1699

cpitclaudel:
Closing this, since it was superseded. Thanks a lot!

PR #1674, opened 2020-02-02
Allow specifying end column and end line for errors

yyoncho:
- backends like lsp-ui could provide that data.

Before:
before

After:
after

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3263,7 +3263,7 @@ non-nil, then only do this and skip per-buffer teardown.)"
                (:constructor flycheck-error-new-at
                              (line column
                                    &optional level message
-                                   &key checker id group
+                                   &key checker id group end-column end-line
                                    (filename (buffer-file-name))
                                    (buffer (current-buffer)))))
   "Structure representing an error reported by a syntax checker.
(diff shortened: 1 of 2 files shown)


cpitclaudel:
Did you see https://github.com/flycheck/flycheck/pull/1400 ?

yyoncho:
> Did you see #1400 ?

Yes - but it is stalled for not being backward compatible(unless I am mistaken) so I went for implementation that does not break existing backends(forgot to mention in the PR that a version with one `region` property). I don't mind rejecting this PR in favour of #1400 - it will for me too.

yyoncho:
ping?

cpitclaudel:
Sorry for the delay (and thanks for the ping).  I think this might be the most reasonable approach.  We can fix the region performance issue at a later point.  Let me just go through #1400 to see what else was relevant in that code.

cpitclaudel:
Closing in favor of #1676.  Thanks for the PR.

> I went for implementation that does not break existing backends

Note that as written, this PR isn't in fact backwards compatible, because the call to (flycheck-error-end-column err) will fail if `err` is created by a package compiled with a previous version of flycheck.

PR #1676, opened 2020-02-11
Add ability to set end column and end line on errors (closes GH-527)

cpitclaudel:
This is a trimmed-down version of #1400 in the spirit of #1674.  The first three commits are cleanup and fixes for the latest Emacs master. Then there's one commit that implements end-line and end-column, and one commit that refactors region calculations to simplify the code.  This version should be backwards-compatible.

@yyoncho, thanks for the ping. @jbaum98, it would be great if you could have a look.  I left the main commit attributed to you (I didn't clearly remember who wrote which parts).  Let me know if you'd prefer being credited in the commit instead of being marked as author.

I'd like a few people to test this in their day-to-day set-ups, with their own checkers.  In particular @kaz-yos who expressed interest in the past, and anyone in @flycheck/contributors who maintains a checker that supports end-lines and end-columns.

Thanks!

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -532,20 +532,20 @@ highlight them according to `flycheck-highlighting-mode'."
   "The highlighting mode for Flycheck errors and warnings.
 
 The highlighting mode controls how Flycheck highlights errors in
-buffers.  The following modes are known:
+buffers when a checker only reports the starting position of an
+error.  The following modes are known:
 
 `columns'
-     Highlight the error column.  If the error does not have a column,
+     Highlight a single character.  If the error does not have a column,
      highlight the whole line.
 
 `symbols'
-     Highlight the symbol at the error column, if there is any,
-     otherwise behave like `columns'.  This is the default.
+     Highlight a full symbol if there is any, otherwise behave like `columns'.
+     This is the default.
 
 `sexps'
-     Highlight the expression at the error column, if there is
-     any, otherwise behave like `columns'.  Note that this mode
-     can be *very* slow in some major modes.
+     Highlight a full expression, if there is any, otherwise behave like
+     `columns'.  Note that this mode can be *very* slow in some major modes.
 
 `lines'
      Highlight the whole line.
--- a/flycheck.el
+++ b/flycheck.el
@@ -3392,125 +3392,80 @@ always greater than BEG."
      ((= end (point-max)) (cons (1- end) end))
      (t (cons end (1+ end))))))
 
-(defun flycheck-error-line-region (err)
-  "Get the line region of ERR.
-
-ERR is a Flycheck error whose region to get.
+(defun flycheck--line-region (pos)
+  "Get the line region of position POS.
 
 Return a cons cell `(BEG . END)' where BEG is the first
 non-whitespace character on the line ERR refers to, and END the
 end of the line."
-  (flycheck-error-with-buffer err
-    (save-restriction
-      (save-excursion
-        (widen)
-        (goto-char (point-min))
-        (forward-line (- (flycheck-error-line err) 1))
-        ;; We are at the beginning of the line now, so move to the beginning of
-        ;; its indentation, similar to `back-to-indentation'
-        (let ((end (line-end-position)))
-          (skip-syntax-forward " " end)
-          (backward-prefix-chars)
-          ;; If the current line is empty, include the previous line break
-          ;; character(s) to have any region at all.  When called with 0,
-          ;; `line-end-position' gives us the end of the previous line
-          (cons (if (eolp) (line-end-position 0) (point)) end))))))
+  (save-excursion
+    (goto-char pos)
+    (forward-line 0)
+    (let ((bol (point))
+          (end (line-end-position)))
+      ;; Move to the beginning of this line's indentation, similar to
+      ;; `back-to-indentation'
+      (skip-syntax-forward " " end)
+      (backward-prefix-chars)
+      ;; If the current line is blank, highlight it in full; if it's
+      ;; empty, include the previous line break character(s) to have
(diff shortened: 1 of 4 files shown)


cpitclaudel:
Turns out just adding new slots at the end of a structure isn't enough either, because pre-compiled packages will create too-short records and feed them to us. I added a commit as a workaround (cc @yyoncho), which rebuilds errors that don't have the right length.

yyoncho:
> I added a commit as a workaround (cc @yyoncho), which rebuilds errors that don't have the right length.

An alternative approach would be to wrap the calls to the new slots in conditional-case.

cpitclaudel:
> An alternative approach would be to wrap the calls to the new slots in conditional-case.

Done, please have a look.  I made the setter do nothing if the slot isn't there.

kaz-yos:
I'm trying to incorporate this into `flycheck-stan`, but right now I am still fixing the developing environment. So don't wait for me if things look ok and I am slow.

purcell:
Really nice work, thanks for tackling this!

flycheck.el
```diff
@@ -3322,113 +3379,93 @@ If the buffer of ERR is not live, FORMS are not evaluated."
...
-    (pcase mode
-      (`columns (or (flycheck-error-column-region err)
-                    (flycheck-error-line-region err)))
-      (`symbols (or (flycheck-error-thing-region 'symbol err)
-                    (flycheck-error-region-for-mode err 'columns)))
-      (`sexps (or (flycheck-error-thing-region 'sexp err)
-                  (flycheck-error-region-for-mode err 'columns)))
-      (`lines (flycheck-error-line-region err))
-      (_ (error "Invalid mode %S" mode)))))
+ERR is a Flycheck error.  It its position is fully specified, use
```
purcell: Small typo here: "it its" -> "if its"

cpitclaudel: Eagle-eyed! Thanks a lot :)

cpitclaudel:
I'll keep using this for a few days on my machine before merging :) Thanks all for testing & reviewing!

flycheck.el
```diff
@@ -1501,6 +1501,13 @@ FILE-NAME is nil, return `default-directory'."
           (file-name-directory file-name)
         (expand-file-name default-directory)))))
 
+(defun flycheck-line-column-to-position (line column)
```
eyal0: Is there a function that does the reverse of this?  [My eglot checker](https://gist.github.com/eyal0/8f58d4209dc8cdb5d9c474b61c49b2dc) already provides the start and end character position.

cpitclaudel: I think the best would be to add a custom constructor, so that your code just gets magically faster once we support both representations without having to compute line and column numbers.  I'll do that in a separate PR, or do you want to do it? The idea is to go through the conversion for now, and at a later point we can change the constructor.

eyal0: I don't feel confident enough at elisp to do this.  Maybe I will try after this PR, if I can get assistance from you in reviewing the code.

purcell: Does this help? https://github.com/purcell/package-lint/blob/cf2ce871549bb816ecb7a9930bdfb77525468f5e/package-lint.el#L201-L207

cpitclaudel: `current-column` looks suspicious in that snippet, since we consider the column number a number of chars in other parts of the code.

cpitclaudel: @eyal0 always happy to help

yyoncho: Just for the record - the LSP protocol returns line + column not point.

purcell: > `current-column` looks suspicious in that snippet, since we consider the column number a number of chars in other parts of the code.

Ah yes, looks like `package-lint` uses zero-based columns for easy compatibility with internal Emacs functions, and then `flycheck-package` uses `flycheck-increment-error-columns` to change them to `flycheck`'s 1-based columns (which presumably originated from that being the convention in most compiler/linter output).

eyal0: @purcell it's not enough to add/subtract one.  `(current-column)` is giving the visual column.  For example, after a single tab by itself on a line, the number might be `8`, even though there is just one character.  Probably it should return 1 or maybe 2.

purcell: UGH. I guess then `(- (point) (line-beginning-position))`....

eyal0: @purcell still wrong, sorry. You need to add one because flycheck assumes that columns begin at one. `(1+ ...)` 😜

cpitclaudel:
:tada:

kaz-yos:
Thank you so much for the great work!

I started working on the corresponding update to the `flycheck-stan` package. As I understand, this PR implemented the error object constructor side.

To put this in use, I am suspecting the `rx-constituents` enhanced in the `flycheck-rx-to-string` function must be extended to include `end-line` and `end-column` that have new match group numbers (for backward compatibility). The relevant part of the current code base is the following.

https://github.com/flycheck/flycheck/blob/master/flycheck.el#L1442-L1450
```lisp
  (let ((rx-constituents
         (append
          `((line . ,(rx (group-n 2 (one-or-more digit))))
            (column . ,(rx (group-n 3 (one-or-more digit))))
            (file-name flycheck-rx-file-name 0 nil)
            (message flycheck-rx-message 0 nil)
            (id flycheck-rx-id 0 nil))
          rx-constituents nil)))
    (rx-to-string form no-group)))
```

Is there ongoing work on this? If not, I'll create an issue and try to work on a PR for this piece.

cpitclaudel:
> Is there ongoing work on this? If not, I'll create an issue and try to work on a PR for this piece.

Yup, see the cpitclaudel_error_ranges branch

kaz-yos:
Thank you so much for your information! I'll look into that branch!

PR #1679, opened 2020-02-17
Configure chaining for Python checkers

fmdkdd:
Rebase of #1486.

I've also changed the chaining to use `warning` level instead of `error`, as I think it is what was intended in the PR.


cpitclaudel:
Works for me! I think we're waiting for @apiology 's CLA, though the amount of code is small enough that we don't necessarily need it.

fmdkdd:
Thanks for the review.

PR #1681, opened 2020-02-24
Add flycheck-position-to-line-column function

eyalsoha:
`flycheck-position-to-line-column` is the inverse of `flycheck-line-column-to-position`

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1508,6 +1508,16 @@ FILE-NAME is nil, return `default-directory'."
     (forward-line (- line 1))
     (min (+ (point) (1- column)) (line-end-position))))
 
+(defun flycheck-position-to-line-column (pos)
+  "Return the line and column of the position POS.
+
+The first line is 1 and the first column is 1."
+  (save-excursion
+    (goto-char pos)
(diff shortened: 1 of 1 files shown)


eyalsoha:
Can someone direct me how to write a unit test for this?

cpitclaudel:
I think the right place to add a test would be in the `test/specs/` folder; but based on our latest conversations I think the best way to go would be to add not just this function, but a constructor for flycheck-error that calls it, too.  I've tried something like that over the week-end; see https://github.com/flycheck/flycheck/tree/cpitclaudel_error_ranges

eyal0:
> I think the right place to add a test would be in the `test/specs/` folder; but based on our latest conversations I think the best way to go would be to add not just this function, but a constructor for flycheck-error that calls it, too. I've tried something like that over the week-end; see https://github.com/flycheck/flycheck/tree/cpitclaudel_error_ranges

Yes this is perfect for me!  I hope that you will have a pull request for this change soon and I will use it!

cpitclaudel:
PR at #1684

PR #1684, opened 2020-02-25
Further work on error ranges

cpitclaudel:
These are client-facing APIs based on the recent range changes:

- dbd63a16 Allow error patterns to mention end-line and end-column
    I don't know how many checkers do that, and I suspect they often do it in compressed form (e.g. `1:5-7` to mean line 1, columns 5 through 7.), so the code assumes `end-line` = `line` if `end-column` is set. cc @kaz-yos

- e4a7446e Add a separate error constructor for checkers that return end positions
    Right now this just converts the position to a line and a column and stores that, but in the future we'll hopefully implement something more efficient.  I think multiple community-maintained checkers already have conversions from offsets to line/columns, so at the very least this should make their lives a bit easier (cc @eyal0)

- dc88bc76 Ensure that maint/flycheck-format works on Emacs master
    This is Emacs bug https://debbugs.gnu.org/cgi/bugreport.cgi?bug=39761

- 1666244f Update flycheck-error-format-position to print end lines and columns
    This is required by the test suite to check end positions.  Btw, I'd like to find a more sustainable way to maintain the test suite in the future.  Ideally, refreshing expected outputs should be straightforward.

- 1c959696 Parse end positions in coq, nix, jsonnet, phpmd, proselint, rust, tslint
    This updates various checkers (I went through all the ones that are packaged with Flycheck) to parse end-lines.  For phpmd though I didn't end up including the end-line though, because it returns giant error spans (over 10 lines).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -69,6 +69,8 @@
   - Add an error explainer for the ``nix-linter`` checker. [GH-1586]
   - Checkers can now specify the exact region covered by an error, using
     the :end-line and :end-column properties. [GH-1400]
+  - Error patterns can now mention ``end-line`` and ``end-column`` in addition
+    to ``line`` and ``column``.
 
 - Improvements
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -68,7 +68,7 @@
     from other files. [GH-1579]
   - Add an error explainer for the ``nix-linter`` checker. [GH-1586]
   - Checkers can now specify the exact region covered by an error, using
-    the :end-line and :end-column properties. [GH-1400]
+    the ``:end-line`` and ``:end-column`` properties. [GH-1400]
   - Error patterns can now mention ``end-line`` and ``end-column`` in addition
     to ``line`` and ``column``.
 
@@ -91,6 +91,9 @@
     to make it easier to change between Python 2 and Python 3 [GH-1113]
   - ``haskell-stack-ghc`` will not try to install GHC thereby freezing Emacs
     [GH-1443]
+  - Checkers that report error positions as a single offset from the start of
+    the file can use the new ``flycheck-error-new-at-pos`` constructor instead
+    of converting that position to a line and a column.
 
 31 (Oct 07, 2017)
 =================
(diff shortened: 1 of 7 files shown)


kaz-yos:
>     * [dbd63a1](https://github.com/flycheck/flycheck/commit/dbd63a1644021cd6929947b822074d72ba343124) Allow error patterns to mention end-line and end-column
>       I don't know how many checkers do that, and I suspect they often do it in compressed form (e.g. `1:5-7` to mean line 1, columns 5 through 7.), so the code assumes `end-line` = `line` if `end-column` is set. cc @kaz-yos

Thank you for the update! The `stanc3` compiler/checker does the `end-line` omission like the following. Your assumption should hold for the cases that `stanc3` gives us.

https://github.com/stan-dev/stan-mode/blob/master/flycheck-stan/examples/example_error_misspelled_type.stanc3out.txt
```
Syntax error in 'examples/example_error_misspelled_type.stan', line 8, column 11 to column 12, parsing error:
   -------------------------------------------------
     6:    vector<lower=0>[J] sigma;
     7:  }
     8:  parameters {
                     ^
     9:    rear mu;
    10:    real<lower=0> tau;
   -------------------------------------------------
Expected top-level variable declaration or "}".
```

cpitclaudel:
@kaz-yos Perfect, thanks for checking that.

fmdkdd:
Man, you're on fire today ;)  Keep it up!

>  Btw, I'd like to find a more sustainable way to maintain the test suite in the future. Ideally, refreshing expected outputs should be straightforward.

Yes, we should have an option to `make integ` that just replaces the expected output with the actual output.

kaz-yos:
I'm not sure if it's possible to submit a commit-specific review or approval, but I reviewed changes in commit dbd63a1 (add group 6 for end-line and group 7 for end-column). It looks good to me.

cpitclaudel:
@flycheck/contributors I'd love to merge this soon, to give it some testing before the release

cpitclaudel:
> I wasn't sure how to review the tests. I guess the only changes are addition of end positions.

That's right :) Given that these pass, I don't think there's too much to double-check there.  Thanks a lot for the review!

cpitclaudel:
Merging this, since I'd like to give this broad testing before the next release.

PR #1685, opened 2020-02-25
Add support for eqc to flycheck-erlang-rebar3-profile

cpitclaudel:
This is a refresh of #1576 ;  @raghavkarol can you have a look? If it looks good to you I'll take care of merging, since @fmdkdd had already reviewed most of this.


cpitclaudel:
Going ahead with this PR since it was just a refresh of an already-reviewed one.

PR #1516, opened 2018-11-16
 Take into account ruleSeverity for TSLint errors

gexplorer:
All rules from TSLint where generating warnings in flycheck. I added the *ruleSeverity* parsing to *level* so TSLint errors generate flycheck errors instead of warnings

PR #1686, opened 2020-02-25
Take ruleSeverity into account for TSLint errors

cpitclaudel:
Rebase and cleanup of #1516 (cc @gexplorer)


cpitclaudel:
Going ahead with since one since it was just a refresh of an existing PR

PR #1691, opened 2020-03-17
Update verilator patterns

cpitclaudel:


cpitclaudel:
We'll need to update the version number in the docker build machine for the tests to pass.

fmdkdd:
The docker image just installs the version that is in the repositories of Ubuntu 18.04, unfortunately.  We could probably try bumping the Ubuntu version used, but 20.04 is just around the corner.

We can just remove verilator from the image so the CI just doesn't test it until we bump.

cpitclaudel:
> The docker image just installs the version that is in the repositories of Ubuntu 18.04, unfortunately. We could probably try bumping the Ubuntu version used, but 20.04 is just around the corner.

I think it's OK to wait for a while; verilator isn't very easy to upgrade outside of using the repos, so we can wait to merge this.

DanielVukelich:
Ubuntu 20.04 LTS released on April 23.  Any chance that this could merge now?

cpitclaudel:
Let's wait until September to get people a bit of time to upgrade?

MikeWalrus:
@cpitclaudel  Why is this still not merged? Even Ubuntu 22.04 LTS has come out. :)

cpitclaudel:
Thanks for the reminder!

PR #1695, opened 2020-03-18
Fix #1507 (allow Emacs Lisp checker to run without a buffer-file-name)

### LINKED ISSUE #1507: Flycheck Emacs Lisp in Org-mode babel source blocks
I want to have Flycheck syntax checking when I am editing Emacs Lisp code in a literal Org-mode file using babel `SRC` blocks.

However the `flycheck-emacs-lisp` checker predicate has the following to say:

> Ensure that we only check buffers with a backing file. For buffers without a backing file we cannot guarantee that file names in error messages are properly resolved, because `byte-compile-file' emits file names relative to the directory of the checked file instead of the working directory. Hence our backwards-substitution will fail, because the checker process has a different base directory to resolve relative file names than the Flycheck code working on the buffer to check.

So whenever I try to launch the checker I get `predicate: nil` in `flycheck-verify-setup`.

This is definitely a drawback in the literate-programming style. Is there a way to work around it so that source blocks can be checked?

cpitclaudel:
CC @matthew-piziak

Closes GH-1507.  This isn't perfect because org mode temporary buffers don't
have a file name at all, so we don't know which buffer they came from.

The original code said this:

> Ensure that we only check buffers with a backing file.  For buffers
> without a backing file we cannot guarantee that file names in error
> messages are properly resolved, because `byte-compile-file' emits file
> names *relative to the directory of the checked file* instead of the
> working directory.  Hence our backwards-substitution will fail, because
> the checker process has a different base directory to resolve relative
> file names than the Flycheck code working on the buffer to check.

The problem is that we create a temp file in /tmp/, and the checker reports
errors relative to that, so we get "xyz" as a file name.  This doesn't match
what we expect (`$PWD/xyz`).  The fix here is to tell the ELisp compiler which
directory to report errors relative to.


matthew-piziak:
Thank you so much @cpitclaudel! What workflow would you recommend for me, who has a literate init file with many source blocks? When I use `flycheck` on this revision I get errors now, but they're mostly of the form "the function `foo` is not known to be defined", and I don't want to put a bunch of `require` statements at the front of every source block. It feels like currently every source block is like its own package.

cpitclaudel:
> What workflow would you recommend for me, who has a literate init file with many source blocks? It feels like currently every source block is like its own package.

It's quite tricky.  Your description is exactly right.

Ideally, you'd want org to create an indirect buffer that has all your snippets in it, and narrow it to the snippet you're editing.  Then Flycheck and Emacs would see the whole context.

Is there a generic way (in org-mode) to extract all snippets that a snippet depends on?

cpitclaudel:
I've updated the PR to make it more robust.  Can you give it a new try and tell me if it still gives you errors (the update won't improve things in terms of spurious errors)

matthew-piziak:
> Ideally, you'd want org to create an indirect buffer that has all your snippets in it, and narrow it to the snippet you're editing. Then Flycheck and Emacs would see the whole context.
> 
> Is there a generic way (in org-mode) to extract all snippets that a snippet depends on?

Well there's **tangling**. If I jump to definition (`xref-find-definition`) in `main.org`, it jumps me to the tangled `main.el` file, even if the definition being jumped to is in the same snippet. If flycheck used `main.el` as the backing file for errors, while using the snippet buffer for error location data, then I think that would do it. Keeping them in sync is another issue though. Maybe I should look into [Org Tanglesync](https://github.com/mtekman/org-tanglesync.el).

>I've updated the PR to make it more robust. Can you give it a new try and tell me if it still gives you errors (the update won't improve things in terms of spurious errors)

Confirmed, still works on 33f2b0178f18b6e25e7b7293c43378d686bb01f3.

cpitclaudel:
> Well there's tangling

Right. Ideally, that temporary buffer that org gives you should be the full tangled file, narrowed to just the snippet that you're editing.  That's a discussion that would be better taken to the org-mode list, though.

>  Confirmed, still works on 33f2b01.

Thanks!

matthew-piziak:
> That's a discussion that would be better taken to the org-mode list.

Quite right. Never did figure out how to use it, to be honest. If only they had a status-based bug tracker…

flycheck.el
```diff
@@ -7759,14 +7769,14 @@ to be checked.  It also sets `package-user-dir' according to
   "Option VALUE filter for `flycheck-emacs-lisp-initialize-packages'."
   (let ((shall-initialize
          (if (eq value 'auto)
-             (or (flycheck-in-user-emacs-directory-p (buffer-file-name))
+             (or (flycheck-in-user-emacs-directory-p
+                  (or buffer-file-name default-directory))
```
fmdkdd: Are we guaranteed that `default-directory` of the temporary Org source buffer the same as the parent Org buffer?

cpitclaudel: Yup, the manual says "When you create a new buffer that
is not visiting a file, via a command like ‘C-x b’, its default
directory is usually copied from the buffer that was current at the time
(*note Select Buffer::)."

fmdkdd: Thanks for enlightening me :+1:

Issue #1519, opened 2018-11-23
flycheck-next-error does not always place point on error


CyberShadow:
## Expected behavior

* `flycheck-next-error` should always place point "on" the error, so that:
  * the error's description is displayed in the minibuffer (as when placing point on the first character of the error)
  * `(flycheck-overlay-errors-at (point))` is non-nil.

## Actual behavior

* At least in certain circumstances, flycheck will place point one character past the error overlay, which causes the above conditions to be false. 

  One situation appears to be when the error occurs on some punctuation - the created overlay will underline the identifier preceding it.

## Steps to reproduce

I can reproduce this behavior with a simple D program from `emacs -Q`:

1. Create a buffer called `test.d`
2. Paste the code: `void main() { int i; i.writeln; }`
3. With point at the beginning of the buffer, invoke `flycheck-next-error` (`C-c ! n`).

## Additional information

* Flycheck version: 32snapshot (package: 20181018.1021)
* GNU Emacs 26.1 (build 1, x86_64-pc-linux-gnu, X toolkit, Xaw scroll bars) of 2018-06-24
* OS: Arch Linux
* `flycheck-verify-setup`:

```
Syntax checkers for buffer test.d in d-mode:

  d-dmd
    - may enable: yes
    - executable: Found at /home/vladimir/bin/dmd

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled
checkers.

--------------------

Flycheck version: 32snapshot (package: 20181018.1021)
Emacs version:    26.1
System:           x86_64-pc-linux-gnu
Window system:    x
```

fmdkdd:
I haven't reproduced yet, but I'm guessing `flycheck-highlighting-mode` is set to `symbols`?  Can you try with another highlighting-mode?

CyberShadow:
Yes, it doesn't occur with `flycheck-highlighting-mode` set to `columns`.

fmdkdd:
Ok, so I'm holding on this until #1400.

cpitclaudel:
Now we (finally!) have #1400 I think we can come back to this.  The problem is that DMD seems to reports errors after their actual location. For example with this:

```
xvoid main() { int i; i.writeln; }
```

The error is on the first character of `main`.  Same for the original example:

```
void main() { int i; i.writeln; }
```

The error is on the `.`.  Our heuristic to highlight symbols then calls `(bounds-of-thing-at-point 'symbol)`, which returns a span covering the word before the point, but not the point.

[...]

PR #1696, opened 2020-03-18
Fall back to columns highlighting when error begins on a non-symbol

cpitclaudel:
Closes GH-1519.

Suggested-by: @CyberShadow

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3450,8 +3450,14 @@ the THING at the column, and END the end of the THING."
             (eq mode 'lines))
         (flycheck--line-region beg)
       (or (pcase mode
-            (`symbols (flycheck-bounds-of-thing-at-point 'symbol beg))
-            (`sexps (flycheck-bounds-of-thing-at-point 'sexp beg)))
+            (`symbols
+             ;; Ensure that we're on a word or symbol.  See
+             ;; https://github.com/flycheck/flycheck/issues/1519
+             (and (< beg (point-max))
+                  (memq (char-syntax (char-after beg)) '(?w ?_))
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -3447,8 +3447,14 @@ the THING at the column, and END the end of the THING."
...
         (flycheck--line-region beg)
       (or (pcase mode
-            (`symbols (flycheck-bounds-of-thing-at-point 'symbol beg))
-            (`sexps (flycheck-bounds-of-thing-at-point 'sexp beg)))
+            (`symbols
+             ;; Ensure that we're on a word or symbol.  See
+             ;; https://github.com/flycheck/flycheck/issues/1519
+             (and (< beg (point-max))
+                  (memq (char-syntax (char-after beg)) '(?w ?_))
+                  (flycheck-bounds-of-thing-at-point 'symbol beg)))
```
CyberShadow: What I had in mind (to avoid depending on the exact behavior of `bounds-of-thing-at-point`):
```suggestion
             (let ((bounds (flycheck-bounds-of-thing-at-point 'symbol beg)))
                 (when (and bounds (< beg (cdr bounds)))
                   bounds))
```
but this looks good too 👍

cpitclaudel: Good point, I like your option too. It feels a tad cleaner to only call bounds-of-things-at-point when we're on a symbol constituent, but I don't have a strong opinion.

cpitclaudel:
Let's give @fmdkdd time to have a look and then we can go ahead and merge this.

CyberShadow:
Thank you!

cpitclaudel:
And thanks to you!

PR #1698, opened 2020-03-19
Use file-truename in check-same-files

### LINKED ISSUE #1634: Temporary file issue with r-lintr
## Bug description
When running in the background, the `r-lintr` checker is unable to reconcile the correct R source path from the temporary one, so Flycheck error messages appear like this:

```
In "../../../private/var/folders/4r/94db7f8j12q1xyjzyrtj6hbwl14jnc/T/flycheck09Z4SY/flycheck.R":
unexpected end of input
```

...and highlight to the first line of the buffer, regardless of where in the buffer the error actually occurs.

## Steps to reproduce
1. With a clean Emacs instance, install and `load-library` the latest `ess` and `flycheck` from Melpa.
2. Open an R script with some style and code errors in it.
3. Disable flymake: `M-x flymake-mode`
4. Enable flycheck: `M-x flycheck-mode`
5. Wait for flycheck to run.

An error like the one above should appear on the first line, _regardless_ of where the _actual_ error is in the file.

Note that `flycheck-compile` works fine -- it spits out the right errors _and_ style mistakes in all the right places.

## Screenshots

Screen Shot 2019-10-18 at 4 55 44 PM

## System configuration
```
Syntax checkers for buffer flycheck.R in ess-r-mode:

First checker to run:

  r-lintr
    - may enable:    yes
    - predicate:     t
    - executable:    Found at /usr/local/bin/R
    - lintr library: present

checkers.

--------------------

Flycheck version: 32snapshot (package: 20190930.1425)
Emacs version:    26.2
System:           x86_64-apple-darwin14.5.0
Window system:    ns
```

Vanilla Emacs, with minimal configuration, on MacOS High Sierra. `.emacs.d/init.el` is:

```emacs-lisp
(require 'package)
(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/"))
(add-to-list 'package-archives '("melpa-stable" . "https://stable.melpa.org/packages/"))
(package-initialize)

(custom-set-variables
 ;; custom-set-variables was added by Custom.
 ;; If you edit it by hand, you could mess it up, so be careful.
 ;; Your init file should contain only one such instance.
 ;; If there is more than one, they won't work right.
 '(package-selected-packages (quote (julia-mode ess flycheck))))
(custom-set-faces
 ;; custom-set-faces was added by Custom.
 ;; If you edit it by hand, you could mess it up, so be careful.
 ;; Your init file should contain only one such instance.
 ;; If there is more than one, they won't work right.
 )
```

Emacs configuration:
- [X] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs

ashiklom:
file-chase-links does not look at parent directory links.
file-truename does.

An example illustrated by #1634:

On MacOS, `/var` is a symbolic link to `/private/var`. Therefore,
`(file-chase-links "/var/file.txt") => "/var/file.txt")`, but
`(file-truename "/var/file.txt") => "/private/var/file.txt"`

As a result, `check-same-files` would incorrectly report these as
different files.

Fixes #1634.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1214,7 +1214,7 @@ emulate inode (see fstat() in emacs/src/w32.c)."
         (and attrs-a attrs-b  ;; Make sure both file-a and file-b exist.
              (equal (nth 10 attrs-a) (nth 10 attrs-b))    ;; inode
              (equal (nth 11 attrs-a) (nth 11 attrs-b))))  ;; filesystem
-      (string= (file-chase-links file-a) (file-chase-links file-b))))
+      (string= (file-truename file-a) (file-truename file-b))))
 
 (defvar-local flycheck-temporaries nil
   "Temporary files and directories created by Flycheck.")


cpitclaudel:
That's what we used to do:

```diff
commit 9dd3f36e54276a5989cd02a4457be7a00dcd2823
Author: Likai Liu <liulk@likai.org>
Date:   Sun Jul 28 00:41:06 2019 -0400

    flycheck-same-files-p improves performance using heuristics
    
    This change improves responsiveness when the buffer has a lot of
    warnings or errors.
    
    https://gist.github.com/liulk/91228e4e61f7e2ba3fce1eeee4915b8c
    
    The total CPU usage for flycheck-handle-signal went down from 86%, to
    59%, to 18%. Out of those, the amount used by flycheck-same-files-p
    went down from 84%, to 47%, to 14%.

diff --git a/flycheck.el b/flycheck.el
--- a/flycheck.el
+++ b/flycheck.el
@@ -1192,14 +1192,20 @@
 
 (defun flycheck-same-files-p (file-a file-b)
-  "Determine whether FILE-A and FILE-B refer to the same file."
-  (let ((file-a (expand-file-name file-a))
-        (file-b (expand-file-name file-b)))
-    ;; We must resolve symbolic links here, since some syntax checker always
-    ;; output canonical file names with all symbolic links resolved.  However,
-    ;; we still do a simple path compassion first, to avoid the comparatively
-    ;; expensive file system call if possible.  See
-    ;; https://github.com/flycheck/flycheck/issues/561
-    (or (string= (directory-file-name file-a) (directory-file-name file-b))
-        (string= (directory-file-name (file-truename file-a))
-                 (directory-file-name (file-truename file-b))))))
+  "Determine whether FILE-A and FILE-B refer to the same file.
+
+Files are the same if (in the order checked):
+
+- They have the same path.  Or,
+- They have the same inode and filesystem numbers, not following symlink.  Or,
+- After chasing symlinks, they result in the same path.
+
+This should work even on w32, where Emacs uses file index to
+emulate inode (see fstat() in emacs/src/w32.c)."
+  (or (string= file-a file-b)
+      (let ((attrs-a (file-attributes file-a))
+            (attrs-b (file-attributes file-b)))
+        (and attrs-a attrs-b  ;; Make sure both file-a and file-b exist.
+             (equal (nth 10 attrs-a) (nth 10 attrs-b))    ;; inode
+             (equal (nth 11 attrs-a) (nth 11 attrs-b))))  ;; filesystem
+      (string= (file-chase-links file-a) (file-chase-links file-b))))
 ```

**Comment by ashiklom** [comment]
Ah, OK. Would a MacOS-specific hack for this specific case be a better option? I.e. If MacOS and starts with `/var`, replace with `/private/var`?

**Comment by cpitclaudel** [comment]
Wait wait, not so fast! :) If we don't find a better solution, your PR is the right way to go (better slow and correct that fast and wrong).  I think hardcoding a specific path isn't the way to go, because we'll get bitten by that again sooner or later.

Can we make sure that the problem is only when a directory is a symlink of another one? In that case, shouldn't the inode test catch the fact that these files are the same?

Otherwise, we could resort to `file-truename` only when all other tests fail and the directory names are the same. (cc @liulk)

**Comment by liulk** [comment]
I think @ashiklom's fix is fine, just replacing file-chase-links with file-truename when the fast cases have been exhausted.

With one suggestion: since file-chase-links is faster but has false negative (never false positive), maybe keep it and then add the file-truename as an or'd case below last the last resort.

```lisp
(or (string= file-a file-b)
    (let ((attrs-a (file-attributes file-a))
          (attrs-b (file-attributes file-b)))
      (and attrs-a attrs-b  ;; Make sure both file-a and file-b exist.
           (equal (nth 10 attrs-a) (nth 10 attrs-b))    ;; inode
           (equal (nth 11 attrs-a) (nth 11 attrs-b))))  ;; filesystem
    (string= (file-chase-links file-a) (file-chase-links file-b))
    (string= (file-truename file-a) (file-truename file-b))
)
```

Though I'm really surprised that the inode approach didn't already work for /private/var/foo vs. /var/foo.

liulk:
Oh I just remembered that the inode approach only works if the file actually exists. But in the error messages parsed by flycheck, the file might not actually exist, so that's why file-chase-links or file-truename might be necessary.

file-chase-links only works if the basename does exist as a symlink but points to a nonexistent target. file-truename works if the basename does not exist at all.

cpitclaudel:
@liulk Thanks a lot for the very fast response!

> Though I'm really surprised that the inode approach didn't already work for /private/var/foo vs. /var/foo.

Me too.  I think we should investigate this. @ashiklom, can you try to have a look maybe?  

>  But in the error messages parsed by flycheck, the file might not actually exist, so that's why file-chase-links or file-truename might be necessary.

Don't we also need file-chase-links to resolve softlinks?  Softlinks don't share their target's inode number.

> Oh I just remembered that the inode approach only works if the file actually exists.

Oh, that makes sense.  But in the particular instance the file does exist, I think?

liulk:
I just remembered that file-attributes corresponds to lstat() not stat(). If the basename is a symlink, then lstat() returns the inode for the symlink itself and won't chase the link.

In that case, let's keep file-chase-links as a fallback to file-attributes, and then file-truename really as the last resort. The rationale is that file-chase-links might generate O(n) iterations where n is the length of the symlink chain, but file-truename would generate O(m*n) where n is the length of the symlink chain, and m is the directory depth of the path.

cpitclaudel:
> In that case, let's keep file-chase-links as a fallback to file-attributes, and then file-truename really as the last resort.

That's probably good.  But I worry that we might be slowing down the case in which the files are different, which is not uncommon for certain checkers. In the long run, it might be good to have a small cache to keep this information.  In fact, wouldn't that be simpler? It's likely quite rare to have a checker return errors mentioning hundreds of different files, while hundreds of errors mentioning the same file is common.

WDYT?

cpitclaudel:
I implemented the memoizing strategy at #1701 .  @ashiklom , could you give it a go and see if it fixes your problem?

liulk:
That looks fine. Hash table should be faster than file-attributes which requires a system call. On the other hand, hash table doesn't pick up symlink changes on the file system. It may not be a big problem though, and I think the performance gain is much more important.

Probably worth emphasizing that setq-local makes the hash table buffer-local, so it gets cleared off once the buffer closes and doesn't slowly leak memory across the lifetime of emacs. I'm sure there are people like me who uses a long running emacs daemon. :)

cpitclaudel:
> Probably worth emphasizing that setq-local makes the hash table buffer-local, so it gets cleared off once the buffer closes

Right, that's exactly why I made if buffer-local (and initialized to nil)

Issue #1673, opened 2020-01-26
cuda-nvcc: Does not work with .h or .cuh files


idcrook:
Thank you for taking the time to report an issue and improve Flycheck.  This template is for **actual bugs you observed**.  If you have trouble setting up Flycheck, or if you have a question, please use the relevant issue template instead.

## Checklist
- [X] I have checked existing issues for potential duplicates before creating this one.
- [X] I have read the [Troubleshooting guide][].

## Bug description
`nvcc` (checker: cuda-nvcc) doesn't run properly on files with `.h` or `.cuh` extensions.  These are header files which contain CUDA code. 

    `nvcc` "fatal   : Don't know what to do with file.cuh"

## Steps to reproduce
1. Open file
2. See error

       `nvcc` "fatal   : Don't know what to do with file.cuh"

## Expected behavior

The `nvcc` command should run preprocessor and compiler on the CUDA header file, as it does with `.cu` files.

## System configuration
```
Syntax checkers for buffer vec3.cuh in cuda-mode:

First checker to run:

  cuda-nvcc
    - may enable: yes
    - executable: Found at /usr/bin/nvcc

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled checkers.

--------------------

Flycheck version: 32snapshot
Emacs version:    28.0.50
System:           x86_64-pc-linux-gnu
Window system:    x

```

Emacs configuration:
- [X] Plain Emacs / Custom configuration

## Additional notes

Created PR #1672 that adds optional flag to `nvcc` that makes it treat source file as CUDA. With this flag set, `nvcc` now successfully runs as a checker on .h and .cuh CUDA files.

[Troubleshooting guide]:  https://www.flycheck.org/en/latest/user/troubleshooting.html

PR #1699, opened 2020-03-19
cuda-nvcc: add explicit language option and suppress file output

idcrook:
- intended to fix https://github.com/flycheck/flycheck/issues/1673
- intended to replace https://github.com/flycheck/flycheck/pull/1672


cpitclaudel:
Great work, thanks a lot.  And thanks for your patience!

PR #1701, opened 2020-03-20
Simplify and speed up flycheck-same-files-p by memoizing file-truename

cpitclaudel:
This version should also be more correct (closes GH-1634, where our tests were
not precise enough).

Reported-by: @ashiklom

CC @liulk: if you still have a copy of your benchmark, it would be good to see how this version performs.


ashiklom:
Just tried this out and it solves the issue! Thanks for the quick turnaround!

cpitclaudel:
Thanks for trying this.  Ideally I'd like @fmdkdd or someone else from @flycheck/contributors to have a look before pushing, but if no one is available we'll go ahead :)

PR #1702, opened 2020-03-20
Fix systemd-analyze checker not adding dummy line numbers

kisaragi-hiu:
#1624 added a new pattern for `systemd-analyze` that allows for missing line numbers to match errors like `foobar.service: Command /home/foo/bar is not executable: No such file or directory` (from #1269). However, these errors would be discarded by `flycheck-relevant-errors-p`, as a dummy line number wasn't added back.

This PR adds them back with `flycheck-fill-empty-line-numbers` so that those errors can show up.


cpitclaudel:
Thanks!

PR #1703, opened 2020-03-20
Add comment about how errors without line numbers are discarded to hopefully save confusion

kisaragi-hiu:
I spent an hour or so figuring out why the systemd-analyze checker isn't returning errors when the `systemd-analyze verify` command clearly does, and finally realized that it's because I didn't add dummy line numbers in the `:error-filter` when I added a new error pattern which has an optional line number in #1624.

(The pull for the issue above is in #1702.)

Add a comment to explain that `flycheck-fill-empty-line-numbers` should be used to add a line number to those errors, and a sentence in the developer documentation to point to where this is implemented, hopefully saving future confusions.


cpitclaudel:
Thanks a lot for taking the time to improve things for everyone; much appreciated!

Issue #1585, opened 2019-06-18
eslint overrides tslint in typescript mode


vincekd:
The `javascript-eslint` checker now runs by default with `typescript-mode` (since https://github.com/DamienCassou/flycheck/commit/8d74ac092dd213a7c5a77135eab903ca872dc717) which overrides my config to use `typescript-tslint`:
```
(with-eval-after-load "flycheck"
  (flycheck-add-mode 'typescript-tslint 'typescript-mode))
```

How do I disable `javascript-eslint` for `typescript-mode`, or make it prefer `tslint`?

Results of `flycheck-verify-setup`:
```
Syntax checkers for buffer test.ts in typescript-mode:

First checker to run:

  javascript-eslint
    - may enable:  yes
    - executable:  Found at /usr/local/bin/eslint
    - config file: found

Checkers that could run if selected:

  typescript-tslint  select
    - may enable:         yes
    - executable:         Found at /usr/bin/tslint
    - configuration file: Found at "/home/user/dev/project/tslint.json"

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled
checkers.
```

cpitclaudel:
You could add a hook to `typescript-mode-hook`:

```
(add-hook 'typescript-mode-hook
    (lambda () (setq-local flycheck-disabled-checkers '(javascript-eslint))))
``` 

But maybe we need a better default? I don't use either checkers these days, so I'll let others chip in.

vincekd:
Also, it seems weird that this breaking change was introduced when `javascript-eslint` does not support typescript out of the box. Need to install special parser/plugins and update config.

cpitclaudel:
Isn't tslint deprecated?

vincekd:
yeah but they just decided a couple of months ago, and there are still a lot of issues to work out for a smooth migration.

https://github.com/palantir/tslint/issues/4534

cpitclaudel:
Is there anything more that we should do here?

zoechi:
I can't get rid of `Symbol’s value as variable is void: javascript-eslint`
when I try https://github.com/flycheck/flycheck/issues/1585#issuecomment-503276929
Any idea what I might do wrong?

Issue #1472, opened 2018-06-01
Huge delay for accesibility of tide's functionality


kreedz:
I have tested this issue on my working project with more than 5000 ts and tsx files. And so huge (30-90 seconds) delay noticed only there. Unfortunately I can't share that project.
Also I have tested this issue on http://github.com/Microsoft/vscode project. There something like 3000 ts files. And delay is smaller - something like 10 seconds.
### Packages that installed
They specified below in my `init.el` config:
 - add-node-modules-path
 -  tide
 - web-mode

### Steps to reproduce
 - Open emacs. Before you open the file make sure that `tslint` is installed locally or globally. For local `tslint` you will need `add-node-modules-path` package (you can find it below in my `init.el` file).
 - Then for the first time open a ts file (e.g. `vscode/src/vs/base/browser/ui/progressbar/progressbar.ts`) and wait until tsserver will do his init stuff.
 - Place pointer on imported variable (e.g. on `TPromise` here `import { TPromise, ValueCallback } from 'vs/base/common/winjs.base';`). Wait some time until that variable will be highlighted (it's tide-hl-mode). So now you can jump to definition by `M-.`. Try to do that.
 - Jump back by `M-,`.
 - Immediately place pointer at some imported variable.
 - Now here I'm **expecting** that variable will be highlighted and it's will be possible to jump to definition.
But **what is happening** instead: variable doesn't highlighted and you can't jump to definition, you have to wait something like 30-90 (or ~10 seconds on vscode project) seconds until do that.

The issue will not occur again in the same file when it happens once.

Same issue happens when I write something like 'asdasd' in the new line and then place pointer at imported variable.
Also experiencing that when open another file from same project and place pointer on imported variable.

This issue happens both on linux and windows machines.
But on linux I'm getting next message when placing pointer on variable and when that variable doesn't highlighted:
`Error from syntax checker typescript-tslint: Bad string format: "doesn't start with `\"'!" [4 times]`
But on windows machine that message doesn't show.

### Outputs and init config from linux
`flycheck-verify-setup`:
```
emacs --version: 25.3.1
Flycheck version: 32snapshot (package: 20180527.918)

flycheck-verify:
Syntax checkers for buffer accessibilityHelp.ts in typescript-mode:

  typescript-tide
    - may enable:        yes
    - predicate:         t
    - Typescript server: running
    - Tide mode:         enabled

  typescript-tslint
    - may enable:         yes
[...]

Issue #1446, opened 2018-04-10
Is it changeable flycheck temporary files path?


johnny-mh:
[Karam test runner](https://karma-runner.github.io/) watching my source directory to rerun test suite.

flycheck makes temporary files to same directory with source code.

so the test runner run tests every key typing and finally process die.

I had use option for `prefix` but it doesn't help.

because the test runner watching entire source directory.

can i change path of flycheck temporary files?

fmdkdd:
You are using the `typescript-tslint` checker, right?

There is no option to change the directory of temporary files that are created using `source-inplace`, because the point of  `source-inplace` is to create files /in the same directory/ as the file being checked.  Some checkers (`tslint` in this case) rely on this behavior in order to correctly check files that have dependencies.

In this case, `source-inplace` was added to fix #1420.

You can try changing the `source-inplace` to `source` in the definition of `typescript-tslint` and see if that works for you (this will create temporary files under `/tmp` instead).

Otherwise, I would recommend to instruct your test runner to ignore some files (most file watchers should have this option for various reasons).

cc @Simplify

everedifice:
@fmdkdd Thank you. I just change `flycheck.el` file that `source-inplace` to `source` on tslint checker. it works nice. but is that OK? Do i change that code when every flycheck updates?

fmdkdd:
> Do i need change that code every flycheck package updates?

If you want to make it permanent, then you could copy and paste the whole checker definition into your own Emacs configuration.  It would override the built-in Flycheck `typescript-tslint`.

But maybe we have to rethink this `source-inplace` afterall.  I'm interested in what @Simplify thinks.

Simplify:
I'll prefer `source-inplace` to stay because of integration with `typescript-tide` and growing number of rules in `tslint` that depend on `tsconfig`. 

@everedifice Temporally files that Flycheck generates are in format `flycheck_<original-filename>` so you can exclude `**/flycheck(*)` in your [karma config](https://karma-runner.github.io/2.0/config/configuration-file.html) i guess?

Deewiant:
Chiming in with another case where the current status quo is painful. I'm using Tide together with Flycheck on an application that was bootstrapped with `create-react-app`, i.e. https://github.com/wmonk/create-react-app-typescript. While running the file watcher (`yarn start` or `npm run start`) and making code changes, eventually the whole server dies due to an error like (paths scrubbed):

```
FatalError:
[...]

PR #1704, opened 2020-03-20
Deprecate typescript-tslint

cpitclaudel:
Closes GH-1446 and closes GH-1472, as eslint supports --stdin, so there are no temp
file problems.  Also closes GH-1585: with tslint being deprecated, we do want to
override it with eslint.


cpitclaudel:
Let's make one more release with the tslint checker in and then remove it.

PR #1706, opened 2020-03-20
Properly handle empty strings from completing-read in -read-error-level

cpitclaudel:
Closes GH-1418.

Reported-by: @phst

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1557,7 +1557,11 @@ Use `flycheck-completing-read-function' to read input from the
 minibuffer with completion.
 
 Show PROMPT and read one of CANDIDATES, defaulting to DEFAULT.
-HISTORY is passed to `flycheck-completing-read-function'."
+HISTORY is passed to `flycheck-completing-read-function'.
+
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -1526,7 +1526,11 @@ Use `flycheck-completing-read-function' to read input from the
 minibuffer with completion.
 
 Show PROMPT and read one of CANDIDATES, defaulting to DEFAULT.
-HISTORY is passed to `flycheck-completing-read-function'."
+HISTORY is passed to `flycheck-completing-read-function'.
+
+Note that `flycheck-completing-read-function' may return an empty
+string instead of nil, even \"\" isn't among the candidates.  See
```
fmdkdd: Missing a word here?
```suggestion
string instead of nil, even when \"\" isn't among the candidates.  See
```

cpitclaudel:
> So with a nil default, -completing-read returned an empty string and we were not handling that properly, is that it?

That's exactly right

flycheck.el
cpitclaudel: Thanks :)

Issue #1461, opened 2018-05-12
Checking uses `exec-path`, but `flycheck-compile` uses $PATH


joewreschnig:
Split off from the discussion of using npx at #1428:

If I have a project-local executable e.g. `eslint` in `node_modules/.bin`, I can add that directory to my `exec-path` to make the checker work (and green in `flycheck-verify-setup`). But, running `flycheck-compile` fails because `compilation-start` shells out, and the directory is not in my `$PATH`.

In my own Emacs Lisp code I've worked around this by always calling `executable-find` before running anything with shells out.

This occurs in Flycheck 20180509.845 (and probably all earlier versions).

fmdkdd:
> In my own Emacs Lisp code I've worked around this by always calling executable-find before running anything with shells out.

So you get the absolute path to the program using `exec-path`, and pass it to the shell, is that it?

joewreschnig:
> So you get the absolute path to the program using exec-path, and pass it to the shell, is that it?

Right. Usually I try to do it as late as possible before calling the command, in case something else also wants to override `exec-path`.

fmdkdd:
Ok, thanks for confirming.  I agree that we should eliminate the discrepancy between `flycheck-compile` and what happens with `flycheck-verify-setup` (which reflects `flycheck-buffer`).  Your method would be a good start I think.

cpitclaudel:
Thanks for the report.  Two years later, I have a PR at #1707

PR #1707, opened 2020-03-20
Improve flycheck-compile

cpitclaudel:
Closes GH-1461


flycheck.el
```diff
@@ -5955,29 +5962,48 @@ use with `compilation-error-regexp-alist'."
...
-                            (cons (flycheck-checker-executable checker) args))
-                   " ")))
+  (let* ((args (flycheck--checker-substituted-shell-command-arguments checker))
+         (program
+          (or (flycheck-find-checker-executable checker)
+              (user-error "Cannot find `%s' using `flycheck-executable-find'"
+                          (flycheck-checker-executable checker))))
+         (wrapped (flycheck--wrap-command program args))
+         (abs-prog
+          (or (executable-find (car wrapped))
```
fmdkdd: I'm confused: The docstring to `flycheck--wrap-command` states that we shouldn't call `executable-find` on its output, but here we are doing it anyway.

cpitclaudel: We shouldn't call `flycheck-executable-find`, but  a plain `executable-find` is fine.  `process-start` does that internally, but we need to do this explicitly here otherwise the shell will do its own executable-find against PATH instead of exec-path.

cpitclaudel: I added a comment to clarify

Issue #929, opened 2016-03-28
Support for checking AWK scripts


ackalker:
It would be nice for Flycheck to have support for checking AWK scripts.
I think this is quite doable, and in a safe way, too, contrary to what one might expect.
Please see [this SO answer](http://stackoverflow.com/a/7212314/588561) for details.

cpitclaudel:
Yup, good call; that SO answer looks convincing. 

We don't have a very rich history of accepting checkers that massage the input too much, but in this case it seems pretty minimal. @lunaryorn will make the final call on that.

Here's an idea: can you try to write a checker definition based on that SO answer? I can help if anything from http://www.flycheck.org/en/latest/_downloads/flycheck.html#Syntax-checker-definitions is unclear. Then we can decide whether it should go into flycheck itself, or as a separate package; but in any case it will be useful!

Cheers,
Clément.

lunaryorn:
@ackalker We'll certainly accept an AWK checker if it's safe—which the SO answer seems to promise, for GNU AWK at least.

 I will not however add it myself; I don't even use awk.  My priorities are elsewhere. Please feel welcome to open a pull request that adds this syntax checker to Flycheck, though, and ask questions if you need help with defining this syntax checker for Flycheck.

Please take a look at our [Contributor's Guide](http://www.flycheck.org/en/latest/contributor/contributing.html) for more information about contributing to Flycheck.

ackalker:
Thanks for the quick replies :) I'll look into it when I have some spare time.

cpitclaudel:
@ackalker did you get around to writing that checker? @unhammer has one at https://gist.github.com/unhammer/bc58677ceb3d72755174537a05486f7c ; maybe we could integrate it in Flycheck?

unhammer:
I'll happily sign the CLA if needed :)

cpitclaudel:
That would be great :) Do you want to open a PR? ^^

unhammer:
I may need some help there. https://www.flycheck.org/en/latest/contributor/contributing.html#pull-requests says I need to install cask and do `make check specs unit`. I did

    $ curl -fsSL https://raw.githubusercontent.com/cask/cask/master/go | python
    $ export PATH="$HOME/.cask/bin:$PATH"

but then I get

    $ make check specs unit
    cask exec emacs -Q --batch -L .  -l maint/flycheck-format.el -f flycheck/batch-check-format
    Cannot open load file: No such file or directory, f
[...]

PR #1708, opened 2020-03-21
Add awk checker using gawk --lint

unhammer:
Re: https://github.com/flycheck/flycheck/issues/929

Should there be a check somewhere for whether gawk is installed?


flycheck.el
```diff
@@ -6737,6 +6738,46 @@ See URL `http://asciidoctor.org'."
...
+  "Remove the repeated file-name/line from the error message of ERR."
+  (setf (flycheck-error-message err)
+        (replace-regexp-in-string
+         "^\\([ \t]*\\)\\(.*\n\\)\\1"
+         "\\2"
+         (replace-regexp-in-string
+          "\ngawk: [^ ]*:"
+          "\n"
+          (flycheck-error-message err))))
+  err)
```
cpitclaudel: We usually prefer using `rx` over direct regexps, but I can make the change if you want :)

flycheck.el
```diff
@@ -6737,6 +6738,46 @@ See URL `http://asciidoctor.org'."
...
+  (seq-do #'flycheck-awk-gawk-fix-message errors)
+  errors)
+
+(flycheck-define-checker awk-gawk
+  "GNU awk's built-in --lint checker."
+  :command ("gawk"
+            ;; Avoid code execution https://github.com/w0rp/ale/pull/1411
+            "--source" "'BEGIN{exit} END{exit 1}'"
+            "-f"
+            source-original
```
cpitclaudel: Could we change this to `source` maybe? Does the current directory matter?

cpitclaudel: If not, we need to add `:predicate flycheck-buffer-saved-p`, since otherwise we'll get errors for the file's on-disk contents which won't match the buffer's contents

unhammer: `source` seems to work when I test :)

flycheck.el
unhammer: rx'ed

PR #1710, opened 2020-03-22
Autoload flycheck-define-checker to prevent unwanted argument expansion

cpitclaudel:
The problem pops up when using this macro in a with-eval-after-load block.  In
that case, the arguments of the macro are expanded, and then the extra quoting
added by the macro itself on top of self-quoting forms breaks things.

As a concrete example, consider this definition:
```elisp
  (flycheck-define-checker test
    "Just a test"
    :command ("echo" "a" source)
    :predicate (lambda () t)
    :error-patterns ((error "Error"))
    :standard-input nil
    :modes (fundamental-mode))
  (push 'test flycheck-checkers))
```

Macro-expansion in the argument of with-eval-after-load will expand the (lambda)
form as well as any macro that might appear in :error-patterns, :modes, etc.
When the macro eventually runs, with its arguments expanded, its assumptions
about receiving its arguments unexpanded will be broken, and all sorts of this
will go wrong (starting with the lambda being quoted a second time).

Autoloading ensures that the macroexpander understands that
flycheck-define-checker is a macro, and doesn't expand its arguments first.

Closes GH-1398.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6589,6 +6589,10 @@ objects)."
 
 
 ;;; Convenience definition of command-syntax checkers
+
+;; This macro is autoloaded to prevent `with-eval-after-load' from expanding its
+;; arguments.  See https://github.com/flycheck/flycheck/issues/1398.
+;;;###autoload
 (defmacro flycheck-define-checker (symbol docstring &rest properties)
   "Define SYMBOL as command syntax checker with DOCSTRING and PROPERTIES.
 


cpitclaudel:
I think @purcell is our goto experts on tricky byte-compilation questions.  @purcell , if you have a minute to look at this, it would be most appreciated.  We've had this long-lasting problem that using `flycheck-define-checker` (a macro) in `with-eval-after-load` or `use-package` creates breakage, because the arguments to the macro are expanded before the macro is called (because the macroexpander doesn't know that `flycheck-define-checker` is a macro).

While we ponder better options, this PR autoloads the macro.  Do you know of a different way to tell the macroexpander about this without loading the full flycheck.el?  Maybe @monnier has ideas too.

Thanks!

purcell:
> the arguments to the macro are expanded before the macro is called (because the macroexpander doesn't know that `flycheck-define-checker` is a macro).

This seems weird to me. I wouldn't think that the code could be either `eval`-ed or `compile`d without knowing whether `flycheck-define-checker` is a macro. But if this workaround helps, I don't see any harm in it. The ultimate solution is to not have unusual quoting/expansion behaviour in the macro, I suspect, as per #1440.

cpitclaudel:
> This seems weird to me. I wouldn't think that the code could be either eval-ed or compiled without knowing whether flycheck-define-checker is a macro. 

Yes, it's a common misconception.  I had the same misunderstanding.  When the macroexpander doesn't know a function or macro, it assumes it's a function and expands its arguments.

To wit:

```elisp
(with-eval-after-load 'xxx
  (unknown-macro (push 123 xyz)))

(defmacro unknown-macro (arg)
  (message "%S" arg))

(provide 'xxx)
````

Prints: `(setq xyz (cons 123 xyz))` (expanded), where as this:

```elisp

(defmacro known-macro (arg)
  (message "%S" arg))

(with-eval-after-load 'yyy
  (known-macro (push 123 xyz)))

(provide 'yyy)
```

prints that: `(push 123 xyz)`

purcell:
Right, yep, thanks for the worked example.

PR #1712, opened 2020-03-25
Clean up config-file and option var declarations

cpitclaudel:
Instead of having the `:safe` property on all config file we can just put it once and for all in `def-config-file-var`; and it should allow `nil`.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1190,6 +1190,10 @@ to a number and return it.  Otherwise return nil."
     (when (and (stringp string) (string-match-p number-re string))
       (string-to-number string))))
 
+(defun flycheck-string-or-nil-p (obj)
+  "Determine if OBJ is a string or nil."
+  (or (null obj) (stringp obj)))
+
 (defun flycheck-string-list-p (obj)
   "Determine if OBJ is a list of strings."
   (and (listp obj) (seq-every-p #'stringp obj)))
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -6912,7 +6917,7 @@ information about the standard library."
   :type '(choice (const "libc++")
                  (const :tag "GNU libstdc++" "libstdc++")
                  (string :tag "Library name"))
-  :safe #'stringp
+  :safe #'flycheck-string-or-nil-p
```
fmdkdd: The choice list should contain nil.

flycheck.el
```diff
@@ -8374,8 +8375,8 @@ Each item is a string with a tag to be given to `go build'."
 Should be a string representing a version, like 1.6 or 1.11.4.
 See `https://staticcheck.io/docs/#targeting-go-versions' for
 details."
-  :type 'string
-  :safe #'stringp
+  :type '(choice (const :tag "Unspecified" nil)
+                 (string :tag "Version"))
```
fmdkdd: Missing a `-string-or-nil-p`?

cpitclaudel: :eagle: :eye:

flycheck.el
cpitclaudel: Good point, fixed.

PR #1668, opened 2020-01-17
Add file extension support for PHP Codesniffer

Sasanidas:
PHP Codesniffer extension support, take the file extension as suggested @cpitclaudel
Thanks!

PR #1715, opened 2020-03-26
Add php-phpcs extension support with conditional.

Sasanidas:
Im sorry for the delay @cpitclaudel , here it is the conditional phpcs extension,
(old pull request  https://github.com/flycheck/flycheck/pull/1668)

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9350,6 +9350,13 @@ See URL `http://pear.php.net/package/PHP_CodeSniffer/'."
             ;; been configured with show_progress enabled
             "-q"
             (option "--standard=" flycheck-phpcs-standard concat)
+            ;; Some files are not detected correctly
+            ;; so it is necessary to pass the extension.
+            (eval
+             (-when-let* ((fname buffer-file-name)
+                          (ext (file-name-extension fname)))
+               (concat "--extensions=" ext)))
+
             ;; Pass original file name to phpcs.  We need to concat explicitly
             ;; here, because phpcs really insists to get option and argument as
             ;; a single command line argument :|


flycheck.el
```diff
@@ -9350,6 +9350,11 @@ See URL `http://pear.php.net/package/PHP_CodeSniffer/'."
             ;; been configured with show_progress enabled
             "-q"
             (option "--standard=" flycheck-phpcs-standard concat)
+            ;; Some files are not detected correctly
+            ;; so it is necessary to pass the extension.
+            (eval
+             (when (file-name-extension (buffer-file-name))
```
cpitclaudel: You need to condition on buffer-file-name as well, since it may be nil.

Something like 

```
(-when-let ((fname buffer-file-name)
            (ext (file-name-extension buffer-file-name))) 
  …)
```

Sasanidas: Umm, what do you think about using (when (buffer-file-name).. like the below option?

cpitclaudel: Sure, that's what the change I suggested does.

flycheck.el
```diff
@@ -9353,7 +9353,7 @@ See URL `http://pear.php.net/package/PHP_CodeSniffer/'."
             ;; Some files are not detected correctly
             ;; so it is necessary to pass the extension.
             (eval
-             (when (file-name-extension (buffer-file-name))
+             (when (buffer-file-name)
                (concat "--extensions=" (file-name-extension (buffer-file-name)))))
```
cpitclaudel: You *also* need the extension, since the buffer might not have when; that's why I suggested `-when-let*`

flycheck.el
```diff
@@ -9353,8 +9353,9 @@ See URL `http://pear.php.net/package/PHP_CodeSniffer/'."
             ;; Some files are not detected correctly
             ;; so it is necessary to pass the extension.
             (eval
-             (when (buffer-file-name)
-               (concat "--extensions=" (file-name-extension (buffer-file-name)))))
+             (-when-let (ext (file-name-extension buffer-file-name))
+               (concat "--extensions=" (file-name-extension buffer-file-name))))
```
cpitclaudel: Sorry, this is still wrong.  You reintroduced the same problem: if buffer-file-name is nil, this will throw an exception.

cpitclaudel: Did you test this code? It sets the variable `ext` to the value of `buffer-file-name`

flycheck.el
```diff
@@ -9350,6 +9350,12 @@ See URL `http://pear.php.net/package/PHP_CodeSniffer/'."
             ;; been configured with show_progress enabled
             "-q"
             (option "--standard=" flycheck-phpcs-standard concat)
+            ;; Some files are not detected correctly
+            ;; so it is necessary to pass the extension.
+            (eval
+             (-when-let (ext  buffer-file-name fn (file-name-extension buffer-file-name))
```
cpitclaudel: Did you test this new code? It sets the variable `ext` to the value of `buffer-file-name`

Sasanidas: yes, i test it with the scratch buffer, file without extension and a file with extension.
No exception throw

cpitclaudel: The syntax on the line with the -when-let is wrong: the fn part and the file-name-extension part are ignored.  Can you fix it? I gave you the correct syntax in my first comment.

Sasanidas: Ok, seems like now with -when-let* it's working

flycheck.el
```diff
@@ -9350,6 +9350,13 @@ See URL `http://pear.php.net/package/PHP_CodeSniffer/'."
             ;; been configured with show_progress enabled
             "-q"
             (option "--standard=" flycheck-phpcs-standard concat)
+            ;; Some files are not detected correctly
+            ;; so it is necessary to pass the extension.
+            (eval
+             (-when-let* ((fname buffer-file-name)
+                          (ext (file-name-extension fname)))
```
cpitclaudel: Almost there, but now you can actually use `ext` instead of `(file-name-extension buffer-file-name)` in the `concat`

Sasanidas: Ok, really thank you for your patience

Sasanidas:
> Thanks, I think we're good to go. Thanks a lot for sticking through the process! Please squash your commits into one and fix the indentation of the code, and I'll be able to merge.

cpitclaudel:
The build failed because of this:

```
/flycheck/flycheck.el:9462: line is over 80 characters
```

flycheck.el
```diff
@@ -9350,6 +9350,11 @@ See URL `http://pear.php.net/package/PHP_CodeSniffer/'."
             ;; been configured with show_progress enabled
             "-q"
             (option "--standard=" flycheck-phpcs-standard concat)
+            ;; Some files are not detected correctly
+            ;; so it is necessary to pass the extension.
+            (eval
+             (-when-let* ((fname buffer-file-name) (ext (file-name-extension fname)))
```
cpitclaudel: Please reintroduce the blank line here

cpitclaudel:
Great work, thanks a lot!

PR #1718, opened 2020-03-28
Prevent redundant calls to display-error-at-point-soon

cpitclaudel:
We still keep the unconditional display in focus-in-hook, since displays such as
pop-ups may need to be re-displayed after focusing back in.

Follow up to #1248.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2646,7 +2646,7 @@ Slots:
     ;; the focus hooks only work on Emacs 24.4 and upwards, but since undefined
     ;; hooks are perfectly ok we don't need a version guard here.  They'll just
     ;; not work silently.
-    (post-command-hook . flycheck-display-error-at-point-soon)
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -4924,29 +4925,45 @@ non-nil."
...
+(defun flycheck--error-display-tick ()
+  "Return point and tick counter of current buffer."
+  (cons (point) (buffer-modified-tick)))
+
+(defvar-local flycheck--last-error-display-tick nil
+  "Value of `flycheck--error-display-tick' when errors were last displayed.")
+
 (defun flycheck-display-error-at-point ()
-  "Display the all error messages at point in minibuffer."
+  "Display the all error messages at point."
```
fmdkdd: Typo?
```suggestion
  "Display all the error messages at point."
```

cpitclaudel: Thanks!

Issue #1383, opened 2018-01-05
Flycheck Pylint output parser error


noxdafox:
It seems Flycheck is getting confused by Pylint output.

```
Suspicious state from syntax checker python-pylint: Flycheck checker python-pylint returned non-zero exit code 4, but its output contained no errors: ************* Module flycheck_scraper
flycheck_scraper.py:9:-1:W:pointless-string-statement:String statement has no effect

Try installing a more recent version of python-pylint, and please open a bug report if the issue persists in the latest release.  Thanks!
```
The file seems to have no pylint errors but a single warning.
```bash
$ pylint3 scraper.py

************* Module scraper
W:  9,-1: String statement has no effect (pointless-string-statement)

...

$ echo $?
4
```

I can reproduce this with the following pylint versions:

- pylint 1.6.5 (Debian Stretch)
- pylint 1.7.4 (Debian Sid)
- pylint 1.8.1 (Pypi) 

Flycheck version: 32snapshot (package: 20171214.1215)

fmdkdd:
The negative column `-1` in the warning fails to match the regexp.  Do you have any idea why `pylint` would output a negative column index?  Seems like a bug to me: there is no code at that column!

noxdafox:
Reported: https://github.com/PyCQA/pylint/issues/1822

PR #1720, opened 2020-03-31
Move pylint to JSON and fix GH-1383

cpitclaudel:
Also remove the syntax error test, because Pylint's JSON output incorrectly
escapes angle brackets (see https://github.com/PyCQA/pylint/pull/3467).

Closes GH-1383.


cpitclaudel:
I don't think it's a very important test, so I'd be in favor of moving forward.

PR #857, opened 2016-01-16
haskell-stack-ghc predicate on stack.yaml

slpopejoy:
For #847. Checks for 'stack.yml', 'stack.yaml' or environment $STACK_YAML.

PR #1721, opened 2020-03-31
Add an :enabled test to haskell-stack-ghc

### LINKED ISSUE #847: haskell-stack-ghc breaks flycheck on non-stack projects
Change #711 breaks cabal-only projects, if you have stack on your PATH.

It appears that flycheck will ignore a checker if it's command isn't on the PATH. Thus haskell-stack-ghc will not fire on systems without `stack` installed. However this is insufficient for users using stack on some projects but not others. 

Workaround is of course to disable the checker and maybe use dir-locals to turn it back on. However I'm hoping there's a way to check for a `stack.yaml` in the project root and enable the checker then?

Thanks, Stuart

cpitclaudel:
Supersedes #857, closes #847. 

This ensures that we only try to run slack when it's available, which also means
that users won't get "No compiler found" messages from Stack if they
don't use it.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8851,13 +8851,13 @@ containing a file that matches REGEXP."
 (defun flycheck-haskell--find-stack-default-directory ()
   "Find a directory to run haskell-stack-ghc.
 
-Return a parent directory with a stack.*.yaml file, or the
+Return a parent directory with a stack*.y[a]ml file, or the
 directory returned by \"stack path --project-root\"."
   (or
    (when (buffer-file-name)
      (flycheck--locate-dominating-file-matching
       (file-name-directory (buffer-file-name))
-      "stack.*\\.yaml\\'"))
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -8925,7 +8920,16 @@ See URL `https://github.com/commercialhaskell/stack'."
     (flycheck-sanitize-errors (flycheck-dedent-error-messages errors)))
   :modes (haskell-mode literate-haskell-mode)
   :next-checkers ((warning . haskell-hlint))
-  :working-directory flycheck-haskell--find-default-directory)
+  :working-directory (lambda (_)
```
fmdkdd: Why not pass the function directly?

cpitclaudel: It doesn't take a checker argument

purcell:
Very nice, Clément!

Issue #1471, opened 2018-05-27
Use Emacs 27's new JSON-parsing primitives


cpitclaudel:
From the changelog at https://github.com/emacs-mirror/emacs/blob/a37166eb80458e87880893e7b8aa98a2a7ce394b/etc/NEWS#L27:

```
** The new configure option '--with-json' adds support for JSON using
the Jansson library.  It is on by default; use 'configure
--with-json=no' to build without Jansson support.  The new JSON
functions 'json-serialize', 'json-insert', 'json-parse-string', and
'json-parse-buffer' are typically much faster than their Lisp
counterparts from json.el.
```

We could feature-check for these and use them if available.

bk322:
(note: one need to `sudo apt install libjansson-dev` on deb-based linux for this to work; also `./configure` shows `Does Emacs use -ljansson? yes` if that was found)

PR #1723, opened 2020-03-31
Use Emacs 27's new JSON parsing primitives when available

cpitclaudel:
Closes GH-1471.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -6450,26 +6450,23 @@ the BUFFER that was checked respectively.
 
 See URL `https://palantir.github.io/tslint/' for more information
 about TSLint."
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -6450,26 +6450,23 @@ the BUFFER that was checked respectively.
 
 See URL `https://palantir.github.io/tslint/' for more information
 about TSLint."
-  (let ((json-array-type 'list))
```
fmdkdd: Was this somehow redundant with the former `json-read`?

cpitclaudel: Yup, exactly.  flycheck-parse-json already sets this variable

Issue #1278, opened 2017-07-08
The issue of perl checker with a large source file


OGAWAHirofumi:
To describe the issue, at first, "perl -wc" can do exit(255) without
reading all of inputed source if there are some kind of warnings.  So,
emacs's (process-send-region) signals 'file-error.

The details of issue is, 

This is partial strace of output of emacs(31773) and perl(31797).

[...]
;; emacs runs perl
31797 execve("/usr/bin/perl", ["/usr/bin/perl", "-w", "-c"], [/* 55 vars */] <unfinished ...>
[...]
;; emacs write source 64k (maximum of pipe size)
31773 write(14, "#!/usr/bin/perl\n#\n# For test:\n# "..., 152975 <unfinished ...>
31773 <... write resumed> )             = 65536
31773 write(14, "EOF\";\n\n                      See"..., 87439) = -1 EAGAIN (Resource temporarily unavailable)
;; perl reading source...
31797 read(0, "#!/usr/bin/perl\n#\n# For test:\n# "..., 8192) = 8192
31797 read(0, "}\n\n# Decide width of summary ima"..., 8192) = 8192
[...]
31773 write(14, "EOF\";\n\n                      See"..., 87439) = 8192
31797 read(0, "printf(\"%s_iops_c.dat\", kdevname"..., 8192) = 8192
31773 read(15,  <unfinished ...>
;; perl outputs warnings/errors
31773 <... read resumed> "Global symbol \"$unit\" requires e"..., 4096) = 283
;; perl exits
31797 exit_group(255)                   = ?
31797 +++ exited with 255 +++
31773 read(15, "", 4096)                = 0
;; emacs close pipes
31773 close(14)                         = 0
31773 close(15)                         = 0
;; ...continue write source, then got error
31773 write(-1, "d) = @_;\n\n    my @normalized_arg"..., 79247) = -1 EBADF (Bad file descriptor)

After this sequence, flycheck-buffer runs (error) part with
'file-error signal from process-send-region, and failed to display
warnings with following error.

Error while checking syntax automatically: (file-error "Writing to process" "Bad file descriptor" #<process flycheck-perl>)

On other hand, command line (or flycheck-compile) outputs this,

$ cat fsperf.pl | perl -wc
Global symbol "$unit" requires explicit package name (did you forget to declare "my $unit"?) at - line 1226.
Global symbol "$unit" requires explicit package name (did you forget to declare "my $unit"?) at - line 1232.
[...]

PR #1724, opened 2020-04-02
Allow checkers to close stdin before reading all input

cpitclaudel:
Fixes GH-1278.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5604,7 +5604,7 @@ function is probably far less efficient than
   ;; Chunk process output on Windows to work around
   ;; https://github.com/flycheck/flycheck/issues/794 and
   ;; https://debbugs.gnu.org/cgi/bugreport.cgi?bug=22344.  The presence of
-  ;; `w32-pipe-buffer-size' denotes an Emacs version (> Emacs 25.1 )where pipe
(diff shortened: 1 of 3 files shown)


cpitclaudel:
Finally got the tests to pass on this one.

flycheck.el
```diff
@@ -5683,7 +5688,19 @@ and rely on Emacs' own buffering and chunking."
...
+                (flycheck-process-send-buffer process)
+              ;; Some checkers exit before reading all input, causing errors
+              ;; such as a `file-error' for a closed pipe, or a plain “no longer
+              ;; connected to pipe; closed it” error for a disconnection.  We
+              ;; report them if needed in `flycheck-finish-checker-process' (see
+              ;; `https://github.com/flycheck/flycheck/issues/1278').
+              (error (process-put process 'flycheck-error err))))
+          ;; Set the actual sentinel and process any events that might have
+          ;; happened while we were sending input.
+          (setf (process-sentinel process) #'flycheck-handle-signal)
```
fmdkdd, line 5701: Why do we need to set the sentinel back to `-handle-signal` in this case?

cpitclaudel: A checker returning errors on the fly could return a few errors as we send it our input, and a few more after we're done, so we need to preserve the handler.

Issue #1729, opened 2020-04-09
Flycheck does not pick up GCC diagnostics that don't have a column number


Deewiant:
## Checklist
- [x] I have checked existing issues for potential duplicates before creating this one.
- [x] I have read the [Troubleshooting guide][].

## Bug description
Flycheck does not report some GCC diagnostics. It looks to me that it's because they don't have a column number. There's at least one such GCC warning I've run into in GCC 9.3.0.

Arguably this is a GCC bug, but practically I think Flycheck should be robust to this and it's probably also easier to address this here.

## Steps to reproduce

1. Use this source code:
```c
#pragma nonsense
int main(int argc, char **argv) { return 0; }
```

2. Open that file in Emacs with flycheck-mode enabled and `c/c++-gcc` as the checker.
3. Observe two warnings reported by Flycheck in e.g. `(flycheck-list-errors)`.
4. Check the output of `(flycheck-compile 'c/c++-gcc)` and see that there are three warnings there, so Flycheck is leaving out one of them:

```
-*- mode: compilation; default-directory: "~/" -*-
Compilation started at Thu Apr  9 09:39:17

/usr/bin/gcc -fshow-column -iquote /tmp/ -Wall -Wextra -x c -S -o /dev/null - < /tmp/asdf.c
<stdin>:1: warning: ignoring #pragma nonsense  [-Wunknown-pragmas]
<stdin>: In function ‘main’:
<stdin>:2:14: warning: unused parameter ‘argc’ [-Wunused-parameter]
<stdin>:2:27: warning: unused parameter ‘argv’ [-Wunused-parameter]

Compilation finished at Thu Apr  9 09:39:17
```

- [x] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

## Expected behavior
I expected all three warnings to be reported by Flycheck.

## System configuration
```
Syntax checkers for buffer asdf.c in c-mode:

First checker to run:

  c/c++-gcc (explicitly selected)
    - may enable:    yes
    - executable:    Found at /usr/bin/gcc
    - next checkers: c/c++-cppcheck

Checkers that could run if selected:

  c/c++-clang  select
    - may enable:    yes
    - executable:    Found at /usr/bin/clang
    - next checkers: c/c++-cppcheck

Checkers that are compatible with this mode, but will not run until properly configured:

  c/c++-cppcheck (disabled)
    - may enable: Automatically disabled!
[...]

PR #1731, opened 2020-04-11
Mark column numbers as optional in GCC error patterns

cpitclaudel:
Fixes GH-1729.


cpitclaudel:
Will merge as soon as I figure out why the Literate Haskell test is broken — I could use some help with that part.

PR #1740, opened 2020-04-22
Fix links to next checkers in flycheck-describe-checker

cpitclaudel:
The links pointed to the definition of the main checker instead of the
documentation of the next checkers.


cpitclaudel:
Thanks!

PR #1742, opened 2020-04-23
Add two new indication modes: left-margin and right-margin

cpitclaudel:
Unlike fringes, margin indicators are not limited to bitmaps, so we can display images or characters or anything else that Emacs can display.  The easiest way to try this out is `M-x flycheck-set-indication-mode` on this branch.

Here's how it looks:

- Fringes (latest Flycheck)

  fringe

- Margins (this branch, with `flycheck-indication-mode` set to `left-margin`)

  margin

### DIFF
--- a/doc/user/error-reports.rst
+++ b/doc/user/error-reports.rst
@@ -117,9 +117,6 @@ wide, but a 16 pixels version is used if the fringe is `wide enough
    distribution please take a look at its documentation if you're unsure about
    the appearance of Flycheck's indicators.
 
-   Note that we discourage you from changing the shape of Flycheck’s fringe
-   indicators.
-
 You can customise the location of these indicators (left or right fringe) with
 `flycheck-indication-mode`, which also lets you turn off these indicators
 completely; additionally, you can move these indicators into the margins instead
@@ -142,14 +139,15 @@ of the fringes:
       Do not indicate errors and warnings in the fringe or in the margin.
 
 The following faces control the colours of fringe and margin indicators.
-However they do not let you change the shape of the indicators—to achieve this
-you'd have to redefine the error levels with `flycheck-define-error-level`.
 
 .. defface:: flycheck-fringe-error
              flycheck-fringe-warning
              flycheck-fringe-info
 
    The icon faces for ``error``, ``warning`` and ``info`` levels respectively.
+
+To change the fringe bitmap or the symbol used in the margins, use the function
+``flycheck-redefine-standard-error-levels``.
 
 Mode line
 =========
--- a/doc/user/error-reports.rst
+++ b/doc/user/error-reports.rst
@@ -138,6 +138,24 @@ of the fringes:
    ``nil``
       Do not indicate errors and warnings in the fringe or in the margin.
 
+By default, Emacs displays fringes, but not margins.  With ``left-margin`` and
+``right-margin`` indication modes, you will need to enable margins in your
+``.emacs``.  For example:
+
+.. code-block:: elisp
+
+   (setq-default left-fringe-width 1 right-fringe-width 8
+                 left-margin-width 1 right-margin-width 0)
+
+If you intend to use margins only with Flycheck, consider using
+``flycheck-set-indication-mode`` in a hook instead; this function adjusts
+margins and fringes for the current buffer.
+
+.. code-block:: elisp
+
+   (setq-default flycheck-indication-mode 'left-margin)
+   (add-hook 'flycheck-mode-hook #'flycheck-set-indication-mode)
+
 The following faces control the colours of fringe and margin indicators.
 
 .. defface:: flycheck-fringe-error
--- a/doc/user/error-reports.rst
+++ b/doc/user/error-reports.rst
@@ -99,11 +99,11 @@ The highlights use the following faces depending on the error level:
    The highlighting face for ``error``, ``warning`` and ``info`` levels
    respectively.
 
-Fringe icons
-============
+Fringe and margin icons
+=======================
 
-In GUI frames Flycheck also adds indicators to the fringe—the left or right
-border of an Emacs window that is—to help you identify erroneous lines quickly.
+In GUI frames, Flycheck also adds indicators to the fringe—the left or right
+border of an Emacs window—to help you identify erroneous lines quickly.
 These indicators consist of a rightward-pointing double arrow shape coloured in
 the colour of the corresponding error level.  By default the arrow is 8 pixels
 wide, but a 16 pixels version is used if the fringe is `wide enough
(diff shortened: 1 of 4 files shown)


cpitclaudel:
The easiest way to try this out is `M-x flycheck-set-indication-mode`, btw

cpitclaudel:
… and this is how it looks:

fringe

margin

mokulus:
#### When you call `flycheck-set-indication-mode` without optional argument then mode is nil and you get user-error.

Try evaluating `(flycheck-set-indication-mode)`. If mode is not specified I believe variable `flycheck-indication-mode` should be used. Possible fix:

```diff
diff --git a/flycheck.el b/flycheck.el
index a4418f01..0161c488 100644
--- a/flycheck.el
+++ b/flycheck.el
@@ -581,23 +581,24 @@ especially if you use margins only in Flycheck buffers."
                                          "left-margin" "right-margin")
                               nil t nil nil
                               (prin1-to-string flycheck-indication-mode)))))
-  (pcase mode
-    ((or `left-fringe `right-fringe)
-     (setq left-fringe-width 8 right-fringe-width 8
-           left-margin-width 0 right-margin-width 0))
-    (`left-margin
-     (setq left-fringe-width 1 right-fringe-width 8
-           left-margin-width 1 right-margin-width 0))
-    (`right-margin
-     (setq left-fringe-width 8 right-fringe-width 8
-           left-margin-width 0 right-margin-width 1))
-    (_ (user-error "Invalid indication mode")))
-  (dolist (win (get-buffer-window-list))
-    (set-window-margins win left-margin-width right-margin-width)
-    (set-window-fringes win left-fringe-width right-fringe-width))
-  (setq-local flycheck-indication-mode mode)
-  (when flycheck-current-errors
-    (flycheck-buffer)))
+  (let ((mode (or mode flycheck-indication-mode)))
+    (pcase mode
+      ((or `left-fringe `right-fringe)
+       (setq left-fringe-width 8 right-fringe-width 8
+	     left-margin-width 0 right-margin-width 0))
+      (`left-margin
+       (setq left-fringe-width 1 right-fringe-width 8
+	     left-margin-width 1 right-margin-width 0))
+      (`right-margin
+       (setq left-fringe-width 8 right-fringe-width 8
+	     left-margin-width 0 right-margin-width 1))
+      (_ (user-error "Invalid indication mode")))
+    (dolist (win (get-buffer-window-list))
+      (set-window-margins win left-margin-width right-margin-width)
+      (set-window-fringes win left-fringe-width right-fringe-width))
+    (setq-local flycheck-indication-mode mode)
+    (when flycheck-current-errors
+      (flycheck-buffer))))
 
 (defcustom flycheck-check-syntax-automatically '(save
                                                  idle-change

```

Sorry for the long diff, but I had to wrap it inside let. It's the only difference.

#### Variable `flycheck-indication-mode` doesn't mention margins.
Try `C-h v flycheck-indication-mode`. Fringes are mentioned but margins are not.

#### Example config

I use straight and use-package. After applying the patch above I can use:
```elisp
  (use-package flycheck
    :straight (:host github
               :repo "flycheck/flycheck"
               :branch "cpitclaudel_margin-indication")
    :hook ((text-mode prog-mode) . flycheck-mode)
    :hook (flycheck-mode . flycheck-set-indication-mode)
    :config
    (setq flycheck-indication-mode 'left-fringe))
```

#### left-margin and git-gutter interfere

`left-margin` doesn't work for me with git-gutter enabled. I think it's because they use the same space and git-gutter takes over. Not sure how to fix it.

#### screenshots

##### left-*fringe* *without* git-gutter
left-fringe-no-git-gutter

##### left-*fringe* *with* git-gutter
left-fringe-with-git-gutter

##### left-*margin* *without* git-gutter
left-margin-no-git-gutter

##### left-*margin* *with* git-gutter - doesn't show
left-margin-with-git-gutter


#### I think left-fringe + git-gutter looks best

If I understand correctly margins use bitmaps and fringes strings, but fringes seems more distinct for me right now. I would consider changing the bitmap to something more visible.

cpitclaudel:
@mokulus Thanks a lot for the feedback.  I've updated the docstring and fixed the issue with `nil`.
(btw, if you settle on fringes, you don't need to call `set-indication-mode`, since Flycheck uses fringes by default, and I'm not planning to change that).

> If I understand correctly margins use bitmaps and fringes strings, but fringes seems more distinct for me right now. I would consider changing the bitmap to something more visible.

It's the reverse: margins use a string currently, and fringes use a bitmap.
The problem with git-gutter is that is uses the margins too, so both symbols get displayed there.  If you set the margin width to 2 I bet you'll see both symbols.  It would be better to override one with the other, but there's no way to do that at the moment, I think.

Alexander-Miller:
Doesn't work on my end, the fringe remains empty: https://imgur.com/a/JeJQGdD

cpitclaudel:
> Doesn't work on my end, the fringe remains empty: https://imgur.com/a/JeJQGdD

That's expected: it should display in the margins.  Did you use `M-x flycheck-set-indication-mode`?

Alexander-Miller:
The margin's definitely there: https://imgur.com/a/3Xx4sFc

doc/user/error-reports.rst
```diff
@@ -117,33 +117,56 @@ wide, but a 16 pixels version is used if the fringe is `wide enough
...
+By default, Emacs displays fringes, but not margins.  With ``left-margin`` and
+``right-margin`` indication modes, you will need to enable margins in your
+``.emacs``.  For example:
+
+.. code-block:: elisp
 
-The following faces control the colours of the fringe indicators.  However they
-do not let you change the shape of the indicators—to achieve this you'd have to
-redefine the error levels with `flycheck-define-error-level`.
+   (setq-default left-fringe 1 right-fringe 8
```
kljohann, line 147:
```suggestion
   (setq-default left-fringe-width 1 right-fringe-width 8
```

doc/user/error-reports.rst
```diff
@@ -117,33 +117,56 @@ wide, but a 16 pixels version is used if the fringe is `wide enough
...
+                 left-margin 1 right-margin 0)
```
kljohann, line 148:
```suggestion
                 left-margin-width 1 right-margin-width 0)
```

mokulus:
@Alexander-Miller How did you install it? Are you sure you used this branch? Maybe try my install script that I posted before.

Alexander-Miller:
I use straight.el as my package manager which clones the packages with git. So all I had to do was pick the right branch and restart.

I used `flycheck-set-indication-mode` which is a new function added in this change, so I'm pretty sure installation is not the problem.

cpitclaudel:
> I used flycheck-set-indication-mode which is a new function added in this change, so I'm pretty sure installation is not the problem.

Yup, that sounds right :) Can you put the point on one of the error underlines and press `C-u C-x =` (or any equivalent, like `M-: (what-cursor-position t)`)? I see this here:

```
There is an overlay here:
 From 433 to 444
  before-string        [Show]
  category             flycheck-error-overlay
  flycheck-error       [Show]
  flycheck-error-index 7
  flycheck-overlay     t
  help-echo            flycheck-help-echo
```

And the before-string is this:

```
#("!" 0 1
  (display
   ((margin left-margin)
    #("»" 0 1
      (face
       (flycheck-fringe-error default))))))
```

What do you get?

(thanks a lot for your help btw :)

doc/user/error-reports.rst
cpitclaudel: Good catch, thanks a lot! I squashed this into a previous commit.

Alexander-Miller:
Overlay is present:

```
There is an overlay here:
 From 896 to 897
  before-string        [Show]
  category             flycheck-error-overlay
  flycheck-error       [Show]
  flycheck-error-index 2
  flycheck-overlay     t
  help-echo            flycheck-help-echo
```

Before-string value is
```
#("!" 0 1
  (display
   ((margin left-margin)
    "")))
```

Looks like the display property is not set correctly.

cpitclaudel:
> Looks like the display property is not set correctly.

Indeed, looks like the error level doesn't define a margin spec.  Are you using another package on top of Flycheck, like LSP maybe?

Can you show what the flycheck-error of the overlay is?  And can you show what `M-: (flycheck-error-level-margin-spec 'warning)` returns?

cpitclaudel:
I just pushed a new commit that sets a default margin spec if none is present in calls to `define-error-level`.  If my guess above is correct, that should fix the issue.

Alexander-Miller:
> Are you using another package on top of Flycheck, like LSP maybe?

It wasn't loaded, and I had nothing else in the fringe/margin.

> If my guess above is correct, that should fix the issue.

It is, I'm seeing the margin indicator with the latest changes :+1: 

> Can you show what the flycheck-error of the overlay is?

I assume this question is obsolete now, or do you still want this info?

cpitclaudel:
>  It is, I'm seeing the margin indicator with the latest changes +1

Great!

> I assume this question is obsolete now, or do you still want this info? 

I think we're good :)

kujohn:
How does one setup emacs so `git-gutter` and `flycheck` can exist? Currently no matter what margin I set, git-gutter always overrides the flycheck indicators.

cpitclaudel:
@kujohn It's a pain: the symbols are there, but there's no notion of priority between them, they just get displayed one after the other, so you only see them if you make the margins wider.  I made a bug report about this on the Emacs tracker but it went nowhere.

laurynas-biveinis:
@kujohn as a workaround it is possible to replace git-gutter with git-gutter-fringe

Issue #1741, opened 2020-04-23
glitch at top and bottom of left arrow in fringe double arrows


tspiteri:
The bitmap contains two lines `#b10011000`, one for the first and one for the last non-zero entries, but they should be `#b11011000`.

cpitclaudel:
It's hard to tell whether this was intended or not, so I'm fine with changing it :) I did that as part of a branch that I'm working on: https://github.com/flycheck/flycheck/pull/1742

Do you want to give it a try, btw? I'm adding support for showing indicators in the margins instead of the fringe.

tspiteri:
Thanks! I'm going to stick to the fringes though, I have just spent quite a bit of time earlier today making the fringe look good for me on HiDPI by creating larger bitmaps of all vanilla Emacs bitmaps, the flycheck bitmap, and the magit bitmaps, and now I'm not going to change to margins :) I only saw the glitch when I was drawing a 16x16 version of the width-8 flycheck bitmap.

(On my system I just assume I know whether it is HiDPI so I don't have any HiDPI smart detection. On my non-HiDPI system I just don't require the bitmap modifying file.)

cpitclaudel:
> Thanks! I'm going to stick to the fringes though, I have just spent quite a bit of time earlier today making the fringe look good for me on HiDPI

That sounds great.  Maybe you'd like to pick up the stalled PR at https://github.com/flycheck/flycheck/pull/1553 ? ;)

tspiteri:
I actually did have a look at #1553, but the comments there talk about detecting HiDPI support. I even dug up the emacs-devel thread mentioned there, and in that thread there was even talk about moving a frame across monitors with different resolutions dynamically. What I did is nothing of the sort: on machines I know I always use in HiDPI mode I change the bitmaps, while on others I don't change the bitmaps.

A hack that could work that I just thought of: flycheck could detect on startup whether `(frame-parameter nil 'left-fringe)` is 16, and if it is 16 then it sets the bitmap to a larger version. This would probably not cause any issues, as it only means that people who themselves customized their fringe to be 16 pixels wide get the hi-res version. I could implement that, but not anything more complex.

cpitclaudel:
> A hack that could work that I just thought of: flycheck could detect on startup whether (frame-parameter nil 'left-fringe) is 16, and if it is 16 then it sets the bitmap to a larger version.

[...]

PR #1744, opened 2020-04-24
Use hi-res fringe bitmap when window's left fringe is 16 pixels wide

tspiteri:
This PR implements [an idea mentioned in issue 1741](https://github.com/flycheck/flycheck/issues/1741#issuecomment-618966444) using [the implementation suggested in the same issue](https://github.com/flycheck/flycheck/issues/1741#issuecomment-619039532).

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3952,23 +3952,32 @@ show the icon."
 ;;; Built-in error levels
 (when (fboundp 'define-fringe-bitmap)
   (define-fringe-bitmap 'flycheck-fringe-bitmap-double-arrow
-    (vector #b00000000
-            #b00000000
-            #b00000000
-            #b00000000
-            #b00000000
-            #b10011000
-            #b01101100
-            #b00110110
-            #b00011011
-            #b00110110
-            #b01101100
-            #b10011000
-            #b00000000
-            #b00000000
-            #b00000000
-            #b00000000
-            #b00000000)))
+    [#b11011000
+     #b01101100
+     #b00110110
+     #b00011011
+     #b00110110
+     #b01101100
+     #b11011000
+     #b00000000])
+  (define-fringe-bitmap 'flycheck-fringe-bitmap-double-arrow-hi-res
+    [#b0000000000000000
+     #b0000000000000000
+     #b1111001111000000
+     #b0111100111100000
+     #b0011110011110000
+     #b0001111001111000
+     #b0000111100111100
+     #b0000011110011110
+     #b0000011110011110
+     #b0000111100111100
+     #b0001111001111000
+     #b0011110011110000
+     #b0111100111100000
+     #b1111001111000000
+     #b0000000000000000
+     #b0000000000000000]
+    nil 16))
 
 (setf (get 'flycheck-error-overlay 'face) 'flycheck-error)
 (setf (get 'flycheck-error-overlay 'priority) 110)
--- a/flycheck.el
+++ b/flycheck.el
@@ -3872,11 +3872,12 @@ The following PROPERTIES constitute an error level:
      property for error level categories is `priority', to
      influence the stacking of multiple error level overlays.
 
-`:fringe-bitmap BITMAP'
+`:fringe-bitmap BITMAPS'
      A fringe bitmap symbol denoting the bitmap to use for fringe
-     indicators for this level.  See Info node `(elisp)Fringe
-     Bitmaps' for more information about fringe bitmaps,
-     including a list of built-in fringe bitmaps.
+     indicators for this level, or a cons of two bitmaps (one for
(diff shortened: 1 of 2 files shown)


flycheck.el
```diff
@@ -3872,11 +3872,13 @@ The following PROPERTIES constitute an error level:
      property for error level categories is `priority', to
      influence the stacking of multiple error level overlays.
 
-`:fringe-bitmap BITMAP'
-     A fringe bitmap symbol denoting the bitmap to use for fringe
-     indicators for this level.  See Info node `(elisp)Fringe
-     Bitmaps' for more information about fringe bitmaps,
-     including a list of built-in fringe bitmaps.
+`:fringe-bitmaps BITMAPS'
```
cpitclaudel, line 3875: Let's keep the keyword as `:fringe-bitmap`, since other code out there depends on this API.   Also, we should continue handling the non-cons case.  Maybe something like this: *A fringe bitmap symbol denoting the bitmap to use for fringe indicators for this level, or a cons of two bitmaps (one for narrow fringes and one for wide fringes).*

flycheck.el
```diff
@@ -3893,8 +3895,8 @@ The following PROPERTIES constitute an error level:
         (plist-get properties :compilation-level))
   (setf (get level 'flycheck-overlay-category)
         (plist-get properties :overlay-category))
-  (setf (get level 'flycheck-fringe-bitmap-double-arrow)
-        (plist-get properties :fringe-bitmap))
+  (setf (get level 'flycheck-fringe-bitmaps-double-arrow)
+        (plist-get properties :fringe-bitmaps))
```
cpitclaudel, line 3899: Here we should check whether the value returned by `plist-get` is a cons, and if not create a cons out of it.

flycheck.el
```diff
@@ -3916,9 +3918,13 @@ The following PROPERTIES constitute an error level:
   "Get the overlay category for LEVEL."
   (get level 'flycheck-overlay-category))
 
-(defun flycheck-error-level-fringe-bitmap (level)
-  "Get the fringe bitmap for LEVEL."
-  (get level 'flycheck-fringe-bitmap-double-arrow))
+(defun flycheck-error-level-fringe-bitmap (level &optional hi-res)
+  "Get the fringe bitmap for LEVEL.
+
+Optional argument HI-RES non-nil means that the retuned bitmap
```
cpitclaudel, line 3924: Typo: `retuned`

flycheck.el
```diff
@@ -3916,9 +3918,13 @@ The following PROPERTIES constitute an error level:
...
+will be the high resoluton version."
```
cpitclaudel, line 3925: Typo: `resoluton`

flycheck.el
```diff
@@ -3945,30 +3951,40 @@ show the icon."
     (error "Invalid fringe side: %S" side))
   (propertize "!" 'display
               (list side
-                    (flycheck-error-level-fringe-bitmap level)
+                    (flycheck-error-level-fringe-bitmap level
+                                                        (= (car (window-fringes)) 16))
```
cpitclaudel, line 3955: This should be `car` or `cadr` depending on whether `side` is `left-fringe` or `right-fringe`:

```
(= 16  (pcase side (`left-fringe (car (window-fringes))) (`right-fringe (cadr (window-fringes)))))
```

flycheck.el
```diff
@@ -3945,30 +3951,40 @@ show the icon."
...
+  (define-fringe-bitmap 'flycheck-fringe-bitmap-double-arrow-hi-res
+    [#b0000000000000000
+     #b0000000000000000
+     #b1111001111000000
+     #b0111100111100000
+     #b0011110011110000
+     #b0001111001111000
+     #b0000111100111100
+     #b0000011110011110
+     #b0000011110011110
```
cpitclaudel, line 3979: No pointy end?

tspiteri: I had tried that, but I didn't like the look.

flycheck.el
```diff
@@ -3977,7 +3996,8 @@ show the icon."
   :severity 100
   :compilation-level 2
   :overlay-category 'flycheck-error-overlay
-  :fringe-bitmap 'flycheck-fringe-bitmap-double-arrow
+  :fringe-bitmaps (cons 'flycheck-fringe-bitmap-double-arrow
```
cpitclaudel, line 3999: Shouldn't these be `:fringe-bitmap` now?

flycheck.el
cpitclaudel: Ok, I tried it and I'm convinced :)

flycheck.el
cpitclaudel: That was an easy fix, so I went ahead and pushed it.  I also updated the docs.

PR #1747, opened 2020-04-25
Restore the 'flycheck-fringe-bitmap-double-arrow property

cpitclaudel:
Reported-by: @tspiteri

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3897,6 +3897,9 @@ The following PROPERTIES constitute an error level:
   (setf (get level 'flycheck-fringe-bitmaps)
         (let ((bitmap (plist-get properties :fringe-bitmap)))
           (if (consp bitmap) bitmap (cons bitmap bitmap))))
+  ;; Kept for compatibility
+  (setf (get level 'flycheck-fringe-bitmap-double-arrow)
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -3897,6 +3897,9 @@ The following PROPERTIES constitute an error level:
   (setf (get level 'flycheck-fringe-bitmaps)
         (let ((bitmap (plist-get properties :fringe-bitmap)))
           (if (consp bitmap) bitmap (cons bitmap bitmap))))
+  ;; Kept for compatibility
+  (setf (get level 'flycheck-fringe-bitmap-double-arrow)
+        (car (get level 'flycheck-fringe-bitmaps)))
```
tspiteri, line 3902: I tested this and the new lsp error levels are narrow while the other errors which lsp leaves unchanged are wide. Why not set it to the whole cons instead of the car?

cpitclaudel: Yup, that's expected.  We don't know what code out there might be doing with this property, so leaving it as a symbol is safer, I think.

cpitclaudel:
We should do a separate PR to lsp-mode to adjust their bitmaps, I think.

PR #1748, opened 2020-04-26
Fix typos

tarsius:


cpitclaudel:
Thanks a lot; pushed as https://github.com/flycheck/flycheck/commit/be66bb78b2da7682abace939b585cfcc02e09fcd (with proper authorship tracking of course). I only kept the `flycheck.el` changes though, since the files in test/ are copied from various upstream projects and I'm not sure it's worth touching them.  WDYT?

flycheck.el
```diff
@@ -7165,7 +7165,7 @@ information about warnings."
 (defun flycheck-c/c++-quoted-include-directory ()
   "Get the directory for quoted includes.
 
-C/C++ compiles typicall look up includes with quotation marks in
+C/C++ compiles typically look up includes with quotation marks in
```
cpitclaudel, line 7168: Good catch, and in fact the "compiles" next to it is a typo as well :)

tarsius:
> since the files in test/ are copied from various upstream projects and I'm not sure it's worth touching them.

If you plan to import newer versions of these files at a later time, then it would probably become a problem if you fixed these typos. If you don't plan to do that, then that makes it a little more desirable to fix these typos, IMO.

cpitclaudel:
OK, I'm sold :) I pushed the remaining fixes as 246e1d4. Thanks a lot!

PR #1749, opened 2020-04-29
Don't sort errors after receiving them from checkers

cpitclaudel:
The error list already sorts errors: no need to redo the work in
flycheck-report-current-errors.  Additionally, the order returned by checkers is
significant in some cases, as when GCC puts notes next to the error they relate
to.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,14 @@
-32-cvs (in development)
+33-cvs (in development)
 =======================
+
+- **Breaking changes**
+
+  - The variable ``flycheck-current-errors`` now contains errors in the order in
+    which they were returned by checkers.  In previous versions of Flycheck,
+    this list was sorted by error position and severity. [GH-1749]
+
+32-cvs (frozen on May 3rd, 2020)
+================================
 
 - Highlights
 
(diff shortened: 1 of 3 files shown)


fmdkdd:
This warrants a changelog entry though.

cpitclaudel:
Good point, I updated the changelog.  I'll merge in a few days,

cpitclaudel:
@yyoncho, just making sure that you don't depend on flycheck-current-errors being sorted before I merge this.

yyoncho:
> flycheck-current-errors

We dont use this function and if this is the only thing that is changed we won't be affected. Generally, the lsp errors might not be sorted since we also act as a hub and we might report errors from several language servers. We could do that sorting on our side if you think that this will improve UE.

cpitclaudel:
> We dont use this function and if this is the only thing that is changed we won't be affected.

Perfect, thanks

> We could do that sorting on our side if you think that this will improve UE.

Nope, no need I think.  Basically we were re-sorting errors every time we got results from a new checker, and sorting them again in the error list — that's unnecessary and it was costly.

I've been working on improving the error list, btw, to make it possible to have distinct error lists backends.  I'll let you know when it's ready, in case you want to have a look (I think it might be a step towards adressing the problems you raised  a while ago)

Issue #1192, opened 2017-01-06
Allow custom actions in error explainers


marsam:
Currently [flycheck-error-list-explain-error](https://github.com/flycheck/flycheck/blob/e02ec2a2/flycheck.el#L4163-L4173) calls `flycheck-display-error-explanation` with the output of the checker's error-explainer, but there are checkers which doesn't provide a command to see the description of a error code.

for example `shellcheck` wiki has the description of its error codes, so I want to make something like:

```diff
modified   flycheck.el
@@ -9124,6 +9124,10 @@ flycheck-shellcheck-supported-shells
   (lambda (errors)
     (flycheck-remove-error-file-names
      "-" (flycheck-dequalify-error-ids errors)))
+  :error-explainer
+  (lambda (err)
+    (-when-let (error-code (flycheck-error-id err))
+      (browse-url (format "https://github.com/koalaman/shellcheck/wiki/%s" error-code))))
   :modes sh-mode
   :predicate (lambda () (memq sh-shell flycheck-shellcheck-supported-shells))
   :verify (lambda (_)
```

or with the `eslint` checker:

```diff
modified   flycheck.el
@@ -7669,6 +7669,10 @@ flycheck-eslint-config-exists-p
                      (flycheck-error-message err))))
             (flycheck-sanitize-errors errors))
     errors)
+  :error-explainer
+  (lambda (err)
+    (-when-let (error-code (flycheck-error-id err))
+      (browse-url (format "http://eslint.org/docs/rules/%s" error-code))))
   :enabled (lambda () (flycheck-eslint-config-exists-p))
   :modes (js-mode js-jsx-mode js2-mode js2-jsx-mode js3-mode rjsx-mode)
   :next-checkers ((warning . javascript-jscs))
```

If you think is a good idea, I can prepare a pull request

fmdkdd:
Yes and yes!  When we added the error-explainer it was mostly with rust in mind, but we made it generic in the hope that other checkers could benefit.  I opened #1134 to discuss that, but it never got traction.

Please do open up a PR 👍

cpitclaudel:
Hi @marsam , do you want to have a new look at this? I think your proposal was quite sound, and the PR should be rather simple (if the error explainer returns nil, we don't need to show the output)

PR #1753, opened 2020-05-02
Allow custom error explainers

cpitclaudel:
This PR has two parts: cleaning up incorrect uses of call-process (typically, by calling it without consulting the checker executable variable), and adding support for URLs as requested in #1192 (cc @marsam)


cpitclaudel:
@marsam good idea, I revised the docs to add an example of an `:error-explainer` and document other properties that we hadn't documented up to this point.

flycheck.el
```diff
@@ -9547,7 +9587,12 @@ See URL `https://eslint.org/'."
        (flycheck-verification-result-new
         :label "config file"
         :message (if have-config "found" "missing or incorrect")
-        :face (if have-config 'success '(bold error)))))))
+        :face (if have-config 'success '(bold error))))))
+  :error-explainer
+  (lambda (err)
+    (let ((error-code (flycheck-error-id err))
+          (url "https://eslint.org/docs/rules/%s"))
+      (and error-code `(url . ,(format url error-code))))))
```
marsam, line 9595: I think we should skip non-builtin rules:
```suggestion
      (unless (cl-find ?/ error-code) ; skip non-builtin rules
        `(url . ,(format url error-code))))))
```

flycheck.el
```diff
@@ -10443,16 +10494,12 @@ expression, which selects linters for lintr."
...
-    (let ((process-environment (append '("LC_ALL=C") process-environment)))
-      (call-process R nil t nil
-                    "--slave" "--restore" "--no-save" "-e"
-                    "library('lintr')")
-      (goto-char (point-min))
-      (not (re-search-forward "there is no package called 'lintr'"
-                              nil 'no-error)))))
+(defun flycheck-r-has-lintr (checker)
+  "Whether CHECKER (R) has installed the `lintr' library."
+  (eq 0 (flycheck-call-checker-process
```
marsam, line 10499: nitpick:
```suggestion
  (zerop (flycheck-call-checker-process
```

cpitclaudel: :thinking: I think that will raise an error if the return value is not a number, but call-process can return strings (and our wrapper can return nil, as well)

cpitclaudel: That being said, eql would be better than eq here. Changed.

flycheck.el
cpitclaudel: Good point, committed.

cpitclaudel: I did keep the `error-code` test though, because we don't want to display a page when the error code is nil.

cpitclaudel:
A plague on github's "commit suggestion" — why doesn't it credit the author properly?

PR #1755, opened 2020-05-04
Improve the UI for disabled checkers

cpitclaudel:
This makes it easier to re-enable a disabled checker or to reset an automatically disabled one.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 33-cvs (in development)
 =======================
 
+- New features and improvements
+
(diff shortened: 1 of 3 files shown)


tarsius:
This results in a new warning:

```
Compiling /home/jonas/.emacs.d/lib/flycheck/flycheck.el...

In toplevel form:
lib/flycheck/flycheck.el:2465:43: Warning: ‘flycheck-disable-checker’ is for
    interactive use only; Directly set ‘flycheck-disabled-checkers’ instead
```

PR #1756, opened 2020-05-04
Tweak the modeline lighter and document it

cpitclaudel:
We can't put a help-echo on the lighter (it's overridden by the default help-echo), so at least make sure that our docstring includes useful info.


cpitclaudel:
> Bold move, to change the default '/' ;)

Yes, a momentous keystroke.  I have more changes to discuss for 33 :)

> The documentation is superb. I'm wondering though if we don't have too many "something went wrong" symbols?

Way too many ^^… but they do have the advantage of being unambiguous, at least.

PR #1760, opened 2020-05-09
Only install external major modes on CI when testing Emacs master

cpitclaudel:
External major modes break all the time, causing spurious failures when building pull requests.  Testing major modes only on master allows us to investigate breakage while reducing spurious failures (master is allowed to fail).


fmdkdd:
Hmm, I guess that we don't strictly need the major modes for the integration tests.  We just need flycheck to be on and to call the correct checker.
It seems that you needed to mock some mode-specific variables though, is that it?

cpitclaudel:
> Hmm, I guess that we don't strictly need the major modes for the integration tests. We just need flycheck to be on and to call the correct checker.

Exactly.  It's still good to test with the real thing, but at least now we do that test in a branch that's allowed to fail, so a PR for a new julia checker doesn't break because racket-mode is broken on MELPA or because xyz-mode decided to drop compatibility with Emacs < 123.4

> It seems that you needed to mock some mode-specific variables though, is that it?

That's it.  Which isn't too bad, because it shows that we were missing a few defvars ^^

bbatsov:
Closing — this predates the move to Eask and GitHub Actions, so it no longer applies to the current CI setup.

PR #1764, opened 2020-05-09
Improve the documentation of set-indication-mode

mssdvd:
I also updated the left-fringe width to 0 px. Is there any reason why it was set to 1?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -593,7 +593,7 @@ MODE is 'left-margin the left-fringe is hidden to save space."
      (setq left-fringe-width 8 right-fringe-width 8
            left-margin-width 0 right-margin-width 0))
     (`left-margin
-     (setq left-fringe-width 1 right-fringe-width 8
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Very nice, thanks a lot.  I left to 1px originally because otherwise there's no visual distinction between the fringe and the margin (I wish there was a way to change the margin's background color, to make ti more visually distinct; maybe we should request this on the Emacs side and change to 0px once it's implemented?)

mssdvd:
Yeah, sure!

cpitclaudel:
Perfect.  I merged the first commit into flycheck-32 and master from the command line, with a small tweak (`left-margin` wasn't quoted).

cpitclaudel:
I'll open a feature request on the Emacs tracker and then I think we're good :)

cpitclaudel:
The feature request is at https://debbugs.gnu.org/cgi/bugreport.cgi?bug=26909 ; there was already one, so I added an extra message.

PR #1768, opened 2020-05-13
Update "known Windows issues" link

chwarr:
It looks like the label has changed from "B-Windows only" to "arch:
windows only".


cpitclaudel:
Thanks a lot for catching the issue and taking the time to fix the link :)

Issue #1758, opened 2020-05-06
pylint false positive attempted relative import beyond top-level package


pedroqueiroga:
This template is for **user support**.  If you are reporting an actual Flycheck bug, please use the relevant issue template instead.

## Checklist
- [x] I have searched existing issues and StackOverflow for solutions.
- [x] I have read the [Flycheck manual][].
- [x] I have read the [Troubleshooting Guide][].
- [x] I have checked my issue has not been reported yet.

## Problem description
python-pylint reports [relative-beyond-top-level] even though I can execute my flask project just fine. If I use pylint alone in the command line it works as expected (no relative-beyond-top-level error), and if I issue the same command as `C-c ! C-c python-pylint` it works as expected too.

example directory tree:
```
├── entry.py
└── pkg
    ├── bar.py
    ├── foo.py
    └── __init__.py
```

foo.py:
```python3
def hello_world():
    return 'Hello World'
```

bar.py:
```python3
from .foo import hello_world


def main():
    assert hello_world() == 'Hello World!'
    return True
```

entry.py:
```python3
from pkg.bar import main

print(main())
```

## Steps to reproduce
1. Open file 'pkg/bar.py'
2. Check current buffer (done automatically) 
3. pylint falsely reports error.

- [x] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

## Expected behavior
No error.

## System configuration
```
Syntax checkers for buffer bar.py<pkg> in python-mode:

First checker to run:

  python-flake8
    - may enable:         yes
    - executable:         Found at /home/pedro/testvenv/bin/python3
    - configuration file: Found at "/home/pedro/test/setup.cfg"
    - `flake8' module:    Found at "/home/pedro/testvenv/lib/python3.7/site-packages/flake8/__init__.py"
    - next checkers:      python-pylint, python-mypy

Checkers that may run as part of the first checker's chain:

  python-pylint
    - may enable:         yes
    - executable:         Found at /home/pedro/testvenv/bin/python3
    - configuration file: Found at "/home/pedro/test/.pylintrc"
    - `pylint' module:    Found at "/home/testvenv/lib/python3.7/site-packages/pylint/__init__.py"
    - next checkers:      python-mypy

  python-mypy
    - may enable:         yes
    - predicate:          t
    - executable:         Found at /home/pedro/testvenv/bin/mypy
    - configuration file: Found at "/home/pedro/test/setup.cfg"

[...]

PR #1771, opened 2020-05-14
Add a :working-directory property to python-pylint

cpitclaudel:
Closes GH-1758.  The algorithm is the same as used by epylint: find the first directory that doesn't have a ``__init__.py`` file.

CC @thisch

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -10528,6 +10528,18 @@ which should be used and reported to the user."
                :filename .path)))
           (car (flycheck-parse-json output))))
 
+(defun flycheck-pylint-find-project-root (_checker)
+  "Find the directory to invoke pylint from.
+
+The algorithm is the same as used by epylint: find the first
+directory that doesn't have a __init__.py file."
+  (locate-dominating-file
+   (if buffer-file-name
+       (file-name-directory buffer-file-name)
+     default-directory)
+   (lambda (dir)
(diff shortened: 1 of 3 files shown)


syl20bnr:
That's a great fix. What's preventing this from being merged ?

cpitclaudel:
One more attempt at this, this time looking for setup.cfg or pyproject.toml before falling back.

g-braeunlich:
Works for me! Thx!
On this occasion:

```lisp
(flycheck-def-config-file-var flycheck-pylintrc python-pylint ".pylintrc")
```
Cannot we change this into
```lisp
(flycheck-def-config-file-var flycheck-pylintrc python-pylint '(".pylintrc" "setup.cfg" "pyproject.toml"))
```
And also add `pyproject.toml` for mypy?
Also can open a new MR, if you prefer.

cpitclaudel:
Thanks for testing and yes, good point; I've added them.

g-braeunlich (approved):

dsedivec:
Thanks for this!  I think this would fix the problem I had with Mypy today.  Repro case, if it helps:

```
~ $ mkdir testpkg
~ $ cd testpkg
testpkg $ touch __init__.py
testpkg $ echo 'import logging' > logging.py
testpkg $ mypy logging.py
logging.py: error: Source file found twice under different module names: 'testpkg.logging' and 'logging'
Found 1 error in 1 file (errors prevented further checking)
```

Issue #1770, opened 2020-05-14
by using temp buffers, flycheck-python-run-snippet and friends circumvent buffer-local virtualenv configurations


glyph:
## Checklist
- [x] I have checked existing issues for potential duplicates before creating this one.
- [x] I have read the [Troubleshooting guide][].

## Bug description

flycheck is meant to "just work" with whatever virtualenv managing thing you have, by respecting `exec-path`.  But it doesn't!  If you've got virtualenvs associated with python buffers, you get `` `xxx' module: Missing; sys.path is… `` errors with a sys.path that looks like it came from the system python.  Unless you have that module installed globally or in your user dir, outside of your virtualenv...

## Steps to reproduce

```elisp
(let ((path-to-use (concat my-virtualenv "/bin"))
    (set (make-local-variable 'process-environment) (copy-tree process-environment))
    (set (make-local-variable 'exec-path) (cons path-to-use (copy-tree exec-path)))
    (setenv "PATH" (concat path-to-use ":" (getenv "PATH"))))

(flycheck-python-run-snippet 'python-flake8 "import sys; print(sys.executable)")
```

- [x] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

## Expected behavior
I expect to see `my-virtualenv/bin/python`; instead I see `/usr/bin/python3` or its moral equivalent.

The root cause here is the use of `with-temp-buffer`, which re-sets all of the buffer-local process-location state that's been carefully & correctly set for each python-mode buffer.

## System configuration

Emacs configuration:
- [x] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs
- [ ] Other shared configuration

## Additional notes
This check affects only the module-is-present check, not (I believe) the actual location of the executable to run so I've worked around it by `python3.8 -m pip install --user flake8`.

[Troubleshooting guide]:  https://www.flycheck.org/en/latest/user/troubleshooting.html

cpitclaudel:
Ouch, thanks so much for the report.  I have made a minimal fix for the flycheck-32 branch at https://github.com/flycheck/flycheck/pull/1772 , and a more extensive fix for master at https://github.com/flycheck/flycheck/pull/1773.  If you have a minute, it would be great if you could test either of these.

PR #1773, opened 2020-05-16
Ensure that flycheck-python-run-snippet is called in the right buffer

cpitclaudel:
Closes GH-1770.  This PR is a more extensive fix for master.


cpitclaudel:
Good point, I added a comment.

PR #1774, opened 2020-05-16
Run error explainers and the functions they return in the right buffers

cpitclaudel:
This is a follow-up to #1770.  Error explainers were run in an arbitrary buffer, but that isn't right: since they typically run processes, they need to be called in the buffer that the error came from, to make sure that they pick up the right executables, etc.

The tricky case is when an error explainer returns a function.  I chose to change the API (call with the errors' buffer current and expect the function to write to stdout), because the one I introduced initially was error-prone and was still very new.


flycheck.el
```diff
@@ -11231,7 +11234,8 @@ Relative paths are relative to the file being checked."
     (lambda ()
       (flycheck-call-checker-process
        'rust nil standard-output t "--explain" error-code)
-      (flycheck--fontify-as-markdown))))
+      (with-current-buffer standard-output
```
fmdkdd, line 11237: Even though this is documented, it would be better as an API if the programmer couldn't do the wrong thing here and forget to write to standard-output.

cpitclaudel: I agree. It's a tricky trade-off.

Either we use the output buffer as the current buffer: then it's trivial to write to the right buffer… but it's also trivial to run the checker process in the wrong buffer (like in #1770).  Or we run in the original buffer and bind standard-output: then it's trivial to run in the right buffer, but also trivial to write in the wrong one.

I think the second option is preferable, because errors are very easy to catch: text is going to appear in the wrong buffer.
On the other hand, option 1 leads to subtle mistakes, like #1770.

Issue #1776, opened 2020-05-17
flycheck-list-errors shows an empty list even when there are errors in the buffer


AndreaOrru:
## Checklist
- [x] I have checked existing issues for potential duplicates before creating this one.
- [x] I have read the Troubleshooting guide.

## Bug description
In Python buffers, with or without LSP enabled, running `flycheck-list-errors` shows an empty list even when there are errors in the buffer. However, weirdly, it's possible to use `flycheck-next-error` and `flycheck-previous-error` to navigate between errors.

## Steps to reproduce
1. Open a Python file.
2. Type something non-sensical to get an error.
3. Open `flycheck-list-errors`
4. Observe an empty list of errors where there should be some.

- [x] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

## Expected behavior
`flycheck-list-errors` should display the errors in the buffer instead of an empty list.

## Screenshots
![image](https://user-images.githubusercontent.com/3673952/82141413-a3cb7300-9870-11ea-8724-6f6af3604cd6.png)

## System configuration
```
Syntax checkers for buffer app.py in python-mode:

First checker to run:

  python-flake8 (explicitly selected)
    - may enable:         yes
    - executable:         Found at /home/andrea/.virtualenvs/gilgamesh/bin/python3
    - configuration file: Not found
    - `flake8' module:    Found at "/home/andrea/.virtualenvs/gilgamesh/lib/python3.8/site-packages/flake8/__init__.py"
    - next checkers:      python-pylint, python-mypy

Checkers that may run as part of the first checker's chain:

  python-mypy
    - may enable:         yes
    - may run:            t
    - executable:         Found at /home/andrea/.virtualenvs/gilgamesh/bin/mypy
    - configuration file: Not found

Checkers that could run if selected:

  lsp  select
    - may enable: yes
    - may run:    t

  python-pycompile  select
    - may enable:    yes
    - executable:    Found at /home/andrea/.virtualenvs/gilgamesh/bin/python3
    - next checkers: python-mypy

Checkers that are compatible with this mode, but will not run until properly configured:

  python-pylint (automatically disabled) reset
    - may enable:         no
    - executable:         Found at /home/andrea/.virtualenvs/gilgamesh/bin/python3
    - configuration file: Not found
[...]

PR #1778, opened 2020-05-17
Save the error list's source buffer before displaying the error list

cpitclaudel:
Closes GH-1776.


AndreaOrru:
Works for me! Brilliant.

cpitclaudel:
Thanks a lot for testing.  I think the link to the issue will provide the perfect summary.

PR #1789, opened 2020-05-27
Update flycheck.el

AppMathIsRealMath:
support verilator 4.030+ error/warning reporting


cpitclaudel:
I think this is covered by #1691 :) Thanks a lot!

PR #1792, opened 2020-05-28
Allow command checkers to specify environment for starting processes

doublep:
This simple change adds possibility for command-based checkers to specify environment (i.e. environment variables) in which their processes are started. This is completely optional and not used by any existing checker. I plan to eventually use this functionality in `flycheck-eldev`. In Eldev it is possible to switch Emacs binary you use, but not as a command line option (command line is parsed by Elisp code, so it's a bit too late), only by modifying an environment variable.

Verified by `make check specs unit`. Also verified that it actually affects process environment e.g. by using this locally:

    :environment (lambda (checker environment) (cons "ELDEV_EMACS=lol" environment))

After this Eldev couldn't launch, as expected.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5700,6 +5700,18 @@ of command checkers is `flycheck-sanitize-errors'.
      or as special symbol or form for
      `flycheck-substitute-argument', which see.
 
+`:environment FUNCTION`
+     A function to build process environment.
+
+     FUNCTION is a function taking two arguments: ENVIRONMENT and
+     CHECKER.  ENVIRONMENT will be initialized to the value of
+     standard variable `process-environment'.  Return value of
+     the function should be a list (possibly even empty) in the
+     same format as `process-environment'.
+
+     This property is optional.  If omitted, processes are
+     started in unmodified `process-environment'.
+
 `:error-patterns PATTERNS'
      A list of patterns to parse the output of the `:command'.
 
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Thanks, the patch is clean and simple.

However, I'm a bit unsure whether we actually need this, at least just yet, since no other checker makes use of this. It has come up a few times, but never really became needed.

(As to picking an emacs version, what I've done in esh is to use a minimal wrapper that calls an Emacs subprocess according the relevant environment variable, I think)

doublep:
> I'm a bit unsure whether we actually need this, at least just yet, since no other checker makes use of this.

Of course, no checker can make use of something that doesn't exist.

> It has come up a few times, but never really became needed.

Aha. But I guess each time a workaround like

> what I've done in esh is to use a minimal wrapper that calls an Emacs subprocess

was used instead of a "clean and simple" feature.

cpitclaudel:
> Of course, no checker can make use of something that doesn't exist.

This is not what I meant.  What I mean is that we've never have a case yet where this was really needed, I think.

> Aha. But I guess each time a workaround like … was used instead of a "clean and simple" feature.

No, not really, at least not in my memory (esh isn't a Flycheck checker)

Besides,

> what I've done in esh is to use a minimal wrapper that calls an Emacs subprocess

Isn't that what you do too? Otherwise, how do you use the environment variable?

bbatsov:
Closing this one after a long look. The patch was clean, but six years on no consumer materialized - including flycheck-eldev, which this was written for and which never ended up building the feature. flycheck-command-wrapper-function covers the env-injection case (and works over TRAMP, which a process-environment binding wouldn't), so I'd rather not carry an API with zero users. If a concrete checker ever needs it, happy to revisit - the right shape today is a small :env property threaded next to the LC_MESSAGES handling, which is about a half-day with specs.

Issue #1787, opened 2020-05-25
Let checkers possibly recover from `suspicious` state


doublep:
Current relevant code is:

    (when (and (not (equal exit-status 0)) (null errors))
      (funcall callback 'suspicious
               (format "bla-bla-bla"  checker exit-status output checker)))

This means that any checker that exits with a non-zero code and without errors triggers `suspicious` state.

`flycheck-eldev` extension heavily rewrites (in its advices) command line of `emacs-lisp` checker. If everything works fine, Emacs eventually gets to compile source file, just as the checker expects, and returns errors that are processed normally or exits with code zero. But if something doesn't work as expected, Flycheck currently blames `emacs-lisp` checker, which is actually not true, because it is `flycheck-eldev` that has caused the mess.

Proposal: let checker _optionally_ define `:handle-suspicious` callback (if not defined, behavior is exactly as now). This callback takes `exit-status` and `output` and returns a list of errors or a string. If return value is a list, `suspicious` status is cancelled and status becomes `finished` as normal, with given list of errors (possibly even empty). If it returns a string, that string is passed to the `callback` and the state becomes `suspicious`.

Additional rationale: I want `flycheck-eldev` require certain version of Eldev. With this proposal it could check available version only when something actually goes wrong and it tries to recover from `suspicous` state: if Eldev is too old, it would return a single error stating its requirement, otherwise fall back to the current behavior.

bbatsov:
Closing: feature request specific to `flycheck-eldev` extension's needs. Stale for 5+ years.

PR #1795, opened 2020-05-31
Allow checkers convert suspicious state to non-standard errors

doublep:
This implements what is proposed in issue #1787.

Rationale: Eldev (for example) can fail before even starting to byte-compile `.el` file. For example, because package headers are missing, because a dependency cannot be downloaded, because Eldev is too old — multitude of reasons. So, it can return nonzero exit status, but there are no byte-compilation warnings/errors in the output. Flycheck treats this as "suspicious", gives an ugly message and eventually disables the checker if it fails like this too often.

With the proposed change I would be able to add

    :handle-suspicious flycheck-eldev--suspicious-state-handler

to the checker definition, with the function being something like this:

    (defun flycheck-eldev--suspicious-state-handler (checker _exit-status output)
      (let ((message (string-trim output)))
        ;; Don't add clarification to a few obvious errors.
        (unless (string-match-p (rx bos "Dependency " (1+ any) " is not available") message)
          (setf message (concat message "\n\n" flycheck-eldev-general-error)))
        `(,(flycheck-error-new-at 1 1 'error message :checker checker))))

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5735,6 +5735,22 @@ of command checkers is `flycheck-sanitize-errors'.
      `flycheck-parse-with-patterns'.  In this case,
      `:error-patterns' is mandatory.
 
+`:handle-suspicious FUNCTION'
+     A function to handle suspicious state: when the process
+     returns non-zero code, but no standard errors (i.e. using
+     `:error-patterns' or `:error-parser') are found.
+
+     The function is called with three arguments: CHECKER,
+     EXIT-STATUS and OUTPUT (as string) with the checked buffer
+     as current.  It should process the output and return a list
+     of non-standard errors that best describe what exactly has
+     failed.  If it cannot do so, it should return symbol
+     `suspicious' to indicate that what has happened is really
+     not expected.
+
+     This property is optional.  If omitted, such state is always
+     treated as suspicious.
+
 `:standard-input t'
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Thanks, that's an interesting approach.  But why not do that on the eldev side?  Could you pass it a flag to say "please report all your errors with a line and a column"?

doublep:
> But why not do that on the eldev side? Could you pass it a flag to say "please report all your errors with a line and a column"?

Because the errors in question have nothing to do with the file Flycheck byte-compiles. They can happen even before the file is looked at. It would be strange to report errors like "foo.el:1: Error: Dependency ‘dashe’ 2.12.1 is not available" when I made a typo in `bar-pkg.el`.

You once mentioned a Haskell checker that can autoinstall dependencies. I'd guess it can also fail for reasons unrelated to the file being checked (when it fails to install dependencies) and then "suspicious" state is triggered.

bbatsov:
I've revived this in #2180 — rebased onto master, plus specs, docs and a changelog entry, with credit to you as co-author. Thanks for the patch and for the patience; the six-year review turnaround is not our proudest stat.

bbatsov:
Superseded by #2180, which landed this with you credited as co-author. Should have closed it then, sorry.

PR #1801, opened 2020-06-15
Insert string explanations into standard-output

### LINKED ISSUE #1800: Error explainer incorrectly writing text to the source code buffer
I'm observing some strange behaviour in a recent version of Flycheck when using the `clang-tidy` error explainer functionality in `lsp-mode.el`.

The problem I see is that the error explainer is writing text to the source code buffer. Inspecting recent PRs, I focused on this one: https://github.com/flycheck/flycheck/pull/1774. Here's what's happening:

1. Imagine an `:error-explainer` lambda that returns a string (https://github.com/flycheck/flycheck/pull/1774/files#diff-374cdd9563b826c712b1dc6a373a0bc1R2046). 

2. Flycheck gets the string from my error explainer at https://github.com/flycheck/flycheck/pull/1774/files#diff-374cdd9563b826c712b1dc6a373a0bc1R5319.

3. Flycheck then checks the type of the explanation it got. As it is a string, it evals this line: https://github.com/flycheck/flycheck/pull/1774/files#diff-374cdd9563b826c712b1dc6a373a0bc1R5614, which is using `insert`. This is basically inserting text in the current buffer, which is the source code buffer.

Is this expected behaviour, or I am missing something? Did you mean to print to the standard output instead (with `print` maybe)? To give you more context, our error explainer is a little bit more complex than the usual ones. It returns a "Loading..." string while we fetch a webpage asynchronously and update the error explainer buffer with a rendered webpage. See https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-cpp.el#L137

Thank you.

danielmartin:
Fixes #1800 

Previously, this was incorrectly inserting the explanation string into
the current buffer, which is the source code buffer where the checker
runs.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5624,7 +5624,7 @@ this error to produce the explanation to display."
            (flycheck-explain-error-mode))
          (cond
           ((functionp explanation) (funcall explanation))
-          ((stringp explanation) (insert explanation))
+          ((stringp explanation) (princ explanation))
           (t (error "Unsupported error explanation: %S" explanation)))
          (display-message-or-buffer standard-output nil 'not-this-window)))))
 


danielmartin:
Test failure `flycheck-define-checker/dockerfile-hadolint/error` seems not caused by these changes. I see there are ongoing changes in another in progress PR to fix this test: https://github.com/flycheck/flycheck/commit/83b1a4386be23381b2c8548970d5e64da7b78b5b

cpitclaudel:
Thanks! You're absolutely right about the test failure.  I plan to fix it today.  Thanks a lot for your patience.

cpitclaudel:
Ouch.  Sorry for dropping the ball on this.  Your work is very much appreciated, @danielmartin .

PR #1767, opened 2020-05-11
Parse optional end column in haskell-hlint

jimenezrick:
Sometimes hlint prints the end column of a hint:

  Main.hs:37:12: Suggestion: Redundant $
  Found:
    $
  Perhaps you should remove it.

  Main.hs:37:14-31: Warning: Redundant bracket
  Found:
    ("Running server")
  Perhaps:
    "Running server"

PR #1804, opened 2020-06-17
Support multi-line suggestions Haskell hlint-3.1.4

dzhus:
As of Hlint 3.1.4, it can produce multi-line multi-column suggestions:
```
src/Main.hs:(302,3)-(307,87): Warning: Redundant <$>
```
This is in addition to standard one-line comments such as
```
src/Main.hs:402:65-69: Warning: Use fmap
```

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9577,17 +9577,20 @@ See URL `https://github.com/ndmitchell/hlint'."
             source-inplace)
   :error-patterns
   ((info line-start
-         (file-name) ":" line ":" column (optional "-" end-column)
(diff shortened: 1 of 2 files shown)


cpitclaudel:
Excellent, thanks a lot! The tests are broken for other reasons, I'll get to that very soon.  See https://github.com/flycheck/flycheck/pull/1767 also for comments on which version of hlint we test

dzhus:
I opened this just to see if it passes some tests as I was too lazy to run them locally – I see the CI image uses `HLint v2.0.11` which doesn't need this feature. I'll test locally instead.

Issue #1805, opened 2020-06-19
Flycheck idle timer clobbers match data


darkfeline:
## Bug description
Flycheck adds a timer `flycheck--handle-idle-trigger` which clobbers match data.

## Steps to reproduce
1. Open scratch buffer
2. Repeatedly evaluate `(match-data)`
3. See how the match data changes

## Expected behavior
Timers should not clobber match data.

## System configuration
```
Syntax checkers for buffer *scratch* in lisp-interaction-mode:

First checker to run:

  emacs-lisp
    - may enable:    yes
    - may run:       t
    - executable:    Found at /usr/bin/emacs
    - next checkers: emacs-lisp-checkdoc

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled
checkers.

--------------------

Flycheck version: 32snapshot (package: 20200610.1809)
Emacs version:    26.3
System:           x86_64-pc-linux-gnu
Window system:    x

```

Emacs configuration:
- [x] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs
- [ ] Other shared configuration

## Additional notes

[Troubleshooting guide]:  https://www.flycheck.org/en/latest/user/troubleshooting.html

PR #1806, opened 2020-06-19
Don't clobber match data in timer

darkfeline:
Fixes https://github.com/flycheck/flycheck/issues/1805


cpitclaudel:
Thanks, excellent catch.  Maybe we should actually put this in `flycheck-buffer-automatically`, to catch similar errors that may pop up from elsewhere?

darkfeline:
I'm not sure. The Emacs manual states that all functions are allowed to change match data, so there's nothing specifically wrong with `flycheck-buffer-automatically` overwriting match data.  It's only timer functions that need to be careful.  (I don't know why Emacs doesn't automatically save match data for timers like it does for process filters.)

cpitclaudel:
> It's only timer functions

And change hooks, right? (we already have a save-match-data for that case).  The fact that we had a `save-match-data` in one of the callers but missed it in the other (idle-trigger) suggests to me that we should just harden `flycheck-buffer-automatically` once and for all

darkfeline:
Okay, sure.  I'm keeping the `save-match-data` in the timer and change hooks though, since I'm paranoid something may get added there in the future that changes match data again.

cpitclaudel:
OK, works for me :)

PR #1809, opened 2020-07-09
Update `haskell-hlint` checker to HLint-3.0

jdnavarro:
HLint v3.0 has been recently [released](https://neilmitchell.blogspot.com/2020/05/hlint-30.html).

- Now `.hlint.yaml` is the default configuration file. (I think this was already supported in the latest releases of 2.0).
- When there is a suggestion or warning it returns a `1` exit code by default. We have to pass `--no-exit-code` to avoid it.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -9527,7 +9527,7 @@ See URL `https://www.haskell.org/ghc/'."
   :next-checkers ((warning . haskell-hlint))
   :working-directory flycheck-haskell--ghc-find-default-directory)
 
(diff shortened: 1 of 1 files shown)


jdnavarro:
The source location [always](https://github.com/ndmitchell/hlint/blob/master/src/GHC/Util.hs#L51) includes the end of the column now. I removed the `optional`s in `:error-patterns`, and it works as expected, but couldn't figure out how to update the tests.

In any case, it still works when leaving the optional end of the column.

bbatsov:
Better late than never. :-)

PR #1811, opened 2020-07-16
Refactor: leverage cl-destructuring-bind

progfolio:
Avoid repeated `plist-get` assignments in favor of `cl-destructuring-bind`.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2095,19 +2095,11 @@ are mandatory.
 Signal an error, if any property has an invalid value."
   (declare (indent 1)
            (doc-string 2))
-  (let ((start (plist-get properties :start))
-        (interrupt (plist-get properties :interrupt))
-        (print-doc (plist-get properties :print-doc))
-        (modes (plist-get properties :modes))
-        (predicate (plist-get properties :predicate))
-        (verify (plist-get properties :verify))
-        (enabled (plist-get properties :enabled))
-        (filter (or (plist-get properties :error-filter) #'identity))
-        (explainer (plist-get properties :error-explainer))
-        (next-checkers (plist-get properties :next-checkers))
-        (file (flycheck-current-load-file))
-        (working-directory (plist-get properties :working-directory)))
-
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -2095,19 +2095,11 @@ are mandatory.
...
-        (explainer (plist-get properties :error-explainer))
-        (next-checkers (plist-get properties :next-checkers))
-        (file (flycheck-current-load-file))
-        (working-directory (plist-get properties :working-directory)))
-
+  (cl-destructuring-bind (&key enabled explainer interrupt modes next-checkers
+                               predicate print-doc start verify working-directory
+                               ((:error-filter filter) #'identity)
+                               &allow-other-keys)
+      properties
```
cpitclaudel, line 2102: This looks pretty good, but I think the original idea was to not depend too much on cl-lib.  Is there a `pcase-let` way to do this?

flycheck.el
```diff
@@ -4183,32 +4175,31 @@ The following PROPERTIES constitute an error level:
...
+    (setf (get level 'flycheck-margin-spec)
+          (or margin-spec
+              (flycheck-make-margin-spec
+               flycheck-default-margin-str
+               (or (get level 'flycheck-fringe-face) 'default))))
+    (setf (get level 'flycheck-margin-continuation)
+          (flycheck-make-margin-spec
+           flycheck-default-margin-continuation-str
+           (or (get level 'flycheck-fringe-face) 'default)))
+    (setf (get level 'flycheck-error-list-face) error-list-face)))
```
cpitclaudel, line 4202: I'm not sure it's a win to separate the put from the get.

flycheck.el
progfolio: Not that I'm aware of. I think pcase-let has a similar facility when extended by `map.el` but I do not have that much experience with it and IIRC it's more verbose and not as flexible as `cl-destructuring-bind`.

flycheck.el
progfolio: Fair enough

PR #1832, opened 2020-10-26
Disable `checkdoc' on Eldev files

doublep:
Don't run `checkdoc` on Eldev files. They are valid Lisp, but don't contain usual comments, so `checkdoc` will never be happy.

It is much easier and cleaner to do this here than in `flycheck-eldev` package. There is already special treatment for Cask.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8584,7 +8584,7 @@ This variable has no effect, if
         (setq flycheck-emacs-lisp-check-declare ,value)))))
 
 (defun flycheck--emacs-lisp-enabled-p ()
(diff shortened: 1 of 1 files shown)


fmdkdd:
Could you shed light on what an eldev file is?

doublep:
It's an alternative (to e.g. Cask) Elisp build tool.

https://github.com/doublep/eldev

https://github.com/flycheck/flycheck-eldev

fmdkdd (approved):

fmdkdd:
Okay, thank you.  Good patch!

Issue #1762, opened 2020-05-09
Correct way to chain checkers to lsp


matthuszagh:
## Checklist
- [ ] I have searched existing issues and StackOverflow for solutions.
- [ ] I have read the [Flycheck manual][].
- [ ] I have read the [Troubleshooting Guide][].
- [ ] I have checked my issue has not been reported yet.

## Problem description
What is the recommended way to chain major mode-specific checkers to checkers used in many major modes? For instance, for python I'm currently using the following:
```
(flycheck-add-next-checker 'lsp 'python-mypy)
```

This works, but invokes the python-mypy checker in non-python modes as well since lsp checker is used in many major modes. For instance, I use the clangd lsp server for C/C++ and when I run `flycheck-verify-setup` in a C file I see

```
Syntax checkers for buffer file.c in c-mode:

First checker to run:

  lsp (explicitly selected)
    - may enable:    yes
    - predicate:     t
    - next checkers: python-mypy
```

python-mypy isn't actually invoked here, which I can verify by attempting to select it explicitly with `flycheck-select-checker`. Nonetheless, this setup doesn't seem ideal.

Please forgive me if this is documented somewhere, I wasn't able to find it in the [documentation section on configuring checker chains](https://www.flycheck.org/en/latest/user/syntax-checkers.html#configuring-checker-chains), or in a google search. Please post a link if I missed something.

## Steps to reproduce
1. Open file '...'
2. Do '....'
3. See error

- [ ] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

## System configuration
```
Paste the output of `M-x flycheck-verify-setup` here.
```

Emacs configuration:
- [ ] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs
- [ ] Other shared configuration

## Additional notes

[Flycheck manual]:  https://www.flycheck.org/en/latest/user/troubleshooting.html
[Troubleshooting guide]:  https://www.flycheck.org/en/latest/user/troubleshooting.html

cpitclaudel:
I don't think you've missed anything — we just don't have a great story for the case of a checker being used by many modes.

I guess in the original spirit of Flycheck there should have been one lsp checker per language server, so more or less one per mode: python-lsp, clangd-lsp, etc.  That would work for LSP, but it doesn't help much for cases like a s

That would allow you to set the chains conditionally.

[...]

Issue #1660, opened 2019-12-27
[Develop] How to setup next-checker?


conao3:
## Checklist
- [x] I have searched existing issues and StackOverflow for solutions.
- [x] I have read the [Flycheck manual][].
- [x] I have read the [Troubleshooting Guide][].
- [x] I have checked my issue has not been reported yet.

## Problem description
How to best solution for adding **next** checker?

## Steps to reproduce
1. Define checker via `flycheck-define-checker` named `indnet-elisp`.
2. `add-to-list` `indent-elisp` to `flycheck-checkers` at last.
3. `flycheck-add-next-checker` `indent-elisp` to `'emacs-lisp` at last.
4. `M-x flycheck-buffer`. Seems like `indent-elisp` checker doesn't run.

## Expected behavior
Add `indent-elisp` as next checker for `emacs-lisp` to ensure run `indent-elisp`.

## Screenshots
None

## System configuration
```
Syntax checkers for buffer flycheck-indent.el in emacs-lisp-mode:

First checker to run:

  emacs-lisp
    - may enable:    yes
    - predicate:     t
    - executable:    Found at /home/conao/.local/bin/emacs
    - next checkers: emacs-lisp-checkdoc, emacs-lisp-package, indent-elisp

Checkers that may run as part of the first checker's chain:

  emacs-lisp-checkdoc
    - may enable:    yes
    - executable:    Found at /home/conao/.local/bin/emacs
    - next checkers: emacs-lisp-package

  emacs-lisp-package
    - may enable: yes

  indent-elisp
    - may enable: yes
    - executable: Found at /home/conao/.local/bin/emacs

Checkers that are compatible with this mode, but will not run until properly configured:

  emacs-lisp-elsa
    - may enable: yes
    - predicate:  nil
    - executable: Found at /home/conao/.local/tools/cask/bin/cask

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled
checkers.

--------------------

Flycheck version: 32snapshot (package: 20191125.1821)
Emacs version:    26.3
System:           x86_64-pc-linux-gnu
Window system:    x
```

Emacs configuration:
- [x] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs
- [ ] Other shared configuration

## Additional notes
I checked this issue by adding below advice and its output.

### checker
``` emacs-lisp
(flycheck-define-checker indent-elisp
  "A indent checker for Elisp."
  :command ("emacs"
[...]

PR #1836, opened 2020-11-13
WIP: Add customization point to retrieve next-checkers

damienrg:
This PR introduce a customization point to retrieve next checkers for a given checker based on `next-checkers-function` property.

Related to:
https://github.com/flycheck/flycheck/issues/1660
https://github.com/flycheck/flycheck/issues/1762


wyuenho:
This is an interesting approach. Can you give a couple of examples of how one would use this to say, chain other checkers to lsp-mode's checker? I imagine python, JS and go devs will run into this issue the most.

yyoncho:
@wyuenho you may wanna check - https://github.com/flycheck/flycheck/issues/1762#issuecomment-750458442

damienrg:
> Can you give a couple of examples of how one would use this to say, chain other checkers to lsp-mode's checker? I imagine python, JS and go devs will run into this issue the most.

Example with `lsp` and the `major-mode`:

```elisp
(defun my-flycheck-lsp-next-checkers (checker)
  (pcase major-mode
    ('sh-mode
     (pcase checker
       ('lsp '(sh-shellcheck sh-bash))
       ('sh-shellcheck '(sh-bash))
       ))
    ('web-mode '(javascript-eslint))
    )
  )

(defcustom my-lsp-flycheck-next-checkers-function nil
  "Variable to hold the function that returns flycheck next checkers for lsp."
  :type '(choice function (const nil)))

(make-variable-buffer-local 'my-lsp-flycheck-next-checkers-function)

;; in my case I only use this function for lsp checker
(defun my-lsp-flycheck-next-checkers (checker)
  "Function called by flycheck to find next checkers for CHECKER."
  (when (functionp my-lsp-flycheck-next-checkers-function)
    (funcall my-lsp-flycheck-next-checkers-function checker)))

(flycheck-set-next-checkers-function 'lsp #'my-lsp-flycheck-next-checkers)

(add-hook 'lsp-after-initialize-hook
          (lambda()
            (setq my-lsp-flycheck-next-checkers-function #'my-flycheck-lsp-next-checkers)))
```

damienrg:
With the previous code to have `(lsp foo bar)`, you would also have to do: `(flycheck-set-next-checkers-function 'foo #'your-function)` because the `flycheck` works.

With the new code, you should be able to do :

```elisp
(defun my-flycheck-lsp-next-checkers (checker)
  (pcase major-mode
    ('sh-mode
     (pcase checker
       ('lsp '(sh-shellcheck sh-bash))
       ('sh-shellcheck '(sh-bash))
       ))
    ('web-mode '(javascript-eslint))
    )
  )
  
  (add-hook 'lsp-after-initialize-hook
          (lambda()
            (setq flycheck-next-checkers-function #'my-flycheck-lsp-next-checkers)))
```

Anyway, I do not think this PR will land and you can use an advice as @yyoncho mentioned.

wyuenho:
Why won't this land? Buffer local checker chain sounds like a great idea to me.

cpitclaudel:
> Why won't this land? Buffer local checker chain sounds like a great idea to me.

Because none of the maintainers have found the time to review the code, unfortunately :/  I'm on the job market atm and finishing a PhD, so not much time for free software.

Additionally, what this PR does can already be achieved with advice; yyoncho posted sample code for that in his previous comment.  With that and the buffer-local obarray options, there's a family of possible design choices that need to be evaluated.

@damienrg , thanks a lot for the PR.  Have you had a look at the advice-based solution?  Is there a reason to prefer this patch over that?  Your example use case would work the same, just with `advice-add` instead of `(setq flycheck-next-checkers-function`.  WDYT?

damienrg:
> wyuenho:
> Why won't this land? Buffer local checker chain sounds like a great idea to me.

I have opened this PR to have feedback about the solution I propose to chain checkers and it is maybe not the right way to do it.

> cpitclaudel:
> @damienrg , thanks a lot for the PR. Have you had a look at the advice-based solution? Is there a reason to prefer this patch over that? Your example use case would work the same, just with `advice-add` instead of `(setq flycheck-next-checkers-function`. WDYT?

Yes it works the same way. The main difference is that advice-based solution means it is not an official feature of `flycheck`.


From a user point of view, I would like to be able to do either:
1. `(setq flycheck-use-checkers '(foo bar baz))`
2. `(setq flycheck-checkers-function #'my-fun)` where `my-fun` returns a list of checkers
3. `(flycheck-use-checkers '(foo bar baz))` that automatically creates the correct chain.


> With that and the buffer-local obarray options, there's a family of possible design choices that need to be evaluated.

I do not believe that changing properties of a checker is the way to go because the user has to change the properties on all the next checkers or `flycheck` should provide `(flycheck-use-checkers '(foo bar baz))`.

Issue #1816, opened 2020-08-05
Flycheck with Tramp over SSH not working for Python files


MemoWalk:
## Checklist
- [ x] I have checked existing issues for potential duplicates before creating this one.
- [ x] I have read the [Troubleshooting guide][].

## Bug description
Flycheck seems not to work correctly with Tramp Mode on remote files. My setup etc. is described in this  [post on Emacs Stackexchange](https://emacs.stackexchange.com/questions/59941/flycheck-trouble-with-tramp-and-python-no-module-named-ssh-home-ubuntu-te).

## Steps to reproduce

- [x ] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

Confer link.

## Expected behavior
Someone from the ´tramp-devel@gnu.org´ offered help (you may contact him, as I understood) and pointed out:

- "[...] it turns out [flycheck.el] doesn't support remote processes. It calls `call-process' (shall
be `process-file'), and `start-process' (shall be `start-file-process')."
- "[...] pylint.el doesn't call any external process. [...] For example, it
uses (aref (tramp-dissect-file-name filename) 3) [...]; [...] it shall use (file-local-name filename) .

## System configuration
Confer link.

Emacs configuration:
- [ x] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs
- [ ] Other shared configuration

## Additional notes

[Troubleshooting guide]:  https://www.flycheck.org/en/latest/user/troubleshooting.html

cpitclaudel:
Hi there, and thanks for the report!

We don't officially support checking over tramp, AFAIK (see https://github.com/flycheck/flycheck/issues/54), though I'd love for us to do this reliably.  Is this something you would be interested in working on?


Also, this part: *pylint.el doesn't call any external process*; can you point to that discussion? I didn't see it in your link.

albinus:
I might be able to help with Tramp integration. Since I'm not a flycheck user, I'd need some guidance. At least a short recipe for testing.

cpitclaudel:
Thanks a lot @albinus , your help would be much appreciated :)  We need someone to drive the process, and I'm swamped with work these days, but once we have a volunteer it should be quick and easy.

[...]

PR #1842, opened 2020-12-06
Fix running checkers on remote hosts via TRAMP.

carrete:
In response to https://github.com/flycheck/flycheck/issues/1816

This patch does three things:

* Update `flycheck-default-executable-find` to search remote hosts when
  `default-directory` is a remote directory. See also the documentation for
  [executable-find](https://www.gnu.org/software/emacs/manual/html_node/elisp/Locating-Files.html).

* Change uses of `call-process` and `start-process` to `start-file-process`
  which is the same as these two functions, but starts the process on a remote
  host when `default-directory` is a remote directory. See also the
  documentation for
  [start-file-process](https://www.gnu.org/software/emacs/manual/html_node/elisp/Asynchronous-Processes.html).

* Passes the local part of `buffer-file-name` to the compile command,
  e.g. strips `/ssh:10.0.0.42`, since the checker is run on the remote machine
  where the file name must be local. See also the documentation for
  [file-local-name](https://www.gnu.org/software/emacs/manual/html_node/elisp/Magic-File-Names.html).


flycheck.el
```diff
@@ -6683,22 +6683,15 @@ shell execution."
...
-         (abs-prog
-          ;; The executable path returned by `flycheck-command-wrapper-function'
-          ;; may not be absolute, so expand it here.  See URL
-          ;; `https://github.com/flycheck/flycheck/issues/1461'.
-          (or (executable-find (car wrapped))
-              (user-error "Cannot find `%s' using `executable-find'"
-                          (car wrapped))))
          (command (mapconcat #'shell-quote-argument
-                             (cons abs-prog (cdr wrapped)) " ")))
+                             (cons program (cdr wrapped)) " ")))
```
carrete, line 6690: This appears to be obsolete. `flycheck-find-checker-executable` calls `flycheck-executable-find` which calls `executable-find` which returns an absolute path.

carrete:
I'll do something about the Travis CI errors

cpitclaudel:
:+1: This is very exciting stuff.  I'm swamped with work right now, but I should have time to look at this later this month.  Hope that's OK!

carrete:
> :+1: This is very exciting stuff.  I'm swamped with work right now, but I should have time to look at this later this month.  Hope that's OK!

Absolutely. I still have to get the tests to pass. I get different results locally 🤔

carrete:
First of all, this only works with Emacs 27.1, the first version to include the second, optional `remote` option to `executable-find`. The `advice` approach is the best I could do because even when `(executable-find executable t)` is behind a version check byte-compilation fails.

Please see my comment about `start-file-process` when let bound. I think it's true, though I'm not certain.

I don't know what's up with the tests. These https://gist.github.com/carrete/8ab881bc3ff1a9c20285f9edf6ce8ac6 are my local test results for Emacs 26.3 and 27.1 which both behave the same. I couldn't get Emacs 25.3 to work.

I have no idea why `flycheck-set-checker-executable/file-not-executable` fails. Apparently warnings.el is seen as executable even though its file permissions say it isn't. I see a recent pull-request, https://travis-ci.org/github/flycheck/flycheck/builds/741547489, that has the same spec failure so I assume the changes in this pull-request aren't to blame. The builds on Travis CI behave differently even though I used [docker-tools](https://github.com/flycheck/docker-tools) and the same build and test commands in .travis.yml. I don't know why the results are different.

I've been using this version in my day job the past week and haven't noticed any problems. At this point I'm not sure what I should try next. Any ideas would be greatly appreciated, @cpitclaudel. No rush. Thanks!

carrete:
I pushed up an additional change that updates the built-in checkers. Sorry, I shouldn't have submitted the pull-request without this. I was distracted by the odd test failures and forgot about it. I still don't get the same results as on Travis CI. The tests behave better there than they do on my laptop. As for the remaining test failures on Travis CI, I believe these are because `source` (possibly more) needs to be tweaked somehow too before it's substituted in the checker definitions.

Blade6570:
Hey, why not replace all `executable-find` to `executable-find-add-remote`? After I replaced it, I did not get any error.

carrete:
@Blade6570 Do you mean tests passed? Provided I understand the issue correctly, the file name of the buffer must also be localized.

yfyyfy:
Based on this pull request, I made further changes so that remote check works on my environment.
I also added support for eslint and stylelint.
Please see #1922 for the details.

carrete:
I needed this recently so I took the opportunity to rebase this on top of the latest upstream, plus add the changes by @yfyyfy to fix the stylelint tests. With these and the move to GitHub Actions the tests now pass.

peterbecich:
Can this be merged? It would help me significantly. Thank you

flycheck.el
```diff
@@ -510,6 +510,20 @@ sandboxes."
...
   :risky t)
 
+;; This can be removed when Emacs 27.1 is the oldest supported version.
+;; See https://github.com/flycheck/flycheck/pull/1842
+(defun executable-find-add-remote (args)
+  "Add optional remote argument t to ARGS when supported."
+  (if (and (not (version< emacs-version "27.1")) (= (length args) 1))
+      (append args '(t))
+    args))
+(advice-add 'executable-find :filter-args 'executable-find-add-remote)
```
jcs090218, line 520: We might need to rename this function with the prefix `flycheck-`. Therefore,

```elisp
(advice-add 'executable-find :filter-args #'flycheck-executable-find-add-remote)
```

flycheck.el
```diff
@@ -510,6 +510,20 @@ sandboxes."
...
+
+(defun buffer-file-local-name (&optional buffer)
+  "Return the local file-name of the current buffer or BUFFER when specified."
+  (let ((file-name (buffer-file-name (or buffer (current-buffer)))))
+    (when file-name (file-local-name file-name))))
```
jcs090218, line 525: And this function too:

```diff
- (defun buffer-file-local-name (&optional buffer)
+ (defun flycheck-buffer-file-local-name (&optional buffer)
```

peterbecich:
Thank you @carrete and @jcs090218

peterbecich:
Any chance this can be merged?

peterbecich:
@marsam , any chance this can be merged? Thank you

zabuxx:
It would be great if this could be merged...

peterbecich:
I see this now: https://github.com/flycheck/flycheck/pull/1922

bbatsov:
Closing — while TRAMP support is a frequently requested feature, partial support for only some checkers creates user confusion. A comprehensive approach covering all (or most) checkers would be needed. We'll track this as a feature request and revisit when a comprehensive plan is available. Thank you for the extensive work on this!

carrete:
Happy to help again if/when this is picked up.

bbatsov:
Glad to hear this. I've been trying to clean up the codebase and the issue tracker a bit lately in preparation for a new release, but afterwards we can revisit the topic again.

PR #1845, opened 2020-12-11
Fix flycheck-may-enable-checker

### LINKED ISSUE #1844: The function flycheck-may-enable-checker fails on non-existing checkers
## Checklist

## Steps to reproduce

Run `(flycheck-may-enable-checker 'gibberish)` in any buffer. The answer is `t`.

## Expected behavior

I would expect it to return false for non-existing checkers.

FredrikMeyer:
The function flycheck-may-enable-checker returns wrong answer on gibberish input / non-existing checkers. See my issue https://github.com/flycheck/flycheck/issues/1844 

This PR adds a check that the checker is valid.

Fixes #1844

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2222,6 +2222,8 @@ Return non-nil if CHECKER may be used for the current buffer, and
 nil otherwise.  The result of the `:enabled' check, if any, is
 cached."
   (and
+   ;; May only enable valid checkers
+   (flycheck-valid-checker-p checker)
    ;; Don't run the :enabled check if the checker is already disabled…
    (not (flycheck-disabled-checker-p checker))
    (or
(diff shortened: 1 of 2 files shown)


cpitclaudel (approved):
Works for me, but… should this be a boolean check or an assertion?

I guess check makes it more convenient to write robust programs independently of whether a package defining a checker is available.  Makes sense.

I (or someone else) need to fix the build; I'll get around to it in a few days.

FredrikMeyer:
> Works for me, but… should this be a boolean check or an assertion?

Hm, I didn't think about that, actually. I guess I have a slight preference for non-crashing behaviour, so I think that's why I went with booleans.

> I guess check makes it more convenient to write robust programs independently of whether a package defining a checker is available. Makes sense.

Good point.

fmdkdd:
I've fixed the build, so you should be able to rebase on master and then we'll have to wait for Travis to do its thing.
(Or maybe I will have found the time to switch the CI to Github actions so we can jump from the sinking ship).

PR #1849, opened 2020-12-18
Set destination to nil in racket checker feature test (#1848).

### LINKED ISSUE #1848: Flycheck corrupts buffer when the underlying checker issues some unhandled warnings
Thank you for taking the time to report an issue and improve Flycheck.  This template is for **actual bugs you observed**.  If you have trouble setting up Flycheck, or if you have a question, please use the relevant issue template instead.

## Checklist

## Bug description
Racket uses [`raco`](https://docs.racket-lang.org/raco/) as its command line tool. As the flycheck [doc](https://www.flycheck.org/en/28/_downloads/flycheck.html#Racket) says, flycheck uses `raco` for error checking. `raco` is also used as the package manager for racket. I customize some racket packages for debugging. Thus I have to install a racket package twice: one is system wide and installed together with racket; the other is local under my home directory. In this case, when `raco` is running, it will use the local version of the package, but gives a warning like:

``` shell
warning: tool "scribble" registered twice: "scribble" and #<path:/usr/share/racket/pkgs/scribble-lib/scribble>
```

Don't worry, this warning is harmless. Everything just works fine.

As the screenshot below shows, when I install two versions of the same racket package and edit a racket file in emacs with `flycheck-mode` enabled, flycheck keeps writing the harmless warning of `raco` into my file. I did some debugging. I guess this is because the first writing causes file change and syntax error, which triggers flycheck to call `raco` again, leading to an infinite loop. It seems flycheck does some redirection or buffer writing. Things that should have been written into standard error are written to my file.

## Steps to reproduce
1. Make some changes to a system wide racket package and install it in user directory (more detailed steps can be provided if necessary)
2. run `raco`(no argument needed), just to make sure you see something like `warning: tool "xxx" registered twice: xxx`
3. Open a racket file in emacs
4. Enable `flycheck-mode`
5. See screenshot

## Expected behavior
Even if the underlying checker reports a warning, the checked buffer should keep untouched.

## Screenshots
flycheck_corrupts_file

## System configuration
```
Syntax checkers for buffer style.rkt in racket-mode:

First checker to run:

  racket
    - may enable:            yes
    - may run:               t
    - executable:            Found at /usr/bin/raco
    - compiler-lib package:  present
    - Geiser Implementation: Using Racket Mode

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled checkers.

--------------------

Flycheck version: 32snapshot (package: 20201214.2154)
Emacs version:    27.1
System:           x86_64-pc-linux-gnu
Window system:    x
```

Emacs configuration:
- [x] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs

## Additional notes
This problem may not limit to `raco`. Just in normal case, it's not that easy to happen.

[Troubleshooting guide]:  https://www.flycheck.org/en/latest/user/troubleshooting.html

chihyang:
This PR fixes #1848.

Sometimes `raco` command will issue a harmless warning to standard
error (eg. when a racket package is installed both in system directory and user
directory). If `destination` is set to `t` in `flycheck-call-checker-process`,
this warning would be improperly redirected to current buffer, causing file
change and syntax error. Then flycheck will call `raco` to check the buffer, and
the warning will be inserted to current buffer again, leading to an infinite
loop.

Setting `destination` to nil avoids problems. Because when
`flycheck-racket-has-expand-p` is called, we don't need `raco`'s output.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -10755,7 +10755,7 @@ See URL `https://github.com/jimhester/lintr'."
 
 (defun flycheck-racket-has-expand-p (checker)
   "Whether the executable of CHECKER provides the `expand' command."
-  (eql 0 (flycheck-call-checker-process checker nil t nil "expand")))
+  (eql 0 (flycheck-call-checker-process checker nil nil nil "expand")))
 
 (flycheck-define-checker racket
   "A Racket syntax checker with `raco expand'.


cpitclaudel (approved):
:+1:

Issue #1852, opened 2020-12-23
flycheck-display-error-explanation inserts explanation at point


mohkale:
Thank you for taking the time to report an issue and improve Flycheck.  This template is for **actual bugs you observed**.  If you have trouble setting up Flycheck, or if you have a question, please use the relevant issue template instead.

## Checklist
- [X] I have checked existing issues for potential duplicates before creating this one.
- [X] I have read the [Troubleshooting guide][].

## Bug description
Pretty sure there's a bug in the definition of `flycheck-display-error-explanation` which is called by `flycheck-explain-error-at-point` and several other functions.

https://github.com/flycheck/flycheck/blob/7ffbbfcb98af22c665f613af391c5357f139f393/flycheck.el#L5615-L5629

The condition block which has a clause that inserts the message into the current buffer is outside of the `with-current-buffer` block causing flycheck to insert it at point. Moving it into the with-selected-buffer block seems to fix the issue.

## Steps to reproduce
1. Open a file, eg. main.py
2. enable flycheck and make sure you have pylint installed
3. enter the following code:

```py
import os

def foo():
    d = foo_bar('baz')
    print(d)
```
4. `foo_bar` should be marked as an error, move point to it and run `M-x flycheck-explain-error-at-point`

- [X] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

## Expected behavior
Either a buffer with the message is shown or the message is shown in the echo area.

## Actual behaviour
the message is inserted directly into my current buffer.

## System configuration
```
Syntax checkers for buffer main.py in python-mode:

First checker to run:

  python-pylint
    - may enable:         yes
    - executable:         Found at /usr/bin/python3
    - configuration file: Not found
    - `pylint' module:    Found at "/home/mohkale/.local/lib/python3.9/site-packages/pylint/__init__.py"
    - next checkers:      python-mypy

Checkers that could run if selected:

  python-pycompile  select
    - may enable:    yes
    - executable:    Found at /usr/bin/python3
    - next checkers: python-mypy

Checkers that are compatible with this mode, but will not run until properly configured:

  python-flake8 (automatically disabled) reset
    - may enable:         no
    - executable:         Found at /usr/bin/python3
    - configuration file: Not found
[...]

PR #1853, opened 2020-12-23
Insert plain error explanations into the right buffer

cpitclaudel:
Fixes GH-1852. See the related discussion in GH-1774.


cpitclaudel:
Duplicate of #1801.

PR #1857, opened 2021-01-09
Add yamllint to YAML checker chain

berquist:
I saw that [yamllint](https://github.com/adrienverge/yamllint) wasn't a part of the YAML checker chain, so this patch adds it. It also structures the chain assuming that you have one of the two main YAML implementations supported (Ruby or JS) run first, then the linter runs, then CWL. It assumes that you would want to run only one of Ruby or JS, not both.


fmdkdd (approved):

fmdkdd:
Sure, thanks.

Issue #1864, opened 2021-02-09
Add an explainer slot to flycheck-error structure


wilkinsonmr:
Thank you for taking the time to improve Flycheck.

## Checklist
- [X] I have checked existing issues for potential duplicates before creating this one.

## Feature description
A clear and concise description of what you want Flycheck to do.

Add a slot to the flycheck-error structure to allow elements from the command output to be passed to the error explainer. For example, some linters include a URL as a pointer to error explanation as part of the error detail. The URL can't always be composed from other slots like in the `eslint` checker, for example. Terraform checker `checkov` does this; it includes a "guideline" URL in its output. 

## Describe alternatives you've considered

Currently, for checkov, I concatenate the explanation elements to the message slot and turn on goto-address mode from within flycheck-errors-mode. For a bandit checker, I set the id slot to the concatenation of the test_id and test_name and build the URL as in eslint, but the ID is then no longer just an error ID. 

## Additional context
Add any other context or screenshots, mockups etc.

cpitclaudel:
Thanks for your work on your Flycheck checkers :)

Adding that slot has been discussed in the past.  One issue is that adding slots breaks ELC compatibility (it was quite painful when we added the end-line and end-column numbers).  

I think we should still do it if it brings significant benefits.  But, have you tried propertizing the error message (or the error ID)? Would that help?

wilkinsonmr:
Thanks to Flycheck for making linting in Emacs so easy!

Good to know this has been considered.  I'm not sure how significant the benefits would be, but it would make taking advantage of the error explainer functionality easier.

I looked at propertizing the message based on your suggestion, but it seems I'd be reimplementing goto-addr (badly).  In addition to tacking on a URL to the message, another way this is being done is to create an additinoal error object for the URL (as its message). Mypy is an example of this approach.

cpitclaudel:
> Thanks to Flycheck for making linting in Emacs so easy!

:)

> I looked at propertizing the message based on your suggestion, but it seems I'd be reimplementing goto-addr (badly).

Can you clarify what you mean by this?

wilkinsonmr:
I assumed you meant to propertize the message with clickable text, but I looked at this some more and seem to have it working.  

I put this code into my parser function 
```el
(put-text-property 0 (length .issue_text) 'explainer-url .more_info .issue_text)
```
where `.more_info` contains a URL from the linter,

and this code in the checker definition 
```el
      :error-explainer
      (lambda (err)
[...]

PR #1867, opened 2021-02-26
Add error-explainer documentation for URLs

wilkinsonmr:
Since the flycheck-error structure does not have a slot for
error-explainer data, document a technique by which URLs can be
passed to the error-explainer function via text properties.

Addresses GH-1864


fmdkdd (approved):
Neat trick.  Thanks!

PR #1826, opened 2020-09-19
Update nix and terraform checkers for recent versions

marsam:
+ Terraform 0.13 introduces a new error message format.
+ Nix 3.x introduces a new format for the error messages. Fixes https://github.com/flycheck/flycheck/issues/1823 cc: @endgame
+ Fix pylint checker tests.

PR #1869, opened 2021-03-17
Add new error pattern for nix unstable

lunik1:
With NixOS/nix#4467 unstable branch of nix changed its error formatting (you're not having deja-vu, this has happened before and was adressed in #1826). This patch adds a new pattern to match the new format.

The pattern added in #1826 is kept in this patch but there may be a case for removing it, it is now only ever used by unreleased nix versions.


fmdkdd (approved):
LGTM, thanks for keeping it up to date!

PR #1870, opened 2021-03-29
Fix flycheck-error-level-interesting-p

KevOrr:
According to its docstring, `flycheck-error-level-interesting-p` should return true if there are no errors as _or more severe than_ `flycheck-navigation-minimum-level`. Currently, however, it only checks if there are any errors exactly as severe as `flycheck-navigation-minimum-level` (via `flycheck-has-current-errors-p`). This patch:

* improves the documentation of
  * `flycheck-has-current-errors-p`
  * `flycheck-has-errors-p`
* introduces two new functions
  * `flycheck-has-current-errors-atleast-p`
  * `flycheck-has-errors-atleast-p`
* Changes `flycheck-error-level-interesting-p` to use `flycheck-has-current-errors-atleast-p`, instead of its "exact" sibling, bringing its actual behavior in line with its documented behavior

I also considered changing `flycheck-has-[current-]errors-p` to use `>=` instead of `eq`, but since it has a public name (I'm assuming `--` is used to signal that a symbol is private), I didn't want to introduce a backwards-breaking change. If instead it should be the case that `flycheck-has-[current-]errors-p` check for errors at least as severe as `level`, then I can edit this PR to modify them directly.


KevOrr:
Oops, I should have run the tests locally before submitting this PR. It seems as though some of the tests rely on the old behavior of `flycheck-error-level-interesting-p`. For example `flycheck-first-error/over-warnings/goes-to-first-error` assumes that by setting `flycheck-navigation-minimum-level` to `'warning`, then `(flycheck-first-error)` will move onto the checkdoc error of level `'info` on line 12 of `errors-and-warnings.el`. Should the checkdoc error indeed be of level `'info` or is this a local configuration thing, and it should be `'warning` or higher?

bbatsov:
Seems like a duplicate of #1994. We should agree which approach to solving the issue is better.

PR #1872, opened 2021-04-12
Update list of `error-explainer`s

Amorymeltzer:
Hadn't been updated since its addition in #1122 (52eb56af)

----

Arguably, the list is long enough that it's not worth listing them individually, and just doing something like "While several checkers produce explanations, the majority do not."  I personally prefer the list for the sake of helpfulness, but ymmv.


bbatsov:
Perhaps we can make this a bullet list?

Amorymeltzer:
@bbatsov Agreed; opened #2051 for that

PR #1875, opened 2021-04-13
Add error-explainer to css-stylelint

Amorymeltzer:
(Initially) Opened as a draft since I'm not really sure if this should be added to the `scss-stylelint` and `less-stylelint` checkers as well.  I think maybe?  Based off <https://stylelint.io/user-guide/usage/options#syntax> it's separate from the `scss` plugin and so on, but I've no experience with scss or less.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8252,7 +8252,12 @@ See URL `http://stylelint.io/'."
   :standard-input t
   :error-parser flycheck-parse-stylelint
   :predicate flycheck-buffer-nonempty-p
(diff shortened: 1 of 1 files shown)


Amorymeltzer:
As above, no idea on whether it should be added for the others as well

bbatsov:
> I think maybe? Based off https://stylelint.io/user-guide/usage/options#syntax it's separate from the scss plugin and so on, but I've no experience with scss or less.

Probably it makes sense for them as well, but it's been a while since I've done anything with scss or less.

PR #1876, opened 2021-04-13
Add error-explainer for markdownlint checker

Amorymeltzer:
Links to the rules defined upstream at
<https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md>

Update url for markdownlint-cli

----

Initially opened as draft as it fails the 80-column check, but I'm not sure what would be preferred given the url is the url


bbatsov:
Can you rebase this and your other similar PRs on top of `master`? (provided you're still interested in them after all this time)

Amorymeltzer:
@bbatsov OMG Yes!  Completely forgot about these.  #1871, #1873, #1875, and this should be all set, I think!

PR #1887, opened 2021-06-05
Add '--follow-imports' configuration option to python-mypy

hexane360:
This PR adds a new configuration option to python-mypy. Setting this to "silent" allows mypy to only report errors from the current file (which makes more sense for flycheck). This could be set in the configuration file instead, but it would change the semantics of running mypy from the command line, so I think it's best to be able to override the option in flycheck.


wyuenho:
You don't need this, there's a [configuration](https://mypy.readthedocs.io/en/stable/config_file.html#confval-follow_imports) for that.

PR #1896, opened 2021-08-10
Drop the pkg-info deps

liuyinz:
Replace the dependence of `pkg-info` with emacs  built-in package `lisp-mnt`

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -10,7 +10,7 @@
 ;; URL: http://www.flycheck.org
 ;; Keywords: convenience, languages, tools
 ;; Version: 32-cvs
-;; Package-Requires: ((dash "2.12.1") (pkg-info "0.4") (let-alist "1.0.4") (seq "1.11") (emacs "24.3"))
+;; Package-Requires: ((dash "2.12.1") (let-alist "1.0.4") (seq "1.11") (emacs "24.3"))
 
 ;; This file is not part of GNU Emacs.
 
@@ -79,6 +79,7 @@
 (require 'rx)                    ; Regexp fanciness in `flycheck-define-checker'
 (require 'help-mode)             ; `define-button-type'
 (require 'find-func)             ; `find-function-regexp-alist'
+(require 'lisp-mnt)              ; `lm-version'
 (require 'json)                  ; `flycheck-parse-tslint'
 (require 'ansi-color)            ; `flycheck-parse-with-patterns-without-color'
 
@@ -89,9 +90,6 @@
 (defvar markdown-hide-markup)                     ;
 (defvar markdown-fontify-code-block-default-mode) ; For rust-error-explainer
 (defvar markdown-fontify-code-blocks-natively)    ;
-
-;; Tell the byte compiler about autoloaded functions from packages
-(declare-function pkg-info-version-info "pkg-info" (package))
 
 
 ;;; Compatibility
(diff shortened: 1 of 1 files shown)


Jturnerusa:
Hello. 

I just wanted to see if there is any update on the status of this PR. It's only changing a single line, and not adding any code that needs to be maintained with the benefit of making Flycheck depend on two less third party packages, both of which have been totally unmaintained for years.

liuyinz:
I have no idea.

cpitclaudel:
Sorry for dropping this.  Happy to merge, but can we confirm that this works with binary installs (`.elc` only)?  I don't know how flycheck is package in every distro, but I wouldn't be surprised if some shipped elc only.

liuyinz:
> Sorry for dropping this. Happy to merge, but can we confirm that this works with binary installs (`.elc` only)? I don't know how flycheck is package in every distro, but I wouldn't be surprised if some shipped elc only.

@Jturnerusa , macos is ok, I'm not familiar with gentoo...

cpitclaudel:
But does lm-version work with an elc-only package?  Probably not, if it uses comments in the file header, right?

liuyinz:
No, it can't. I test it.

Jturnerusa:
Yeah unfortunately the `lm-version` function requires the library's source to be installed.

I am not totally sure about this and haven't tested it yet, but I think `pkg-info` also requires source files to get the library and package version information. From what I can tell `pkg-info` ends up calling into `epl` functions, then `epl` ends up calling into the function `package-buffer-info` in `package.el` which also is parsing version info from headers similar to how `lm-version` works, and the function even uses some `lisp-mnt` functions itself in the process!

I am not great at reading elisp source so I may be wrong about this, and will see what I can do to test this out. If I am wrong then I guess this patch could potentially break for some users if they only have elc files available after installation (do any package managers do this?).

liuyinz:
LGTM.
Actually I almost don't check version mannually, just curious about dependencies to make this patch. If as @Jturnerusa said,
`package-buffer-info` parse headers as similar as `lm-version`, then it would fail with only binary as well, right?

Jturnerusa:
> How about we take this patch but with the call to the version function in ignore-errors?

What exactly do you mean here? I can't seem to find this right now but if we need another PR and patch with a different function call or whatever I can certainly help! 

Thanks a lot for responding and taking this into consideration by the way!

cpitclaudel:
Yep @liuyinz, I think @Jturnerusa is right and it's likely that pkg-info doesn't even help with the case I'm worried about ^^.

@Jturnerusa , my suggestion was to wrap the function into a call to `ignore-errors`.  But now that I read the docstring it doesn't really need to be done that way.

That being said, the new output isn't the same as the previous one.  I think it's probably fine: most code that I can find only using this function actually doesn't expect the most general output.  For example on my machine I see `Flycheck version: 33snapshot (package: 20220314.27)`.

But still, we're going to break code like https://github.com/simon-katz/nomis-public-dev-setup/blob/2e01484be1267ea1e2c280f61e47eaecc7cf065d/emacs-configuration/emacs-installation/emacs-init-files/nomis-flycheck.el

I'm not sure whether it's worth it.

liuyinz:
update a little function, remove two old deps.
how gratifying !

bbatsov:
I think it's much better to use `package-get-version` instead. See https://github.com/clojure-emacs/cider/blob/master/cider-util.el#L386 for an example. 

But in general I think it's a great idea to shed `pkg-info`, mostly because it depends on the big abandonware package EPL.

PR #1898, opened 2021-08-19
When works with erb on web-mode, enable eruby checker.

zw963:
When open a erb file use web-mode, eruby check not enabled default, this PR fix this.

image


bbatsov:
Moot now: the eruby checkers were removed from Flycheck a while back (47ae804d), so there's nothing left to add web-mode to. Thanks for the patch, and sorry it sat unanswered.

PR #1900, opened 2021-09-03
Add variable for temp file suffix

Corentin-pro:
Added `flycheck-temp-suffix` for the temporary file created by flycheck.

Coding in python seems to require such temporary file (because of pylint?). When coding with frameworks like Flask, which is quiet popular, the debug mode of such framework automatically reload on python source file changes. This can happen also with some javascript tools.

There is no work around with Flask for instance to avoid reloading when flycheck temporary file are generated (no configuration possible).

Appending `.temp` on the temporary file for instance will not trigger the auto reload mechanism (extension sensitive) while still being valid for the linter. This change is just adding an option for user to fix the issue while only adding 7 lines of code.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -868,6 +868,12 @@ function."
   :package-version '(flycheck . "0.19")
   :risky t)
 
+(defcustom flycheck-temp-suffix "flycheck"
+  "Sufffix for temporary files created by Flycheck."
+  :group 'flycheck
+  :type 'string
+  :risky t)
+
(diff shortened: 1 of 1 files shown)


Corentin-pro:
As CLA assistant is not working let me right it here:

I, the contributor, agree to licence my contributions to the Flycheck project under the terms of the GPL 3.0 and any later version, and to license my contributions to the documentation of the Flycheck project under the terms of the Creative Commons Attribution-ShareAlike 4.0 International license.

PR #1907, opened 2021-09-21
erlang compiler now knows about columns

massemanet:
As of sometimes in 2020(?) the erlang compiler includes column numbers in errors & warnings.

This PR prevents the erlang checkers from failing on that. Also fixes the tests.


bbatsov:
Closing in favor of #1914.

Issue #1913, opened 2021-10-27
Erlang/OTP 24 added column numbers to output, breaking pattern matching.


juster:
## Bug description
Flycheck behaves strangely after upgrading Erlang. The errors are mostly matched but only line 1 of the file has any error indicator. I traced this to improper error message parsing. This is due to Erlang/OTP changing the format of their error/warning messages. Column numbers were added after the line numbers as well as carets (^) for column indicators!

Example of trying to compile your test error.erl from the terminal:
```
error.erl:7:1: head mismatch
%    7| error_func() ->
%     | ^

error.erl:3:2: Warning: export_all flag enabled - all functions will be exported
%    3| -compile(export_all).
%     |  ^

```

## Steps to reproduce

1. Upgrade Erlang to 24.0 or higher.
2. Open an .erl file in Emacs and enable flycheck.
3. Type poorly across multiple lines.
4. Look for error indicators on the wrong line.

## Emacs configuration:

- [ ] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [X] Doom Emacs
- [ ] Other shared configuration

## Addition Notes

This is mentioned in the [release notes for Erlang 24.0](https://www.erlang.org/news/148). Notice the second bullet point:

> Compiler warnings and errors now include column numbers in addition to line numbers.

## Pull Request Coming

I have a pull request I will tag with this issue and send it. The fix is pretty simple.

bbatsov:
Fixed by #1914.

PR #1914, opened 2021-10-27
Fix Erlang/OTP 24 which added columns to output

juster:
Column numbers were added to warnings and errors. This breaks the
pattern matching for Erlang 24.0 and later. Column numbers are optional
in the pattern so that earlier versions of Erlang will still match
properly.

Fixes GitHub issue #1913.


masse-wg2:
Seems the CI is broken?

tsloughter:
Any updates on this? Should the PR be reopened so maybe CI passes if someone can't kick it to run again?

cpitclaudel (approved):
Looks good, thanks.  I would suggest not testing for the version of Erlang in the test; instead, just assume the columns are there.  We usually test only the latest version of each tool.

bbatsov:
I think Travis CI is dead at this point and needs to be replaced with something else (e.g. GHA).

PR #1917, opened 2021-10-30
Fixed flycheck-rust-cargo-has-command-p for newer Cargo versions

### LINKED ISSUE #1916: `flycheck-rust-cargo-has-command-p` returns `nil` incorrectly
Thank you for taking the time to report an issue and improve Flycheck.  This template is for **actual bugs you observed**.  If you have trouble setting up Flycheck, or if you have a question, please use the relevant issue template instead.

## Checklist

## Bug description
Running `(flycheck-rust-cargo-has-command-p "run")` (or any other Cargo command) returns `nil` when it should return `t`. This is because `flycheck-rust-cargo-has-command-p` uses `cargo --list` as a basis, which returns descriptions for each subcommand (ever since https://github.com/rust-lang/cargo/pull/9848). This means `(member command ...)` returns `nil`.

## Steps to reproduce
1. Run `(flycheck-rust-cargo-has-command-p "run")` with a sufficiently recent (>= 1.56) version of Cargo installed.
2. Observe that it returns `nil`

## Expected behavior
The command should return `t` when the command is installed

Emacs configuration:
- [ ] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [x] Doom Emacs

[Troubleshooting guide]:  https://www.flycheck.org/en/latest/user/troubleshooting.html

PatrickNorton:
Should fix #1916.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -11431,7 +11431,8 @@ versions inferior to 1.25)."
 
 Execute `cargo --list' to find out whether COMMAND is present."
   (let ((cargo (funcall flycheck-executable-find "cargo")))
-    (member command (mapcar #'string-trim-left
+    (cl-some (lambda (x) (string-prefix-p command x))
+             (mapcar #'string-trim-left
                             (ignore-errors (process-lines cargo "--list"))))))
 
 (defun flycheck-rust-valid-crate-type-p (crate-type)
--- a/flycheck.el
+++ b/flycheck.el
@@ -11431,9 +11431,9 @@ versions inferior to 1.25)."
 
 Execute `cargo --list' to find out whether COMMAND is present."
   (let ((cargo (funcall flycheck-executable-find "cargo")))
-    (cl-some (lambda (x) (string-prefix-p command x))
-             (mapcar #'string-trim-left
-                            (ignore-errors (process-lines cargo "--list"))))))
+    (seq-find (lambda (x) (string-prefix-p command x))
+              (mapcar #'string-trim-left
+                      (ignore-errors (process-lines cargo "--list"))))))
 
 (defun flycheck-rust-valid-crate-type-p (crate-type)
   "Whether CRATE-TYPE is a valid target type for Cargo.
--- a/flycheck.el
+++ b/flycheck.el
@@ -11431,9 +11431,10 @@ versions inferior to 1.25)."
 
 Execute `cargo --list' to find out whether COMMAND is present."
   (let ((cargo (funcall flycheck-executable-find "cargo")))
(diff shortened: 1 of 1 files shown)


bergey:
This looks good to me.  I encountered the same bug today, and wrote an equivalent fix, but `string-prefix-p` is simpler & more efficient than what I wrote.

rymndhng:
I've also run into this, and this fix works for me.

gerlacdt:
Patch works for me too, thx!

ngoldbaum:
I just ran into this and would appreciate it if this could be merged.

Gastove:
Also just smacked into this. It seems like we've got a couple decent solutions; @cpitclaudel, any chance you or another maintainer could give this a review?

cpitclaudel:
LGTM, but Flycheck uses `seq`, not `cl-lib`, so it would be a bit better to use `seq-find`.  Additionally, if `cargo --list` prints `foo-blah`, won't the new code conclude that `foo` is available?

ngoldbaum:
maybe #1948 is a better fix?

PatrickNorton:
I've changed the code to use `seq-find`, but I think that #1948 does have a more robust solution overall—I hadn't thought about what would happen if one command were to be a substring of another.

flycheck.el
```diff
@@ -11431,8 +11431,9 @@ versions inferior to 1.25)."
 
 Execute `cargo --list' to find out whether COMMAND is present."
   (let ((cargo (funcall flycheck-executable-find "cargo")))
-    (member command (mapcar #'string-trim-left
-                            (ignore-errors (process-lines cargo "--list"))))))
+    (seq-find (lambda (x) (string-prefix-p command x))
+              (mapcar #'string-trim-left
+                      (ignore-errors (process-lines cargo "--list"))))))
```
cpitclaudel, line 11436: Would something like this work?

```suggestion
    (member command 
      (mapcar (lambda (line)
                (replace-regexp-in-string "\\s-*\\(\\S-+\\).*\\'" "\\1" line))
              (ignore-errors (process-lines cargo "--list"))))))
```

cpitclaudel:
Unfortunately the code in #1948 will corrupt the match data.  I've posted an untested piece of code that might do the trick; can someone check?  Thanks!

cpitclaudel:
Thanks!  I'll wait for one more person to approve this or a few days to pass and then I'll merge it.  Sorry for letting this sit so long and thanks a lot for being more reactive than me.

PatrickNorton:
No worries! Waiting for one more approval sounds fine. You happened to catch me during some down time; I wish I was usually as responsive as I am today :).

PR #1920, opened 2021-11-08
isolate the snippet of source code used in formatted error messages

### LINKED ISSUE #1919: flycheck’s tooltips can be displayed sdrawkcab when the source contains bidi directional formatting characters
## Checklist

## Bug description
Source code containing bidi directional formatting characters can mess up the direction of the error message itself.

I imagine that localized error messages might cause the same problem as well, though I haven’t any easy way to check.

## Steps to reproduce

1. Start with a source file containing tricky comments:

```rust
fn main() {
    let is_admin = false;
    /*‮ } ⁦if is_admin⁩ ⁦ begin admins only */
        println!("You are an admin.");
    /* end admins only ‮ { ⁦*/
}
```
2. Rust compiler reports errors
3. Hover over the first comment, or move point to it, to see the sdrawkcab messages. See the screenshot as well.

## Expected behavior
The text direction of the error message is not affected by the content of the source code it arose from.

## Screenshots
Screenshot from 2021-11-07 23-50-47

## The Fix
I will make a pull request for this, but for the moment here is the change I made to fix this:

```lisp
(defun flycheck-error-format-message-and-id (err &optional include-snippet)
  "Format the message and id of ERR as human-readable string.

If INCLUDE-SNIPPET is non-nil, prepend the message with a snippet
of the text that the error applies to (such text can only be
determined if the error contains a full span, not just a
beginning position)."
  (let* ((id (flycheck-error-id err))
         (fname (flycheck-error-filename err))
         (other-file-p (and fname (not (equal fname (buffer-file-name))))))
    (concat (and other-file-p (format "In %S:\n" (file-relative-name fname)))
            (and include-snippet
                 (-when-let* ((snippet (flycheck-error-format-snippet err)))
                   (flycheck--format-message "`\N{LEFT-TO-RIGHT MARK}\N{FIRST STRONG ISOLATE}%s\N{POP DIRECTIONAL ISOLATE}\N{LEFT-TO-RIGHT MARK}': " snippet)))
            (or (flycheck-error-message err)
                (format "Unknown %S" (flycheck-error-level err)))
            (and id (format " [%s]" id)))))
```

Likely the other fields ought to have the same treatment.

## System configuration
I’ve got some `lsp-mode` stuff going on as well, which isn’t reflected here.

```
Syntax checkers for buffer flycheck.el in emacs-lisp-mode:

First checker to run:

  emacs-lisp
    - may enable:    yes
    - may run:       t
    - executable:    Found at /home/db48x/bin/emacs
    - next checkers: emacs-lisp-checkdoc

Checkers that may run as part of the first checker's chain:

  emacs-lisp-checkdoc
    - may enable: yes
    - executable: Found at /home/db48x/bin/emacs

checkers.

--------------------

Flycheck version: 32snapshot (package: 20210825.1804)
Emacs version:    29.0.50
System:           x86_64-pc-linux-gnu
Window system:    x
```

Emacs configuration:
- [x] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs

## Additional notes
https://en.wikipedia.org/wiki/Bidirectional_text
https://www.w3.org/International/questions/qa-bidi-unicode-controls

db48x:
lest bidi formatting characters cause the error message to be rendered
sdrawkcab. In principle we should probably do this to the filename and
even the error message, since those could be suspect as well. As
ordinary RTL text doesn’t cause any problem, it is in practice not so
important. Note also that if flycheck is to be localized, this format
string must be localized as well (or the localization system must
support such substitutions directly).

Fixes #1919.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -3864,6 +3864,14 @@ most.  It defaults to 20."
            "\\s-+" " " (buffer-substring beg (min end (point-max))))
           (or max-length 20) nil nil t))))))
 
+;; use isolates and direction markers to make it less likely for text
+;; taken from the source to render the error message unreable
+;; (sdrawkcab).
+(defconst flycheck--format-snippet
+  (concat "`\N{LEFT-TO-RIGHT MARK}\N{FIRST STRONG ISOLATE}"
+          "%s\N{POP DIRECTIONAL ISOLATE}\N{LEFT-TO-RIGHT MARK}': ")
+  "Format string to use on a snippet of source code related to the error.")
+
 (defun flycheck-error-format-message-and-id (err &optional include-snippet)
   "Format the message and id of ERR as human-readable string.
 
@@ -3877,7 +3885,7 @@ beginning position)."
     (concat (and other-file-p (format "In %S:\n" (file-relative-name fname)))
             (and include-snippet
                  (-when-let* ((snippet (flycheck-error-format-snippet err)))
-                   (flycheck--format-message "`%s': " snippet)))
+                   (flycheck--format-message flycheck--format-snippet snippet)))
             (or (flycheck-error-message err)
                 (format "Unknown %S" (flycheck-error-level err)))
             (and id (format " [%s]" id)))))


bbatsov:
This landed on master as ddca9f77 ([#1919] in the v36 changelog). Thanks!

PR #1925, opened 2021-12-06
Fix Buttercup tests on Emacs 28.

phst:
- On Emacs 28, Checkdoc has a new customization option
  ‘checkdoc-column-zero-backslash-before-paren’.

- The header line format is slightly different.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1,6 +1,6 @@
 ;;; flycheck.el --- On-the-fly syntax checking -*- lexical-binding: t; -*-
 
-;; Copyright (C) 2017-2020 Flycheck contributors
+;; Copyright (C) 2017-2021 Flycheck contributors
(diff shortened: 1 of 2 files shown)


phst:
Thanks! Probably you'll also need to merge this. The CI check doesn't seem to finish, though.

bbatsov:
@phst Sadly Travis CI is dead for a while and this job will never finish. We'll likely have to replace it.

phst:
Fair, but you still need to merge this PR if you want to accept it, I don't have the necessary permissions.

bbatsov:
I literally can't press the merge button due to this required check - that's the problem. :-) @fmdkdd @cpitclaudel Any suggestions?

cpitclaudel:
I have removed the dependency on travis; thanks for the ping

PR #1929, opened 2021-12-17
Add -gencode option for cuda-nvcc

### LINKED ISSUE #1926: cuda flycheck support "gencode" option
Thank you for taking the time to improve Flycheck.

## Checklist

## Feature description
An important option to the nvcc compiler is "gencode", so that the generated source code will match the architecture.  It should be configurable because the default will often be old and lead to compiler errors.

eyal0:
This fixes #1926


eyal0:
Can we consider a merge on this please?

PR #1934, opened 2022-02-03
Don't break python2 yet.

ebobby:
Last commit to these lines changed the command from `python` to `python3`
because Python2 was EOL. And that's right, but it doesn't mean it is not used :)

Every major distribution still uses `python` to refer to whatever version it has
installed, either 3 or 2. Plus, pyenv shims do work with `python` to select
whatever the current version may be.

Also, whenever we get Python 4, I am sure `python` will point to it.


wasamasa:
https://github.com/flycheck/flycheck/issues/1055

PR #1936, opened 2022-02-28
Linters should be stateless and efficient (fixing call to R linter)

hyiltiz:
The call to R should respect `inferior-R-args` variable. If not, then at least please do not call the R slave session with `--restore`, which would try to load `./RData`, and depending on what `PWD` is, that could try to load gigabytes of workspace image into the lint, including variables and functions. That introduces unacceptable lag (every keystore requires over 10s of wait time whenever the linter is invoked then killed off, loading gigabytes of image). What's worse is that a linter should start stateless and thus be predictable; loading arbitrary images from a "current directory or from a default fallback location" is a very bad behavior for a linter.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -10779,14 +10779,14 @@ expression, which selects linters for lintr."
   "Whether CHECKER (R) has installed the `lintr' library."
   (eql 0 (flycheck-call-checker-process
           checker nil nil nil
-          "--slave" "--restore" "--no-save" "-e"
+          "--slave" "--no-restore" "--no-save" "-e"
           "library('lintr')")))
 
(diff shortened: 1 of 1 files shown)


cpitclaudel (approved):
LGTM.  Pinging @dholm since he wrote the code (albeit a long time ago).
Maybe change the commit message to mention the R checker?

hyiltiz:
Anything that I can do before this can be merged?

cpitclaudel:
> Anything that I can do before this can be merged?

Just the commit message, then I'll merge

hyiltiz:
That works?

cpitclaudel:
Yep, thanks!

PR #1946, opened 2022-05-28
Fix(lintr): make user-defined lintr-linters can be correctly passed to the R process.

milanglacier:
if user directly calls

```elisp
(setq flycheck-lintr-linters
      "linters = with_defaults(line_length_linter=line_length_linter(200),commented_code_linter = NULL)")

```

with something like this in their config

in deed this will have no effect.

Explicitly specifying`"linters=" flycheck-lintr-linters` makes things changed


bbatsov:
I'm fine with the proposed change, but it needs a changelog entry.

bbatsov:
@milanglacier Friendly ping :-)

PR #1948, opened 2022-06-09
FIX: detect cargo sub-commands

hrls:
Looks like ` flycheck-rust-cargo-has-command-p` was failed with current `cargo --list` output due to optional extra comments line per each sub-command. My output of `cargo --list` contains the following line:
```
    clippy               Checks a package to catch common mistakes and improve your Rust code.
```

  
  
Now it's fixed. Flycheck detects `clippy` and works correctly with `rust-clippy` checker.
Screen Shot 2022-06-09 at 17 53 23

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -11433,9 +11433,12 @@ versions inferior to 1.25)."
   "Whether Cargo has COMMAND in its list of commands.
 
 Execute `cargo --list' to find out whether COMMAND is present."
-  (let ((cargo (funcall flycheck-executable-find "cargo")))
-    (member command (mapcar #'string-trim-left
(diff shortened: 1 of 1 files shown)


hrls:
https://github.com/flycheck/flycheck/issues/1916

hrls:
This PR is similar to https://github.com/flycheck/flycheck/pull/1917 but regexp is more strict than simple string prefix check.

cpitclaudel:
Thanks, and sorry for the delay.  I think the code should probably save and restore the match data, or else it will silently corrupt it.  The alternative would be to use a string replacement instead of a match.

PR #1950, opened 2022-06-12
feat: left arrows bitmaps for fringes

hrls:
Just reversed original ones. Useful for right fringes. Now this can be configured as simple as:
```elisp
(flycheck-redefine-standard-error-levels
     nil
     (cons 'flycheck-fringe-bitmap-double-arrow-left
           'flycheck-fringe-bitmap-double-arrow-left-hi-res))
```

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4327,7 +4327,17 @@ show the indicator."
    #b00110110
    #b01101100
    #b11011000]
-  "Bitmaps used to indicate errors in the fringes.")
+  "Bitmaps used to indicate errors in the left fringes.")
+
+(defconst flycheck-fringe-bitmap-double-left-arrow
+  [#b00011011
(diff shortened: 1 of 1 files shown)


cpitclaudel (approved):
Neat, thanks!

cpitclaudel:
Very nice first contribution, thanks!  Maybe we can update the manual to mention this somewhere (or, even better, use this by default if we detect that we're displaying in the right fringe?

hrls:
I'm not sure which exact function is supposed to detect this. Looks like we can add a simple flip into `flycheck-set-indication-mode` when mode is `right-fringe` and current fringe bitmaps are `flycheck-fringe-bitmap-double-arrow{,-hi-res}` as by default. First condition is easy, but the second one … 🤔

PR #1955, opened 2022-07-18
Add flycheck-python-mypy-python-executable

wyuenho:
If mypy is installed globally in an isolated virtualenv, such is the case for `pre-commit` or `pipx` installed mypy, mypy will not be able to discover the installed packages in your project, even if working-directory is set, because the virtualenv for the project present in the working directory may very well be completely different.

Due to this complication, hardcoding `python_executable` in mypy's configuration is not feasible either, because it often needs to be discovered dynamically by invoking something like `poetry run which python`, so it needs to be supplied on the command line.

This PR allows users of flycheck to configure the python executable to perform mypy import discovery as they see fit.

### DIFF
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -1061,6 +1061,11 @@ to view the docstring of the syntax checker.  Likewise, you may use
          Set to ``null-device`` to disable writing cache directories
          entirely.
 
+      .. defcustom:: flycheck-python-mypy-python-executable
+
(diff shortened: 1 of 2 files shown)


wyuenho:
@cpitclaudel @bbatsov @marsam @jcs090218 can somebody take a look at this please?

marsam (approved):
LGTM. Thank you!

PR #1968, opened 2022-08-02
Set run_name to __main__ when running python modules

gaetano-guerriero:
By default run_name is set to package.__main__ when running a python
package with runpy.

This causes issues with libraries checking "if __name__ ==
'__main__':" in their __main__.py file (flake8 5 is such an example).

Setting run_name to '__main__' is more similar to running the package
with "python -m ...", check also "_run_module_as_main" in runpy .


cpitclaudel (approved):

cpitclaudel:
Thanks a lot!

PR #1970, opened 2022-08-15
Require emacs >= 25.3

wyuenho:
Just to make it clear which version flycheck actually supports

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -10,7 +10,7 @@
 ;; URL: http://www.flycheck.org
 ;; Keywords: convenience, languages, tools
 ;; Version: 33-cvs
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Isn't this a bit much? While our testing situation is admittedly not great, we haven't taken any steps recently to break versions < 25.3, so disconnecting users on outdated versions seems unnecessary?

bbatsov:
Emacs 24 is really ancient (9.5 years old), so the proposal does make sense to me. 25.3 is exactly 5 years old at this point, so it makes for a reasonable target.

jcs090218:
I second this; this sounds reasonable. I would go even further to support Emacs `26.1` or above.

PR #1972, opened 2022-08-15
Refine flycheck-display-errors lifecycle

wyuenho:
This PR fixes a 7-8 year-old bug where the `flycheck-display-error-at-point-timer` has been calling `flycheck-display-errors-function` with error overlays marked for deletion, so even when an error has been fixed, this particular code path will still display the fixed error in the echo area.

Reproduction:

0. Install flycheck, do not configure it at all.
1. Turn on `flycheck-mode` on any elisp buffer.
2. Make a syntax error such as
    ```elisp
    (de |fun foo ())
    ```
3. Observe flycheck has highlighted `de` and `fun` and there's a `reference to free variable 'fun'` message displayed in the echo area.
4. Press backspace to fix the error so it becomes `(defun foo ())`
5. Observe the highlighting on  is gone on the symbol, but the message is being displayed in the echo area.

Expectation:

After the error has been fixed, the stale error message should not continue to display in the echo area.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -428,6 +428,18 @@ If set to nil, do not display errors at all."
   :package-version '(flycheck . "0.13")
   :risky t)
 
+(defcustom flycheck-clear-displayed-errors-function #'flycheck-clear-displayed-error-messages
+  "Function to hide error message displayed by `flycheck-display-errors-function'.
+
+If set to a function, it will be called with no arguments to
+clear all displayed errors at point."
+  :group 'flycheck
+  :type '(choice (const :tag "Clear displayed error messages"
+                        flycheck-clear-displayed-error-messages)
+                 (function :tag "Clear displayed errors function"))
+  :package-version '(flycheck . "34.2")
+  :risky t)
+
 (defcustom flycheck-help-echo-function #'flycheck-help-echo-all-error-messages
   "Function to compute the contents of the error tooltips.
 
@@ -2968,7 +2980,7 @@ Slots:
     ;; the focus hooks only work on Emacs 24.4 and upwards, but since undefined
     ;; hooks are perfectly ok we don't need a version guard here.  They'll just
     ;; not work silently.
-    (post-command-hook . flycheck-maybe-display-error-at-point-soon)
+    (post-command-hook . flycheck-display-error-at-point-soon)
     (focus-in-hook     . flycheck-display-error-at-point-soon)
     (focus-out-hook    . flycheck-cancel-error-display-error-at-point-timer)
     (post-command-hook . flycheck-hide-error-buffer)
@@ -3374,6 +3386,7 @@ current syntax check."
     (flycheck-stop))
   (flycheck-delete-all-overlays)
   (flycheck-clear-errors)
+  (flycheck-clear-displayed-error-messages)
   (flycheck-error-list-refresh)
   (flycheck-hide-error-buffer))
 
@@ -5545,6 +5558,11 @@ non-nil."
   (when flycheck-display-errors-function
     (funcall flycheck-display-errors-function errors)))
 
+(defun flycheck-clear-displayed-errors ()
+  "Clear errors using `flycheck-clear-displayed-errors-function'."
+  (when flycheck-clear-displayed-errors-function
+    (funcall flycheck-clear-displayed-errors-function)))
+
 (defvar-local flycheck-display-error-at-point-timer nil
   "Timer to automatically show errors.")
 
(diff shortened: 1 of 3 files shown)


cpitclaudel:
I think I see the error; however, I don't think this fix is right: `(flycheck-delete-marked-overlays)` is delayed to avoid flickering, and indeed this patch reintroduces flickering when a change does not fix an error (the overlays disappear and subsequently reappear).

Additionally, I'm not positive that not showing the error is the right thing to do: if I fix one error but another one persists (until the checker returns), shouldn't we still display that error?

cpitclaudel:
Thanks a lot!  I'm traveling this month but I can get to this early September.  I hope that's OK.

wyuenho:
@cpitclaudel have you had a chance to test this out?

jimmy-wong-beamery:
@cpitclaudel ping again

wyuenho:
@cpitclaudel ping again

wyuenho:
@cpitclaudel have you had a chance to take a look at this yet?

jcs090218:
Can you rebase this, so we test it with our latest CI? Thanks!

cc @bbatsov for the review, another maintainer has the push permission.

bbatsov:
I'd also suggest squashing the related commits together for the review.

wyuenho:
Rebased. I'm not sure what Eask is and why it is failing to download packages and breaking the tests.

I'll squash after review so you have a history to look at.

bbatsov:
> Rebased. I'm not sure what Eask is and why it is failing to download packages and breaking the tests.

It's replacement for Cask that's actively maintained.

flycheck.el
```diff
@@ -3329,6 +3341,7 @@ current syntax check."
     (flycheck-stop))
   (flycheck-delete-all-overlays)
   (flycheck-clear-errors)
+  (flycheck-clear-displayed-error-messages)
```
bbatsov, line 3344: Isn't this something that `flycheck-clear-errors` should normally be doing? I find it weird to have to clear the errors and the displayed error messages separately.

flycheck.el
```diff
@@ -5493,6 +5506,11 @@ non-nil."
   (when flycheck-display-errors-function
     (funcall flycheck-display-errors-function errors)))
 
+(defun flycheck-clear-displayed-errors ()
+  "Clear errors using `flycheck-clear-displayed-errors-function'."
+  (when flycheck-clear-displayed-errors-function
```
bbatsov, line 5511: Do we really need this to be customizable? Why would users want a different behaviour?

jcs090218:
The CI is failing due to the upstream changes, I've opened the PR for the fix, see #2007.

@bbatsov Can you review and merge this? Thanks!

Sorry for the inconvenience! 😓

bbatsov:
The fix is merged.

cpitclaudel:
This is a subtle part of flycheck.  I'm very thankful for the changes and I'm sorry I haven't gotten around to reviewing them :'(  Hopefully @bbatsov and @jcs090218 can test / review and report.

flycheck.el
wyuenho: Not all error messages are displayed the same way. There are many packages that adjusts how the messages are displayed, so use childframes, others use overlays. Displaying and clearing messages have to be a symmetric operation operating on the same data type in order to be reliable.

flycheck.el
wyuenho: Same reason as https://github.com/flycheck/flycheck/pull/1972#issuecomment-1216166601. Clearing the errors data structures and displayed messages have to be distinct in order to prevent flickering.

flycheck.el
wyuenho: Also, see a usage example here

https://github.com/flycheck/flycheck-inline/pull/22/files

wyuenho:
@bbatsov have I addressed your questions?

bbatsov:
@wyuenho Can you rebase on `master`, so I can see what's the current state of the CI?

flycheck.el
```diff
@@ -454,6 +454,18 @@ If set to nil, do not display errors at all."
...
+(defcustom flycheck-clear-displayed-errors-function #'flycheck-clear-displayed-error-messages
+  "Function to hide error message displayed by `flycheck-display-errors-function'.
+
+If set to a function, it will be called with no arguments to
+clear all displayed errors at point."
+  :group 'flycheck
+  :type '(choice (const :tag "Clear displayed error messages"
+                        flycheck-clear-displayed-error-messages)
+                 (function :tag "Clear displayed errors function"))
+  :package-version '(flycheck . "33")
```
bbatsov, line 466: This should be 34, now that 33 is out.

bbatsov:
Before we can merge this we also need:

- a changelog entry
- some entry in the user docs, so people would actual find out about this new configuration variable

flycheck.el
```diff
@@ -5605,6 +5613,14 @@ Hide the error buffer if there is no error under point."
       (save-selected-window
         (quit-window nil window)))))
 
+(defun flycheck-clear-displayed-error-messages ()
+  "Clear error messages displayed by `flycheck-display-error-messages'."
+  (unless (null flycheck--last-displayed-message)
```
bbatsov, line 5618: Maybe use `when` here instead?

flycheck.el
```diff
@@ -5605,6 +5613,14 @@ Hide the error buffer if there is no error under point."
...
+    (if (and (stringp flycheck--last-displayed-message)
+             (equal (current-message) flycheck--last-displayed-message))
```
bbatsov, line 5620: Why not do a string comparison instead?

flycheck.el
```diff
@@ -5561,6 +5565,8 @@ and if the echo area is not occupied by minibuffer input."
   "Flycheck error messages"
   "Major mode for extended error messages.")
 
+(defvar flycheck--last-displayed-message nil)
```
bbatsov, line 5568: I'd add a docstring here.

bbatsov:
I left a bit of extra feedback. You'll also need to rebase this on top of `master`, as I was doing a lot of changes lately.

bbatsov:
@wyuenho Friendly ping :-)

wyuenho:
@bbatsov please take a look now. I'm not sure what's happening with the tests, the failures make no sense to me. This is as far as I can push this.

flycheck.el
wyuenho: Because it is the result returned from `display-message-or-buffer`, which can be a string or a window.

wyuenho:
@bbatsov should be good to go now

Issue #1983, opened 2022-11-09
Mypy error messages are truncated if mypy config has `pretty = true`


spookylukey:
## Checklist
- [X] I have checked existing issues for potential duplicates before creating this one.
- [X] I have read the [Troubleshooting guide][].

## Bug description
If mypy config has `pretty = true`, then messages appear truncated. For example, for this file:
```python
def myfunc(a: int, b: str) -> int:
    return a


def foo() -> None:
    print(myfunc(1, 1))
```

Normally flycheck shows this:
```
Argument 2 to "myfunc" has incompatible type "int"; expected "str" (python-mypy)
```

If you add `pretty = true` to mypy.ini, however, you get this:

```
Argument 2 to "myfunc" has incompatible (python-mypy)
```

In some cases it gets truncated to just two or three words of the original message, making it very hard to decipher.

## Steps to reproduce
As described.

- [X] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

## Expected behavior
You should get the full error message regardless of the the `pretty` setting.

## System configuration
```
Syntax checkers for buffer flycheck-mypy-bug.py in python-mode:

First checker to run:

  python-flake8
    - may enable:         yes
    - executable:         Found at /home/luke/devel/clients/datapane/datapane-hosted/dp-server/.venv/bin/flake8
    - configuration file: Not found
    - next checkers:      python-pylint, python-mypy

Checkers that may run as part of the first checker's chain:

  python-mypy
    - may enable:         yes
    - may run:            t
    - executable:         Found at /home/luke/devel/clients/datapane/datapane-hosted/dp-server/.venv/bin/mypy
    - configuration file: Found at "/home/luke/devel/clients/datapane/datapane-hosted/dp-server/mypy.ini"

Checkers that could run if selected:

  python-pycompile  select
    - may enable:    yes
    - executable:    Found at /home/luke/devel/clients/datapane/datapane-hosted/dp-server/.venv/bin/python3
    - next checkers: python-mypy

  python-pyright  select
    - may enable: yes
    - executable: Found at /home/luke/.local/bin/pyright

Checkers that are compatible with this mode, but will not run until properly configured:

  python-coverage
    - may enable: yes
    - may run:    nil
    - report:     Could not find coverage file. (Hint: use ‘M-x python-coverage-select-coverage-file’ to choose manually.)

The following syntax checkers are not registered:
  - bazel-build-buildifier
  - bazel-module-buildifier
[...]

PR #1984, opened 2022-11-09
Fix message truncation for python-mypy due to pretty formatting

spookylukey:
If the the user has "pretty = true" in their mypy config it can result in truncation of the messages reported. We invoke mypy with --no-pretty to override this.

Minimum version is bumped to 0.730 (2019-09-26) because this is the version in which --pretty and --no-pretty were added:
https://mypy-lang.blogspot.com/2019/09/mypy-730-released.html

Fixes GH-1983 and closes GH-1983

### DIFF
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -1046,7 +1046,7 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       .. note::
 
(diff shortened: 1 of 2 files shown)


cpitclaudel (approved):
Thanks!

PR #1987, opened 2022-11-18
option to control whether to display error at point after checking

taquangtrung:
Hi, 

Currently, if there is an error at the current point of the current buffer, Flycheck will always display it after finishing syntax checking.

I'm using [flycheck-posframe](https://github.com/alexmurray/flycheck-posframe) and it is quite inconvenient when the popup window automatically displays and hides the current text (when the error message spans several lines).

So, I introduced an option `flycheck-display-errors-at-point-after-checking` in this PR. When it is set to `nil`, Flycheck will not automatically display the error after finishing syntax checking.

Can you take a look at this PR and merge it if possible?

Thanks!


bbatsov:
I'm OK with the proposed change, but I'd suggest to documented the config variable and it's potential use-case somewhere, otherwise no one will ever find it. This also needs a changelog entry.

taquangtrung:
@bbatsov : I just added the documentation on changes and use cases. Can you check if these are ok?

bbatsov:
> I just added the documentation on changes and use cases. Can you check if these are ok?

They look OK, but I'd also suggest to add this somewhere in the online docs, so more people would actually see it.

taquangtrung:
@bbatsov : Thanks for the suggestions! I've just updated it too!

PR #1994, opened 2023-01-17
Fix `flycheck-error-level-interesting-p` function

benzanol:
Previously, errors below the minimum severity could get marked as interesting even if errors above the minimum severity existed, as long as no errors which were exactly the minimum severity existed.
This means that if you set your minimum navigation severity to `warning`, and your file had only errors and hints, navigating would still jump to hints even though they are below the minimum severity.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -4668,6 +4668,18 @@ level."
           (push (cons level 1) counts-by-level))))
     counts-by-level))
 
+(defun flycheck-has-at-least-errors-p (errors level)
+  "Check if there are ERRORS at least as severe as LEVEL."
+  (let ((severity (flycheck-error-level-severity level)))
+    (seq-some (lambda (e) (>= (flycheck-error-level-severity
+                                  (flycheck-error-level e))
+                                 severity))
+                 errors)))
+
+(defun flycheck-has-at-least-current-errors-p (level)
+  "Check if there are current errors at least as severe as LEVEL."
+  (flycheck-has-at-least-errors-p flycheck-current-errors level))
+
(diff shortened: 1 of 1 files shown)


bbatsov:
Can you illustrate this with some example, so I can understand better the problem you're trying to fix here?

benzanol:
> Can you illustrate this with some example, so I can understand better the problem you're trying to fix here?

Set `flycheck-navigation-minimum-level` to `warning`.
Open a file which has errors with level `error` (more than warning) as well as `info` (less than warning,) but no warnings.
For example, the following rust code will have an error on `4` and an info on `&str`:
```rust
fn main() {
    let a: &str = 4;
}
```
`flycheck-previous-error` and `flycheck-next-error` will navigate to the info and to the error, when it should just go to the error, because of the minimum navigation level.
This is fixed by the PR.

Hope this helps!

bbatsov:
Got it. I'd suggest adding some spec covering this, if there are already some tests related to the `flycheck-navigation-minimum-level` behavior, so we won't run into regressions down the road. The change also needs a changelog entry.

bbatsov:
I've also noticed that the older #1870 tackles the same problem, so you can check the solution there as before we proceed. (at a glance it seemed pretty similar to me)

bbatsov:
@benzanol Friendly ping :-)

bbatsov:
Closing — this was fixed differently in master (commit 4615773d) which strictly enforces `flycheck-navigation-minimum-level` by comparing error severity levels. Thank you for the contribution!

Issue #1957, opened 2022-07-19
Replacing Travis CI


bbatsov:
Travis CI has been discontinued and we need some alternative. It seems pretty much all PRs are block on this. That's the final build that ever ran:

![image](https://user-images.githubusercontent.com/103882/179763085-381fa563-ef0f-4e21-8bdb-3536cf878f71.png)

Should we just remove it for now while we're in the process of identifying a replacement?

cpitclaudel:
Yes, let's remove it.  For now I've removed the branch protection.
FWIW, I think it's good to run the flycheck tests, but the tests that verify that we can parse the output of each and every checker are too much work (because they depend on the exact phrasing of each message.

bbatsov:
Supposedly there's still some Travis CI that can be used with relatively few changes - https://docs.travis-ci.com/user/migrate/open-source-repository-migration Perhaps this should be explored at some point. I've noticed that one major issue with the tests is that we have to install a ton of tools to check their output. I assume those are the tests that you're referring to.

jcs090218:
I would love to work on this if flycheck would like to use [Eask](https://github.com/emacs-eask/cli) for their CI tests. Eask is very similar to Cask, so it shouldn't be hard to do it.

BTW, Eask has built-in support to read files like Eask.24, Eask.25.1, etc. Multiple Eask-file co-exists is possible and should make CI tests a bit easier for this package.

bbatsov:
These days there's also Eldev, which looks similar to Eask. Not sure what advantages Eask might have over Eldev.

jcs090218:
I know Eldev, but I am not very familiar with it. Can you explain the advantages of using Eldev?

Eask is similar to Cask if we compare two files:

- Example Cask file: https://github.com/flycheck/flycheck/blob/master/Cask (from this repo)
- Example Eask file: https://github.com/emacs-lsp/lsp-mode/blob/master/Eask

The commands are similar as well, 

```bash
$ cask eval
$ cask install

$ eask eval
$ eask install
```

Eask is intended to build similar to Cask but with better cross-platform capability (using node), simpler wrapper, etc.

bbatsov:
Well, I can't speak for Eask (having never used it), but compared to Cask I like that `eldev` supports things like linting and running tests with various tools out-of-the-box. It also have configuration that's written in Elisp, instead of some DSL, which for me is a feature. 

[...]

PR #2002, opened 2023-02-13
ci: Start off by adding Eask-file and basic tests

jcs090218:
For #1957.

From what `@bbatsov` has suggested in https://github.com/flycheck/flycheck/issues/1957#issuecomment-1427094106, I have moved the entire workflows from Travis CI to GHA. 

#### Major changes

- Add tests for macOS and Windows (basic tests for Windows only)
- Add Eask-file
- Add three workflow files
  - `.github/workflows/docs.yml`, equivalent to `- JOB="build-manual"`
  - `.github/workflows/lint-python.yml`, equivalent to `- JOB="lint-python"`
  - `.github/workflows/test.yml`
- Replace `Cask` with `Eask` in Makefile
- Fix `buttercup` and `ert` tests (it was broken)

#### Questions (from me)

- Should we remove any Travis CI related files, `.travis.yml` and `.travis`?

I want to see what maintainers think before moving on.

cc @cpitclaudel @bbatsov


jcs090218:
Tests aren't running in this repo, but you can see it here https://github.com/jcs-PR/flycheck/actions.

flycheck.el
```diff
@@ -10,7 +10,7 @@
 ;; URL: http://www.flycheck.org
 ;; Keywords: convenience, languages, tools
 ;; Version: 33-cvs
-;; Package-Requires: ((dash "2.12.1") (pkg-info "0.4") (let-alist "1.0.4") (seq "1.11") (emacs "24.3"))
+;; Package-Requires: ((emacs "26.1") (dash "2.12.1") (pkg-info "0.4") (let-alist "1.0.4") (seq "1.11"))
```
jcs090218, line 13: The default version on Ubuntu 18.04 is still `25.x`, but it will only support until April 2023. I hope this is reasonable changes. LMKWYT?

Related to #1970.

bbatsov: Yeah, that's fine IMO.

bbatsov:
> Should we remove any Travis CI related files, .travis.yml and .travis

Yeah, we can safely remove all the legacy Travis-related files.

flycheck.el
cpitclaudel: No strong feelings, but would there be a downside to using 25.3?  I have a few old machines on which I use 25.

bbatsov: I guess as a middle ground we can target 25.x, but omit it from the CI. At any rate - I don't feel strongly about this.

jcs090218: > No strong feelings, but would there be a downside to using 25.3? I have a few old machines on which I use 25.

There won't be a downside using `25.x` (user-wise), although Eask only supports `26.1` or above.

I would encourage people to upgrade to `26.1` or above. Emacs has significantly improved throughout the years. Like: 

1. Some useful macros `when-let*` and `if-let*`
2. `let-alist` and `seq ` are built-in now
3. `set` and `make-local-variable` can replace with `defvar-local` or `setq-local`.
4. It's faster

, etc

In terms of maintenance, it's better and easier for us. Overall, it's just a nice Emacs version to have/target.

@cpitclaudel LMKWYT? If you still prefer `25.x`, I will perform the change. ;)

cpitclaudel: Flycheck is too large (I think) to play fast an loose with its supported version set; we have lots of folks that depend on us, and we have traditionally supported a lot of relatively old versions.  I think sticking with 25.x for now would be safer.

(I agree with you that 26 is nicer)

jcs090218: 👍 I have reverted and now supports `25.1`!

bbatsov:
We should probably also add badges for each workflow to the README.

jcs090218:
I think those badges aren't created yet.

- https://github.com/flycheck/flycheck/actions/workflows/docs.yml/badge.svg
- https://github.com/flycheck/flycheck/actions/workflows/lint-python.yml/badge.svg
- https://github.com/flycheck/flycheck/actions/workflows/test.yml/badge.svg

#### In markdown (broken images)

[Docs](https://github.com/flycheck/flycheck/actions/workflows/docs.yml)
[Lint Python](https://github.com/flycheck/flycheck/actions/workflows/lint-python.yml)
[CI](https://github.com/flycheck/flycheck/actions/workflows/test.yml)

Do you want me to add them beforehand?

 I am thinking of creating another PR for this since this PR is already huge. Of course, we can do this (documenting) in the same PR as well.

jcs090218:
Let me know what else I need to do to get this proceeds! Either one of these states,

1. We are still reviewing, please be patience.
2. Let's start documenting in this PR
3. Anything else?

Like `@cpitclaudel` has commented in https://github.com/flycheck/flycheck/issues/1957#issuecomment-1427091763. Is the bar too high? too low? etc.

bbatsov:
> I am thinking of creating another PR for this since this PR is already huge. Of course, we can do this (documenting) in the same PR as well.

Fine by me.

bbatsov (approved):

PR #2007, opened 2023-02-26
ci: Fix arguments that contain whitespaces

jcs090218:
There is an upstream change from [Eask](https://github.com/emacs-eask/cli), update this to fix CI.

And, I realized the buttercup tests was never completely run on Windows,

```
Ran 3 out of 113 specs, 0 failed, in 5.56s.
```

This PR should fix that as well.

### DIFF
--- a/test/specs/test-melpa-package.el
+++ b/test/specs/test-melpa-package.el
@@ -63,7 +63,7 @@ version."
           (when version
             (let* ((name (format "flycheck-%s" version))
                    (url (format "http://melpa.org/packages/%s.tar" name)))
-              (with-timeout (5)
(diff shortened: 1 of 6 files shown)


test/specs/test-melpa-package.el
```diff
@@ -63,7 +63,7 @@ version."
           (when version
             (let* ((name (format "flycheck-%s" version))
                    (url (format "http://melpa.org/packages/%s.tar" name)))
-              (with-timeout (5)
+              (with-timeout (30)
```
jcs090218, line 66: Not sure why this is reverted. 😕

bbatsov (approved):

PR #2009, opened 2023-03-05
markdownlint-cli: Add default config file names and rule selection in args

berquist:
* The default configuration filenames are taken from the project readme.
* `--enable` and `--disable` allow specifying names of rules without putting them in a configuration file.

Please let me know if you'd like this tested.


jcs090218 (approved):
I've a quick glance, LGTM and CI passes. However, I don't have the push permission, so we will have to wait for other maintainers to review. ;)

cpitclaudel (approved):

cpitclaudel:
Thanks!

Issue #2010, opened 2023-03-15
How do I customize the mark char used on the border?


LionyxML:
## Checklist
- [X] I have searched existing issues and StackOverflow for solutions.
- [ ] I have read the [Flycheck manual][].
- [ ] I have read the [Troubleshooting Guide][].
- [X] I have checked my issue has not been reported yet.

## Problem description
I can't find anything about it, sorry if this is trivial.

I moved my flycheck marks to the right so it won't be conflicting with git-gutter.

The thing is I use it on the border, and not the fringe, so I can still use it on the terminal.

And I can't find where I can customize this chars, since they're "pointing" to the right, and I want it to change to something else:

My right border:
![image](https://user-images.githubusercontent.com/16169950/225319526-05b9e0bb-2ae6-4f23-bc50-4ffdd7722683.png)

Everything I manage to find about this mark is related to bitmap on the fringe, not a char on the border. Is this even customizable?

## Steps to reproduce
1. make an error
2. move the flycheck marks to the right

- [ ] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

## Expected behavior

Emacs configuration:
- [X] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs
- [ ] Other shared configuration

## Additional notes

[Flycheck manual]:  https://www.flycheck.org/en/latest/user/troubleshooting.html
[Troubleshooting guide]:  https://www.flycheck.org/en/latest/user/troubleshooting.html

LionyxML:
I am kind of suspecting this is a bug.

There's a `flycheck-fringe-bitmap-double-left-arrow` and its high res version, but they seem to be unused in the code.

At least there's nothing that forces it to be used when the margin or fringe is set to the `right`.

https://github.com/flycheck/flycheck/blob/master/flycheck.el#L4409

I think it should've been something around here to conditionally choose the right "arrow".

LionyxML:
Seemed to be an easy fix. I just opened this PR :)

PR #2012, opened 2023-03-19
Fix arrows directions

LionyxML:
When setting fringe or border to 'right', the provided marker kept pointing right.

This aims to change the default behaviour of
`flycheck-redefine-standard-error-levels` to make into account the selected `flycheck-indication-mode`.

So if the left-margin or left-fringe is set, everything stays the same, buf if they're set to the right-margin or rigt-fringe an arrow pointing to the left will be used.

Diff may show other little fixes, this was done by `make format`, I thought there's nothing bad about it and kept it on this commit.

On macOS eask was trickie to set up, it required an ignore since the `shut-up` package keeps telling the ls of macOS has no --dired (even tough it could be smarter and point to gls (gnu ls)). 

The docker image also gave me an error:
```
docker build --build-arg EMACS_VERSION=26.2 --tag tools-and-emacs:26.2 -f .travis/tools-and-emacs .
[+] Building 0.0s (1/2)                                                                                                                       
 => ERROR [internal] load build definition from tools-and-emacs                                                                          0.0s
 => => transferring dockerfile: 54B                                                                                                      0.0s
------
 > [internal] load build definition from tools-and-emacs:
------
failed to solve with frontend dockerfile.v0: failed to read dockerfile: error from sender: resolve : lstat .travis: no such file or directory
```

In order to test it I manually byte compiled flycheck, moved the .elc to my .emacs.d/ and messed with:
```
(setq-default flycheck-indication-mode 'right-margin)
(add-hook 'flycheck-mode-hook #'flycheck-set-indication-mode)
```

Cycling trough left-margin, right-margin, left-fringe, right-fringe and opening a toy file in python-mode:

right-margin:
image

left-margin:
image

left-fringe:
image

right-fringe:
image

Also nicely working on terminal Emacs on margin modes:
image

This aims to complete the issue #2010

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -2163,7 +2163,7 @@ Signal an error, if any property has an invalid value."
                 (lambda ()
                   ;; Run predicate in the checker's default directory
                   (let ((default-directory
-                          (flycheck-compute-working-directory symbol)))
+                         (flycheck-compute-working-directory symbol)))
                     (funcall predicate)))))
           (real-enabled
            (lambda ()
@@ -2171,7 +2171,7 @@ Signal an error, if any property has an invalid value."
                  (or (null enabled)
                      ;; Run enabled in the checker's default directory
                      (let ((default-directory
-                             (flycheck-compute-working-directory symbol)))
+                            (flycheck-compute-working-directory symbol)))
                        (funcall enabled)))
                (lwarn 'flycheck
                       :warning "%S is no valid Flycheck syntax checker.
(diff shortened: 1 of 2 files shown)


cpitclaudel:
This patch will only work when the customization of the indication mode happens before Flycheck is loaded, right?  I worry that this will be confusing for users.

LionyxML:
Hello there @cpitclaudel ! 

Hmm, Sorry, I am not sure I am following you. :( 

Also, I've got for no reason still some checkers pointing to the wrong side when dealing with `lsp-mode` starting and re-starting.

bbatsov:
@LionyxML Can you rebase on top of `master` and remove whitespace changes that are not relevant to your patch, so it's easier to examine the suggested changes?

LionyxML:
Nice. As soon as I have some free time I'll take a look into it.

PR #2029, opened 2023-08-11
test: Emacs 29.1

jcs090218:
Test for Emacs 29.1.


jcs090218:
There is an error in current Windows snapshot, so we can ignore that for now.

@bbatsov Can you take a look? Thank you!

marsam (approved):

Issue #1757, opened 2020-05-06
Support for org-lint


zoechi:
Thank you for taking the time to improve Flycheck.

## Checklist
- [x] I have checked existing issues for potential duplicates before creating this one.

## Feature description
A clear and concise description of what you want Flycheck to do.

`M-x` `org-lint` provides a list of issues in *.org files.
This should be integrated with flycheck.

## Describe alternatives you've considered
I only found https://emacs.stackexchange.com/questions/29496/automatically-run-org-lint-through-flycheck

## Additional context
Add any other context or screenshots, mockups etc.

cpitclaudel:
Neat idea.  Would you like to give it a try?
The solution on stackexchange isn't ideal because it'll freeze your Emacs; a better approach would be to do handle org-lint like checkdoc.

zoechi:
I'd love to do it, but I'm still a Emacs rookie and spend way too much time tweaking my config already.

cpitclaudel:
I should be able to help :) We have pretty decent documentation at https://www.flycheck.org/en/latest/developer/developing.html, I think.

I think it won't be a very hard project.  The definition for checkdoc is about this big:

```elisp
(flycheck-define-checker emacs-lisp-checkdoc
  "An Emacs Lisp style checker using CheckDoc.

The checker runs `checkdoc-current-buffer'."
  :command ("emacs" (eval flycheck-emacs-args)
            "--eval" (eval (flycheck-sexp-to-string
                            (flycheck-emacs-lisp-checkdoc-variables-form)))
            "--eval" (eval flycheck-emacs-lisp-checkdoc-form)
            "--" source)
  :error-patterns
  ((warning line-start (file-name) ":" line ": " (message) line-end))
  :modes (emacs-lisp-mode)
  :enabled flycheck--emacs-lisp-enabled-p)
```

And the definition of the bit of lisp that we run in the emacs subprocess is quite small too:

```elisp

(defconst flycheck-emacs-lisp-checkdoc-form
  (flycheck-prepare-emacs-lisp-form
    (unless (require 'elisp-mode nil 'no-error)
      (require 'lisp-mode))
    (require 'checkdoc)
    (let ((source (car command-line-args-left))
          (process-default-directory default-directory))
      (with-temp-buffer
        (insert-file-contents source 'visit)
        (setq buffer-file-name source)
        (setq default-directory process-default-directory)
        (with-demoted-errors "Error in checkdoc: %S"
[...]

PR #2031, opened 2023-09-03
Support org lint

RlckAstley:
Adding a pull request to support org-lint. The code was originally written by @ several years ago @czikus  and proposed by @zoechi, see https://github.com/flycheck/flycheck/issues/1757

CI/CD and make specs fails with `(seq-difference flycheck-checkers checkers)' to be `equal' to `nil', but instead it was `(org-lint)' which does not match because: (different-types (org-lint) nil). I'm having a hard time understanding why this is happening. From what I can tell, my style and definition matches spec. I've spent a few hours trying to figure this out, but need to move on. Will try to get on gitter, but seems a bit dead. Also, I'm on liberachat #emacs under RickAstley.  Apologies for a premature pull, but I need some help and am a noob.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -10134,6 +10134,16 @@ string is a module to `use' in Perl."
   :safe #'flycheck-string-list-p
   :package-version '(flycheck . "32"))
 
+  (flycheck-define-checker org-lint
+    "Org buffer checker using `org-lint'."
+    :command ("emacs" (eval flycheck-emacs-args)
+              "--eval" (eval flycheck-org-lint-form)
+              "--" source)
+    :error-patterns
+    ((error line-start line ": " (message) line-end))
+    :modes (org-mode))
+
+
 (flycheck-define-checker perl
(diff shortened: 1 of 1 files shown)


cpitclaudel:
> I'm having a hard time understanding why this is happening. From what I can tell, my style and definition matches spec. I've spent a few hours trying to figure this out, but need to move on

Sorry for the delay.  All checkers that are defined should also appear in the list flycheck-checkers.  I think this is what this test is complaining about.

RlckAstley:
> > I'm having a hard time understanding why this is happening. From what I can tell, my style and definition matches spec. I've spent a few hours trying to figure this out, but need to move on
> 
> Sorry for the delay. All checkers that are defined should also appear in the list flycheck-checkers. I think this is what this test is complaining about.

Hey cpitclaudel, no prob. Sorry for the noobish pull, I'll check on this and see.

bbatsov:
@RlckAstley Any updates here?

PR #2037, opened 2023-10-06
Replace pkg-info with package-get-version

Thaodan:
As suggested in https://github.com/flycheck/flycheck/pull/1896#issuecomment-1190724395

Only downside it was "only" added in Emacs 27.1.

A workaround would to use compat on older Emacs.


bbatsov (approved):

flycheck.el
```diff
@@ -1287,7 +1287,7 @@ If the version number could not be determined, signal an error,
 if called interactively, or if SHOW-VERSION is non-nil, otherwise
 just return nil."
   (interactive (list t))
-  (let ((version (pkg-info-version-info 'flycheck)))
+  (let ((version (package-get-version)))
```
bbatsov, line 1290: I think we should make this conditional, so we don't have to require Emacs 27 for this trivial function. E.g. we can just add the version as a variable and fall back to it in the absence of this function.

flycheck.el
```diff
@@ -2645,7 +2655,7 @@ to enable disabled checkers.")))
       (fill-region-as-paragraph (point) end)))
 
   (princ "\n\n--------------------\n\n")
-  (princ (format "Flycheck version: %s\n" (flycheck-version)))
+  (princ (format "Flycheck version: %s\n" (flycheck--pkg-version)))
```
bbatsov, line 2658: I guess you don't really need to change this, as `flycheck-version` still will return this as a string.

bbatsov:
You'll need to rebase this on top of `master`.

Issue #2024, opened 2023-07-12
Terraform-tflint command line arguments have changed


arcsteveio:
Thank you for taking the time to report an issue and improve Flycheck.  This template is for **actual bugs you observed**.  If you have trouble setting up Flycheck, or if you have a question, please use the relevant issue template instead.

## Checklist
- [x] I have checked existing issues for potential duplicates before creating this one.
- [x] I have read the [Troubleshooting guide][].

## Bug description
The latest version of tflint 0.47.0 changed how the command line arguments work such that it no longer accepts a file to lint from the command line. Instead you need to prefix the files with a `--filter`. 

Here's the output from flycheck-compile:

```
-*- mode: compilation; default-directory: "~/dev/examples/storage/s3/" -*-
Compilation started at Wed Jul 12 12:44:27

/usr/local/bin/tflint --format\=json /Users/arcsteveio/dev/examples/storage/s3/main.tf
{"issues":[],"errors":[{"message":"Command line arguments support was dropped in v0.47. Use --chdir or --filter instead.","severity":"error"}]}
Compilation exited abnormally with code 1 at Wed Jul 12 12:44:27
```

If I take that same command and put a `--filter` in front of the file it appears to work fine:

```
❯ /usr/local/bin/tflint --format\=json --filter=/Users/arcsteveio/dev/examples/storage/s3/main.tf
{"issues":[],"errors":[]}%
```

## Steps to reproduce

Assuming Flycheck and tflint are installed.

1. Open a Terraform file.
2. Change the file and save it.
3. See error: 

```
Suspicious state from syntax checker terraform-tflint: Flycheck checker terraform-tflint returned 1, but its output contained no errors: {"issues":[],"errors":[{"message":"Command line arguments support was dropped in v0.47. Use --chdir or --filter instead.","severity":"error"}]}
Try installing a more recent version of terraform-tflint, and please open a bug report if the issue persists in the latest release.  Thanks!
```

- [x] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

## Expected behavior
Flycheck uses tflint and marks up my buffer with handy errors to fix.

## Screenshots
n/a

## System configuration
```
First checker to run:

  terraform
    - may enable:    yes
    - executable:    Found at /usr/local/bin/terraform
    - next checkers: terraform-tflint

Checkers that may run as part of the first checker's chain:

  terraform-tflint
    - may enable: yes
    - may run:    t
[...]

PR #2041, opened 2023-11-06
fix tflint invocation

dod38fr:
Closes: #2024

tflint can no longer be run on individual files. With this commit, tflint is run on all files of the directory of the buffer file.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -12255,7 +12255,8 @@ information about tflint."
 See URL `https://github.com/terraform-linters/tflint'."
   :command ("tflint" "--format=json"
             (option-list "--var-file=" flycheck-tflint-variable-files concat)
-            (option-list "--chdir=" source-directory concat))
+            (option-list "--chdir=" source-directory concat)
+            (option-list "--filter=" source-original concat))
   :error-parser flycheck-parse-tflint-linter
   :predicate flycheck-buffer-saved-p
   :modes terraform-mode)
--- a/flycheck.el
+++ b/flycheck.el
@@ -12253,10 +12253,8 @@ information about tflint."
   "A Terraform checker using tflint.
 
 See URL `https://github.com/terraform-linters/tflint'."
-  :command ("tflint" "--format=json"
-            (option-list "--var-file=" flycheck-tflint-variable-files concat)
-            (option-list "--chdir=" source-directory concat)
-            (option-list "--filter=" source-original concat))
+  :command ("tflint" "--format=json" "--force"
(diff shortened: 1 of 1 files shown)


mbauhardt:
pls approve :pray:

flycheck.el
```diff
@@ -12255,7 +12255,7 @@ information about tflint."
 See URL `https://github.com/terraform-linters/tflint'."
   :command ("tflint" "--format=json"
             (option-list "--var-file=" flycheck-tflint-variable-files concat)
-            source-original)
+            (option-list "--chdir=" source-directory concat))
```
mbauhardt, line 12258: Do you think it make sense to apply `--filter` as well.

See https://github.com/terraform-linters/tflint/pull/1687
```
$ tflint --chdir=modules/production --filter=main.tf
// replacement of "tflint modules/production/main.tf"
```

mbauhardt, line 12258:
```suggestion
            (option-list "--filter=" buffer-file-name concat))
```

dod38fr: Good point.

dod38fr: I'd rather leave the `--chdir` option in place. At worst, it does nothing.

mbauhardt:
The changes above works sometimes and sometimes not.
I don't know exactly why.

Looking at the sources and search for `source-original`

>  source-original
     Return the path of the actual file to check, or an empty
     string if the buffer has no file name.
    Note that the contents of the file may not be up to date
     with the contents of the buffer to check.  Do not use this
     as primary input to a checker, unless absolutely necessary.
     When using this symbol as primary input to the syntax
     checker, add `flycheck-buffer-saved-p' to the `:predicate'.

I did not found anything regarding `source-directory`. 
`eval: source-directory` gives me `/builddir/emacs-29.1/gtk3/` which is not the path to the buffer I want to check.

dod38fr:
@mbauhardt you're right. Turns out the `--path` and `--filter` are not needed. What was missing is the `--force` option.

Could you try again ?

mbauhardt:
works pretty well. :)
Attention, the upstream has changed and fixed the actual bug you reported.
But somehow the list of warnings/errors does not refresh automatically.
With your changes the linter behaves faster (live).

morphalus (approved):
It works with `"--force"` added

jonasagx (approved):

webframp (approved):
Confirming this also works for me. Redefined the tflint checker in my local config with these changes for now until this is merged.

bbatsov (approved):

PR #2044, opened 2023-11-17
Also enable nix checkers in nix-ts-mode

purcell:
nix-ts-mode is a treesitter major mode for Nix, which doesn't have nix-mode as its parent.


jcs090218 (approved):

bbatsov (approved):

bbatsov:
Thanks!

PR #2058, opened 2024-02-26
Fix a bunch of "reference to free variable" warnings

Hi-Angel:
These two commits are pretty simple, but they fix 11 warnings, which is a lot. Most of the existing warnings in flycheck come from that single function.

I did not runtime-test that though as I don't have/don't know anything about that "statix" app, so testing is appreciated.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -11196,23 +11196,22 @@ information about statix."
                     end-column .at.to.column))
 
             (let-alist err
-              (let ((diagnostic (car .diagnostics)))
-                (flycheck-error-new-at
-                 start-line
-                 start-column
-                 (pcase .severity ("Error" 'error)
-                        ("Warn" 'warning)
-                        (_ 'warning))
-                 (format "%s: %s" .note message)
-                 :id (format "%s%02d" (pcase .severity
-                                        ("Error" "E")
-                                        ("Warn" "W")
-                                        (_ "")) .code)
-                 :checker checker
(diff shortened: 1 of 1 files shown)


bbatsov:
The suggestions seem reasonable, but I see that your second commit nested a couple of `let-alist` calls, which the original code way trying to avoid for some reason. (perhaps that's not a good idea)?

bbatsov:
According to the Emacs docs the nesting is fine (see https://www.gnu.org/software/emacs/manual/html_node/elisp/Association-Lists.html), so I'll merge your PR. 

Might be a good idea to add some byte-compilation job to our CI to detect such problems automatically. // @jcs090218

Hi-Angel:
Oh, thanks! I was just writing a comment about that 😊 Yeah, I was going to say that the comment does not mention why it was avoiding to nest `let-alist`, but I presume the original author @lunik1 was afraid that dot-variables in the embedded `let-alist` would refer to the outer one *(which is not the case as you mentioned here 😊)*

lunik1:
It was a while ago so I don't exactly remember. Maybe I misinterpreted
```
Nesting let-alist inside each other is allowed, but the code in the inner let-alist can’t access the variables bound by the outer let-alist.
```
as meaning the inner `let-alist` couldn't access the outer `let-alists`'s variables even if rebound by `setf`/`let`. Anyway, thanks for the fix.

PR #2060, opened 2024-02-29
docstrings should not exceed 80 characters in width

gutron:
The docstring generated for flycheck-stylelintrc includes each of its checkers, ending up 87 characters long.  I restricted its definition to print only the first three (of 4) checkers, which is enough to fix it, though we might want to consider putting this limit in the macro that generates the docstring.

Separately, the docstring for flycheck-cuda-extended-lambda included an overly-long line, which I wrapped.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8120,8 +8120,9 @@ See URL `https://github.com/CSSLint/csslint'."
 (defconst flycheck-stylelint-args '("--formatter" "json")
   "Common arguments to stylelint invocations.")
 
-(flycheck-def-config-file-var flycheck-stylelintrc
-    (css-stylelint scss-stylelint sass-stylelint less-stylelint) nil)
+(let ((print-length 3))
+  (flycheck-def-config-file-var flycheck-stylelintrc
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -8120,8 +8120,9 @@ See URL `https://github.com/CSSLint/csslint'."
 (defconst flycheck-stylelint-args '("--formatter" "json")
   "Common arguments to stylelint invocations.")
 
-(flycheck-def-config-file-var flycheck-stylelintrc
-    (css-stylelint scss-stylelint sass-stylelint less-stylelint) nil)
+(let ((print-length 3))
```
bbatsov, line 8123: Might be good at some comment explaining the need for this here, as without the PR context it won't be obvious.

bbatsov:
> I restricted its definition to print only the first three (of 4) checkers, which is enough to fix it, though we might want to consider putting this limit in the macro that generates the docstring.

Agreed.

flycheck.el
gutron: Good point. I added a comment.

PR #2065, opened 2024-03-07
Support buffers without backing files in Ruff

cpitclaudel:
I think this problem is broader than Ruff.

I've update the docstring of `:standard-input`, but we should review all relevant checkers to fix the issue globally (most of them have correct patterns already, but for the remaining ones like mix or stylelint, someone needs to run the tool to see what it does when given no filename)


bbatsov:
The suggested changes look good to me. Please, mention them somewhere in the docs as well, and in the changelog.

Issue #1912, opened 2021-10-23
scss-stylelint checker doesn't work due to --syntax option removed from stylelint v14


yadex205:
Thank you very much for making Flycheck :bow:

## Checklist
- [x] I have checked existing issues for potential duplicates before creating this one.
- [x] I have read the [Troubleshooting guide][].

## Bug description
`scss-stylelint` checker with Stylelint v14 always reports `The "syntax" option is no longer available. You should install an appropriate syntax, e.g. postcss-scss, and use the "customSyntax" option`.

## Steps to reproduce
1. Use `scss-stylelint` checker with Stylelint v14

- [x] I have read https://emacs.stackexchange.com/questions/28429/how-do-i-troubleshoot-emacs-problems

## Expected behavior
`scss-stylelint` handles Stylelint correctly.

## System configuration
```
Syntax checkers for buffer ___.scss in scss-mode:

First checker to run:

  scss-stylelint
    - may enable:         yes
    - may run:            t
    - executable:         Found at /___/node_modules/.bin/stylelint
    - configuration file: Not found

Checkers that are compatible with this mode, but will not run until properly configured:

  scss-lint (automatically disabled) reset
    - may enable:         no
    - executable:         Not found
    - configuration file: Not found

  sass/scss-sass-lint (automatically disabled) reset
    - may enable:         no
    - executable:         Not found
    - configuration file: Not found

  scss (automatically disabled) reset
    - may enable: no
    - executable: Not found

Flycheck Mode is enabled.  Use C-u C-c ! x to enable disabled
checkers.

--------------------

Flycheck version: 32snapshot
Emacs version:    27.2
System:           x86_64-pc-linux-gnu
Window system:    nil

```

Emacs configuration:
- [x] Plain Emacs / Custom configuration
- [ ] Spacemacs
- [ ] Doom Emacs
- [ ] Other shared configuration

## Additional notes
`scss-stylelint` and similar checkers seem using stylelint's `--syntax` option which is removed from Stylelint v14. (https://github.com/stylelint/stylelint/blob/main/CHANGELOG.md#1400) So do you have any plans to support Stylelint v14?

BTW I wrote a temporary checker like below for my `.emacs.d/init.el` and saw it works as I expected.

```
(flycheck-define-checker general-stylelint
  "A checker for CSS and related languages using Stylelint"
  :command ("stylelint"
            (eval flycheck-stylelint-args)
            (option-flag "--quiet" flycheck-stylelint-quiet)
[...]

PR #1944, opened 2022-05-23
fix(stylelint): stylelint v14 uses incompatible cli arguments

Fuco1:
Detect the major version of stylelint, and for verion 14 don't pass
the --syntax argument.

Fixes #1912

PR #2066, opened 2024-03-13
fix(stylelint): remove support for deprecated command line arguments

justinbarclay:
Removes support for versions of stylelint older than 14 by not passing in the `--syntax` argument.

Fixes https://github.com/flycheck/flycheck/issues/1912

I am not sure if this is in good form, and if not, I am happy to close it, but I've forked @Fuco1's [work](https://github.com/flycheck/flycheck/pull/1944) and removed the support for using the major version of stylelint to change the command behavior.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,12 @@ Bugs fixed
 
 - [#2057]: Revert the extraction of ``flycheck-version`` with ``lm-version``.
 
+----------
+Changes
+----------
+
+- **(Breaking)** [#2066]: Remove support for versions of ``stylelint`` older than v14.
+
 34.1 (2024-02-18)
 ======================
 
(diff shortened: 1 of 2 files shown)


bbatsov:
Thanks for picking this up! A bit of feedback:

- it's best to define the functions before their first usage
- probably those should be marked as private API (`flycheck--`). 
- don't forget to add a changelog entry

justinbarclay:
Thanks for the feedback @bbatsov, everything should be done now. 

I kind of guessed at the changelog entry - if this was a bug fix or a breaking change due to removing support for older versions of styelint. I went with breaking change, but I'm happy for any editorial feedback there 😅

bbatsov:
It's fine as it is. Thanks!

PR #2070, opened 2024-04-10
Add a new checker 'r' for R

kunhtkun:
This is a rather rudimentary checker using the builtin `parse` function of R. It can only report syntax/parsing errors, but it does not depend on any third-party libraries. So hopefully this checker would still be useful.

As for testing, I saw: https://github.com/flycheck/flycheck/blob/1cc041503deb31e16819a1bdda342820c38ef5f8/doc/contributor/contributing.rst?plain=1#L93-L95

However, it seems that tests for the `r-lintr` checker are still written with ERT in `flycheck-test.el`, so I again added an ERT test for this checker. Please let me know if this is not good.

Thanks!


bbatsov:
Looks good to me. Thanks!

PR #2072, opened 2024-04-19
Drop outdated integration testing instructions

mavit:
These instructions can’t be followed, because they refer to a Travis-related Dockerfile that no-longer exists, building on container images “rebuilt each month” that actually haven’t been updated for years.

### DIFF
--- a/doc/contributor/contributing.rst
+++ b/doc/contributor/contributing.rst
@@ -106,36 +106,6 @@ Run ``make help`` to see a list of all available targets.  Some common ones are:
 .. _Eask: https://emacs-eask.github.io/
 .. _Buttercup: https://github.com/jorgenschaefer/emacs-buttercup
 
-Running all the integration tests
-=================================
-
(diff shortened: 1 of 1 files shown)


bbatsov:
Fair point. I'm not sure what we should do about those tests in general. @cpitclaudel Any thoughts on this?

berquist:
> Fair point. I'm not sure what we should do about those tests in general. @cpitclaudel Any thoughts on this?

Keeping them would be a large maintenance burden, and many are failing right now, but if you are able to engage people from the community to help, I think you should keep them.

I also think providing a Docker image that contains the dependencies is a good idea and requires less ongoing maintenance burden.  (I started on a Nix flake with a dev shell and it seems to be working well.)

bbatsov:
> I also think providing a Docker image that contains the dependencies is a good idea and requires less ongoing maintenance burden. (I started on a Nix flake with a dev shell and it seems to be working well.)

Yeah, I guess that'd be nice.

PR #2075, opened 2024-05-24
Add flycheck-chktex-extra-flags

slotThe:
Add an option for the user to customise which options `chktex` is being called with. This can be used to, for example, disable certain checkers. It's essentially a duplicate of `tex-chktex-extra-flags` (which I would much rather use directly, but the consensus seems to be to create `flycheck-*` variables instead).

##### [Add flycheck-chktex-extra-flags](https://github.com/flycheck/flycheck/commit/9a5a6456b91ad979ae47b6ad2e601b6fce52851c)

> This works like tex-chktex-extra-flags in that the user may specify arbitrary flags to be passed to chktex.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@ New Features
 - [#2071]: Add a new checker ``perl-perlimports``, for cleaning up Perl import statements.
 - [#1972]: New defcustom ``flycheck-clear-displayed-errors-function`` to
   customize how error messages are to be cleared.
+- [#2075]: Add the ``flycheck-chktex-extra-flags`` option to the ``tex-chktex`` checker.
 
 -----------
 Bugs fixed
--- a/flycheck.el
+++ b/flycheck.el
@@ -12602,12 +12602,25 @@ See URL `https://github.com/terraform-linters/tflint'."
   :predicate flycheck-buffer-saved-p
   :modes terraform-mode)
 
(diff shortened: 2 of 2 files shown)


bbatsov:
> This can be used to, for example, disable certain checkers. It's essentially a duplicate of tex-chktex-extra-flags (which I would much rather use directly, but the consensus seems to be to create flycheck-* variables instead).

I'm not sure I understand what you mean here. All Flycheck variables are prefixed with `flycheck-` for consistency.

slotThe:
> > This can be used to, for example, disable certain checkers. It's essentially a duplicate of tex-chktex-extra-flags (which I would much rather use directly, but the consensus seems to be to create flycheck-* variables instead).
>
> I'm not sure I understand what you mean here. All Flycheck variables are prefixed with `flycheck-` for consistency.

This was just me rambling about philosophical points, you can safely ignore it :)

(What I meant was that `flycheck-chktex-extra-flags` is an exact copy of `tex-chktex-extra-flags`, so what I would have intuitively done is to not introduce a new variable and let users customise `tex-chktex-extra-flags` instead of `flycheck-chktex-extra-flags`. This may or may not introduce some coupling between `tex-mode.el` and flycheck, but I feel like that's already there anyways)

slotThe:
(I rebased on top of `master`, since I saw there were some conflicts in CHANGES)

bbatsov:
Got it. Thanks!

Issue #2068, opened 2024-04-02
python-ruff problem with ignore-per-file extensions


twmr:
## Checklist

- [x] I have checked existing issues for potential duplicates before creating this one.
- [x] I have read the [Troubleshooting guide][].

## Bug description

When the `python-ruff` checker is used in a buffer for which special ignore-rules apply (can be set via e.g. `[lint.extend-per-file-ignores]` in `ruff.toml`), these ignore-rules are not considered.

It seems as if this is due to the value of `default-directory`, which is used for running `ruff`.
When I run `flycheck-compile` on a python file in `mypkg/` in my project, it shows that

`ruff check --config /tmp/ruff-example/ruff.toml --output-format\=text --stdin-filename /tmp/ruff-example/mypkg/mymod.py - < /tmp/ruff-example/mypkg/mymod.py`

is called in `/tmp/ruff-example/mypkg`. 

My `ruff.toml` file exists in the project root: `/tmp/ruff-example`.

Since ruff is **not** called in the project-root, my special ignore extensions:
```toml
[lint.extend-per-file-ignores]
"__init__.py" = ["E402", "D104"]
"mypkg/*.py" = [
  "D100",
  "D101",
  "D102",
  "D103",
]
```
are not considered. Therefore, a D103 warning (missing docstring in public function) is reported by flycheck/python-ruff for every public function in `mymod.py` without a docstring.

## Expected behavior

I expect that ruff specific settings in my project ruff.toml/pyproject.toml are not ignored.

## System configuration

```
Syntax checkers for buffer mymod.py in python-ts-mode:

First checker to run:

  python-ruff
    - may enable:         yes
    - executable:         Found at /home/thomashisch/miniconda3/bin/ruff
    - configuration file: Found at "/tmp/ruff-example/ruff.toml"
    - next checkers:      python-mypy

Checkers that may run as part of the first checker's chain:

  python-mypy
    - may enable:         yes
    - may run:            t
    - executable:         Found at /home/thomashisch/micromamba/envs/py_secs/bin/mypy
    - configuration file: Not found

Checkers that could run if selected:

  python-pylint  select
    - may enable:         yes
    - executable:         Found at /home/thomashisch/micromamba/envs/py_secs/bin/python
    - configuration file: Found at "/home/thomashisch/.pylintrc"
    - `pylint' module:    Found at "/home/thomashisch/micromamba/envs/py_secs/lib/python3.12/site-packages/pylint/__init__.py"
[...]

PR #2076, opened 2024-05-29
Use project working-directory for python-ruff (#2068)

corecode:
Without having working-directory set to the project root, ruff will be confused about path names.


bbatsov:
So, ruff can't be used without a project?

wyuenho:
`locate-dominating-file` will return `nil` when a Python project file is not found, in which case it's just going to be exactly the same as it is now. It'll work. This just takes care of the case when the use doesn't want to write custom Elisp to discover the ruff config file in order to set the defcustom `flycheck-python-ruff-config` because by default it's pointing to a file relative to the CWD, which is by default `default-directory`, which is the same as the Python file in the buffer as opposed to the project root, which is where most people put all of their config files.

PR #2077, opened 2024-06-25
Update vhdl-ghdl error-patterns

cyrilarnould:
The error message format has changed going from ghdl 4.0.0 to 4.1.0. This also allows differentiating between errors and warnings.


bbatsov:
You should mention this change in the changelog.

PR #2078, opened 2024-06-28
ruff: --output-format=text replaced with --output-format=concise

berquist:
ruff 0.5.0 changed this flag in https://github.com/astral-sh/ruff/pull/12010.


manufactory:
Works here with `ruff 0.5.0`

bbatsov:
What will happen to people using older versions of `ruff` with this change?

wyuenho:
--output-format has been deprecated since 0.2.0. The original PR took too long to merge, so it wasn't discovered as it continued to work up until 0.4.5.

bbatsov:
I see. Well, this has to be rebased on top of `master` and a changelog entry should be added. I'll wait to see if @berquist will do this soon, otherwise I'll just do it myself.

PR #2081, opened 2024-07-03
Escape single quote in flycheck-chktex-extra-flags docstring.

bodea:
Without the escape, byte compilation emits the following error:

  Error: custom-declare-variable `flycheck-chktex-extra-flags' docstring has
         wrong usage of unescaped single quotes (use \= or different quoting)


bbatsov:
Thanks!

PR #2083, opened 2024-07-26
yaml-actionlint: only allow inside GitHub workflows dir

berquist:
Right now, with the default chain and priority setup, the `yaml-actionlint` checker is running first for all YAML files.  For anything outside of a repository's `.github/workflows` (not `.github`!) directory, this checker doesn't make sense.

Without modifying the YAML checker chain, this patch stops `actionlint` from running on files outside this directory.  I also tried to handle Windows but I'm not sure if `.github\\workflows` is the correct path.

I think this checker should be added to the chains after `yaml-yamllint` but wanted to make this change first.


bbatsov:
> I think this checker should be added to the chains after yaml-yamllint but wanted to make this change first.

Agreed.

PR #2086, opened 2024-08-29
Fix PyMarkdown configuration lookup

odontomachus:
The PyMarkdown config file lookup variable was set to flycheck-markdown-markdownlinkt-cli-config, so the PyMarkdown configuration was not found. The PyMarkdown configuration defaults were not set, set according to https://github.com/jackdewinter/pymarkdown/blob/main/docs/advanced_configuration.md.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -10878,7 +10878,7 @@ See URL `https://docs.astral.sh/ruff/'."
             (id (one-or-more (any alpha)) (one-or-more digit) " ")
             (message (one-or-more not-newline))
             line-end))
-  :working-directory flycheck-python-find-project-root 
+  :working-directory flycheck-python-find-project-root
   :modes (python-mode python-ts-mode)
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -11319,15 +11319,16 @@ See URL `https://github.com/markdownlint/markdownlint'."
...
+    '(".pymarkdown" ".pymarkdown.yml" "pyproject.toml")
   :package-version '(flycheck . "34"))
 
 (flycheck-define-checker markdown-pymarkdown
   "Markdown checker using PyMarkdown.
 
 See URL `https://pypi.org/project/pymarkdownlnt/'."
   :command ("pymarkdown"
-            (config-file "--config" flycheck-markdown-markdownlint-cli-config)
+            (config-file "--config" flycheck-markdown-markdown-config)
```
bbatsov, line 11331: Shouldn't this say `flycheck-markdown-pymarkdown-config`?

PR #2088, opened 2024-10-20
Fix markdown for XML and YAML docs

dunmaksim:
Hello!

This PR fix markdown for `xml-xmllint` docs: https://www.flycheck.org/en/latest/languages.html#syntax-checker-xml-xmllint

fix-markdown


bbatsov:
Thanks!

PR #2096, opened 2024-12-01
Make flycheck-explain-error-at-point copy the full error message instead of the last one

lima-limon-inc:
When `flycheck-explain-error-at-point` is used, only the latest error in the messages list is copied to the kill ring.

With this patch, full messages list will be copied into a the kill ring as a single string.

Is the original behavior intended? If so, why is that the case.

And an aside, Is there an alternative to `flycheck-list-errors` but for a single error? I'd like the error at point to be pasted into a new buffer. Is this possible?

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -5699,8 +5699,10 @@ universal prefix arg, and only the id with normal prefix arg."
   (let ((messages (delq nil (seq-map (or formatter #'flycheck-error-message)
                                      (flycheck-overlay-errors-at pos)))))
     (when messages
-      (seq-do #'kill-new (reverse messages))
-      (message (string-join messages "\n")))))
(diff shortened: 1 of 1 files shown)


bbatsov:
Hmm, I thought that successful `kills` were supposed to accumulate, so perhaps that's just a bug. 

Or the idea was that the last message would be the most relevant. @cpitclaudel Any thoughts on this one?

bbatsov:
For the record on why this closed: you were right about the problem, and it was worse than the title suggests. As well as pasting only one message, a destructive `nreverse` meant the echo area showed only the first, and the text went to `message` as a format string, so a `%` in a message was read as a directive. All three are fixed in #2287, with specs. Written fresh because the patch here has unbalanced parens and doesn't compile. Thanks for spotting it.

PR #2101, opened 2024-12-20
Byte compile elisp only when (trusted-content-p) is true

mavit:
See also https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=b5158bd191422e46273c4d9412f2bf097e2da2e0


bbatsov:
Superseded by #2160, which takes the same `trusted-content-p` approach and incorporates the test-trust binding and the byte-compile-only gate placement from this PR. Thanks for the work here — sorry it fell through the cracks back then; you had this right from the start.

PR #2105, opened 2024-12-31
feat: add options for configuring jsonnet

DinoChiesa:
The Jsonnet tool can take arguments specifying the include directories, and external variables. 
This change allows specification of those options. 

There are no tests in this change. 

I tried to build and add tests for this but:
- The existing tests, executed via `make check specs unit` as directed in the Contributinig doc, do not run successfully. Even the format checks fail.
- if I skip the checks for formatting, and run `make SELECTOR='(language jsonnet)' integ`, the existing jsonnet tests (there are just two) also do not run. They get skipped. I don't know why. 
- if I alter flycheck-test.el to include `(require 'jsonnet-mode)` and modify Eask to add jsonnet-mode as a dependency, the jsonnet tests run, but do not succeed. I tried to get them to succeed, but could not

I spent too much time trying to figure this out, and ran into too many dead-ends.  I love flycheck, but the obstacles to  contributing tests are formidable. 
 
I will gladly add tests if you can provide a master in which the existing jsonnet tests run and succeed.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -10046,10 +10046,10 @@ See URL `https://stedolan.github.io/jq/'."
 For example (\"./lib\") ."
   :type '(repeat (directory :tag "Include directory"))
   :safe #'flycheck-string-list-p
-  :package-version '(flycheck . "20241230"))
+  :package-version '(flycheck . "35.0-snapshot"))
 
 (flycheck-def-args-var flycheck-jsonnet-command-args jsonnet
-  :package-version '(flycheck . "20241230"))
+  :package-version '(flycheck . "35.0-snapshot"))
 
 (flycheck-define-checker jsonnet
   "A Jsonnet syntax checker using the jsonnet binary.
--- a/flycheck.el
+++ b/flycheck.el
@@ -10040,11 +10040,26 @@ See URL `https://stedolan.github.io/jq/'."
           (zero-or-more not-newline) line-end))
   :modes (json-mode js-json-mode json-ts-mode))
 
+(flycheck-def-option-var flycheck-jsonnet-include-paths nil jsonnet
+  "a list of include paths to specify to the jsonnet binary, via -J .
+
+For example (\"./lib\") ."
+  :type '(repeat (directory :tag "Include directory"))
+  :safe #'flycheck-string-list-p
+  :package-version '(flycheck . "20241230"))
+
+(flycheck-def-args-var flycheck-jsonnet-command-args jsonnet
+  :package-version '(flycheck . "20241230"))
+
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -10040,11 +10040,26 @@ See URL `https://stedolan.github.io/jq/'."
           (zero-or-more not-newline) line-end))
   :modes (json-mode js-json-mode json-ts-mode))
 
+(flycheck-def-option-var flycheck-jsonnet-include-paths nil jsonnet
+  "a list of include paths to specify to the jsonnet binary, via -J .
+
+For example (\"./lib\") ."
+  :type '(repeat (directory :tag "Include directory"))
+  :safe #'flycheck-string-list-p
+  :package-version '(flycheck . "35.0-snapshot"))
```
bbatsov, line 10049: This should say just 35.0.

bbatsov:
Please, mention your changes in the changelog and make sure all tests are passing. E.g.

>       documents all syntax checker options  Expected `(seq-difference all-options documented-options)' to be `equal' to `nil', but instead it was `(flycheck-cuda-compiler-options)' which does not match because: (different-types (flycheck-cuda-compiler-options) nil). (0.83ms)

This means you haven't updated the docs.

bbatsov:
FYI - there was a breakage in the unit test setup that was fixed a few days ago https://github.com/flycheck/flycheck/commit/ca2c98e2bead4773355d5252b2bb6ffb0b58c48f

I'm assuming it was the root case of your problems. In your case if you the docstrings follow the proper format and the new options are documented the tests should pass.

DinoChiesa:
I've merged the commit you mentioned.  
I've updated CHANGES.rst. 
The tests still do not run .  (and the existing jsonnet tests don't _pass_ even when I hack them, see #2102 )
Also there is no clear documentation on just how to run the tests, or I am looking in the wrong place.

I hope you all will accept this change.

bbatsov:
```
Documentation Syntax checkers Options documents all syntax checker options
FAILED: Expected `(seq-difference all-options documented-options)' to be `equal' to `nil', but instead it was `(flycheck-jsonnet-command-args flycheck-jsonnet-include-paths)' which does not match because: (different-types (flycheck-jsonnet-command-args flycheck-jsonnet-include-paths) nil).
```

This means you forgot to update accordingly https://github.com/flycheck/flycheck/blob/master/doc/languages.rst

DinoChiesa:
ok I've updated languages.rst with latest commit.

bbatsov:
And we're good to go now! Thanks for working on this!

Issue #1020, opened 2016-07-25
ANSI colour escape codes with python-pylint


owlas:
# ANSI colour escape codes with python-pylint

I have just enabled flycheck for the first time, which has detected my python-pylint executable. However, I am getting the following error that seems to arise because ANSI colours are not being escaped properly and are messing up message parsing:

```shell
Suspicious state from syntax checker python-pylint: Checker python-pylint returned non-zero exit code 16, but no errors from output: ^[[7;33m************* Module flycheck_mymodule^[[0m
flycheck_mymodule.py:1:0:^[[1mC^[[0m:^[[1mmissing-docstring^[[0m:^[[1mMissing module docstring^[[0m
```

My setup:
```python
  python-flake8
    - predicate:  nil
    - executable: Not found

  python-pylint
    - predicate:  t
    - executable: Found at /Users/oliverlaslett/miniconda2/bin/pylint

  python-pycompile
    - predicate:  t
    - executable: Found at /Users/oliverlaslett/miniconda2/bin/python
```

```shell
GNU Emacs 24.5.1 (x86_64-apple-darwin15.2.0) of 2015-12-20 on elcapitanvm.local
```

```shell
Flycheck version: 26snapshot (package: 20160419.712)
```

This was reproduced by:
```shell
$ emacs -Q file.py
$ M-x package-initialize
$ M-x global-flycheck-mode
```

lunaryorn:
@owlas Thank you for reporting this issue.

However, Flycheck doesn't "escape" (i.e. parse?) ANSI colour codes, because syntax checkers shouldn't be printing colourised output in the first place.  It simply doesn't make sense to write ANSI colours in non-TTY environments.

Do you happen to have a pylint configuration somewhere that enables coloured output?

owlas:
Interesting. I just `pip install`ed pylint and used it out of the box. I'm using Iterm2 on mac. Could the emulator be formatting the output?

owlas:
Ok, very sorry, your comment has helped me to find the problem.

By testing a file in a new project, I've found that the project (from GH) I was using has a custom `.pylintrc` file that is enabled colour output.

owlas:
Would it be reasonable to make a feature request that colour codes are escaped properly?

lunaryorn:
@owlas I'd not implement such a request myself, and my vote would also go against merging a corresponding pull request.

[...]

PR #2106, opened 2025-01-01
Update flycheck-define-checker scala :command to run w/o ansi color codes

imdaveho:
ansi color codes for colorizing output should be opt-in; this is reflected in these issues:

* https://github.com/scalameta/metals/issues/2621
* https://github.com/flycheck/flycheck/issues/1020

It is also not necessarily trivial to detect and add ansi-color support depending on the implementation of emacs. For me, with doom emacs, this is not easy or transparent: https://github.com/doomemacs/doomemacs/issues/8216

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -12093,7 +12093,7 @@ See URL `https://github.com/sasstools/sass-lint'."
   "A Scala syntax checker using the Scala compiler.
 
 See URL `https://www.scala-lang.org/'."
-  :command ("scalac" "-Ystop-after:parser" source)
+  :command ("scalac" "-Ystop-after:parser" "-color" "never" source)
   :error-patterns
   ((error line-start (file-name) ":" line ": error: " (message) line-end))
   :modes scala-mode


cpitclaudel:
Thanks!  Can you clarify which versions of scalac have that flag?  Mine doesn't recognize it.
If compatibility is a concern, I suppose we could use a separate parsing function that strips ansi codes.
But in any case the Scala checker should probably move to Scala-CLI, right?

imdaveho:
> Thanks! Can you clarify which versions of scalac have that flag? Mine doesn't recognize it. If compatibility is a concern, I suppose we could use a separate parsing function that strips ansi codes. But in any case the Scala checker should probably move to Scala-CLI, right?

seems like this option was added in scala 3: https://docs.scala-lang.org/scala3/guides/migration/options-new.html

cpitclaudel:
> Would this be complicated?

No, very easy.  Just one line, in fact:

```
  :error-parser flycheck-parse-with-patterns-without-color
```

Can you try that?

> is there advice on writing a hook or something that can amend the :command key directly

I think you want `(put 'scala 'flycheck-command …)`

>  Isn't scala-cli using scalac under the hood for compliation anyway?

Yes, but it's aware of dependencies and project structure.

imdaveho:
> I think you want `(put 'scala 'flycheck-command …)`

:facepalm: wow, that was...how did I miss that?!

> No, very easy. Just one line, in fact:
>
> ` :error-parser flycheck-parse-with-patterns-without-color`
>
> Can you try that?

hmm :confused: not sure why, but that didn't work. I tried the put method as well as overriding the existing scala check with a new one with the field set as such:

Screenshot From 2025-01-03 15-06-20

cpitclaudel:
Are we sure the problem is just the highlighting?  The pattern expects a filename at the beginning of a line, but the output has a line starting with `[E019] Syntax Error` before the file name.

imdaveho:
> Are we sure the problem is just the highlighting? The pattern expects a filename at the beginning of a line, but the output has a line starting with `[E019] Syntax Error` before the file name.

Yeah I think so. I even wrote a manual function to strip out the ansi codes and output the before/after to the message buffer just to check if it is working. As you can see from the screenshot, the initial output contains all the ansi colors, then it runs them through the function and strips them out, but after even sending the `clean-output` to `flycheck-parse-with-patterns` the output remains the original for some reason. From this I conclude that:

1) We know that the function associated with `:error-parser` does get called
2) For some reason, when `flycheck-parse-with-patterns` is called, the original output is used instead

Screenshot From 2025-01-03 15-18-36

cpitclaudel:
Thanks.  Could you paste the raw output here? (M-x base64-encode-region if we worry that Github will mangle it)

imdaveho:
> Thanks. Could you paste the raw output here? (M-x base64-encode-region if we worry that Github will mangle it)

output:
https://pastebin.com/hxYtyDg7

config.el snippet:
https://pastebin.com/kVKW6ymz

cpitclaudel:
The problem here (in addition to the colors, which can be fixed by changing the error parser) is that the error regexps are wrong.

Just to be 100% sure, are you saying that your patch is enough to make things work on your machine?

The issue I'm seeing is that even with `-color never`, Scala3 uses the following error format:

```
-- [E008] Not Found Error: /home/cpc/scala3-ansi-colors/src/main/scala/Main.scala:5:43 --------------------------
5 |def msg =  "I was compiled by Scala 3. :)" / 2
  |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |           value / is not a member of String
one error found
```

This doesn't match the existing error patterns, which were written for Scala2:

```
  :error-patterns
  ((error line-start (file-name) ":" line ": error: " (message) line-end))
```

So the patterns need to be updated, too.  The problem is that the new format is quite unwieldy to parse with a regex, because the error message is multiple lines down under the quote.  I asked, and there's no alternative printer, so I suppose we should update the regexes...

imdaveho:
> The problem here (in addition to the colors, which can be fixed by changing the error parser) is that the error regexps are wrong.
> 
> Just to be 100% sure, are you saying that your patch is enough to make things work on your machine?
> 
> The issue I'm seeing is that even with `-color never`, Scala3 uses the following error format:
> 
> ```
> -- [E008] Not Found Error: /home/cpc/scala3-ansi-colors/src/main/scala/Main.scala:5:43 --------------------------
> 5 |def msg =  "I was compiled by Scala 3. :)" / 2
>   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   |           value / is not a member of String
> one error found
> ```
> 
> This doesn't match the existing error patterns, which were written for Scala2:
> 
> ```
>   :error-patterns
>   ((error line-start (file-name) ":" line ": error: " (message) line-end))
> ```
> 
> So the patterns need to be updated, too. The problem is that the new format is quite unwieldy to parse with a regex, because the error message is multiple lines down under the quote. I asked, and there's no alternative printer, so I suppose we should update the regexes...

umm just to clarify, you're saying that the issue is with the error-patterns that are not matching, therefore the ansi-codes are not being replaced, therefore whether stripping them or applying ansi-color, neither is working unless we update the patterns?

cpitclaudel:
I'm saying that the error patterns won't match Scala 3's new output format, regardless of ansi coloring.  So stripping the ansi codes isn't enough: we also need to change the patterns.  At least that's how I read the code, but I could be wrong; hence the question above (Just to be 100% sure, are you saying that your patch is enough to make things work on your machine?).

I asked the Scala folks and there is not other output format at the moment, unfortunately.  So we'd need somewhat complicated patterns.

bbatsov:
As Clément noted, the underlying problem is Scala 3's new compiler output format, which our error patterns don't understand — disabling colors alone doesn't fix the parsing. Closing in favor of #2179, which tracks proper Scala 3 output support.

PR #2107, opened 2025-01-04
Add -Xcompiler option for cuda-nvcc.

eyal0:
The arguments provided to `-Xcompiler` are passed to nvcc's underlyring compiler.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@ New Features
 - [#1972]: New defcustom ``flycheck-clear-displayed-errors-function`` to
   customize how error messages are to be cleared.
 - [#2075]: Add the ``flycheck-chktex-extra-flags`` option to the ``tex-chktex`` checker.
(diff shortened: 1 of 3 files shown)


eyal0:
Here's what it looks like by default:

image

Here it is if I add the compiler option `-Wshadow=local`, which turns on a warning from `nvcc` to the compiler (I think `gcc`) to warn about the variable `int x;` on line 5 which is hiding the one on line 2.

The default is `-Wall` and `-Wextra`, which is consistent with the warnings enabled on other c++ checkers.

bbatsov:
You also need to update the docs for the tests to pass.

bbatsov:
I'm referring to that particular failure. 

>       documents all syntax checker options  Expected `(seq-difference all-options documented-options)' to be `equal' to `nil', but instead it was `(flycheck-cuda-compiler-options)' which does not match because: (different-types (flycheck-cuda-compiler-options) nil). (0.83ms)

eyal0:
Let's see if I did it right this time!

PR #2111, opened 2025-01-31
Ruby 3.4 compatibility

aaronjensen:
Ruby 3.4's new parser includes the ruby path in the first error:

```
Suspicious state from syntax checker ruby: Flycheck checker ruby returned 1, but its output contained no errors: /some-pathruby/3.4.1/bin/ruby: -:3: syntax error found (SyntaxError)
  1 | require_relative "../../automated_init"
  2 | 
> 3 | do
    | ^~ unexpected 'do', ignoring it
  4 | context "Request" do
  5 |   context "Unauthorized" do
```

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -11690,7 +11690,8 @@ See URL `https://www.ruby-lang.org/'."
   ((error line-start "SyntaxError in -:" line ": " (message) line-end)
    (warning line-start "-:" line ":" (optional column ":")
             " warning: " (message) line-end)
-   (error line-start "-:" line ": " (message) line-end))
+   ;; Ruby 3.4 includes the interpreter path when emitting syntax errors
+   (error line-start (optional (one-or-more (not (any ":"))) ": ") "-:" line ": " (message) line-end))
   :modes (enh-ruby-mode ruby-mode ruby-ts-mode)
   :next-checkers ((warning . ruby-chef-cookstyle)))
 


bbatsov (approved):

bbatsov:
What's about the format for warnings? Hasn't in changed as well?

flycheck.el
```diff
@@ -11690,7 +11690,7 @@ See URL `https://www.ruby-lang.org/'."
   ((error line-start "SyntaxError in -:" line ": " (message) line-end)
    (warning line-start "-:" line ":" (optional column ":")
             " warning: " (message) line-end)
-   (error line-start "-:" line ": " (message) line-end))
+   (error line-start (optional (one-or-more (not (any ":"))) ": ") "-:" line ": " (message) line-end))
```
bbatsov, line 11693: Won't something like `(optional interpreter ":")` have the same result here?

bbatsov, line 11693: Maybe you can add a comment explaining here the optional presence of the interpreter path.

aaronjensen:
As far as I can tell, the format for warnings has not changed. I'm wondering if I should submit an improvement suggestion to add a mode where Ruby can print one syntax error per line. Its current output is fine for humans but awful to parse if you want to highlight more than one syntax error.

bbatsov:
You can probably do this here https://github.com/ruby/prism

aaronjensen:
Done: https://github.com/ruby/prism/issues/3455

Thanks for the merge @bbatsov

PR #2114, opened 2025-02-25
Update flycheck-add-overlay/right-position-in-narrowed-buffer output

manphiz:
- Output changed when running under Emacs 30.1

### DIFF
--- a/test/flycheck-test.el
+++ b/test/flycheck-test.el
@@ -2119,7 +2119,10 @@
     (should (= (length (flycheck-overlays-in (point-min) (point-max))) 2))
     ;; Remove restrictions and test that all errors are reported
     (widen)
-    (should (= (length (flycheck-overlays-in (point-min) (point-max))) 4))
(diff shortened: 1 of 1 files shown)


bbatsov:
I see the next test is disabled on Emacs > 29 and I'm guessing it should be updated as well. I'm also wondering if we shouldn't actually make an update to the underlying function that's returning this different result to make sure it's behaving consistently across Emacs versions.

manphiz:
> I see the next test is disabled on Emacs > 29 and I'm guessing it should be updated as well. I'm also wondering if we shouldn't actually make an update to the underlying function that's returning this different result to make sure it's behaving consistently across Emacs versions.

That would be great, but I guess it's also understandable that the handling in different versions of Emacs differ.  So it seems that matching the output in different versions of Emacs is more inline with real world usage, though at the expense of maintainability.  But I don't have strong feeling for either :)

PR #2123, opened 2025-05-24
For checker `python-ruff` default `output-format` to `text` instead o…

nordlow:
…f `concise` which is no longer supported

Verified using ruff version 0.0.291.


bbatsov:
Indeed.

nordlow:
Is flycheck tagged regularly so I can install a new release nightly with the right ELPA channel?

PR #2132, opened 2025-09-21
Add new option flycheck-shellcheck-infer-shell

jacksonrayhamilton:
Hello,

At my company we want to use ShellCheck to check our scripts, but we also want to support a breadth of OSes.  The choice was made to use Zsh as the default shell because it comes standard in macOS, is much newer than the Bash that comes installed on macOS, and could easily be installed on other OSes.  However since ShellCheck doesn't support Zsh, we are adding the `# shellcheck shell=bash` directive and trying to write mostly-portable code.  Not sure if this sounds unwise but it's where we're at.

What this looks like in practice is that our shell files start like this:

```sh
#!/usr/bin/env zsh
# shellcheck shell=bash
...
```

ShellCheck's documentation indicates that this should be possible: https://github.com/koalaman/shellcheck/wiki/Directive#shell.  And when I test with the `shellcheck` command via CLI, it works.

However, in FlyCheck's checker, currently there are two checks that prevent me from checking such code.

1. The `zsh` shebang will fail the member check for `flycheck-shellcheck-supported-shells`.  I can easily work around that with `(add-to-list 'flycheck-shellcheck-supported-shells 'zsh)`.
2. However, on top of that, the `sh-shellcheck` checker is defined to unconditionally pass the `--shell=zsh` argument in this case, which causes a program error, "Unknown shell: zsh"

The only solution that I'm aware of is for the checker's `:command` to be customizable with a new option that removes the `--shell` arg, allowing for ShellCheck to infer the shell according to the shebang or any `shell` directives in the file.

Let me know if this sounds reasonable to you or if you'd prefer a different approach.  Thanks!


jacksonrayhamilton:
I revisited this to try and clean it up before begging for attention.

- Added missing docs for the change (fixing most CI failures)
- Hopefully assigned the new integration test a better identifier than originally
- Updated the original integration test to expect output like the latest stable ShellCheck provides
- Verified that the integration tests pass on my machine

CI still reports some failures in the matrix, but they are the same jobs that failed in the last merged PR: https://github.com/flycheck/flycheck/actions/runs/19103063205/job/54579613912.  So I'm assuming it's not critical that those jobs pass right now.

@bbatsov any feedback?

# Actual ShellCheck output

Screenshot 2025-11-15 at 10 46 03 AM

# Passing ShellCheck integration tests

Screenshot 2025-11-15 at 11 16 07 AM

bbatsov:
Looks good to me. Thanks!

PR #2165, opened 2026-03-21
Fix org-lint line number handling

aaronjensen:
Corrects this issue:

```
Error from syntax checker org-lint: Wrong type argument: number-or-marker-p, #("7" 0 1 (org-lint-marker #<marker at 178 in [D] Some File>))
```

May be Emacs 31 specific. Referenced in #2161

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -8849,7 +8849,10 @@ has access to all installed packages and user configuration."
                                (pcase e
                                  (`(,_n [,line ,_trust ,desc ,_checker])
                                   (flycheck-error-new-at
-                                   line nil 'info desc
+                                   (if (stringp line)
+                                       (string-to-number line)
+                                     line)
+                                   nil 'info desc
                                    :checker checker))
                                  (_
                                   (flycheck-error-new-at


dmgerman:
While we wait for it to be addressed, this is how I patched my configuration. it a work around until the bug is fixed.

* flycheck: patch locally until addressed

In Emacs 31, ~org-lint~ returns a propertized string for the line number
instead of an integer.  flycheck 36.0 passes it raw to
~flycheck-error-new-at~, which expects an integer.  The fix (adding a
~string-to-number~ conversion) is tracked as an uncommitted patch in the
flycheck repo; apply it here until it lands upstream.

    #+begin_src emacs-lisp :exports both
      (if (and (version<= "31.0.6" emacs-version)
               (string= flycheck-version "36.0"))
          (setf (flycheck-checker-get 'org-lint 'start)
                (lambda (checker callback)
                  (condition-case err
                      (let ((errors
                             (delq nil
                                   (mapcar
                                    (lambda (e)
                                      (pcase e
                                        (`(,_n [,line ,_trust ,desc ,_checker])
                                         (flycheck-error-new-at
                                          (if (stringp line)
                                              (string-to-number line)
                                            line)
                                          nil 'info desc
                                          :checker checker))
                                        (_
                                         (flycheck-error-new-at
                                          1 nil 'warning
                                          (format "Unexpected org-lint format: %S" e)
                                          :checker checker))))
                                    (org-lint)))))
                        (funcall callback 'finished errors))
                    (error (funcall callback 'errored
                                    (error-message-string err))))))
        (warn "flycheck org-lint: patch not applied (emacs %s, flycheck %s) — check if still needed"
              emacs-version flycheck-version)))
    #+end_src

RicardoSwitchNetwork:
What about using `(flycheck-string-to-number-safe line) nil 'info desc`?

PR #2167, opened 2026-04-13
Add luau-analyze syntax checker

redguardtoo:
Hi, I add a new syntax check for luau programming language, using cli tool `luau-analyze`

https://github.com/luau-lang/luau

Luau is Roblox’s official scripting language, derived from Lua 5.1, optimized for game development with optional static typing, fast execution, and deep integration into Roblox Studio. It supports dynamic typing by default, but developers can add type annotations for better safety and maintainability.

I added unit test and updated doc. Run `make spec` in project root and looks ok. But I'm not sure how the unit test for luau script works. Not sure luau-analyze is included in any Linux distribution. Let me know what should I do next?

screenshot of working checker,
image

### DIFF
--- a/doc/languages.rst
+++ b/doc/languages.rst
@@ -761,6 +761,16 @@ to view the docstring of the syntax checker.  Likewise, you may use
 
       Check syntax with the `Lua compiler <https://www.lua.org/>`_.
 
+.. supported-language:: Luau
+
+   Flycheck checks Lua with `luau-analyze`.
+
+   .. syntax-checker:: luau-analyze
+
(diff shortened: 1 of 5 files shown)


bbatsov:
Luau support just landed in #2334 - thanks for kicking this off, the checker exists because of your proposal. The merged version diverged on the details the tool forced (the plain formatter turns out to hardcode W0 on every diagnostic, so severities and rule ids only exist in the default format, and source-original checks the file on disk rather than the buffer), so it was easier to redo it fresh than to iterate here. Credit's in the changelog.

PR #2191, opened 2026-07-19
Run syntax checkers over TRAMP

bbatsov:
Flycheck can now check buffers visiting remote files over TRAMP, originally reported in #1816. The remote awareness lives in the shared process/executable/temp-file helpers rather than in each checker, so command checkers work over TRAMP without per-checker changes (which is where the earlier #1842 stalled). Checkers run on the remote host, so their tools have to be installed there.

`global-flycheck-mode` enables Flycheck in remote buffers by default. Since each remote check spawns a process over the network, `flycheck-check-syntax-automatically-remote` limits the automatic triggers there to `save` and `mode-enabled`; set it to `t` for the same triggers as local buffers.

Builds on #1842, thanks to @carrete for the groundwork there.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,15 @@
 ``master`` (unreleased)
 =======================
 
+- [#1816]: Flycheck now runs syntax checkers over TRAMP.  When a buffer
+  visits a remote file, command checkers execute on the remote host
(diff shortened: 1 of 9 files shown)


carrete:
Awesome!

PR #2192, opened 2026-07-20
Add project-wide diagnostics to the error list

bbatsov:
Flycheck has always been per-buffer. This adds a whole-project view to the error list without scanning files you haven't opened: it keeps the cross-file errors that checkers like `tsc`, `cargo check` and `mypy` already report and `flycheck-relevant-errors` currently drops, and aggregates them with every open buffer's errors. Press `P` in the error list to toggle between the current buffer and the project (keyed by `project-current`, or the checker's working directory as a fallback).

Cross-file errors reflect the last check that reported them and clear when that buffer is re-checked; without a resident server there's no way to know another file changed. A few calls I'd want your take on: the `project-current` keying, the `P` toggle instead of a separate command, and storing exactly what the buffer reports so its own errors match in both scopes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 ``master`` (unreleased)
 =======================
 
+- The error list can now show whole-project diagnostics.  Press ``P``
+  (``flycheck-error-list-toggle-scope``) to switch it between the current
+  buffer and the project, where it aggregates the errors of every open
+  Flycheck buffer together with the cross-file errors that checkers like
+  ``tsc``, ``cargo check`` and ``mypy`` report but the per-buffer view
+  drops.  A buffer's project is Emacs' ``project-current`` when found, and
+  the checker's working directory otherwise.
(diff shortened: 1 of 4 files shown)


cpitclaudel:
This is really nice, thanks @bbatsov .

Btw, there was a branch that I worked a lot on a while ago and never got around to merging: https://github.com/flycheck/flycheck/compare/cpitclaudel_error-list-revamp

The branch start at [Add a new error list backend: flycheck-error-tree](https://github.com/flycheck/flycheck/commit/5adb7821a430b616d692d2fa920461da5655c53a).  The idea was to make error list browsing both much more performance efficient (it uses hashes instead of linear scans to find the right error in the list), and much more flexible (the configuration is a set of projection function, and it displays a tree of results).

I wonder how easy it would be to rebase this branch these days and complete it.  Unfortunately I don't remember too much about it.  I suppose it would work nicely with the move to displaying errors from all buffers.

bbatsov:
Thanks for the heads up! I'll see what I can do about this - I've already started to incorporate some of your ideas into the current work stream.

PR #2194, opened 2026-07-21
python-ruff JSON output format for robust parsing

rhoit:
Switch from text parsing to JSON format because the standard text output structure changed and no longer includes explicit error codes. Using JSON prevents regex parsing failures on unstructured text.


bbatsov:
Thanks! The ruff checker already runs with `--output-format=json` and a dedicated parser (`flycheck-parse-ruff`), so the switch to JSON is in place. Closing as done.

rhoit:
thank you for your amazing work!

PR #2233, opened 2026-07-29
Fix below-style annotations trapping vertical motion

bbatsov:
A user reported that `flycheck-annotate-mode`'s `below` style breaks navigation: `next-line` needs an extra press to get past the annotated line, and `evil-next-visual-line` can't move past it at all.

The cause is that the multi-line message hung off the annotated line's own newline, so its extra screen rows counted as part of that line for visual-line motion. Evil made it worse - it pulls point back off the newline the annotation sits on, so every downward press just re-landed on the same line.

Anchoring the block on the start of the *following* line fixes both cases: its rows now belong to that line's buffer position, so `next-line` and `evil-next-visual-line` land there cleanly. The rendering is unchanged. I verified the behavior in a scripted GUI Emacs (plain and Evil) and the eol/sideline styles were already fine.


wbolster:
> A user reported

that would've been me! 🙃

thanks, it works now, also in evil-mode

PR #2257, opened 2026-07-30
Fix current version variable value

saulotoledo:
Fixes the value of the `flyway-version` variable, which is currently outdated. The value of this variable is useful for scripts.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -1508,7 +1508,7 @@ Only has effect when variable `global-flycheck-mode' is non-nil."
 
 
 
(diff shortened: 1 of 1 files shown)


saulotoledo:
Updated to 39.0 due to https://github.com/flycheck/flycheck/pull/2258
@bbatsov, [this line](https://github.com/flycheck/flycheck/blob/19f8887e4564c998b2deff273654c6e3fa1c5df4/flycheck.el#L13) is still 38.3

bbatsov:
Thanks for fixing this!

PR #2263, opened 2026-07-31
Document running Eglot alongside a separate lint server

bbatsov:
The LSP section reads as a choice between three integrations, so it isn't obvious you can let Eglot run the language server you edit with while Flycheck lints from a different one, or have both report into Flycheck at once. That's what people are after when they want ESLint's or Tailwind's server on top of their main one.

The part worth writing down is that clearing the `exclusive` options isn't enough to get both. Each bridge chains to the first other checker supporting the mode in `flycheck-checkers` order, and both bridges sit at the end of that list, so they chain to command checkers and never to each other. Connecting them needs an explicit `flycheck-add-next-checker`, and I've documented which one starts the chain and how to pin it.

Arguably the bridges should just chain to each other when both are on and non-exclusive, rather than needing the manual line. Left that alone for now.

### DIFF
--- a/doc/user/syntax-checks.rst
+++ b/doc/user/syntax-checks.rst
@@ -314,6 +314,73 @@ In short: reach for ``flycheck-lsp-mode`` when you just want a linter that
 happens to speak LSP, and for ``flycheck-eglot-mode`` (see above) when you want
 a full language server whose diagnostics Flycheck should own.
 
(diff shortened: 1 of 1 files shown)


bbatsov:
Superseded by the chaining fix: turning both bridges on now just works, so the manual wiring this documented is no longer the story. Docs folded into that PR.

PR #2285, opened 2026-08-03
python-ruff correct config priority

rhoit:
All of the above rules apply equivalently to pyproject.toml, ruff.toml, and .ruff.toml files. If Ruff detects multiple configuration files in the same directory, the .ruff.toml file will take precedence over the ruff.toml file, and the ruff.toml file will take precedence over the pyproject.toml file.

source: https://docs.astral.sh/ruff/configuration/#config-file-discovery

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -15719,7 +15719,7 @@ Requires Flake8 3.0 or newer. See URL
                   (warning . python-mypy)))
 
 (flycheck-def-config-file-var flycheck-python-ruff-config python-ruff
-                              '("pyproject.toml" "ruff.toml" ".ruff.toml"))
+                              '(".ruff.toml" "ruff.toml" "pyproject.toml"))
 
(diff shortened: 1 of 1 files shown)


flycheck.el
```diff
@@ -15719,7 +15719,7 @@ Requires Flake8 3.0 or newer. See URL
                   (warning . python-mypy)))
 
 (flycheck-def-config-file-var flycheck-python-ruff-config python-ruff
-                              '("pyproject.toml" "ruff.toml" ".ruff.toml"))
+                              '(".ruff.toml" "ruff.toml" "pyproject.toml"))
```
bbatsov, line 15724: Might be a good idea to add some comment about the order here, so someone doesn't change it accidentally in the future.

rhoit: I have added the comment in the following commit.

PR #2302, opened 2026-08-04
Proselint improvements

mavit:
Here are a series of improvements for the `proselint` checker:

- Fix proselint version detection.
- End position was sometimes off by one in older proselint versions.
- ~Add a fixer.~


bbatsov:
Can you rebase on top of master? I've addressed the Emacs snapshot failures there.

bbatsov:
Thanks - the version-probe fix matters more than the PR lets on: proselint 0.16 rejects the old `--json -` invocation outright, so the checker is completely broken against it today. I'd like to land that part (plus the end-position correction) right away.

The fixer I'd hold off on. I tested both generations: 0.16 returns `replacements: null` for everything I could trigger, and on 0.14 the only check that populates it is consistency.spacing, whose "replacement" is a regex pattern - the exact case you're filtering out. So as far as I can tell no released proselint emits a literal replacement the fixer could apply. Drop the fixer bits and keep the probe + end-position fixes and this can go in (after the rebase). Two notes for when the fixer becomes real: build fixes with flycheck--make-fix so they carry the check-start tick, and a single filtered pattern won't cover replacements being regexes in general.

mavit:
Okay, this now contains just the updated version detection and the end position workaround.  I'll move the fixer to a follow-up pull request.

That the version detection for 0.16 is broken, tells us that not many people are using it yet, I think.  This wouldn't be surprising, given that distributions are reluctant to take it because of its incompatibility with previous versions.

flycheck.el
```diff
@@ -15948,6 +15949,20 @@ See URL `https://proselint.com/' for more information about proselint."
...
+;; further than it actually does.  This is fairly trivial until we try
+;; to apply a replacement, at which point it causes us trouble.
+;; https://github.com/amperser/proselint/issues/1048
+(defun flycheck--proselint-off-by-one-correction (buffer end-pos)
+  "Return an offset if END-POS of BUFFER doesn't look like the end of an error."
+  (save-excursion
+    (with-current-buffer buffer
+      (if (char-equal
+           (char-syntax (char-before (- end-pos 1)))
+           (char-syntax (char-before end-pos)))
```
bbatsov: `char-before` returns nil at the start of the buffer and `char-syntax` won't take nil, so this signals `wrong-type-argument` for any diagnostic ending at position 1 or 2:

```
end-pos 1 -> ERROR wrong-type-argument
end-pos 2 -> ERROR wrong-type-argument
end-pos 5 -> -1
```

This runs in the parser, so a diagnostic covering the first character of the buffer takes down the whole check rather than just that one error. Binding both `char-before` results and returning 0 when either is nil covers it.

Separately, `save-excursion` sits outside `with-current-buffer`, so it saves point in the wrong buffer. Nothing here moves point, so it's cosmetic.

flycheck.el
```diff
@@ -15969,13 +15984,13 @@ absent from the table has not been probed yet.")
...
-                       nil nil nil "--version"))
+                       nil nil nil "version"))
                flycheck--proselint-old-args-by-host))
     (if (gethash host flycheck--proselint-old-args-by-host)
-        ;; Proselint versions <= 0.14.0:
-        (list "--json" "-")
-      ;; Proselint versions >= 0.16.0
-      (list "check" "--output-format=json"))))
+        ;; Proselint versions >= 0.16.0
+        (list "check" "--output-format=json")
```
bbatsov: This flips what the cache means: `t` is now new proselint (>= 0.16). The docstring still says each value is `t` for old (<= 0.14.0) and nil for new, so it states the opposite of the code, and the variable name reads backwards now too. Either invert what you store, or rename it and fix the docstring.

bbatsov (commented):
Thanks for splitting this. The version detection is the urgent half and I want it in: 0.16 rejects the old invocation outright, so the checker is dead against it today. Comments inline, and it'll need a changelog entry.

bbatsov:
Any chance you can pick this up? The version detection is the urgent half - the checker is dead against 0.16 today. If you're short on time I'll fix the two review points myself and land it with your commit.

mavit:
I can look at this again on Tuesday, after the public holiday here.  Otherwise, please feel free to merge and fix up yourself.

flycheck.el
mavit: Fixed.

flycheck.el
mavit: Fixed.

What this a theoretical thing, by the way, or were you actually able to trigger it?  I'm not sure that any real diagnostics can end that early in the text.

flycheck.el
```diff
@@ -18194,9 +18211,9 @@ absent from the table has not been probed yet.")
     (when (eq (gethash host flycheck--proselint-old-args-by-host 'unknown) 'unknown)
       (puthash host
                ;; Probe on the host the check will run on (remote over TRAMP).
-               (zerop (process-file
-                       (or flycheck-proselint-executable "proselint")
-                       nil nil nil "--version"))
+               (not (zerop (process-file
+                            (or flycheck-proselint-executable "proselint")
+                            nil nil nil "version")))
```
mavit: Hang on, I just realised why I wrote it this way in the first place: https://github.com/dense-analysis/ale/issues/5042#issuecomment-3548580562

I think the answer is to inspect the output of `proselint --version` for a version number.

flycheck.el
```diff
@@ -18188,15 +18205,26 @@ absent from the table has not been probed yet.")
...
+    ;; Probe on the host the check will run on (remote over TRAMP).
+    (process-file (or flycheck-proselint-executable "proselint")
+                  nil t nil "--version")
+    (goto-char (point-min))
+    (search-forward-regexp
+     (rx line-start
+         (zero-or-one "Proselint ")
+         (group digit (zero-or-more (any digit ".")) digit)
+         line-end)))
+  (match-string 1))
```
bbatsov, line 18220: This sits outside `with-temp-buffer`, so it extracts from whatever buffer is current when the check runs, not from the probe output. Move it inside the form.

flycheck.el
```diff
@@ -18188,15 +18205,26 @@ absent from the table has not been probed yet.")
...
 (defvar flycheck-proselint-executable)
 
+(defun flycheck--proselint-version ()
+  "Return the version number of Proselint, as a string."
+  (with-temp-buffer
+    ;; Probe on the host the check will run on (remote over TRAMP).
+    (process-file (or flycheck-proselint-executable "proselint")
+                  nil t nil "--version")
+    (goto-char (point-min))
+    (search-forward-regexp
```
bbatsov, line 18215: Without the no-error argument this signals when the probe prints nothing recognisable, or the executable is missing, and that takes the whole check down. Returning nil and treating it as new proselint would be the safer fallback.

flycheck.el
bbatsov: Triggered by calling the function directly with an end position of 1, not from real proselint output. I don't know whether proselint ever reports a diagnostic that early, but since this runs inside the parser the guard is cheap insurance.

flycheck.el
mavit: Fixed.

flycheck.el
mavit: Fixed.

PR #2338, opened 2026-08-10
Fix problems caused by emacs-lisp-checkdoc buffer name handling.

gynamics:
checkdoc use buffer-file-name to decide whether or not checking file comments, this check should not be applied to temporary buffers. However, flycheck always use a temporary file, so file comment check is always applied, and the buffer file name in check message is set to the name of temporary file, which can be confusing.

To fix this problem, add source-original to the checker command list. If it is nil or empty string, we don't set buffer-file-name to indicate checkdoc. Otherwise we set buffer-file-name to it, so that file comment check produce messages with correct file name.

### DIFF
--- a/flycheck.el
+++ b/flycheck.el
@@ -13788,6 +13788,7 @@ See Info Node `(elisp)Byte Compilation'."
     (require 'checkdoc)
 
     (let ((source (car command-line-args-left))
+          (source-original (cadr command-line-args-left))
           ;; Remember the default directory of the process
           (process-default-directory default-directory))
       ;; Note that we deliberately use our custom approach even despite of
@@ -13796,7 +13797,8 @@ See Info Node `(elisp)Byte Compilation'."
       ;; for us.
       (with-temp-buffer
         (insert-file-contents source 'visit)
-        (setq buffer-file-name source)
+        (setq buffer-file-name
+              (unless (string-empty-p source-original) source-original))
         ;; And change back to the process default directory to make file-name
         ;; back-substitution work
         (setq default-directory process-default-directory)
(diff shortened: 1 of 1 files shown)


bbatsov:
Can you clarify a bit the exact nature of the problem (e.g. with some screenshot). I assume you're not seeing the right buffer name in the messages, but I want to make sure I understand this correctly.

gynamics:
No, this fix prevents checkdoc's `checkdoc-file-comments-engine` being performed on a temporary buffer which is not a file. This follows the behavior of `checkdoc`, as in `checkdoc.el:542`
```lisp
    ;; check the comments
    (setf (nth 0 status)
          (cond
           ((not buffer-file-name) "Not checked")
           ((checkdoc-file-comments-engine) "Errors")
           (t "Ok")))
```
Currently, since we set `buffer-file-name` to `source` manually, `checkdoc-file-comments-engine` is always performed, but usually it doesn't make sense to a temporary buffer, e. g. an org indirect edit buffer. So I suggest we should judge whether or not setting `buffer-file-name` by `source-original`.

gynamics:
For example, currently I get messages like this in org indirect buffer, which is not a file:

image

With this fix applied, this message won't disturb me.

bbatsov:
Good catch, and checkdoc.el:542 explains the intent well. The problem: with `buffer-file-name` nil, checkdoc labels its output `#<buffer  *temp*>`, which our error pattern can't parse - so these buffers lose *all* checkdoc output, docstring checks included. Silence just looks like the noise is gone.

Keep `buffer-file-name` as is and disable only the file-comments engine instead:

```elisp
(when (string-empty-p source-original)
  (advice-add 'checkdoc-file-comments-engine :override #'ignore))
```

Same effect as what checkdoc does on its own, but the labels stay parseable. A spec for the file-less case would be welcome too.

gynamics:
> checkdoc labels its output `#<buffer *temp*>`, which our error pattern can't parse

I have modified the error pattern to match `#<buffer  *temp*>`, which is the expected temporary buffer label. It works correctly now.

image

PR #2363, opened 2026-08-24
Add :verify for emacs-lisp checker

mavit:
This is fairly prominent, because the following message is displayed for the `*scratch*` buffer on startup if `global-flycheck-mode` is enabled:

> Flycheck: no syntax checker for lisp-interaction-mode can run here; C-c ! v shows why

Before this patch, it didn’t actually show why.


flycheck.el
```diff
@@ -14707,6 +14707,26 @@ See Info Node `(elisp)Byte Compilation'."
       (flycheck-sanitize-errors errors))))
   :modes (emacs-lisp-mode lisp-interaction-mode)
   :enabled flycheck--emacs-lisp-byte-compile-enabled-p
+  :verify
+  (lambda (_)
+    (list (flycheck-verification-result-new
+           :label "`flycheck--emacs-lisp-enabled-p'"
```
bbatsov: The label is an internal function name, which won't mean anything to someone reading the verify buffer. Something like `checkable buffer` with the reason in the message (`no, autoloads file` / `no, package manifest` / `no, dir-locals file`) would say what's actually wrong. A small helper returning that reason, with `flycheck--emacs-lisp-enabled-p` built on top of it, keeps the two from drifting.

flycheck.el
```diff
@@ -14707,6 +14707,26 @@ See Info Node `(elisp)Byte Compilation'."
...
+           :message (if (flycheck--emacs-lisp-enabled-p) "yes" "no")
+           :face (if (flycheck--emacs-lisp-enabled-p) 'success '(bold warning)))
+          (flycheck-verification-result-new
+           :label "trusted content"
+           :message (cond ((not (fboundp 'trusted-content-p))
+                           "Trusted content feature is new in Emacs 30")
```
bbatsov: On Emacs < 30 trust isn't a factor, so a warning-faced row here reads like a problem. I'd skip the row when `trusted-content-p` isn't bound, and when it is bound but returns nil, point at the remedy, e.g. "no, see `trusted-content`".

bbatsov (commented):
Thanks, this was a real gap. Could you add a spec or two (`flycheck-verify-generic-checker` returns the rows, so `test/specs/languages/test-emacs-lisp.el` can check the messages directly) and a changelog entry under New Features? A couple of comments inline.

flycheck.el
mavit: My thought was that the security risk was worth drawing attention to, but you're probably right that this isn't Flycheck's problem to solve.

flycheck.el
mavit: My thought was that, since this is for Emacs lisp, the user would be comfortable clicking through to read the source to see the definition for themselves.  Still, you're right that that's cumbersome.

mavit:
All comments addressed.

test/specs/languages/test-emacs-lisp.el
```diff
@@ -159,7 +178,30 @@
...
+                  :to-equal '(bold warning)))
+
+        (expect (flycheck-may-use-checker 'emacs-lisp-checkdoc) :not :to-be-truthy)
+        (let ((verification-result
+               (car
+                (seq-filter
+                 (lambda (r)
+                   (string-equal (flycheck-verification-result-label r)
+                                 "checkable content"))
+                 (flycheck-verify-generic-checker 'emacs-lisp)))))
```
bbatsov: This block queries `emacs-lisp` again rather than `emacs-lisp-checkdoc`, so the checkdoc side of the verify output isn't covered. Once that's switched it's good to go.

mavit: I've corrected this to check the `emacs-lisp-checkdoc` checker, and added the required verifier to make that test pass.

PR #2376, opened 2026-08-27
Run checkers in Emacs 31’s experimental new `markdown-ts-mode`

mavit:


test/specs/languages/test-proselint.el
```diff
@@ -2,6 +2,8 @@
 ;;; Code:
 (require 'flycheck-buttercup)
 (require 'test-helpers)
+(when (version< "31" emacs-version)
+  (require 'markdown-ts-mode))
```
bbatsov: Worth making this `(require 'markdown-ts-mode nil t)`, here and in test-markdown.el. The mode is experimental, and without NOERROR a build that doesn't ship it fails the whole spec file to load instead of skipping the specs that need it.

bbatsov (commented):
Thanks, the mode coverage looks complete to me. Comment inline, and it needs a changelog entry since it's user visible.

The red Emacs 29 jobs aren't yours: a bazel-mode release broke dependency installation for every PR. Fixed in #2375, so a rebase on master clears them.

test/specs/languages/test-proselint.el
mavit: Done.

mavit:
Changelog entry added.

PR #2377, opened 2026-08-27
Use PATTERN make variable to filter tests

mavit:
The Contributor’s Guide says:

> `make specs` runs all Buttercup specs for Flycheck. Set PATTERN to run only specs matching a specific regular expression, e.g. `make PATTERN='^Mode Line'` specs to run only tests for the mode line.

Make this actually work, by passing the pattern to Buttercup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ OPTIPNG = optipng
 
 # Program options
 EMACSOPTS =
-PATTERN = .*
+PATTERN = \\\`
 # Internal variables
 EMACSBATCH = $(EMACS) -Q --batch -L . -L test/specs $(EMACSOPTS)
 RUNEMACS =
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ compile:
 
 .PHONY: specs
 specs: compile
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,7 +24,7 @@ OPTIPNG = optipng
 
 # Program options
 EMACSOPTS =
-PATTERN = .*
+PATTERN = \\\`
```
bbatsov, line 27: Could you add a short comment on why this isn't just `.*`? Took me a minute to work out: `eask exec` re-expands its arguments, so `.*` globs into filenames and buttercup then tries to open `.readthedocs.yaml` as a directory. Without a note someone will simplify it back and break `make specs`.

bbatsov (commented):
Good catch, the docs have promised this for a while. Checked it here: the default still runs everything, and `PATTERN='^Mode Line'` filters to 15 specs. One thing inline.

The red Emacs 29 jobs aren't yours: a bazel-mode release broke dependency installation for every PR. Fixed in #2375, so a rebase on master clears them.

Makefile
mavit: I've split this into a separate commit.

PR #2400, opened 2026-09-14
Restrict `perlimports` fixer to simple cases

mavit:
The `perlimports` command has gained a new `--sort` setting that creates zero-context multi-hulk diffs that `flycheck-perl-perlimports-parse-diff` can’t generate a replacement for.  Don’t try.


mavit:
An alternative approach I tried, which was modify `flycheck-perl-perlimports-parse-diff` to apply the diffs to a copy of the buffer and examine the outcome, was not successful.  The diffs have zero context, which triggered several bugs in `diff-apply-buffer`, including https://debbugs.gnu.org/cgi/bugreport.cgi?bug=81395.  Additionally, some of the diffs are fake, pretending to apply to whole lines for simplicity, when in reality that's not what the buffer contents looks like.

bbatsov:
Looks right. Before I merge: a parse spec showing that a two-hunk diff comes back with no fix (a literal JSON string next to the existing perlimports parse test in test-perl.el is enough), and a CHANGELOG entry under Bugs fixed, e.g. `perl-perlimports` no longer offers a quick fix for a multi-hunk diff, which perlimports emits with `--sort` and which can't be applied as one replacement.

